PR #19, opened 2015-01-07
Add coveralls support

Callek:
This PR adds coverage back into balrog job runs (its currently covered in jenkins) and exports the report to coveralls.io

Also adds rednose (so we get pretty travis output) and adjusts IRC notification format along the same lines as the rest of our stuff as @petemoore has done.

Latest coveralls from my fork is at https://coveralls.io/builds/1703126  (I can also squash these commits if the many-commit format is not wanted for balrog)

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -23,3 +23,4 @@ omit =
     */test_*
     */vendor/*
     */site-packages/*
+    */site.py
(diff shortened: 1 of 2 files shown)


bhearsum:
r+

petemoore:
Awesome, thanks @Callek for taking care of this!

PR #20, opened 2015-01-28
adding New Relic support to balrog.wsgi

superawesome:
Having some occasional trouble with this app now that it's under heavy load, this should help us track down what's going on.

### DIFF
--- a/balrog.wsgi
+++ b/balrog.wsgi
@@ -1,7 +1,19 @@
 import logging
 from os import path
+from os import getenv
 import site
 import sys
+
+try:
+    import newrelic.agent
+except ImportError:
+    newrelic = False
+if newrelic:
+    newrelic_ini = getenv('NEWRELIC_PYTHON_INI_FILE', False)
+    if newrelic_ini:
+        newrelic.agent.initialize(newrelic_ini)
+    else:
+        newrelic = False
 
 mydir = path.dirname(path.abspath(__file__))
 
@@ -42,3 +54,6 @@ application.config['SENTRY_PROCESSORS'] = ['auslib.util.sentry.SanitizeHeadersPr
 
 if application.config['SENTRY_DSN']:
     sentry.init_app(application)
+
+if newrelic:
+    application = newrelic.agent.wsgi_application()(application)


balrog.wsgi
```diff
@@ -1,8 +1,20 @@
 import logging
 from os import path
+from os import getenv
 import site
 import sys
 
+try:
+    import newrelic.agent
```
bhearsum: Looks like this is a compiled library. It should get added to the requirements/compiled.txt.

balrog.wsgi
```diff
@@ -42,3 +54,6 @@ application.config['SENTRY_PROCESSORS'] = ['auslib.util.sentry.SanitizeHeadersPr
 
 if application.config['SENTRY_DSN']:
     sentry.init_app(application)
+
+if newrelic:
+    application = newrelic.agent.wsgi_application()(application)
```
bhearsum: Should we get this set-up for the Admin application as well? It runs out of admin.wsgi.

balrog.wsgi
superawesome: I'd kinda rather not at this point. These changes are intended to pull in New Relic support if it's available, not require it to be available. In our production environment, it is, so it gets imported. However on a developer's laptop, it may not be (and even if it was, they wouldn't necessarily have a valid license key to use with it), and I don't think we should require it. It's inherently an optional component.

If we do want to require it, I think requirements/prod.txt is probably a better place- signifying that it should exist for production, but not is not critical for other environments.

balrog.wsgi
superawesome: My vote is no... we can, but I don't envision much value there. I don't foresee us ever being concerned about performance on the admin interface (usage being so much lower and limited to internal people/tools), whereas we already are for the update side.

balrog.wsgi
bhearsum: Hm, fair point.

bhearsum:
Let's do it! Should I be filing a bug about getting newrelic packages/config installed on the webheads, and pushing this latest code to production?

superawesome:
The server-side stuff is already present, all we need is a push and an apache restart and data will show up. A bug for the push would be good, but that's all we'll need. :)

bhearsum:
Sounds good. I'm going to wait until Monday to merge this, but I'll file a bug right afterwards.

PR #21, opened 2015-02-26
Bug 1137367 - Filter release API response by product, version, name

rail:


bhearsum:
Is there a reason that you didn't add filtering support to names_only mode? It seems to me that we would want to be able to use names_only + name_prefix....

rail:
I can add that too.

bhearsum:
Please do! r=me with that.

PR #23, opened 2015-08-03
bug 1187938: Add a new class to support Settings

tarekziade:


auslib/test/admin/views/test_releases.py
```diff
@@ -689,3 +701,38 @@ def testPostRevisionRollbackBadRequests(self):
...
+
+        select_rel = dbo.releases.t.select()
+        r = select_rel.where(dbo.releases.name=='settings')
+        r = r.execute().fetchall()
+        self.assertEquals(len(r), 1)
+        rec = r[0]
+        self.assertEquals(rec['name'], 'settings')
+        self.assertEquals(rec['version'], '1')
+        self.assertEquals(rec['product'], 'settings')
+        self.assertDictEqual(byteify(json.loads(rec['data'])),
```
bhearsum: Unfortunately, we can't use Python 2.7 specific features such as assertDictEquals, as we run on Python 2.6 in production :(.

auslib/test/blobs/test_settings.py
```diff
@@ -0,0 +1,44 @@
+import unittest
+from xml.dom import minidom
+
+from auslib.blobs.settings import SettingsBlob
+
+
+_DATA = """\
+{
+    "name": "fake",
+    "schema_version": 1001,
```
bhearsum: This schema version needs updating :)

auslib/blobs/settings.py
```diff
@@ -0,0 +1,43 @@
...
+        }
+    }
+
+    def __init__(self, **kwargs):
+        Blob.__init__(self, **kwargs)
+        if "schema_version" not in self:
+            self["schema_version"] = 2000
+
+    def shouldServeUpdate(self, updateQuery):
+        return False
```
bhearsum: I don't think you want False here...it means any update requests that point at Settings blobs will always get nothing back. I'm pretty sure you want to always return True here, since the client is going to be smart enough to only ping Kinto when the last_modified has changed.

bhearsum:
Sorry for the delay in reviewing this - I wanted to chat with Nick first. One good thing that came out of that conversation is that we realized that even if we decide to put all Go Faster updates in the same ping to Balrog, we can still use separate Blobs for each different type of thing. That means we can land and use this for Settings without blocking on a decision about a consolidated update ping.

I gave this is a go in a Vagrant environment and it should be good to go once the few things above are fixed. We still need to deal with data submission, whitelist settings, etc, but we should be able to start playing in aus4-dev once this lands.

auslib/test/admin/views/test_releases.py
tarekziade: > A normal assertEquals should be fine here.

Sure

> What's the purpose of the byteify'ing?

To compare bytes with bytes, since the json encoding/decoding returns unicode when you initially have strings. e.g.

```python
>>> a = {'aé': '1'}; b = {u'aé': u'1'}
>>> a == b
False
```

auslib/blobs/settings.py
tarekziade: I was not sure what that flag meant. Will put True :)

auslib/test/admin/views/test_releases.py
bhearsum: Ah. We should probably be doing that everywhere...the unittest module is good enough to cast it for us when it can, but I bet that won't work with Python 3...

bhearsum:
Looks like byteify still has Python 2.7 specific code in it, per the failing travis job at https://travis-ci.org/mozilla/balrog/jobs/74569333. Once that's fixed I'm happy to merge this and we can start poking at dev!

tarekziade:
oops sorry I thought I switched to 2.6 but I did not.. fixing

PR #26, opened 2015-09-22
Bug 1201538 - Implement IMEI whitelisting for b2g fota updates; r=bhe…

rwood-moz:
…arsum

### DIFF
--- a/auslib/AUS.py
+++ b/auslib/AUS.py
@@ -6,7 +6,6 @@ import logging
 from auslib.global_state import dbo
 from auslib.log import cef_event, CEF_ALERT
 
-
 def isSpecialURL(url, specialForceHosts):
     if not specialForceHosts:
         return False
@@ -65,6 +64,14 @@ class AUS:
         rules = sorted(rules, key=lambda rule: rule['priority'], reverse=True)
         rule = rules[0]
         self.log.debug("Matching rule: %s" % rule)
+
+        # If the rule requires a whitelist, check via our whitelist blob
+        if rule['whitelist']:
+            self.log.debug("Matching rule requires a whitelist")
+            release = dbo.releases.getReleases(name=rule['whitelist'], limit=1)[0]
+            whitelistBlob = release['data']
+            if not whitelistBlob.shouldServeUpdate(updateQuery):
+                return None, None
 
         # There's a few cases where we have a matching rule but don't want
         # to serve an update:
--- a/auslib/blobs/whitelist.py
+++ b/auslib/blobs/whitelist.py
@@ -1,0 +1,39 @@
+from auslib.blobs.base import Blob
+
+import logging
+log = logging.getLogger(__name__)
+
+
+class WhitelistBlobV1(Blob):
+    format_ = {
+        'name': None,
+        'schema_version': None,
+        'whitelist': [
+            { 'imei': None }
+        ]
+    }
+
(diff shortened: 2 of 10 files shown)


auslib/blobs/whitelist.py
```diff
@@ -0,0 +1,42 @@
...
+                'lastModified="%(last_modified)s"/>')
+
+
+class WhitelistBlobV1(Blob):
+    format_ = {
+        'name': None,
+        'schema_version': None,
+        'whitelist': [
+            { 'imei': None }
+        ]
```
bhearsum: Ah, this is an interesting format for the whitelist! In my head I was thinking it would just be a list, but this looks more flexible -- it allows us to whitelist based on multiple parts of the incoming query in the future. Great!

auslib/AUS.py
```diff
@@ -66,6 +65,14 @@ def evaluateRules(self, updateQuery):
         rule = rules[0]
         self.log.debug("Matching rule: %s" % rule)
 
+        # If the rule requires a whitelisted imei, check via our whitelist blob
+        if rule['whitelist']:
+            self.log.debug("Matching rule requires whitelisted IMEI")
```
bhearsum: Nit: Don't mention IMEI here - that aspect is encapsulated within the whitelist blob. Out here it's just a generic whitelisting column.

auslib/blobs/whitelist.py
```diff
@@ -0,0 +1,42 @@
+import re
+from auslib.blobs.base import Blob
+
+import logging
+log = logging.getLogger(__name__)
+
+INDENT = 4 * ' '
+SETTING_TMPL = ('<setting id="%(id)s" '
```
bhearsum: Looks like these are leftover from copying settings.py - I think they can die?

bhearsum:
This looks pretty darn good! There's a few nits inline above, and Travis caught some existing tests that are now failing -- it looks like some existing test data needs to be updated to include whitelist, eg: https://github.com/mozilla/balrog/blob/master/auslib/test/test_AUS.py#L9

Once that stuff is fixed I'll get this merged!

auslib/blobs/whitelist.py
```diff
@@ -0,0 +1,42 @@
...
+        self.log.debug("Checking to see if IMEI '%s' is whitelisted", requestIMEI)
+        if any(listItem['imei'] == requestIMEI for listItem in self['whitelist']):
+            self.log.debug("IMEI is whitelisted")
+            return True
+        self.log.debug("IMEI not whitelisted")
+        return False
+
+    def shouldServeUpdate(self, updateQuery):
+        self.log.debug(updateQuery)
+        requestIMEI = updateQuery['IMEI']
```
bhearsum: One thing to be careful of here: not every update query type will have an IMEI present. Probably safer to use updateQuery.get() here...

auslib/AUS.py
rwood-moz: Ah good point, ok thanks

auslib/blobs/whitelist.py
rwood-moz: Good call, thx

auslib/blobs/whitelist.py
rwood-moz: Ok right, thanks

PR #28, opened 2015-10-09
Bug 1213420: Support a default build target entry for GMP and SystemAddons releases.

Mossop:


auslib/blobs/gmp.py
```diff
@@ -32,14 +32,15 @@ def __init__(self, **kwargs):
...
 
     def getResolvedPlatform(self, vendor, platform):
-        try:
+        if platform in self['vendors'][vendor]['platforms']:
             return self['vendors'][vendor]['platforms'][platform].get('alias', platform)
-        except KeyError:
-            raise BadDataError("No platform '%s' in vendor '%s'", platform, vendor)
+        if "default" in self['vendors'][vendor]['platforms']:
+            return "default"
+        raise BadDataError("No platform '%s' in vendor '%s'", platform, vendor)
```
nthomas-mozilla: This could say "No platform '%s' or default in vendor '%s'"

auslib/test/blobs/test_gmp.py
```diff
@@ -49,6 +49,11 @@ def setUp(self):
                     "filesize": 666,
                     "hashValue": "666",
                     "fileUrl": "http://evil.com/fire"
+                },
+                "default": {
+                    "filesize": 20,
+                    "hashValue": "20",
```
nthomas-mozilla: I'd prefer to have different values for filesize and hashValue, to avoid accidental test passes.

auslib/test/blobs/test_gmp.py
```diff
@@ -67,6 +80,9 @@ def testGetVendorsForPlatformOnlyInOne(self):
     def testGetResolvedPlatform(self):
         self.assertEquals("q", self.blob.getResolvedPlatform("c", "q2"))
 
+    def testGetResolvedPlatformDefault(self):
+        self.assertEquals("default", self.blob.getResolvedPlatform("d", "q2"))
+
```
nthomas-mozilla: Lets add a test to ensure that we prefer a specified platform over default when both are present, eg 
  self.assertEquals("r", self.blob.getResolvedPlatform("d", "r"))

PR #30, opened 2015-12-07
Bug 886543, supported sending xml about OS desupport

nthomas-mozilla:
Here is support for .... er ... desupporting an OS.


bhearsum:
This looks good to me, r+

PR #33, opened 2015-12-14
bug 1232436: modernize Balrog's toolchain.

bhearsum:
OK, this enables tox+taskcluster-github+setup.py, and fixes some issues flake8 picked up. Travis is going to fail because of the stuff I'm removing, and Taskcluster-github doesn't work until it's been merged. I did a manual kick-off of the new CI task here, which shows it passing: https://tools.taskcluster.net/task-inspector/#BoLfk7lASgW3g4lOh-bxuw/0.

There's a couple of parts of this that may not be obvious:
1) The switch to SQLAlchemy's null(), because flake8 doesn't know how to not-complain about comparing columns to None. null() returns the dialect (aka DB) specific version of NULL, and has the exact same effect as comparing to None.
2) Fixing up of MaybeCacher tests. Before that landed the constructor arguments were removed. I have no idea how those tests have been passing up until now...


auslib/admin/views/releases.py
```diff
@@ -116,7 +118,9 @@ def changeRelease(release, changed_by, transaction, existsCallback, commitCallba
                     cef_event("Bad input", CEF_WARN, errors=msg, release=rel)
                     return Response(status=400, response=msg)
                 if 'hashFunction' in releaseInfo['data'] and hashFunction and hashFunction != releaseInfo['data']['hashFunction']:
-                    msg = "hashFunction '%s' doesn't match the one on the release object ('%s') for release '%s'" % (hashFunction, releaseInfo['data']['hashFunction'], rel)
+                    msg = "hashFunction '%s' doesn't match the one on the release object ('%s') for release '%s'".format(
+                        hashFunction, releaseInfo["data"]["hashFunction"],
+                    )
                     cef_event("Bad input", CEF_WARN, errors=msg, release=rel)
```
nthomas-mozilla: Do you need %(release) in the msg, so that it can be expanded by cef_event() ?

ui
```diff
@@ -1 +1 @@
-Subproject commit ac8b624d4095ecdefd59cf5fe8b55f5397d5cec8
+Subproject commit 9b1abc3624aaf2abc6736fdb1ac3246899c440d2
```
nthomas-mozilla: Why does this change here ?

nthomas-mozilla:
This all looks good to me, please do get rail to check over the tox and other things he knows better though.
The MaybeCacher tests not failing is strange, I can't see why they'd be working either. Presumably they fail if you add an assertEqual(True, False) ? Would be good to understand this.

ui
bhearsum: Subrepos are hard...I had the ui subrepo checked out to include https://bugzilla.mozilla.org/show_bug.cgi?id=1109071 at the time of this commit. I'll fix that.

auslib/admin/views/releases.py
bhearsum: Whoops, I didn't realize I'd removed that. cef_event won't fill in log messages AFAIK, so I need to fix up the formatting (including to switch %s to {} style).

bhearsum:
Here's the latest tc-gh run: https://tools.taskcluster.net/task-inspector/#R5HSqTjCQkm1TT1s9unnlA/0

Makefile
```diff
@@ -1,68 +0,0 @@
-VIRTUALENV ?= /usr/bin/virtualenv
```
rail: DIAF, Makefile!

setup.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+setup(
+    name="balrog",
+    version="1.0",
+    description="Mozilla's Update Server",
+    author="Ben Hearsum",
+    author_email="ben@hearsum.ca",
+    packages=["auslib"],
+    install_requires=[
```
rail: Pycharm thanks you! :)

tox.ini
```diff
@@ -0,0 +1,58 @@
...
+    Paste
+    pyflakes
+    pytest
+    pytest-cov
+    requests
+# compiled deps
+    mysql-python==1.2.3
+    jinja2==2.5.5
+    sqlalchemy==0.7.1
+    simplejson==2.0.9
```
rail: I wonder is there a better way to specify these, so both `tox` and `python setup.py test` would install them... Not a blocker though

rail:
:shipit:

bhearsum:
I dug into the cache test non-failures a bit more. Turns out that they weren't even getting run in Travis (eg: https://travis-ci.org/mozilla/balrog/jobs/90759587 has no mention of them), which explains why they weren't failing. I think that's because there's no __init__ in auslib/test/util. Looks like pytest doesn't care about that, but I'm going to add one anyways for consistency.

bhearsum:
I also noticed that the rules tests weren't running after comparing the travis.yml to taskcluster.yml. I got those going again, and fixed up a few more issues that they uncovered. Here's the latest run: https://tools.taskcluster.net/task-inspector/#HAIscCEyTq-SD834xYs6_A/0

bhearsum:
Thanks for the find_packages tip! I had to use a slightly different invocation to make it work, but it looks good now.

rail:
lgtm

PR #34, opened 2015-12-16
bug 1164289: reverting to prior rules and releases is busted.

bhearsum:
I'm pretty sure that this has never worked with the new UI. When we added support to the backend for the new UI, we started accepting JSON input in addition to regular form data (https://github.com/mozilla/balrog/commit/4338af9338012201e4fc355c778305bc09e905b7). Most of the backend endpoints have Forms associated with them, which handle looking at request.form/args/data/json, but the History endpoints don't so they were missed in the conversion (eg: https://github.com/mozilla/balrog/blob/4338af9338012201e4fc355c778305bc09e905b7/auslib/admin/views/rules.py#L277).

This was never caught in testing because the tests continued to use form data instead of json data. This patch fixes the tests + endpoints to use request.json. It's a bit yucky because request.json can be None in certain cases, so we have to check that before we dive into it.


auslib/test/admin/views/test_releases.py
```diff
@@ -697,11 +697,11 @@ def testPostRevisionRollback(self):
 
     def testPostRevisionRollbackBadRequests(self):
         # when posting you need both the rule_id and the change_id
```
nthomas-mozilla: s/rule_id/release/ ?

auslib/test/admin/views/test_releases.py
```diff
@@ -684,7 +684,7 @@ def testPostRevisionRollback(self):
         assert row['name'] == 'd'  # one of the fixtures
 
         url = '/releases/d/revisions'
-        ret = self._post(url, {'change_id': change_id})
+        ret = self._post(url, json.dumps({'change_id': change_id}), content_type="application/json")
```
nthomas-mozilla: Maybe a JSONTestMixin._post() would be helpful to avoid repetition of content_type.

nthomas-mozilla:
Looks good.

PR #35, opened 2015-12-16
Try to fix MANIFEST.in to include all of auslib/migrate

bhearsum:
This worked fine locally, but my latest pull request is failing because the balrog sdist that gets generated doesn't include auslib/migrate/versions. I'm guessing this is because they are .py files, but there's no __init__.py. Since it really isn't a proper Python module I think the best idea is to include them in the MANIFEST.in...let's see if that works.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,1 +1,2 @@
-include auslib/migrate/migrate.cfg
+recursive-include auslib/migrate *
+recursive-exclude * *.py[co]


bhearsum:
So that's weird...the taskcluster run for this passed shortly before I merged, but now it's pending again?

bhearsum:
And it failed, because I deleted the branch. I know this works, however, because PRs such as https://github.com/mozilla/balrog/pull/34 are now passing.

I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1233042 on this taskcluster-github issue.

PR #36, opened 2015-12-16
Upgrade most of the vendor packages

bhearsum:
This turned out to be quite trivial - no code changes needed. I did notice that the ausilb/util/testing.py tests were failing after I emptied out the vendor lib (and forgot to re-add retry), and I realized that we probably don't have any use for the snippet comparison gunk anymore, so I ripped it out. I can revert that and switch it to redo if there's still a use case for it though.

Because I'm not looking at upgrading binary packages such as SQLAlchemy I thought it best not to take the latest version of sqlalchemy-migrate, but everything else in the vendor lib has been updated to the latest available version.


nthomas-mozilla:
Removing auslib/test/util/test_comparison.py looks like collateral damage, lets add that back. As for removing the server vs server testing code, that's probably fine.

Any special checks needed for deploying this ? I'm not sure how complete our test coverage is.

bhearsum:
Good catch on the comparison tests.

I intend to do some extra testing in dev before pushing to prod, probably in combination with a few other patches. I'll do some reruns of various types of update submissions, and at least one update verify test. I think that will hit the vast majority of the code. I'm happy to do more if you have some ideas though.

PR #37, opened 2015-12-24
Make whitelisting work like other rule checks to allow for fallthrough when it fails

bhearsum:
Right now, if an update query fails a rule's whitelist check, no more rules are processed and it gets no update at all. This sucks, because it means we can't have any sort of fallback for non-whitelisted users.

This pull request fixes that; it moves the processing to the Rules class, which means that rules with a whitelist are only returned from getRulesMatchingQuery if they pass the whitelist check, or have no whitelist. I originally put this in evaluateRules, but it ended up being cleaner in Rules (thanks for the suggestion, @catlee!)

I haven't done any testing beyond the unit tests, but it looks like it should work!


auslib/db.py
```diff
@@ -789,7 +789,22 @@ def getRulesMatchingQuery(self, updateQuery, fallbackChannel, transaction=None):
...
+            # considered "matching" if it passes the whitelist check.
+            # The decision about matching or not is delegated to the whitelist blob.
+            if rule.get("whitelist"):
+                self.log.debug("Matching rule requires a whitelist")
+                try:
+                    whitelist = dbo.releases.getReleaseBlob(name=rule["whitelist"], transaction=transaction)
+                    if whitelist and not whitelist.shouldServeUpdate(updateQuery):
+                        continue
+                # It shouldn't be possible for the whitelist blob not to exist,
+                # but just in case...
```
catlee: does shouldServeUpdate exist for all blob types? could that ever raise an exception we shoudl handle here?

catlee:
r+ assuming we're handling exceptions properly above

auslib/db.py
bhearsum: It exists for all Blobs. Any which don't override will inherit the NotImplementedError it raises though: https://github.com/mozilla/balrog/blob/master/auslib/blobs/base.py#L113

The expectation is that all blob types should override it, but since Python doesn't enforce that....I guess we need to do something. I'm tempted to have the base implementation return False, it feels weird to be handling a NotImplementedError.

bhearsum:
What do you think of this now, @catlee?

catlee:
r+

PR #38, opened 2015-12-29
Switch Balrog to Docker + uWSGI

bhearsum:
This patch adds everything needed to run Balrog in Docker containers. I did a bit more re-org/cleanup along with it.

The general approach is to run the Balrog apps within Docker with uWSGI. I've created uWSGI ini files and new .wsgi files for this purpose (the old ones have been retained for compat until we switch over production). I spoke with CloudOps last week and they said they typically run one Docker image per host in --net host mode, so that's what I'm optimizing for here. I've moved all the non-secret parts of the config into the new wsgi files (no reason not to AFAICT); the secrets parts are grabbed from the environment. A mysql Docker image is also needed to host the database for local development. I wrote a small script that sets all of these up.

Other changes:
* Get rid of .ini examples, because they're unused in Docker
* Include ui/dist in Balrog package (because it's part of the admin app)
* Add binary requirements to setup.py list
* Move standalone scripts + sample data to scripts/ directory, to cleanup top level a bit
* Update sample data
* Kill Vagrantfile + associated junk (mostly puppet stuff)
* Update README; remove outdated EXAMPLE and DATA_FORMAT files

Originally, I killed the "vendor" dir and old .wsgi files, but I'd like to be able to merge this without breaking the current production, so I added them back. Once we've transitioned the infra we can kill them.


rail:
lgtm, couldn't find anything wrong. :+1:

PR #40, opened 2015-12-31
Allow rules to have alias'

bhearsum:
This patch adds "alias" support for rules. I decided to go with single alias per rule, and to not allow alias' to be used in history operations. I think both of these things make things simpler and more clear, but I could be convinced otherwise.

Most of this patch should be pretty straightforward. I included a change to one of the testing mixins that removes an assertion about the mime type of the response. When I added the testGetRefusesAlias test for the history view that assertion was failing, and doing so in a way that was eating the actual error. I don't really think it's adding much value (the other tests that have been using it try to parse the response as JSON anyways, so they'll fail if the response is otherwise).

### DIFF
--- a/auslib/admin/views/forms.py
+++ b/auslib/admin/views/forms.py
@@ -9,6 +9,10 @@ import logging
 log = logging.getLogger(__name__)
 
 
+# If present, rule alias' must be a string containing at least one non-numeric character.
+RULE_ALIAS_REGEXP = "(^.*[^0-9].*$|^$)"
+
+
 class DisableableTextInput(TextInput):
     """A TextInput widget that supports being disabled."""
 
@@ -108,7 +112,7 @@ class RuleForm(Form):
     backgroundRate = IntegerField('Background Rate', validators=[Required(), NumberRange(0, 100)])
     priority = IntegerField('Priority', validators=[Required()])
     mapping = SelectField('Mapping', validators=[])
-    alias = NullableStringField('Alias', validators=[Length(0, 50), Regexp("(^.*[^0-9].*$|^$)")])
+    alias = NullableStringField('Alias', validators=[Length(0, 50), Regexp(RULE_ALIAS_REGEXP)])
     product = NullableStringField('Product', validators=[Length(0, 15)])
     version = NullableStringField('Version', validators=[Length(0, 10)])
     buildID = NullableStringField('BuildID', validators=[Length(0, 20)])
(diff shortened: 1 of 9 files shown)


auslib/admin/views/forms.py
```diff
@@ -3,12 +3,16 @@
...
 from wtforms.widgets import TextInput, FileInput, HiddenInput
-from wtforms.validators import Required, Optional, NumberRange, Length
+from wtforms.validators import Required, Optional, NumberRange, Length, Regexp
 
 import logging
 log = logging.getLogger(__name__)
 
 
+# If present, rule alias' must be a string containing at least one non-numeric character.
+RULE_ALIAS_REGEXP = "(^.*[^0-9].*$|^$)"
```
catlee: this is pretty complex. can we say an alias has to start with a letter, and contain only alphanumeric characters, plus maybe -? i.e. "[a-zA-Z][-a-zA-Z0-9]+" ?

auslib/db.py
```diff
@@ -637,6 +637,7 @@ class Rules(AUSTable):
     def __init__(self, metadata, dialect):
         self.table = Table('rules', metadata,
                            Column('rule_id', Integer, primary_key=True, autoincrement=True),
+                           Column('alias', String(50), unique=True),
```
catlee: does unique imply an index? i.e. querying by alias will be fast?

auslib/db.py
```diff
@@ -796,22 +797,25 @@ def getRulesMatchingQuery(self, updateQuery, fallbackChannel, transaction=None):
                 self.log.debug(r)
         return matchingRules
 
-    def getRuleById(self, rule_id, transaction=None):
-        """ Returns the unique rule that matches the give rule_id """
-        rules = self.select(where=[self.rule_id == rule_id], transaction=transaction)
+    def getRule(self, id_or_alias, transaction=None):
+        """ Returns the unique rule that matches the give rule_id or alias."""
+        rules = self.select(
+            where=[(self.alias == id_or_alias) | (self.rule_id == id_or_alias)],
```
catlee: would it be better to figure out if id_or_alias is an id or alias first? IIRC, you can only refer to a single index per table per query, so this could be slow unless we add a new index for both columns, or figure out which column we want to query first.

auslib/admin/views/forms.py
bhearsum: Yeah, that should be fine. It's probably an anti-feature to allow lots of special characters anyways!

auslib/db.py
bhearsum: Good point. I don't think it would be hugely slower to use the current query, but it should be easy and cheap to figure out what it is...so we may as well.

auslib/db.py
bhearsum: Doesn't look like one was created when I upgraded my local db. I'll update the migrate script to add it.

bhearsum:
I think I've addressed all of your comments, @catlee.

catlee:
lgtm

bhearsum:
From IRC:
16:00 <@catlee> bhearsum: for https://github.com/mozilla/balrog/pull/40, would it make sense to have getRule() use the same regexp as the form to determine if it's a rule id or alias?
16:00 <@catlee> or maybe write a function for it

Going to investigate this before merging.

bhearsum:
@catlee - How's this?

catlee:
yep, looks good!

PR #41, opened 2016-01-06
Update permissions list to match new API

bhearsum:
Turns out this didn't get updated when the new API was added. This patch fixes that (adds /releases, updates the rollback endpoints), and also fixes a bug in the Docker container creation script that I found while trying to test this (prevents the removal of the initial admin permission in the database).


catlee:
r+

PR #42, opened 2016-01-11
Switch Balrog to JSONSchemas for blob validation

bhearsum:
This is not ready to be merged, but it's a fully working POC that switches all of the Balrog blobs over to JSONSchemas, and removes their old-style formats. I used YAML for the schemas so we could continue to comment them.

I had to make a few minor changes to other code to make this work, including:
* lower()'ing True/False values before putting them in XML. Prior to this patch, we used to add them as "true" or "false" strings. We might be able to continue doing that, using a jsonschema enum to validate. The current way felt a bit cleaner to me, but I'm not tied to it.
* Kill base Blob tests. These did nothing except validate the isValidBlob method, which is now dead.
* Get rid of ISE 500 handler for tests. This causes lots of tracebacks to get eaten, and makes debugging harder. Turning this off caused a test to fail, which in turn caused me to find a bug where some parts of a release can be updated even if an exception is raised in other parts! I commented out that test for now, and intend to fix it in a different patch.

@rail, do you feel comfortable giving some feedback on this?


auslib/blobs/base.py
```diff
@@ -88,17 +46,28 @@ def createBlob(data):
...
+        # raises exception if broken
+        jsonschema.validate(self, schema)
+        return True
+
+    def getSchema(self):
+        if self.jsonschema in self.cached_schemas:
+            return self.cached_schemas[self.jsonschema]
+
+        schema = yaml.load(open(path.join(path.dirname(path.abspath(__file__)), "schemas", self.jsonschema)))
+        self.cached_schemas[self.jsonschema] = self.cached_schemas
```
rail: Shouldn't it be 
```python
self.cached_schemas[self.jsonschema] = schema
```

bhearsum: Yeah, this caching was busted in multiple ways. I fixed it to use the actual global cache in my latest commits.

auslib/test/admin/views/test_releases.py
```diff
@@ -382,22 +451,26 @@ def testLocaleGet404(self):
...
-        data = json.dumps(dict(complete=dict(filesize=1)))
-        with mock.patch('auslib.global_state.dbo.releases.addLocaleToRelease') as r:
-            r.side_effect = Exception("Fail")
-            ret = self._put('/releases/a/builds/p/l', data=dict(data=data, product='a', version='c', data_version=1, schema_version=1))
-            self.assertStatusCode(ret, 500)
-            ret = dbo.releases.t.select().where(dbo.releases.name == 'a').execute().fetchone()
-            self.assertEqual(ret['product'], 'a')
-            self.assertEqual(ret['version'], 'a')
-            self.assertEqual(json.loads(ret['data']), dict(name='a', hashFunction="sha512", schema_version=1))
+    # FIXME: We shouldn't rely on 500 to validate behaviour. This test should fake a 400 instead.
```
bhearsum: Don't praise me too much :). I'm not going to fix this as part of this patch unless you insist.

bhearsum:
OK, this is ready for proper review now. Since you first looked at it I added some additional validation in the schemas and fixed the caching.

bhearsum:
This is ready to land now. I'm not going to merge it until https://bugzilla.mozilla.org/show_bug.cgi?id=1238944 is ready, because they both require a config update, and I don't want to put the "master" branch in limbo for too long.

PR #43, opened 2016-01-12
Fix options handling

bhearsum:
Turns out this is somewhat related to bug 1237263, in that it likely broke when we switched to the new UI. This patch fixes it by switching "options" to a JSONStringField (which will mark the field as erroring if it can't interpret the value), and actually _enabling_ validation on the form when making changes to a permission.

I also:
* Killed the custom widget for ExistingPermissionForm, because we don't render anything with WTForms anymore.
* Fixed the /rules permission which was added recently, which I mistakenly didn't allow "method" to be configured on.


rail:
:+1:

PR #45, opened 2016-01-13
Improve coverage

bhearsum:
While working on other patches I noticed that coverage is enabled for auslib/migrate, which is a bit silly. This patch fixes that, moves coverage to a config, and enables branch coverage for better accuracy. Still at 90% coverage, woo!


bhearsum:
@rail - does this look okay to you?

rail:
:+1:

PR #46, opened 2016-01-18
bug 1238944: enable caching for balrog admin app

bhearsum:
This turned out to be a bit more involved than I thought. The main issue that came up was that because we're enabling caching on a rw app, we need to make sure that we're updating the cache whenever the database changes (otherwise you end up with an outdated cache). It's also important to note that I'm explicitly _not_ enabling the blob_version cache. This is because our admin app runs in multiple wsgi processes, so even though one will be updated whenever it changes, the rest won't notice until they expire. Enabling the blob_version cache would probably end up causing a lot of collisions because of this. Maybe we should look at using memcache down the road, or maybe that would just shunt the network traffic elsewhere?

The other thing worth noting here is the weirdness I encountered after adding the deepycopy() for objects in the cache. Blobs are a dict subclass, and when they get copied, each piece of instance data on them is copied. This turns out to be an issue when self.log is a stream, which is unpickable. Moving it to a class level attribute fixed it at the cost of some repeated code.

I've done as much testing as I can with my local Docker image. I intend to do a bunch more after this + https://bugzilla.mozilla.org/show_bug.cgi?id=1240545 are in dev. Production funsize sends a lot of traffic to Balrog dev, and that should help verify both that I haven't broken anything, and that the cache is actually helping.


bhearsum:
@catlee - I found a slightly different solution to the logging problem. Turns out that no matter where you do it (__new__, metaclass, __init__) the key is that you need to explicitly put it into the class attributes, so I just did this in plain-old Blob.__init__. Was there anything else in here you wanted to see addressed?

catlee:
Looks much better this way!

Does __class__.__name__ include the full module path?

bhearsum:
It doesn't....but we should probably include that - it's in __module__.

PR #47, opened 2016-01-21
Add jsonschema for new DesupportBlob

bhearsum:
I missed this one because it was buried in the apprelease.py file...


rail:
:shipit:

PR #49, opened 2016-01-25
Improve blob validation errors

bhearsum:
The current state of things is that a massive jsonschema error (with the entire blob in it) will be returned if any part of the schema fails. This is really quite useless on the UI, and this patch fixes it up. I had to add a custom exception because there's post-processing of the errors to be done, and I didn't want to repeat that everywhere that we're validating blobs. I also renamed isValid to validate to make it more clear that it raises ("isValid" sounds like it should return True or False, not raise). There's a small UI patch to cope with the returning of a list instead of a string too. With both of them, we get errors like: http://people.mozilla.org/~bhearsum/sattap/bea114d5.png.

### DIFF
--- a/auslib/blobs/base.py
+++ b/auslib/blobs/base.py
@@ -76,8 +76,6 @@ class Blob(dict):
         errors = [e.message for e in validator.iter_errors(self)]
         if errors:
             raise BlobValidationError("Invalid blob! See 'errors' for details.", errors)
-        else:
-            return True
 
     def getSchema(self):
         def loadSchema():
--- a/auslib/blobs/base.py
+++ b/auslib/blobs/base.py
@@ -11,6 +11,12 @@ from auslib.AUS import isSpecialURL
 from auslib.global_state import cache
 
 
+class BlobValidationError(ValueError):
+    def __init__(self, message, errors, *args, **kwargs):
+        self.errors = errors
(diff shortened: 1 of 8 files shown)


auslib/blobs/base.py
```diff
@@ -59,13 +64,20 @@ def __init__(self, *args, **kwargs):
...
+        # Normal usage is to use .validate(), but errors raised by it return
+        # a massive error message that includes the entire blob, which is way
+        # too big to be useful in the UI. Instead, we iterate over the
+        # individual errors (which are all single sentences which contain
+        # the name of the failing property and why it failed), and return those.
+        errors = [e.message for e in validator.iter_errors(self)]
+        if errors:
+            raise BlobValidationError("Invalid blob! See 'errors' for details.", errors)
+        else:
+            return True
```
rail: This return looks artificial to me, it's made for tests only... Maybe delete the `else` block and use `self.assertIsNone(self.blob.validate())` instead? Unfortunately there is no opposite of `this.assertRaises()`

bhearsum: Yeah, that's a good point. I usually just use self.blob.validate() for cases like that, and add a comment saying an exception will be raised if there's an error.

rail:
:+1:

PR #51, opened 2016-02-08
Kill version in releases table.

bhearsum:
I also improved the UI in a couple of places:
1) Use "Product: foo" in the main product display, rather than just showing the bolded product.
2) Title the "View Data" modal with release name rather than product name.

### DIFF
--- a/auslib/test/admin/views/test_releases.py
+++ b/auslib/test/admin/views/test_releases.py
@@ -685,7 +685,7 @@ class TestReleaseHistoryView(ViewTest, JSONTestMixin):
 
     def testPostRevisionRollback(self):
         # Make some changes to a release
-        data = json.dumps(dict(detailsUrl='blah', fakePartials=True, schema_version=1))
+        data = json.dumps(dict(detailsUrl='beep', fakePartials=True, schema_version=1))
         ret = self._post(
             '/releases/d',
             data=dict(
@@ -696,7 +696,7 @@ class TestReleaseHistoryView(ViewTest, JSONTestMixin):
         )
         self.assertStatusCode(ret, 200)
 
-        data = json.dumps(dict(detailsUrl='blah', fakePartials=False, schema_version=1))
+        data = json.dumps(dict(detailsUrl='boop', fakePartials=False, schema_version=1))
         ret = self._post(
             '/releases/d',
             data=dict(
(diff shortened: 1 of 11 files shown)


auslib/test/admin/views/test_releases.py
```diff
@@ -735,39 +691,34 @@ def testPostRevisionRollback(self):
...
         count, = query.execute().first()
-        self.assertEqual(count, 2 * 2)
+        self.assertEqual(count, 2)
 
-        # Oh no! We prefer the version 222.0
         row, = table.history.select(
-            where=[table.history.version == '222.0'],
+            where=[table.history.product == 'd'],
             limit=1
         )
```
nthomas-mozilla: I think we ought to add some sort of reverse sorting here, to ensure we revert the most-recent change.

auslib/test/admin/views/test_releases.py
```diff
@@ -779,11 +730,10 @@ def testPostRevisionRollback(self):
 
         query = table.history.t.count()
         count, = query.execute().first()
-        self.assertEqual(count, 2 * 2 + 1)
+        self.assertEqual(count, 3)
 
         row, = table.select(where=[table.name == 'd'])
-        self.assertEqual(row['version'], '222.0')
-        self.assertEqual(row['data_version'], 6)
+        self.assertEqual(row['data_version'], 4)
```
nthomas-mozilla: Lets add a check that the release blob has been reverted properly. Ideally we can tell the difference between test setup --> True --> False --> <reverted value>, so might need to swap to something numerical over boolean.

nthomas-mozilla:
r+, but please make testPostRevisionRollback a little more robust prior to merging. And just to confirm - a submission script that still sends a version will be fine, that'll be ignored ?

bhearsum:
I'll definitely fix up that test first. That's a very good point about the submission script though...I think it may cause a 400. I'll look into that as well.

bhearsum:
Looks like passing random unknown POST arguments doesn't break anything, so we're good on that front: http://people.mozilla.org/~bhearsum/sattap/2a89a05f.png. I'll be updating the submission script anyways though, of course. I'll push a fix for testPostRevisionRollback in a bit.

bhearsum:
How's this @nthomas-mozilla?

nthomas-mozilla:
Looks good to me.

PR #52, opened 2016-02-12
Add docker-compose support

rail:
This adds [docker-compose](https://docs.docker.com/compose/) support for running local docker instances.

`sample-data.sql` is outdated, so I created a new file with only permissions left in place.. We should find a better solution for this, but this is another issue.


bhearsum:
On sample-data.sql, I think one thing that may help is to write a unit test that makes sure it can be successfully imported. Had that existed, I would've been forced to update the sample data before landing my most recent schema change...

bhearsum:
I'm still not completely sure if it's better to have everyone building their images locally, or pulling a central one, but using docker compose is definitely better than my script. Let's start with this, and iterate from there.

rail:
One of the pros of building locally is the fact you fail early and this way we can keep Dockerfile up to date. The same pro is also a con, because you may end up dealing with docker instead of code, but this is dev after all. ;)

As an option we can add another yml file with `s/build:.*/image: bhearsum/balrog:latest` and use in cases like that.

PR #53, opened 2016-02-12
Add test for sample data + update it.

bhearsum:
@rail - this is a quick fix for the sample data. I still don't think we should be packaging it with the library...it's too big.

### DIFF
--- a/auslib/test/test_db.py
+++ b/auslib/test/test_db.py
@@ -1,6 +1,7 @@
 import logging
 import mock
 import os
+from os import path
 import simplejson as json
 import sys
 from tempfile import mkstemp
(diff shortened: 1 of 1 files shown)


auslib/test/test_db.py
```diff
@@ -546,6 +547,24 @@ def testMultiplePrimaryHistoryDeleteRollback(self):
             self.assertEquals(len(ret), 5, msg=ret)
 
 
+class TestSampleData(unittest.TestCase, MemoryDatabaseMixin):
+    """Tests to ensure that the current sample data (used by Docker) is
+    compatible with the current schema."""
+    sample_data = path.join(__file__, "..", "..", "scripts", "sample-data.sql")
+    sample_data = "/home/bhearsum/repos/master/balrog/scripts/sample-data.sql"
```
rail: no such file or directory :P

bhearsum: Oops...I thought I took that out! No wonder it was passing :D. I'll fix this.

rail: otherwise lgtm

rail:
:shipit:

PR #54, opened 2016-02-29
Bug 1113111 - tracebacks when a platform exists in a blob but has no …

nurav:
Changed `dict[key]` access to `dict.get(key, default)`.

### DIFF
--- a/auslib/blobs/apprelease.py
+++ b/auslib/blobs/apprelease.py
@@ -58,12 +58,16 @@ class ReleaseBlobBase(Blob):
 
     def getBuildID(self, platform, locale):
         platform = self.getResolvedPlatform(platform)
-	if locale not in self['platforms'].get(platform, {}) \
+        if locale not in self['platforms'].get(platform, {}) \
                                           .get('locales', {}):
             raise BadDataError("No such locale '%s' in platform '%s'" % (locale, platform))
         try:
             return self['platforms'][platform]['locales'][locale]['buildID']
         except KeyError:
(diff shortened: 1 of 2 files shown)


bhearsum:
This seems okay, can you add a test to ensure we don't regress it again? You'll need something similar to https://github.com/mozilla/balrog/blob/master/auslib/test/blobs/test_apprelease.py#L72

nurav:
Sure! On it.

bhearsum:
I was doing some final testing of this and I realized that there's still a case that's not handled here, specifically if a KeyError is hit on line 67. Can you fix that as well and add a test?

There's a few pep8 issues to deal with:
auslib/blobs/apprelease.py:61:1: E101 indentation contains mixed spaces and tabs
	if locale not in self['platforms'].get(platform, {}) \
^
auslib/blobs/apprelease.py:61:1: W191 indentation contains tabs
	if locale not in self['platforms'].get(platform, {}) \
^
auslib/blobs/apprelease.py:62:1: E101 indentation contains mixed spaces and tabs
                                          .get('locales', {}):
^
auslib/blobs/apprelease.py:62:43: E127 continuation line over-indented for visual indent
                                          .get('locales', {}):
                                          ^

bhearsum:
I triggered tests on this in https://tools.taskcluster.net/task-inspector/#et5cZZCIRPyWrMz1c9DyBQ/0, which passed. The code looks good to me as well. Thank you very much!

PR #55, opened 2016-03-02
Add support for Dockerflow requirements (minus mozlog)

bhearsum:
This should make Balrog compatible with all of the Dockerflow requirements except for Mozlog (which I'll address in a follow-up). Here's a summary:
* Add Docker Build/Push job to Taskcluster graph. This is working fine, and has already pushed to https://hub.docker.com/r/mozilla/balrog/ once. It won't work at the moment because I had an issue after I locked down the scopes a bit. https://bugzilla.mozilla.org/show_bug.cgi?id=1252895 is going to fix that, so I won't be merging this until that's ready.
* Change the Dockerfile to generate production-ready docker images. It no longer copies unnecessary files to the image, and no longer installs Balrog to the system. We only ever did that to get the dependencies installed, and I've switched that to be done with "pip install -r requirements.txt".
* Add support for __version__, __heartbeat__, and __lbheartbeat__ to the admin & public apps. I had to switch where the wsgi app was mounted to make this work with uwsgi (previously mounted the app to /api, which prevented the __ endpoints from working)
** __version__ looks for version.json to find that information, and I've added a stub to the repo. A proper one is generated as part of the docker build process.
* Look at $PORT instead of hardcoding listen addresses
* Don't set REMOTE_USER in production uwsgi; only set it for dev.
* Fix mysql docker image to work on Mac (not related to this...but probably fixes bug 1247491)


mostlygeek:
This is looking really good. 

Do you have a plan for release tagging for the container? For example, do you plan to build `master` into a `:latest` container tag, or into a `balrog:release-$YYYY-MM-DD` typed timestamped tag?

bhearsum:
I don't have a strong preference. It seems to be pretty standard to have a "latest" tag available for Docker images, but I can see the case for having a dated one too. Let me know what works best for you, and we can do that :)

mostlygeek:
The convention I've been going with is to 

1. `master` => `:latest`
2. `{tagname}` => `:{tagname}`

Then we can deploy tags. Tag names can be a YYYY-MM-DD-HH if you want :)

bhearsum:
Ah, I'd forgotten that we talked about deploying tags. That's a good idea though, I'll see what I can do there.

bhearsum:
OK, so Taskcluster-Github doesn't support triggering events based on creation of releases/tags yet. I'm going to get a bug on file for that. In the meantime, I'm going to update this to always push master -> :latest, and other branches to :$branch.

Does having tags block us from going to production?

bhearsum:
https://bugzilla.mozilla.org/show_bug.cgi?id=1253619 for Taskcluster-Github supporting "release" events.

version.txt
```diff
@@ -0,0 +1 @@
+2.0
```
nthomas-mozilla: We also have a version defined in auslib/__init__.py, providing auslib.version. Can we squash the two use-cases into one ?

auslib/test/admin/views/test_dockerflow.py
```diff
@@ -0,0 +1,34 @@
...
+            ret = self.client.get("/__heartbeat__")
+            self.assertEqual(ret.status_code, 200)
+            self.assertEqual(cr.call_count, 1)
+
+    def testHeartbeatWithException(self):
+        with mock.patch("auslib.global_state.dbo.rules.countRules") as cr:
+            cr.side_effect = Exception("kabom!")
+            # Because there's no web server between us and the endpoint, we recieve
+            # the Exception directly instead of a 500 error
+            self.assertRaises(Exception, self.client.get, "/__heartbeat__")
```
nthomas-mozilla: Will the actual endpoint go through the 500 handler and return <updates></updates> ? Does that matter to the monitoring system ?

s/kabom/kaboom/, or is that a reference I missed ?
s/recieve/receive/

auslib/admin/base.py
```diff
@@ -59,3 +62,40 @@ def add_security_headers(response):
 app.add_url_rule("/releases/<release>/revisions", view_func=ReleaseHistoryView.as_view("release_revisions"))
 app.add_url_rule("/history/diff/<type_>/<change_id>/<field>", view_func=DiffView.as_view("diff"))
 app.add_url_rule("/history/view/<type_>/<change_id>/<field>", view_func=FieldView.as_view("field"))
+
+
+# Endpoints required by CloudOps as part of the Dockerflow spec: https://github.com/mozilla-services/Dockerflow
```
nthomas-mozilla: Can these be imported from a common file rather than duplicated for admin and client side ?

version.txt
bhearsum: Good point, I don't see any reason why these can't be merged.

auslib/test/admin/views/test_dockerflow.py
bhearsum: I think that will matter to the monitoring system...if this gets caught and converted to the empty update xml it ends up as a 200 instead of 500. I thought I had dealt with this, but maybe I forgot about it after realizing I couldn't get a get() request to return 500...I'll look into this more.

uwsgi/admin.ini
```diff
@@ -11,12 +11,16 @@ check-static = /app/ui/dist
...
-# Always set REMOTE_USER
-# TODO: Where to put this when this config is shared by production?
-route = ^/.*$ addvar:REMOTE_USER=balrogadmin
-
-# Skip further routing for requests to /api
-route-if = startswith:${REQUEST_URI};/api continue:
+# Because the Angular app and the WSGI app are both hosted at /, we need to
+# set-up some fancy routing to cope. Specifically, we need to take any URIs
+# that should be for the admin app, and direct them there. These are:
+# Anything coming in to /api, which also needs to be stripped away before
```
nthomas-mozilla: I don't grok the setup here, but want to check if we are setting ourselves up for future pain.

nthomas-mozilla:
Only basic review of taskcluster/docker stuff as not my area of expertise. 
* I suggest adding another TODO to scripts/push-dockerimage.sh about dated tags
* does the new requirements.txt mean we will drop vendor/ ?  Would be great if we add hashes to requirements.txt and leverage http://pip.readthedocs.org/en/latest/reference/pip_install/#hash-checking-mode

I'd like to understand the developer workflow. Do we copy auslib into the container for both prod and dev cases ? If so, does that need to be redone after each code modification ? Pretty sure I missed something here.

bhearsum:
I'll add that TODO, and yeah, we can kill vendor/ once we're done with our webops deployment, we won't be using it for local dev nor cloudops deployments.

As for the workflow, production will always have the latest code copied into the images that are built on pushes to the "master" branch. This happens at https://github.com/mozilla/balrog/pull/55/files#diff-3254677a7917c6c01f55212f86c57fbfR16 in the Dockerfile.

Dev images will be built from the same Dockerfile, but docker-compose overrides /app with a bind mount, so your local balrog repo will always be used instead of the code that was copied in when the image was built. uwsgi also autoreloads whenever code changes, which is a nice enhancement over the old Vagrant set-up! So the workflow for testing is:
* Create branch
* Write code
* "docker-compose up"
* Test in browser
* Fix stuff in local code (uwsgi autoreloads)
* Repeat last 2 steps until done.

uwsgi/admin.ini
bhearsum: So long as we're hosting the UI and the wsgi app in the same place I think there's risk of pain, I think. Admittedly, serving them both out of / instead of putting the wsgi app in /api is worse, but I couldn't find a way to keep it in /api and make the Dockerflow endpoints work :(. Ultimately, I think a better solution is to host them on different domains, eg: aus-api.mozilla.org and aus-admin.mozilla.org.

auslib/admin/base.py
bhearsum: Probably....I'll take a stab at this.

bhearsum:
Since Nick's review I have made the following changes:
* Added (currently commented out) hashes to requirements.txt. Because of our dependency on Python 2.6, I have to comment out "functools32" in requirements.txt (otherwise installations fails on 2.6, because that module is not available). However, doing this means that there's no way to provide the hash when we try to install it on Python 2.7 (because it's commented out), so that ends up failing. I intend to desupport Python 2.6 a week or so before we switch production to the CloudOps stack and uncomment all of the hash verification stuff at that time.
* Centralize creation of Dockerflow endpoints
* Stop eating 500s on __heartbeat__
* Update Docker build/push script to create branch-based tags as well as date based ones. Each push will get a tag for its branch name (or "latest" for master) as well as a $branch-YYYY-MM-DD-HH:MM tag. Once https://bugzilla.mozilla.org/show_bug.cgi?id=1253619 is fixed we can probably replace the date based tags with ones named after the Git tag names.

If this looks good to you both, I'll merge it after https://bugzilla.mozilla.org/show_bug.cgi?id=1255439 is fixed, and it should create our first mozilla/balrog:latest image!

Dockerfile
```diff
@@ -6,17 +6,13 @@ ENV DEBIAN_FRONTEND=noninteractive
 
 RUN apt-get -q update && \
     apt-get -q --yes install \
-      mysql-client \
-      nodejs-legacy \
-      npm && \
+      mysql-client && \
```
mostlygeek: why do you need mysql-client in the container?

auslib/dockerflow.py
```diff
@@ -0,0 +1,40 @@
...
+                "commit": "unknown",
+            })
+
+    @app.route("/__heartbeat__")
+    def heartbeat():
+        """Per the Dockerflow spec:
+        Respond to /__heartbeat__ with a HTTP 200 or 5xx on error. This should
+        depend on services like the database to also ensure they are healthy."""
+        # Counting the rules should be a trivial enough operation that it won't
+        # cause notable load, but will verify that the database works.
```
bhearsum: How often will this this endpoint get hit? I'm tempted to continue querying a real table because it's a bit more robust.

mostlygeek: Probably dozens of times a minute. It's not high traffic. If you set a `Cache-Control: max-age=20` header we can rate limit we don't have to worry about a DOS. Then you can check whatever you think is appropriate. :)

Dockerfile
mostlygeek: I would recommend a a dev specific one. Then you don't have to worry about mixing Dockerfile logic for different environments, like the `CMD` and `ENTRYPOINT` instructions. You may want to leave these out for a dev container.

Dockerfile
```diff
@@ -6,17 +6,13 @@ ENV DEBIAN_FRONTEND=noninteractive
...
-
-WORKDIR /app/ui
-RUN npm install
-
-WORKDIR /app
+# Copying Balrog to /app instead of installing it means that production can run
+# it, and we can bind mount to override it for local development.
+COPY auslib requirements.txt setup.py ui uwsgi version.json /app/
+# Add --require-hashes after python2.6 support is dropped, and uncomment everything in requirements.txt
+RUN pip install -r requirements.txt
```
mostlygeek: I'm back/forth on this Dockerflow requirement as well. It's there so the container can be treated like any linux executable. 

how about: 

run.sh
```
#!/bin/sh

if [ $1 == "public" ]; then
   exec <cmd to run uswgi public>
elif [ $1 == "admin" ]; then
   exec <cmd to run uwsgi admin>
else
   echo "unknown mode: $1"
   exit 1
fi
```

The `exec` is important as it will replace the shell with another program. This means signals `SIGHUP`, `SIGKILL`, etc will pass to uwsgi correctly.

```

In the Dockerfile, the default is to run the public facing app
...
ENTRYPOINT ["/app/bin/run.sh"]
CMD ["public"]
```

So to run the container:
public: `docker run <options> balrog:latest`
admin: `docker run <options> balrog:latest admin`

mostlygeek:
excellent! R+ :+1:

scripts/push-dockerimage.sh
```diff
@@ -0,0 +1,34 @@
...
+
+branch_tag="${branch}"
+if [ "$branch" == "master" ]; then
+    branch_tag = "latest"
+fi
+date_tag="${branch}-${date}"
+docker build -t mozilla/balrog:${branch_tag} .
+docker tag -t mozilla/balrog:${branch_tag} mozilla/balrog:${date_tag}
+docker login -e $dockerhub_email -u $dockerhub_username -p $dockerhub_password
+docker push mozilla/balrog:${image_tag}
```
nthomas-mozilla: One more ${image_tag} to fix here.

requirements.txt
```diff
@@ -0,0 +1,44 @@
...
+sqlalchemy==0.7.1 \
+#    --hash=sha256:5ef95d19c31a8cd3905c697be0a7e94e70ab1926ecd4159c3e6c1cf01fc3c492
+sqlalchemy-migrate==0.7.2 \
+#    --hash=sha256:7f55c768ee26f143fedb11b365a4c57c3bbdb5211e049c7f04b8ae8107c28333
+tempita==0.5.1 \
+#    --hash=sha256:0ebe6938ca7401db79bac279849fffcb5752029150bcb6f3c3edbe7aa9a077d8
+uwsgi==2.0.11.2 \
+#    --hash=sha256:0b889b0b4d2dd3f6625df28cb0b86ec44a68d074ede2d0dfad0b91e88914885c
+Werkzeug==0.11.2 \
+#    --hash=sha256:eb4f98994b40a8e2edce289e811c9a38880ae53eb9ff6f031d0e243a69b0fcbf --hash=sha256:b6a93e909b31e3e14659584ff7329864275905ddc1b8653861bdfe1e597250f4
```
nthomas-mozilla: Double hash here ??

nthomas-mozilla:
This all looks good to me, from a code and the dev workflow point of view. Just one thing though ... I pulled f4ddb51 down to my mac and hit an error:

```
$ docker-compose up
Creating balrog_balrogdb_1
Creating balrog_balrogadmin_1
Creating balrog_balrogpub_1
ERROR: Container command not found or does not exist.
```
Have you tried this on your mac ? 'docker-compose build' fails to install requirements for balrogadmin:
```RequirementParseError: Expected ',' or end-of-list in blinker==1.4 cef==0.5 at  cef==0.5```
but I think you mention above.

requirements.txt
bhearsum: Yeah, that's intended. Werkzeug has a wheel and a tarball available for this version (https://pypi.python.org/pypi/Werkzeug/0.11.2), so I wanted to make sure both were viable installation candidates.

scripts/push-dockerimage.sh
bhearsum: Thanks for catching that. I think I might need a second push as well; I've been waiting for https://bugzilla.mozilla.org/show_bug.cgi?id=1255439 before testing this script again.

bhearsum:
Hm, that might be a Python 2.6 only error. I'm willing to bet I didn't test this on a Docker/Docker Compose that was run with Python 2.6. I'll look into this more.

bhearsum:
Hm, I can't reproduce that issue. I suspect I've rebuilt my images and containers completely from scratch somewhere along the way though, so maybe that makes a difference? I'm curious to know more about your configuration - particularly the Python version and whether or not "--no-cache" makes a difference.

bhearsum:
This might be a pip version issue...it looks like they only recently added support for special handling of comment lines that follow continuations: https://github.com/pypa/pip/commit/b58d2c9f34d869db96fc1f47e43039ef576f300f. Maybe try removing the comment lines completely...if that works I can just remove the hashes completely for now.

nthomas-mozilla:
I sorted out my issues running docker-compose by forcing docker to pull a new image for python:2.7, so r+ from me.

bhearsum:
https://bugzilla.mozilla.org/show_bug.cgi?id=1255439 landed in production today, I'm going to merge this.

PR #56, opened 2016-03-03
Bug 1127875 - add ability to mark releases as read only

nurav:
Defined a new endpoint and permission `/releases/:name/read_only`, and added column `read_only` to the releases table.

Releases can be marked as read only by anyone with the `/releases/:name/read_only` permission, but can be set to read write again only by an admin account.

### DIFF
--- a/auslib/admin/views/forms.py
+++ b/auslib/admin/views/forms.py
@@ -1,7 +1,7 @@
 import simplejson as json
 
 from flask_wtf import Form
-from wtforms import StringField, IntegerField, SelectField
+from wtforms import StringField, IntegerField, SelectField, BooleanField
 from wtforms.widgets import TextInput, FileInput, HiddenInput
 from wtforms.validators import Required, Optional, NumberRange, Length, Regexp
 
@@ -147,4 +147,11 @@ class CompleteReleaseForm(Form):
     name = StringField('Name', validators=[Required()])
     product = StringField('Product', validators=[Required()])
     blob = JSONStringField('Data', validators=[Required()], widget=FileInput())
+    data_version = IntegerField('data_version', widget=HiddenInput())
+
+
+class ReadOnlyForm(Form):
+    name = StringField('Name', validators=[Required()])
+    product = StringField('Product', validators=[Required()])
+    read_only = BooleanField('Read Only', validators=[Required()])
     data_version = IntegerField('data_version', widget=HiddenInput())
--- a/auslib/admin/views/forms.py
+++ b/auslib/admin/views/forms.py
@@ -153,5 +153,5 @@ class CompleteReleaseForm(Form):
 class ReadOnlyForm(Form):
     name = StringField('Name', validators=[Required()])
     product = StringField('Product', validators=[Required()])
-    read_only = BooleanField('Read Only', validators=[Required()])
+    read_only = BooleanField('read_only')
     data_version = IntegerField('data_version', widget=HiddenInput())
--- a/auslib/admin/views/forms.py
+++ b/auslib/admin/views/forms.py
@@ -152,6 +152,5 @@ class CompleteReleaseForm(Form):
 
 class ReadOnlyForm(Form):
     name = StringField('Name', validators=[Required()])
-    product = StringField('Product', validators=[Required()])
     read_only = BooleanField('read_only')
     data_version = IntegerField('data_version', widget=HiddenInput())
--- a/auslib/admin/views/releases.py
+++ b/auslib/admin/views/releases.py
@@ -334,16 +334,20 @@ class ReleaseReadOnlyView(AdminView):
     @requirepermission('/releases/:name/read_only')
     def _put(self, release, changed_by, transaction):
         form = ReadOnlyForm()
+
+        if not form.validate() and not request.json:
+            cef_event("Bad input", CEF_WARN, errors=form.errors)
+            return Response(status=400, response=json.dumps(form.errors))
         is_release_read_only = dbo.releases.isReadOnly(release)
 
-        if form.read_only:
+        if form.read_only.data:
             if not is_release_read_only:
-                dbo.releases.updateRelease(release, read_only=True, changed_by=changed_by, old_data_version=form.data_version, transaction=transaction)
+                dbo.releases.updateRelease(release, changed_by, form.data_version.data, read_only=True, transaction=transaction)
         else:
             if is_release_read_only:
                 # Only an admin user can unset the read_only field once it's set to True
-                if dbo.permissions.hasUrlPermission(changed_by, 'admin'):
-                    dbo.releases.updateRelease(release, read_only=False, changed_by=changed_by, old_data_version=form.data_version, transaction=transaction)
+                if dbo.permissions.hasUrlPermission(changed_by, 'admin', 'PUT'):
+                    dbo.releases.updateRelease(release, changed_by, form.data_version.data, read_only=False, transaction=transaction)
                 else:
                     msg = "%s is not allowed to set releases as read-write" % changed_by
                     cef_event("Unauthorized attempt to mark release as read-write", CEF_ALERT, msg=msg)
--- a/auslib/admin/views/releases.py
+++ b/auslib/admin/views/releases.py
@@ -303,6 +303,39 @@ class SingleReleaseView(AdminView):
         return Response(status=200)
 
 
+class ReleaseReadOnlyView(AdminView):
+    """/releases/:release/read_only"""
+
+    def get(self, release):
+        releases = dbo.releases.getReleases(name=release, limit=1)
+        is_release_read_only = dbo.releases.isReadOnly(name=release, limit=1)
+        if not releases:
+            return Response(status=404,
+                            response='Requested release does not exist')
+        release = releases[0]
+        return jsonify({
+            'read_only': is_release_read_only
+        })
+
+    @requirelogin
+    @requirepermission('/releases/:name/read_only')
+    def _put(self, release, changed_by, transaction):
+        req_read_only = request.form.get('read_only')
+        is_release_read_only = dbo.releases.isReadOnly(release)
+        old_data_version = request.form.get('data_version')
+
+        if req_read_only and not is_release_read_only:
+            dbo.releases.changeReadOnly(name=release, read_only=True, changed_by=changed_by, old_data_version=old_data_version, transaction=transaction)
+        elif not req_read_only and is_release_read_only:
+            # Only an admin user can unset the read_only field once it's set to True
+            if dbo.permissions.hasUrlPermission(changed_by, 'admin'):
+                dbo.releases.changeReadOnly(name=release, read_only=False, changed_by=changed_by, transaction=transaction)
+            else:
+                msg = "%s is not allowed to set releases as read-write" % changed_by
+                cef_event("Unauthorized attempt to mark release as read-write", CEF_ALERT, msg=msg)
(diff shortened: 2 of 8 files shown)


auslib/admin/views/releases.py
```diff
@@ -303,6 +303,39 @@ def _delete(self, release, changed_by, transaction):
         return Response(status=200)
 
 
+class ReleaseReadOnlyView(AdminView):
+    """/releases/:release/read_only"""
+
+    def get(self, release):
+        releases = dbo.releases.getReleases(name=release, limit=1)
+        is_release_read_only = dbo.releases.isReadOnly(name=release, limit=1)
```
bhearsum: I don't think you need to do both of these lookups. If you just do the isReadOnly one inside of a try/except block, you can figure out whether or not it exists by catching an exception.

auslib/admin/views/releases.py
```diff
@@ -303,6 +303,39 @@ def _delete(self, release, changed_by, transaction):
...
+                            response='Requested release does not exist')
+        release = releases[0]
+        return jsonify({
+            'read_only': is_release_read_only
+        })
+
+    @requirelogin
+    @requirepermission('/releases/:name/read_only')
+    def _put(self, release, changed_by, transaction):
+        req_read_only = request.form.get('read_only')
```
bhearsum: It would be good to use a Form for this, like we do for everything else (see auslib/admin/views/forms.py). The Form can handle validation and coercing the passed value to bool for you.

auslib/admin/views/releases.py
```diff
@@ -303,6 +303,39 @@ def _delete(self, release, changed_by, transaction):
...
+        is_release_read_only = dbo.releases.isReadOnly(release)
+        old_data_version = request.form.get('data_version')
+
+        if req_read_only and not is_release_read_only:
+            dbo.releases.changeReadOnly(name=release, read_only=True, changed_by=changed_by, old_data_version=old_data_version, transaction=transaction)
+        elif not req_read_only and is_release_read_only:
```
bhearsum: Both of these variables flipping state in these two conditionals is a bit hard too read. How about something like this instead:
if req_read_only:
  if not is_release_read_only:
    ...
else:
  if is_release_read_only:

auslib/admin/views/releases.py
```diff
@@ -303,6 +303,39 @@ def _delete(self, release, changed_by, transaction):
...
+            # Only an admin user can unset the read_only field once it's set to True
+            if dbo.permissions.hasUrlPermission(changed_by, 'admin'):
+                dbo.releases.changeReadOnly(name=release, read_only=False, changed_by=changed_by, transaction=transaction)
+            else:
+                msg = "%s is not allowed to set releases as read-write" % changed_by
+                cef_event("Unauthorized attempt to mark release as read-write", CEF_ALERT, msg=msg)
```
bhearsum: Do you need to return here? It looks like this will end up making no changes but still returning 201 as it stands right now.

auslib/admin/views/releases.py
```diff
@@ -303,6 +303,39 @@ def _delete(self, release, changed_by, transaction):
...
+        return Response(status=201, response='read_only changed')
+
+
 class ReleaseHistoryView(HistoryAdminView):
     """/releases/:release/revisions"""
```
bhearsum: You should return the state of the read only column in any other views that return an entire release (or releases) as well.

auslib/db.py
```diff
@@ -1087,11 +1094,27 @@ def localeExists(self, name, platform, locale, transaction=None):
...
+        return row['read_only']
+
+    def _proceedIfNotReadOnly(self, name, limit=None, transaction=None):
+        if self.isReadOnly(name, limit, transaction):
+            raise ReadOnlyError("Release '%s' is read only" % name)
+
+    def changeReadOnly(self, name, read_only, changed_by, old_data_version=None, transaction=None):
+        where = [self.name == name]
+        what = dict(read_only=read_only)
+        self.update(where=where, what=what, changed_by=changed_by, old_data_version=old_data_version, transaction=transaction)
```
bhearsum: Did you implement this as its own method because it would be hard to enforce the read-only condition if we supporting updating read_only in updateRelease?

auslib/migrate/versions/011_add_read_only_releases.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+def upgrade(migrate_engine):
+    metadata = MetaData(bind=migrate_engine)
+
+    releases = Table('releases', metadata, autoload=True)
+    releases_history = Table('releases_history', metadata, autoload=True)
+
+    read_only = Column('read_only', Boolean)
+    read_only.create(releases)
```
bhearsum: What is this going to get set to for existing releases?

auslib/test/admin/views/test_releases.py
```diff
@@ -783,3 +783,22 @@ def testSettings(self):
...
+    def testReadOnlySetTrue(self):
+        data = dict(read_only=True, product='Firefox', data_version=1)
+        self._put('/releases/b/read_only', data=data)
+        read_only = dbo.releases.isReadOnly(name='b')
+        self.assertEqual(read_only, True)
+
+    def testReadOnlySetFalseWithoutPermissionForProduct(self):
+        data = dict(read_only=False, product='Firefox', data_version=1)
+        ret = self._put('/releases/b/read_only', username='bob', data=data)
+        self.assertStatusCode(ret, 401)
```
bhearsum: These tests look fine, but we need a few more. At least:
* Verify that an admin can unset the read only flag
* Verify that a non-admin user can *set* the read only flag (testReadOnlySetTrue uses an admin user). This probably required tweaking the permissions in https://github.com/mozilla/balrog/blob/master/auslib/test/admin/views/base.py#L15.
* Verify that a non-admin user who can set the read-only flag can't unset it. I think this is broken at the moment because of the missing "Return" statement in the view code. This case should probably expect a 403.
* Verify that other views fail when trying to modify a read-only release. These should probably be in existing classes.

auslib/admin/views/forms.py
```diff
@@ -148,3 +148,10 @@ class CompleteReleaseForm(Form):
     product = StringField('Product', validators=[Required()])
     blob = JSONStringField('Data', validators=[Required()], widget=FileInput())
     data_version = IntegerField('data_version', widget=HiddenInput())
+
+
+class ReadOnlyForm(Form):
+    name = StringField('Name', validators=[Required()])
+    product = StringField('Product', validators=[Required()])
```
bhearsum: Is there a reason "product" is required here?

auslib/admin/views/releases.py
```diff
@@ -297,12 +305,56 @@ def _delete(self, release, changed_by, transaction):
             cef_event("Bad input", CEF_WARN, errors=form.errors)
             return Response(status=400, response=json.dumps(form.errors))
 
-        dbo.releases.deleteRelease(changed_by=changed_by, name=release['name'],
-                                   old_data_version=form.data_version.data, transaction=transaction)
+        try:
+            dbo.releases.deleteRelease(changed_by=changed_by, name=release['name'],
+                                       old_data_version=form.data_version.data, transaction=transaction)
+        except ReadOnlyError as e:
+                msg = "Couldn't update release: %s" % e
```
bhearsum: Typo fix: "delete" instead of "update" here.

auslib/admin/views/releases.py
```diff
@@ -297,12 +305,56 @@ def _delete(self, release, changed_by, transaction):
...
 
 
+class ReleaseReadOnlyView(AdminView):
+    """/releases/:release/read_only"""
+
+    def get(self, release):
+        try:
+            is_release_read_only = dbo.releases.isReadOnly(name=release, limit=1)
+        except:
+            return Response(status=404,
```
bhearsum: Careful with the exceptions you catch here - make sure you're only catching the one that indicates this is a 404. With a bar "except" you'll catch general problems (eg: if the database is down), and return a 404 instead of a 500 for them.

auslib/admin/views/releases.py
```diff
@@ -297,12 +305,56 @@ def _delete(self, release, changed_by, transaction):
...
+        return jsonify({
+            'read_only': is_release_read_only
+        })
+
+    @requirelogin
+    @requirepermission('/releases/:name/read_only')
+    def _put(self, release, changed_by, transaction):
+        form = ReadOnlyForm()
+
+        if not form.validate() and not request.json:
```
bhearsum: What's the "and not request.json" part for? I don't see it getting used again. I think WTForms handles finding data in the request; I don't think you need to touch request.json yourself.

auslib/admin/views/releases.py
```diff
@@ -297,12 +305,56 @@ def _delete(self, release, changed_by, transaction):
...
+        else:
+            if is_release_read_only:
+                # Only an admin user can unset the read_only field once it's set to True
+                if dbo.permissions.hasUrlPermission(changed_by, 'admin', 'PUT'):
+                    dbo.releases.updateRelease(release, changed_by, form.data_version.data, read_only=False, transaction=transaction)
+                else:
+                    msg = "%s is not allowed to set releases as read-write" % changed_by
+                    cef_event("Unauthorized attempt to mark release as read-write", CEF_ALERT, msg=msg)
+                    return Response(status=403, response=msg)
+        return Response(status=201, response='read_only changed')
```
bhearsum: This flows much better now, great!

bhearsum:
This is shaping up really nicely. There's a few small things to address still, but you're almost there, great job!

auslib/test/admin/views/test_releases.py
```diff
@@ -783,3 +800,68 @@ def testSettings(self):
...
+    def testReadOnlyGet(self):
+        ret = self._get('/releases/b/read_only')
+        is_read_only = dbo.releases.t.select(dbo.releases.name == 'b').execute().first()['read_only']
+        self.assertEqual(json.loads(ret.data)['read_only'], is_read_only)
+
+    def testReadOnlySetTrueAdmin(self):
+        data = dict(name='b', read_only=True, product='Firefox', data_version=1)
+        self._put('/releases/b/read_only', username='bill', data=data)
+        read_only = dbo.releases.isReadOnly(name='b')
+        self.assertEqual(read_only, True)
```
bhearsum: While testing with the UI and some manual requests to the API I realized that these tests are not quite valid....you're passing a Python dict rather than JSON data. Although the tests pass, they don't simulate real requests. You should dump the data to json before passing it, like we do in other tests, eg: https://github.com/mozilla/balrog/blob/master/auslib/test/admin/views/test_releases.py#L27

You should end up with some failures like this after doing so:
__________________________________________________________________________________________________________________ TestReadOnlyView.testReadOnlySetTrueAdmin __________________________________________________________________________________________________________________

self = <auslib.test.admin.views.test_releases.TestReadOnlyView testMethod=testReadOnlySetTrueAdmin>

    def testReadOnlySetTrueAdmin(self):
        data = json.dumps(dict(name='b', read_only=True, product='Firefox', data_version=1))
        self._put('/releases/b/read_only', username='bill', data=data)
        read_only = dbo.releases.isReadOnly(name='b')
>       self.assertEqual(read_only, True)
E       AssertionError: False != True

auslib/test/admin/views/test_releases.py:816: AssertionError


Sorry for not catching this earlier...shouldn't be too difficult to fix though.

bhearsum: Turns out this is PEBKAC! These tests are fine, my manual tests were using the wrong endpoints. Ignore this!

PR #58, opened 2016-03-17
[rule-modals] This api change enables channel autocomplete

njirap:
Closes-bug: #1257293

### DIFF
--- a/auslib/admin/views/rules.py
+++ b/auslib/admin/views/rules.py
@@ -18,18 +18,25 @@ class RulesAPIView(AdminView):
 
     def get(self, **kwargs):
         rules = dbo.rules.getOrderedRules()
-        count = 0
-        _rules = []
-        for rule in rules:
-            _rules.append(dict(
-                (key, value)
-                for key, value in rule.items()
-            ))
-            count += 1
-        ret = {
-            "count": count,
-            "rules": _rules,
-        }
+        if request.args.get('channels_only'):
+            channels = []
+            for rule in rules:
+                if rule['channel'] is not None:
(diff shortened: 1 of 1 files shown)


auslib/admin/views/rules.py
```diff
@@ -18,18 +18,25 @@ class RulesAPIView(AdminView):
...
-            _rules.append(dict(
-                (key, value)
-                for key, value in rule.items()
-            ))
-            count += 1
-        ret = {
-            "count": count,
-            "rules": _rules,
-        }
+        if request.args.get('channels_only'):
```
bhearsum: I can see why you took this approach since it's what we do for releases, but I don't think it will scale well here - this method could quickly clog up with "if request.args.get("$foo_only")" blocks. I think a better approach is to look for a single argument that accepts a list of columns to return. So a request like:
/rules?columns=product;channel

would return something like:
{
  "count": 2,
  "rules": [
    {"product": "Firefox", "channel": "release"},
    {"product": "Firefox", "channel": "beta"}
  ]
}

Please make sure to add some tests for the new functionality as well.

njirap: Ok got it.. So if I push directly to mozilla do I create a new branch or just direct to master..?

bhearsum: You only have read access to this repository, so continue pushing to your own to update this pull request. You may want to create a branch in your repo, though.

njirap: ok thanks!

njirap:
Moved this to a new branch https://github.com/mozilla/balrog/pull/61

PR #59, opened 2016-03-17
Implement Mozlog and remove CEF logging

bhearsum:
This branch implements mozlog for Balrog and fully replaces CEF logging with it. The JsonLogFormatter is taken from https://raw.githubusercontent.com/mozilla/testpilot/master/testpilot/base/logging.py. I'm told it's going to be made available independent at some point, so we'll probably switch to that when it is.

Logging configuration has changed a little bit here as well, with Mozlog formatting being the default, but overridable with an environment variable (which we use for the Docker images).

One thing we're losing here is granularity of errors (CEF_* all become a WARNING here) - should I be doing something different here, @jvehent? My assumption is that we're probably going to be running at log level WARNING in production, so I perhaps I should be switching some of these to ERROR or CRITICAL messages?

There's a couple of new TODOs here. One of them is about e-mail notification for rules/permission changes. I'll be doing that in bug 1251338. The other is a note about the requestid that BalrogLogger generates. That's a new bug I found that I don't intend to fix here.

### DIFF
--- a/auslib/db.py
+++ b/auslib/db.py
@@ -17,7 +17,6 @@ import migrate.versioning.api
 from auslib.global_state import cache, dbo
 from auslib.AUS import isForbiddenUrl
 from auslib.blobs.base import createBlob
-from auslib.log import cef_event, CEF_ALERT
 from auslib.util.comparison import string_compare, version_compare
 
 import logging
@@ -1253,20 +1252,22 @@ class Permissions(AUSTable):
         return True
 
 
+# TODO: replace these with e-mails about rules/permissions changes?
+# https://bugzilla.mozilla.org/show_bug.cgi?id=1251338
 def getHumanModificationMonitors(systemAccounts):
     # Long lines from "what" get truncated to avoid printing out massive
     # release blobs ty the logs.
     def onInsert(table, who, what):
         if who not in systemAccounts:
-            cef_event('Human modification', CEF_ALERT, user=who, what=what, table=table.name, type='insert')
+            pass
 
     def onDelete(table, who, where):
         if who not in systemAccounts:
-            cef_event('Human modification', CEF_ALERT, user=who, where=where, table=table.name, type='delete')
+            pass
 
     def onUpdate(table, who, where, what):
         if who not in systemAccounts:
-            cef_event('Human modification', CEF_ALERT, user=who, what=what, where=where, table=table.name, type='update')
+            pass
     return onInsert, onDelete, onUpdate
 
 
--- a/auslib/log.py
+++ b/auslib/log.py
@@ -1,16 +1,17 @@
-from logging import Logger
+import logging
+
+import json
+import socket
+import sys
+import traceback
 
 from flask import request
 
-import cef
 
 log_format = "%(asctime)s - %(levelname)s - PID: %(process)s - Request: %(requestid)s - %(name)s.%(funcName)s#%(lineno)s: %(message)s"
 
-# Needs to be set by entry points.
-cef_config = {}
 
-
-class BalrogLogger(Logger):
+class BalrogLogger(logging.Logger):
 
     def makeRecord(self, name, level, fn, lno, msg, args, exc_info, func=None, extra=None):
         if extra is None:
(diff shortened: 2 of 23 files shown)


auslib/log.py
```diff
@@ -26,42 +26,118 @@ def makeRecord(self, name, level, fn, lno, msg, args, exc_info, func=None, extra
...
+
+    This log formatter outputs JSON format messages that are compatible with
+    Mozilla's standard heka-based log aggregation infrastructure.
+
+    See also:
+    https://mana.mozilla.org/wiki/display/CLOUDSERVICES/Logging+Standard
+    https://mana.mozilla.org/wiki/pages/viewpage.action?pageId=42895640
+
+    Adapted from:
+    https://github.com/mozilla-services/mozservices/blob/master/mozsvc/util.py#L106
```
nthomas-mozilla: Lets change master to a specific revision, or otherwise note what revision you used.

nthomas-mozilla:
Seems fine to me from a code inspection point of view. I'm curious what the actual logging looks like, and to hear the answer to the levels question. Fair to say we might have missed some runtime errors because test coverage is minimal ?

bhearsum:
Here's some output after commenting out LOG_FORMAT=plain from docker-compose.yml:
```
balrogadmin_1 | {"EnvVersion": "2.0", "Type": "SingleReleaseView", "Severity": 7, "Fields": {"message": "processing PUT request to /releases/B2G-dogfood-fota-20151130130150", "requestid": 139850289015120}, "Timestamp": 1458309773690236928, "Logger": "Balrog", "Hostname": "2abf3b0d38a9", "Pid": 11}
balrogadmin_1 | {"EnvVersion": "2.0", "Type": "Releases", "Severity": 7, "Fields": {"message": "Looking for releases with:", "requestid": 139850289015120}, "Timestamp": 1458309773706931968, "Logger": "Balrog", "Hostname": "2abf3b0d38a9", "Pid": 11}
balrogadmin_1 | {"EnvVersion": "2.0", "Type": "Releases", "Severity": 7, "Fields": {"message": "name: B2G-dogfood-fota-20151130130150", "requestid": 139850289015120}, "Timestamp": 1458309773707062016, "Logger": "Balrog", "Hostname": "2abf3b0d38a9", "Pid": 11}
balrogadmin_1 | {"EnvVersion": "2.0", "Type": "Releases", "Severity": 7, "Fields": {"message": "product: None", "requestid": 139850289015120}, "Timestamp": 1458309773707262976, "Logger": "Balrog", "Hostname": "2abf3b0d38a9", "Pid": 11}
balrogadmin_1 | {"EnvVersion": "2.0", "Type": "AUSTransaction", "Severity": 7, "Fields": {"message": "Attempting to execute SELECT releases.`read_only` \nFROM releases \nWHERE releases.name = %s", "requestid": 139850289015120}, "Timestamp": 1458309773713052928, "Logger": "Balrog", "Hostname": "2abf3b0d38a9", "Pid": 11}
balrogadmin_1 | {"EnvVersion": "2.0", "Type": "auslib.blobs.apprelease.ReleaseBlobV4", "Severity": 7, "Fields": {"requestid": 139850289015120}, "Timestamp": 1458309773713998080, "Logger": "Balrog", "Hostname": "2abf3b0d38a9", "Pid": 11}
balrogadmin_1 | {"EnvVersion": "2.0", "Type": "SingleReleaseView", "Severity": 4, "Fields": {"message": "Bad input: Couldn't update release: name in database (B2G-dogfood-fota-20151130130150) does not match name in blob (B2G-mozilla-b2g37_v2_2-nightly-latest)", "requestid": 139850289015120}, "Timestamp": 1458309773754620160, "Logger": "Balrog", "Hostname": "2abf3b0d38a9", "Pid": 11}
balrogadmin_1 | {"EnvVersion": "2.0", "Type": "AUSTransaction", "Severity": 7, "Fields": {"message": "exc is:", "traceback": "Uncaught exception:\nNone\nNone\n", "requestid": 139850289015120, "error": "None"}, "Timestamp": 1458309773755002112, "Logger": "Balrog", "Hostname": "2abf3b0d38a9", "Pid": 11}
```

I'm not exactly sure what you're getting at re: missing runtime errors. Which test coverage are you referring to being minimal?

auslib/db.py
```diff
@@ -1275,20 +1274,22 @@ def hasUrlPermission(self, username, url, method, urlOptions={}, transaction=Non
...
 
     def onDelete(table, who, where):
         if who not in systemAccounts:
-            cef_event('Human modification', CEF_ALERT, user=who, where=where, table=table.name, type='delete')
+            pass
 
     def onUpdate(table, who, where, what):
         if who not in systemAccounts:
-            cef_event('Human modification', CEF_ALERT, user=who, what=what, where=where, table=table.name, type='update')
+            pass
```
jvehent: Why not log those as `info` in the meantime?

auslib/log.py
```diff
@@ -26,42 +26,118 @@ def makeRecord(self, name, level, fn, lno, msg, args, exc_info, func=None, extra
                 # Request object, whose id is actually what we want.
                 # Without this we end up with the id of the proxy object, which
                 # is static for the life of the application.
+                # TODO: this doesn't seem to be 100% unique.
+                # Sometimes requests that happen around the same time will
+                # end up with the same value here. Possibly only when the
+                # query strings are the same, or srcip?
                 requestid = id(request._get_current_object())
```
jvehent: add a random?
```python
>>> iv = os.urandom(8)
>>> print("".join("{:02x}".format(c) for c in iv))
a6ddfb5d76f21422
```

jvehent:
> One thing we're losing here is granularity of errors (CEF_* all become a WARNING here) - should I be doing something different here, @jvehent? My assumption is that we're probably going to be running at log level WARNING in production, so I perhaps I should be switching some of these to ERROR or CRITICAL messages?

@mostlygeek would know better, but my understanding is we log *everything* coming out of the application. If you set the log level to DEBUG, then we'll log at that level. We don't post-filter on log level in the logging pipeline. 
I generally recommend logging at INFO level, and since you have everything set to WARNING, we will be fine there. Log volume isn't a concern.

Changing some of those levels from WARNING to ERROR or CRITICAL is up to you. I think a violation that gets blocked is fine as a WARNING. An application that should have succeeded but failed is an ERROR, and something massively wrong that's causing the app to crash is a CRITICAL (like a DB connection failure, for example). Feel free to adapt those to your liking.

auslib/db.py
bhearsum: Could do, I've been treating email notification as a must have before we transition, so I didn't think it mattered. It's probably worthwhile logging them in additional to e-mail though, so I'll do this.

auslib/log.py
bhearsum: I might be missing something, but I'm not sure how that helps? If I generate a new random number for each call to make_record, each log entry (even from within the same request) will end up with a different id. I don't want to block on this in any case...I just didn't want to forget that this needs to get fixed at some point.

bhearsum:
I'm a bit hesistant to log at DEBUG by default - that enables some _extremely_ spamming stuff in auslib/db.py (hundreds of lines per request).

In any case, it sounds like as long as all these former cef events get logged in some way, we're good. I'll have another look over them to see if some should be switched to ERROR.

jvehent:
> I'm a bit hesistant to log at DEBUG by default - that enables some extremely spamming stuff in auslib/db.py (hundreds of lines per request).

Sorry, that was unclear, I'm absolutely **not** suggesting you log at DEBUG by default, that would be madness :sob: Just log at WARNING or INFO and we'll be fine.

r+ for me

mostlygeek:
For production we should/will have the log level set to INFO.

auslib/db.py
bhearsum: Actually, I just noticed that my email notification patch does this, so I'll leave this stubbed for now. That'll be up for review as soon as this one merges.

bhearsum:
This most recent commit changes to the default log level to match what it'll be in prod. While I was poking at this some more I also noticed that the COPY commands in the Dockerfiles are wrong - they copied the contents of directories such as "auslib" without including the dirname. This was never noticed in dev because of the bind mounts, it would've busted production though.

bhearsum:
Recent landings (especially my Dockerflow patch) have caused some really ugly history. I'm going to force-push a rebased version of this branch to avoid that here.

bhearsum:
I followed up with Nick on IRC about his "test coverage" comment, and I'll be adding a few tests around the JsonLogFormatter to address that. I'm going to merge this in the meantime because it contains Dockerfile fixes that are preventing @mostlygeek from using the current Docker image. I'll send a new PR with the tests when they're ready.

PR #60, opened 2016-03-17
Bug 1134190 - balrog throws ISE 500 instead of 400 when data_version is wrong

aksareen:
Added an exception handler at web layer to catch OutdatedDataError case.

### DIFF
--- a/auslib/admin/views/base.py
+++ b/auslib/admin/views/base.py
@@ -45,27 +45,38 @@ def requirepermission(url, options=['product']):
     return wrap
 
 
+def catchOutDataError(messages):
+    def wrap(f):
+        def decorated(*args, **kwargs):
+            try:
+                return f(*args, **kwargs)
+            except OutdatedDataError as e:
+                msg = "Couldn't preform the request %s. Outdated Data Version. old_data_version doesn't match current data_version: %s" % (messages, e)
+                cef_event("Bad input", CEF_WARN, errors=msg)
+                return Response(status=400, response=json.dumps({"data": e.args}))
+        return decorated
+    return wrap
+
+
 class AdminView(MethodView):
 
     def __init__(self, *args, **kwargs):
         self.log = logging.getLogger(self.__class__.__name__)
         MethodView.__init__(self, *args, **kwargs)
 
+    @catchOutDataError("POST")
     def post(self, *args, **kwargs):
         self.log.debug("processing POST request to %s" % request.path)
         with dbo.begin() as trans:
             return self._post(*args, transaction=trans, **kwargs)
 
+    @catchOutDataError("Update")
     def put(self, *args, **kwargs):
         self.log.debug("processing PUT request to %s" % request.path)
-        try:
-            with dbo.begin() as trans:
-                return self._put(*args, transaction=trans, **kwargs)
-        except OutdatedDataError as e:
-            msg = "Couldn't Update. Outdated Data Version: %s" % e
-            cef_event("Bad input", CEF_WARN, errors=msg)
-            return Response(status=400, response=json.dumps({"data": e.args}))
+        with dbo.begin() as trans:
+            return self._put(*args, transaction=trans, **kwargs)
 
+    @catchOutDataError("Delete")
     def delete(self, *args, **kwargs):
         self.log.debug("processing DELETE request to %s" % request.path)
         with dbo.begin() as trans:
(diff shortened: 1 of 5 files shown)


auslib/admin/views/releases.py
```diff
@@ -241,8 +241,12 @@ def _put(self, release, changed_by, transaction):
                 msg = "Couldn't update release: %s" % e
                 cef_event("Bad input", CEF_WARN, errors=msg)
                 return Response(status=400, response=json.dumps({"data": e.args}))
+            except OutdatedDataError as e:
+                msg = "Couldn't update release: %s" % e
+                cef_event("Bad input", CEF_WARN, errors=msg)
+                return Response(status=400, response=json.dumps({"data": e.args}))
             data_version += 1
-            return Response(json.dumps(dict(new_data_version=data_version)), status=200)
+            return Response(json.dumps(dict(new_data_version=data_version)), status=200)
```
bhearsum: You've got the general idea here, but I think we can deal with this more globally if you do this in auslib/admin/views/base.py rather than in each individual view. Can you give that a shot?

auslib/test/admin/views/test_releases.py
```diff
@@ -51,6 +51,41 @@ def testReleasePostUpdateExisting(self):
...
+                            }
+                        }
+                    }
+                }
+            }
+        }
+        """))
+
+        load = json.loads(ret.data)
+        print json.dumps(load)
```
bhearsum: This print can go away, we don't want it landed.

auslib/test/admin/views/test_releases.py
```diff
@@ -51,6 +51,41 @@ def testReleasePostUpdateExisting(self):
...
+        self.assertEqual(ret.data, json.dumps(dict(new_data_version=2)), "Data: %s" % ret.data)
+        data = json.dumps(dict(detailsUrl='blah', fakePartials=True, schema_version=1))
+        ret3 = self._put('/releases/d', data=dict(data=data, product='d', data_version=1))
+        self.assertStatusCode(ret3, 400)
+
```
bhearsum: Looks like this test passes even without your changes...can you look into that?

auslib/admin/views/releases.py
aksareen: yeah I can do that.so I'm making the code more DRY by adding the exception handlers in base class itself. anything else I can try other than this ?

bhearsum:
This is pretty close now! All the code you've got looks good, but we should handle these errors for POST and DELETE requests as well - they may also raise OutdatedDataError at times. Once that's done, I think this is ready to merge.

auslib/admin/views/base.py
```diff
@@ -45,27 +45,38 @@ def decorated(*args, **kwargs):
     return wrap
 
 
+def catchOutDataError(messages):
+    def wrap(f):
+        def decorated(*args, **kwargs):
+            try:
+                return f(*args, **kwargs)
+            except OutdatedDataError as e:
+                msg = "Couldn't preform the request %s. Outdated Data Version. old_data_version doesn't match current data_version: %s" % (messages, e)
```
bhearsum: Typo here, should be "perform" not "preform".

auslib/admin/views/base.py
```diff
@@ -45,27 +45,38 @@ def decorated(*args, **kwargs):
     return wrap
 
 
+def catchOutDataError(messages):
```
bhearsum: Better to name this "catchOutdatedDataError", to match the actual Exception name.

auslib/admin/views/base.py
```diff
@@ -45,27 +45,38 @@ def decorated(*args, **kwargs):
...
         self.log = logging.getLogger(self.__class__.__name__)
         MethodView.__init__(self, *args, **kwargs)
 
+    @catchOutDataError("POST")
     def post(self, *args, **kwargs):
         self.log.debug("processing POST request to %s" % request.path)
         with dbo.begin() as trans:
             return self._post(*args, transaction=trans, **kwargs)
 
+    @catchOutDataError("Update")
```
bhearsum: You're using an HTTP method name above ("POST"), and then switching to a non-HTTP name here and for delete(). I'm fine with either, but please be consistent.

aksareen: Fixing the typos

bhearsum:
All the code here looks fine, just need to fix a couple of style issues and we can merge it!

aksareen:
fixing the typos . next one will be final commit

bhearsum:
Looks great now, thank you very much!

PR #61, opened 2016-03-18
Autocomplete more fields

njirap:


bhearsum:
We talked on IRC a little bit about this, and the approach of having /rules return a vastly different response based on the "columns" arg being present doesn't make sense. My example from #58 is the right type of response to return if you're going to get the products/channels from the "rules" endpoint. Doing it this way means the UI needs to do a bit of postprocessing to find the unique products/channels.

If you want to avoid doing that in the UI, you could look at adding endpoints such as "/rules/products", which could return a response like:
{"products": ["Firefox", "Thunderbird"]}

Either way, please add tests.

njirap:
The corresponding ui PR -> https://github.com/mozilla/balrog-ui/pull/19/commits

njirap:
The endpoint of the form "/rules/*foo*" is already taken by 'id_or_alias' for single rule view. So I had the option of adding "/rules/*column*" but that would mean that particular endpoint is not consistent in the form of its response, it can return either a response with one rule item or a column with many values.. So in the spirit of endpoint response consistency I let "/rules/id_or_alias" be and created "/rules/column/*column*" which returns the aforementioned response like:

{"products": ["Firefox", "Thunderbird"]}

.. I hope this is proffesional.

bhearsum:
Thanks for catching the potential conflict. Using a collection to hold these is a good workaround! I've got a few comments in-line, and you'll need to merge the latest from the upstream master branch, as this is currently unmergable.

auslib/admin/base.py
```diff
@@ -51,6 +52,7 @@ def add_security_headers(response):
 app.add_url_rule("/rules", view_func=RulesAPIView.as_view("rules"))
 # Normal operations (get/update/delete) on rules can be done by id or alias...
 app.add_url_rule("/rules/<id_or_alias>", view_func=SingleRuleView.as_view("rule"))
+app.add_url_rule("/rules/column/<column>", view_func=SingleRuleColumnView.as_view("rule_columns"))
```
bhearsum: This should be "/rules/columns/<column>" - collections are plural, not singular.

auslib/admin/views/releases.py
```diff
@@ -507,3 +507,22 @@ def _post(self, changed_by, transaction):
         )
         response.headers['Content-Type'] = 'application/json'
         return response
+
+
+class SingleReleaseColumnView(AdminView):
+    """ /releases/column/:column"""
+
+    def get(self, column):
+        releases = dbo.releases.getReleases()
```
bhearsum: This is an extremely expensive query, and should be replaced with getReleaseInfo(). The main difference between those two methods is that the latter will return Release.data columns, which are extremely big, and not something you could filter on anyways.

auslib/admin/views/releases.py
```diff
@@ -507,3 +507,22 @@ def _post(self, changed_by, transaction):
...
+        column_values = []
+
+        for release in releases:
+            for key, value in release.items():
+                if key == column and value is not None:
+                    column_values.append(value)
+        column_values = list(set(column_values))
+        ret = {
+            "count": len(column_values),
+            column + "s": column_values,
```
bhearsum: No need to pluralize here. Adding an "s" won't work for some columns anyways (eg: Rules.alias becomes aliass).

auslib/test/admin/views/test_releases.py
```diff
@@ -865,3 +865,11 @@ def testReadOnlyNonAdminCanSetFlagButNotUnset(self):
...
         ret = self._put('/releases/b/read_only', username='bob', data=data_unset)
         self.assertStatusCode(ret, 403)
+
+
+class TestSingleColumn_JSON(ViewTest, JSONTestMixin):
+
+    def testGetReleasesSingleColumn(self):
+        ret = self._get("/releases/column/product")
+        got = json.loads(ret.data)
+        self.assertEquals(got["count"], 4)
```
bhearsum: Please check the entire return value here, not just the count. I'd also like to see an additional test that makes sure that filtering on a column that doesn't exist causes a 404.

ui
```diff
@@ -1 +1 @@
-Subproject commit ab5c0c9c9bd82d96b3adb86bf07bb0c31839e0a1
```
njirap: This change completely refused not to be added..I couldn't get rid of it..

bhearsum: Submodules are weird. Most likely what you need to do is "cd ui && git checkout master && git pull" to fix this. Let me know if you still have trouble...I'm sure we can find some way to sort it out.

auslib/test/admin/views/test_releases.py
```diff
@@ -870,3 +870,16 @@ def testSettings(self):
...
         self.assertEqual(json.loads(ret.data), json.loads(data['blob']))
+
+
+class TestSingleColumn_JSON(ViewTest, JSONTestMixin):
+
+    def testGetReleasesSingleColumn(self):
+        expected_product = ["a", "c", "b", "d"]
+        expected = dict(count=4, product=expected_product)
+        ret = self._get("/releases/columns/product")
+        self.assertEquals(json.loads(ret.data), expected)
```
bhearsum: It looks like this test is going to fail sometimes due to unpredictable ordering of the list. You've got "expected_product" sorted already, so you probably need to sort ret["product"] before comparing. You might be better off with two asserts, one for ret["count"], and one for ret["product"]. Other than that, this patch looks good!

ui
njirap: The submodule commit revision is now the latest master one..

PR #63, opened 2016-03-21
Implement core functionality for scheduled changes

mozbhearsum:
This is a pretty good chunk of the underlying database code that will be needed for https://bugzilla.mozilla.org/show_bug.cgi?id=1246675. It implements all of the under the hood stuff except for the permissions check done right before enacting a change. The guts of the ScheduledChangesTable are pretty hairy right now - I need to do an optimization pass on it. I'm mostly looking for a higher level review of the implementation, and to bounce a couple of things off of you, specifically:
* You'll notice that enactChange() is calling baseTable.update/insert. This is because there's no way for this class to know what the base table's generic update or insert method is. I still need to replace methods such as updateRule with an override of "update" to make sure that scheduled changes aren't bypassing any Rule-specific business logic that may be in there.
* In thinking about how to implement the permissions check at enacting time, I'm starting to think that we should move permissions down to the database level, rather than the API level. This would probably end up replacing the URLs with a different string that maps to logical operations like "update" or "delete". Then we'd decorate each Rules/Releases/Permissions method with @requirepermission rather than decorating the web layer.


nthomas-mozilla:
Some feedback on 21374e7:
- re enactChange(), it seems sensible to keep all the validation logic we have in updateRule/updateRelease/updatePermission. Perhaps we should extract the validation out, so we can run it at submission time and provide immediate feedback. Also at merge time.
- re permissions check when enacting a change - I guess you mean in db.py, rather than in the database itself! I think db.py is OK so long as we still bubble up errors in the response. I'm not following why we need API change though, something to do with knowing the HTTP method down in db.py ? 
- Would ScheduledChangeTable.__init__() (and other table handling) be simplified by using a sc_ prefix for all columns not from the other table, eg sc_data_version, sc_telemetry_product. Or does that just make trouble for the history table around id & data_version ? Alternatively rename all imported columns with a prefix, or serialize the imported row into json. These may all be different cans of worms, but it took me a while to grok all the back and forth.
- In mergeUpdate(), the loop over scheduled_changes looks like it will modify the what arg, accumulating all the changes as it goes.

bhearsum:
Regarding enactChange(), I'm not certain I'm following your comments (or more likely, I didn't explain it well). I don't intend to get rid of any of the validation logic currently being done in updateRule & friends, merely rename those methods to "update", "insert", etc. to make it a generic interface. Is that how you were reading it? I think it's a good idea to look at extracting the validation logic though, so I'll give that a shot.

The API change to the permissions is indeed because of encapsulation. When you start looking at doing permission checks from within the ScheduledChangeTable method, it becomes immediately very weird to be thinking about URLs.

I think it's a good idea to look at prefixing one or the other columns in the ScheduledChangeTable. Off the top of my head, I lean towards prefixing the baseTable ones, but I'll give it some more thought. As you can see, we already have to rename at least data_version, so I don't think it will cause trouble around the id or data_version. Doing this would even let us have scheduled changes for scheduled changes ;).

I'll fix up mergeUpdate - thanks for the catch!

As I mentioned on IRC, the additional commits around the web API were not intended...I've been meaning to develop that on a separate branch so the parts could be individually reviewable/landable. I'm going to risk a force push to remove them, and create a new PR when they're closer to ready.

bhearsum:
1a5337a deals with your 3rd and 4th points, as well as:
* Adding some update/delete stuff that I missed before
* Refactoring the existing methods to be much cleaner -- adding a prefix to base table columns helped a lot with that.

@nthomas-mozilla - could you clarify what you meant in your first point re: validation logic?

nthomas-mozilla:
re enactChange() - your original comment about modifying the names of updateRule() etc seems fine to me. If, for some reason, overriding AUSTable.update wasn't desirable you could always pick a standard entry point for the scheduled table to call to. I tend to get confused after about 2 subclasses or indirections either way 😄 

Seems that I went off on a tangent because updateRelease() has a bunch of stuff which should maybe be on the API layer (or maybe only the API layer if we're duplicating). That's where the comment about splitting validation come from, plus the assumption we'd want to do that in both places (not so sure about that duplication now). updateRule() is pretty sparse, and updatePermission() will naturally fallout of moving the permission decoration.

If the permissions checks move down to the db layer to make enacting changes easier, is it still easy/possible to make sure we reject changes when submitted to the API ? Some sort of dryrun mode, perhaps ?

bhearsum:
Moving permissions down to the database level shouldn't have any effect on the enforcement of permissions through the API. We'll have to bubble the errors back up (maybe in a custom exception) and catch them at the web layer (maybe in AdminView) to make sure we're still returning appropriate responses. I think the tests are pretty good in this area though, so it shouldn't be hard to verify...

bhearsum:
TODO list for myself (feel free to add to it)
* [ ] Move permissions to the db layer & rename
* [ ] Rename updateRule & co. to update/insert/delete

bhearsum:
I opened #80 for the permissions rework, since it's a pretty isolated change.

bhearsum:
I think the latest batch of changes addresses everything from above, though I'm still 100% clear about the updateRelease parts. I merged #80 into this, so https://github.com/bhearsum/balrog/compare/db-permissions...rule-changes may be a useful view when comparing. Aside from that, the new stuff here is:
* Adjust Rules/Releases/Permissions to override insert/update/delete, as talked about in earlier comments.
* Check permissions of user who scheduled a change, right before enacting it. This required me to add a "dryrun" mode to update/insert, because that's where those checks happen.
* Don't rely on "dbo" object to do permission checks. https://bugzilla.mozilla.org/show_bug.cgi?id=1261061 has some background on this.

Let me know if you want a walkthrough of anything or otherwise want to chat about this or #80. Both are pretty substantial patches.

auslib/admin/views/releases.py
```diff
@@ -233,9 +233,8 @@ def _put(self, release, changed_by, transaction):
         if dbo.releases.getReleases(name=release, limit=1):
             data_version = form.data_version.data
             try:
-                dbo.releases.updateRelease(name=release, blob=blob,
-                                           product=form.product.data, changed_by=changed_by,
-                                           old_data_version=data_version, transaction=transaction)
+                dbo.releases.update({"name": release}, {"data": blob, "product": form.product.data}, changed_by=changed_by,
```
nthomas-mozilla: Nit, In some of these insert/update/delete calls we have where=blah style arguments, which helps readability.

auslib/db.py
```diff
@@ -665,9 +718,243 @@ def rollbackChange(self, change_id, changed_by, transaction=None):
...
+                conditions[col] = what[col]
+
+        self._validateConditions(conditions)
+
+        # Use the appropriate base table methods in dry run mode to ensure the
+        # user has permission for the change they want to schedule
+        if columns.get("data_version"):
+            self.baseTable.update(base_table_where, columns, changed_by, columns["data_version"], transaction=transaction, dryrun=True)
+        else:
+            self.baseTable.insert(changed_by, transaction=transaction, dryrun=True, **columns)
```
nthomas-mozilla: We talked in person about factoring this type of permission check out, but now I'm wondering if the variability of the arguments makes that not very useful.

auslib/db.py
```diff
@@ -665,9 +718,243 @@ def rollbackChange(self, change_id, changed_by, transaction=None):
...
+        what = {}
+        for col in sc:
+            if col.startswith("base_"):
+                what[col[5:]] = sc[col]
+
+        # The scheduled change is marked as complete first to avoid it being
+        # updated unnecessarily when the base table's update method calls
+        # mergeUpdate. If the base table update fails, this will get reverted
+        # when the transaction is rolled back.
+        self.update([self.sc_id == sc_id], {"complete": True}, changed_by=sc["scheduled_by"], old_data_version=sc["data_version"], transaction=transaction)
```
nthomas-mozilla: In person we talked about moving the complete=True check into the select, so that the db does the work when the # of scheduled changes grows.

auslib/db.py
```diff
@@ -665,9 +718,243 @@ def rollbackChange(self, change_id, changed_by, transaction=None):
...
+        where = []
+        for col in self.base_primary_key:
+            where.append((getattr(self, "base_%s" % col) == old_row[col]))
+
+        scheduled_changes = self.select(where=where, transaction=transaction)
+        if not scheduled_changes:
+            self.log.debug("No scheduled changes found for update; nothing to do")
+            return
+        for sc in scheduled_changes:
+            if sc["complete"]:
```
nthomas-mozilla: (moving this to the right place) In person we talked about moving the complete=True check into the select, so that the db does the work when the # of scheduled changes grows.

auslib/test/test_db.py
```diff
@@ -637,6 +638,376 @@ def testHistoryGetChangeWithDataVersionWithNonPrimaryKeyColumn(self):
...
+                super(TestTable2, self).__init__(db, "sqlite", scheduled_changes=True, history=True, versioned=True)
+
+        table = TestTable2("fake", self.metadata)
+        self.metadata.create_all()
+        what = {"fooid": 2, "when": 4532}
+        self.assertRaisesRegexp(ValueError, "Missing primary key column", table.scheduled_changes.insert, changed_by="bob", **what)
+
+    def testInsertWithIncompatibleConditions(self):
+        what = {"foo": "blah", "when": "abc"}
+        self.assertRaisesRegexp(ValueError, "Cannot parse", self.sc_table.insert, changed_by="bob", **what)
```
nthomas-mozilla: I'm not sure what this one is testing - it appears to only have a badly formed and unused foo. The name makes me think of testValidateConditionsNotAllowedWhenAndOther().

auslib/test/test_db.py
```diff
@@ -637,6 +638,376 @@ def testHistoryGetChangeWithDataVersionWithNonPrimaryKeyColumn(self):
...
+        self.assertEquals(history_row.data_version, 2)
+        self.assertEquals(history_row.base_fooid, 1)
+        self.assertEquals(history_row.base_foo, "bb")
+        self.assertEquals(history_row.base_bar, "barbar")
+        self.assertEquals(history_row.base_data_version, 1)
+
+    def testUpdateWithBadConditions(self):
+        where = [self.sc_table.sc_id == 1]
+        what = {"telemetry_product": "boop", "telemetry_channel": "boop", "telemetry_uptake": 99}
+        self.assertRaisesRegexp(ValueError, "Invalid combination of conditions", self.sc_table.update, where, what, changed_by="bob", old_data_version=1)
```
nthomas-mozilla: This is kinda sneaky! To swap the type of condition you must pass when: None, yes ?

auslib/test/test_db.py
```diff
@@ -637,6 +638,376 @@ def testHistoryGetChangeWithDataVersionWithNonPrimaryKeyColumn(self):
...
+        self.assertEquals(row.bar, "barbar")
+        self.assertEquals(row.data_version, 2)
+        self.assertEquals(history_row.foo, "aa")
+        self.assertEquals(history_row.bar, "barbar")
+        self.assertEquals(history_row.changed_by, "bob")
+        self.assertEquals(history_row.data_version, 2)
+        self.assertEquals(sc_row.complete, True)
+
+    def testEnactChangeNoPermissions(self):
+        # TODO: May want to add something to permissions api/ui that warns if a user has a scheduled change when changing their permissions
```
nthomas-mozilla: And we'd get a notification at the time it would have been applied.

auslib/test/test_db.py
```diff
@@ -898,36 +1270,36 @@ def testUpdateRule(self):
...
         what['mapping'] = 'd'
-        self.paths.updateRule(changed_by='bill', id_or_alias=1, what=what, old_data_version=1)
+        self.paths.update(where={"rule_id": 1}, what=what, changed_by="bill", old_data_version=1)
 
         rules = self.paths.t.select().where(self.paths.rule_id == 1).execute().fetchall()
         copy_rule = dict(rules[0].items())
         rule = self._stripNullColumns([copy_rule])
 
-        expected = [dict(rule_id=1, priority=100, backgroundRate=100, version='3.5', buildTarget='d', mapping='d', update_type='z', data_version=1)]
+        expected = [dict(rule_id=1, priority=100, backgroundRate=100, version='3.5', buildTarget='d', mapping='d', update_type='z', data_version=2)]
```
nthomas-mozilla: omg, how did this pass before ?

auslib/test/test_db.py
nthomas-mozilla: Ah, it's a timestamp parsing test, so lets just tweak to testInsertWithMalformedConditions, or something.

nthomas-mozilla:
Modulo the few comments above, and a few things we noted down in person, I'm very happy with this.

auslib/admin/views/releases.py
mozbhearsum: Good point, I'll use this style everywhere.

auslib/test/test_db.py
mozbhearsum: Yeah. Sorry, bad test name there!

auslib/test/test_db.py
mozbhearsum: Yeah. In order to make sure it's not possible for the conditions to get into an invalid state, you need to remove "when" and add the telemetry ones. Is that awkward to you? An alternative here could be not to allow the type of condition to change (require the user to remove the old scheduled change and add a new one).

nthomas-mozilla: I don't mind if the UI &API take care of this, and it's written down somewhere.

auslib/test/test_db.py
mozbhearsum: I spent some time trying to find out, and I'm not sure :(. I may come back to this though...it's very concerned.

auslib/db.py
mozbhearsum: We could reduce some duplication here, but the only thing that would truly be factored out (that is to say: making a decision in one place instead of many) is the data version check to decide if its an update or insert. I could go either way on this one.

mozbhearsum:
Latest push addresses most of review comments, still a couple of things to do though:
* Look at defining nullable scheduled changes columns in austable/concrete classes.
* Improve some comments
* Look at doing dryruns even in AUSTable methods - maybe even doing them with transactions that are subsequently rolled back? I'm not sure this should block here, so I might leave it for a follow-up up...

mozbhearsum:
The latest push does all the comment fixing I wanted, and adds a test to make sure that we can't create a scheduled change with a row that would be rejected by the base table due to a null column. It turns out that we already retain nullability of base table columns (except for PKs and data_version, which is by design), which I think was the only reason we were looking at defining column nullability in a separate data structure on the base table? I could be wrong about that, though, so correct me if I'm wrong.

I didn't look at enhancing dryruns for the reasons noted above.

@nthomas-mozilla - as long as you're happy with the scope of this patch, I think it's ready for final review. If you want to insist on enhancing dryruns or something else that we talked about, I understand though!

mozbhearsum:
Like before, this is built on top of my db-permissions branch, so https://github.com/mozbhearsum/balrog/compare/db-permissions...rule-changes may be the best view of it.

auslib/db.py
```diff
@@ -668,9 +728,259 @@ def rollbackChange(self, change_id, changed_by, transaction=None):
...
+        else:
+            self.baseTable.insert(sc["scheduled_by"], transaction=transaction, **what)
+
+    def mergeUpdate(self, old_row, what, changed_by, transaction=None):
+        """Merges an update to the base table into any changes that may be
+        scheduled for the affected row. If the changes are unmergable
+        (meaning: the scheduled change and the new version of the row modify
+        the same columns), an UpdateMergeError is raised."""
+
+        # pyflakes this should be "is False", but that's not how SQLAlchemy
```
nthomas-mozilla: Nit, 'pyflakes *thinks* this should be'

auslib/test/test_db.py
```diff
@@ -638,6 +639,384 @@ def testHistoryGetChangeWithDataVersionWithNonPrimaryKeyColumn(self):
...
+        self.assertEquals(row.data_version, 1)
+        self.assertEquals(row.base_foo_name, "i'm a foo")
+        self.assertEquals(row.base_foo, "123")
+        self.assertEquals(row.base_bar, "456")
+        self.assertEquals(row.base_data_version, None)
+
+    def testInsertWithNullableColumn(self):
+        what = {"bar": "abc", "when": 34567}
+        # TODO: we should really be checking directly for IntegrityError, but AUSTransaction eats it.
+        self.assertRaisesRegexp(TransactionError, "IntegrityError", self.sc_table.insert, changed_by="bob", **what)
```
nthomas-mozilla: Nit, maybe testInsertWithNonNullableColumn.

nthomas-mozilla:
I looked at the diffs for the later revisions and liked what I saw. The big white bird has completely wiped my brain re nullability - it sounds like you've discovered it was already handled properly.

r+

mozbhearsum:
Code-wise, this is ready to go now, but it needs to land after #80 is safely in production. We also need to do a schema upgrade prior to it hitting admin prod.

mozbhearsum:
This is ready to go now that the bulk of #80 has landed. This patch now includes the removal of the last bits of the old permissions system as well.

PR #64, opened 2016-03-22
Add test that ensures JsonLogFormatter is used when appropriate.

bhearsum:
Since I've taken this code from elsewhere I think it's safe to assume that format() works fine, so I'm just testing configure_logging(). Incidentally, this discovered that JsonLogFormatter would've busted production because calling super(JsonLogFormatter, self) doesn't work in Python 2.6 (because logging.Formatter is still an old-style class).

### DIFF
--- a/auslib/log.py
+++ b/auslib/log.py
@@ -138,7 +138,7 @@ def safer_format_traceback(exc_typ, exc_val, exc_tb):
 def configure_logging(stream=sys.stdout, formatter=JsonLogFormatter, format_=log_format, level=logging.DEBUG):
     logging.setLoggerClass(BalrogLogger)
     handler = logging.StreamHandler(stream)
-    formatter = JsonLogFormatter(fmt=format_)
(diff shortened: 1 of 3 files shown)


auslib/log.py
```diff
@@ -138,7 +138,7 @@ def safer_format_traceback(exc_typ, exc_val, exc_tb):
 def configure_logging(stream=sys.stdout, formatter=JsonLogFormatter, format_=log_format, level=logging.DEBUG):
     logging.setLoggerClass(BalrogLogger)
     handler = logging.StreamHandler(stream)
-    formatter = formatter(fmt=format_)
+    formatter = JsonLogFormatter(fmt=format_)
```
nthomas-mozilla: This looks it will regress the LOG_FORMAT=plain setting docker uses. Otherwise fine.

bhearsum: Whoops, good catch. That was me fiddling around while writing tests, and not reverting.

PR #65, opened 2016-03-22
bug 1251338: Send e-mail when changes are made to rules or permissions.

bhearsum:
Not too much to say about this one. I've done local tests in Docker and it appears to work fine. Because it's implemented in-process and synchronously, it may slow down requests that change Rules or Permissions...but those are pretty few, and mostly done by humans. Maybe once we have the Balrog Agent (or something else?) we should queue this sort of thing up in the DB and send mail out of process?


bhearsum:
Callek pointed out on IRC that we'll eat errors and never send mail if anything goes wrong. They'll be logged at least, but they won't prevent the change from happening. I could go either way on this...

nthomas-mozilla:
I think I'd prefer to not block changes on email failing in the worst case, ie omg I have to change something right now. Ideally the mail should robust and it would be good to surface errors somehow. Either through the UI (useful for admins but not super robust for error reporting) or something watching the logs and alerting (harder to ignore, hopefully!). Would a local mail server avoid needing that ? It would hopefully avoid lag from being synchronous but may shift errors to some other place to check (ie not somewhere RelEng can get to). Maybe some retries around connectivity errors too.

I'm curious what the emails look like. For inserts it shows all the new data I guess, but for update just the changed data without the old content, and delete only the id or alias ? I suspect my first request after landing would be to know what values changed or were removed.

It may also be useful to allow comments/justifications for why changes are being made. So an extra bit of UI which is 'Reason for change'.

bhearsum:
Here's some examples of mail:
UPDATE:
Details of change:
balrogadmin
{'comment': u'Bug 1234277 - Create SHA1 watersheds', 'product': u'Firefox', 'buildID': u'<20151223004004', 'whitelist': None, 'backgroundRate': 100, 'mapping': u'Firefox-mozilla-aurora-nightly-20151223004004', 'priority': 150, 'alias': None, 'version': None, 'headerArchitecture': None, 'update_type': u'minor', 'buildTarget': None, 'locale': None, 'osVersion': u'Windows_NT', 'distribution': None, 'channel': u'aurora123', 'distVersion': None}
[<sqlalchemy.sql.expression._BinaryExpression object at 0x7f173e3e32d0>]

DELETE:
Details of change:
balrogadmin
[<sqlalchemy.sql.expression._BinaryExpression object at 0x7f173f9981d0>]

INSERT:
Details of change:
balrogadmin
{'comment': u'OS Blocking rule for de-supported OS on Release channel.', 'product': u'ueoauoeaueou', 'buildID': None, 'locale': None, 'whitelist': None, 'backgroundRate': 100, 'mapping': u'No-Update', 'priority': 15012, 'alias': None, 'version': None, 'headerArchitecture': None, 'update_type': u'minor', 'buildTarget': None, 'distribution': None, 'osVersion': u'Darwin 6,Darwin 7,Darwin 8,Darwin 9,Windows_95,Windows_98,Windows_NT 4,Windows_NT 5.0,Windows_NT 5.1.0,Windows_NT 5.1.1,GTK 2.0.,GTK 2.1.,GTK 2.2.,GTK 2.3.,GTK 2.4.,GTK 2.5.,GTK 2.6.,GTK 2.7.,GTK 2.8.,GTK 2.9.,GTK 2.10.,GTK 2.11.,GTK 2.12.,GTK 2.13.,GTK 2.14.,GTK 2.15.,GTK 2.16.,GTK 2.17.', 'channel': u'release*', 'distVersion': None}


So, clearly I need to fix the WHERE clauses to actually print something useful at the very least.

bhearsum:
As for blocking on error, I've been working on the assumption that we'd have a local SMTP server in deployments rather than a remote one. Is that going to be the case, @mostlygeek? Regardless, maybe there's a timeout or something that we can set to limit to make sure requests don't take excessively long.

In the medium term I think it's better if these aren't sent in process. One idea I had was to make Balrog Agent do it, once it exists. This would probably involved a new table to that it would poll for mail that needs sending, and we could probably build some UI around that for status and such.

I was also wondering whether it might be better to use Pulse as a notification system...but we'd have to implement that through Balrog Agent (or similar) as well to avoid doing it in process, so it may not have any benefit over doing it via e-mail.

mostlygeek:
I will setup SMTP@localhost that will proxy/send to SES. I would recommend keeping the app logic simple and let the SMTP server (postfix) take care of the retry/queuing/etc.

bhearsum:
I've updated this PR to provide much better information in the e-mails that get sent. Here's examples: https://gist.github.com/bhearsum/08808cfcdb4f9a6bcb84c3e5d2751a19. @nthomas-mozilla, are there other improvements you want made to this? Are you OK with the current implementation given that we'll be routing through a local MTA?

nthomas-mozilla:
Your implementation in Balrog seems fine for local MTA, but I'd like to know from @mostlygeek how errors bubble up to *someone* to handle.

Re the email formatting, INSERT looks fine, although sorting in same column order as the table/UI would be nice to have. 
DELETE is pretty opaque, could you show the data which is going to be removed ? Right now it's lots of text wrapped around the rule_id.
For UPDATE it would be really nice to have something said 'the current rule looks like this' and 'and these are the new values', or combine that like this
  rule_id:   XYZ,     unchanged
  alias:      'foo' --> 'bar'
  priority:   100,      unchanged
Is that easyish ? I'm completely optimizing for my own use case here, which is being nosy about Balrog changes via email, so the email should be complete and avoid me going to look up more details. What are OpSec are interested in having ? Who is the intended audience for these emails ?

mostlygeek:
@nthomas-mozilla so the local proxy will deliver to an SES endpoint. Bounce/delivery failure notifications go to SNS which can go into an SQS queue. Something just needs to process messages in the SQS queue.

bhearsum:
I think I can include the information you're looking for the in the mail - I'll give it a shot. I don't see any downside to including them. As far as OpSec goes, my understanding is that this is intended for people like us. Eg: if we get mail about a rule change at a surprising time, it would prompt us to investigate further. I'm thinking of them similar to the release automation notification mail. @jvehent - does that match what's in your head?

jvehent:
Yep, that's accurate. Ideally, if the notification can contain a diff, that's better.

bhearsum:
@nthomas-mozilla - I updated this with your suggestions for better e-mail bodies. This had the nice side effect of removing the ugly introspection code. Here's samples of the new mails: https://gist.github.com/bhearsum/78107204a9afedf10dee3c559e8e3134

nthomas-mozilla:
That's looking a lot nicer. For the update case could we show the before/after by using this code ?
+                for k in row:
+                    if query.parameters[k] != row[k]:
+                        row[k] = '%s ---> %s' % (row[k], query.parameters[k])
+                    else:
+                        row[k] = '%s  (%s)' % (row[k], Unchanged())

Or use difflib with a crazy big context setting ?

bhearsum:
Could do. I think there's a dictdiff module as well, I'll see what I can find.

bhearsum:
In the end, I think your formatting suggesting is the best bet. dictdiff output is meant more for software than humans (examples on http://dictdiffer.readthedocs.org/en/latest/), and converting everything to text just to use difflib seems like overkill. The latest here ends up with messages like this: https://gist.github.com/bhearsum/f7269e7522e8e478ea607a05f598095b

The only thing that's less than ideal here is that because of the repr() some strings are printed as unicode (like the ones in the where clauses, which come in via the request), while others are strings (namely the ones coming back from the db). I tried to work around this by doing quoting by hand (instead of relying on repr), but that ended up with either ints being quoted, or a lot of type checking.

nthomas-mozilla:
👍  Thanks for patience with all the changes.

Further down the track we could make sure we're making release builds non-modifiable in post_release, and notify if anyone attempts to change them, or makes them modifiable again.

bhearsum:
That's a fantastic idea. I'll get something on file so we don't forget!

bhearsum:
https://bugzilla.mozilla.org/show_bug.cgi?id=1266063 for mailing on ro -> rw.

PR #66, opened 2016-03-24
Bug 1113689 - Releases api returns rule_ids field

njirap:


auslib/db.py
```diff
@@ -946,7 +946,22 @@ def getReleaseInfo(self, product=None, limit=None,
             column = [self.name]
         else:
             column = [self.name, self.product, self.data_version, self.read_only]
+
         rows = self.select(where=where, columns=column, limit=limit, transaction=transaction)
+        j = join(dbo.releases.t, dbo.rules.t, dbo.releases.name == dbo.rules.mapping)
+        ref_list = select([dbo.releases.name, dbo.rules.rule_id]).select_from(j).execute().fetchall()
```
bhearsum: This query is a good start! As I mentioned on IRC, we need to include rules that point at whitelists as well. Ideally, we'd get them all from one query. If that's not possible, you could do a second query and merge the results.

auslib/db.py
```diff
@@ -946,7 +946,22 @@ def getReleaseInfo(self, product=None, limit=None,
...
+        for ref in ref_list:
+          rel_name, rule_id = ref
+          try:
+            ref_dict[rel_name].append(rule_id)
+          except KeyError:
+            ref_dict[rel_name] = [rule_id]
+
+        for row in rows:
+          row['rule_ids'] = ref_dict[row['name']]
+
```
bhearsum: Looks like there's at least one issue with this code that the test is catching. Have a look at the log on https://tools.taskcluster.net/task-graph-inspector/#CuxnWYRESRm-t8rmHEPpuw/YJHyOs7lQGGmoQ0v8F_u_g/0 for more details.

You'll definitely need to update the tests because you're changing the format of the return value here. Please also add tests to make sure that the new functionality works correctly.

auslib/db.py
```diff
@@ -946,7 +946,27 @@ def getReleaseInfo(self, product=None, limit=None,
...
+                try:
+                    ref_dict[rel_name].append(rule_id)
+                except KeyError:
+                    ref_dict[rel_name] = [rule_id]
+
+            for row in rows:
+                try:
+                    row['rule_ids'] = ref_dict[row['name']]
+                except KeyError:
+                    row['rule_ids'] = []
```
bhearsum: I think the logic of these two loops can be improved. If you nest them, you should be able to get rid of the "ref_dict" intermediary structure. I also think using ifs instead of try/except would make this read better. Eg:
if "rule_ids" not in row:
  row["rule_ids"] = ...
else:
  row["rule_ids"] = []

auslib/test/admin/views/test_releases.py
```diff
@@ -967,3 +967,49 @@ def testReadOnlyNonAdminCanSetFlagButNotUnset(self):
...
+        self.assertFalse(rule_id in not_mapped_rel['rule_ids'])
+
+        dbo.rules.t.insert().execute(id=rule_id, priority=100, version='3.5', buildTarget='d',
+                                     backgroundRate=100, mapping=rel_name, update_type='minor', data_version=1)
+
+        releases = self._get("/releases")
+        releases_data = json.loads(releases.data)
+        mapped_rel = next(rel for rel in releases_data['releases'] if rel['name'] == rel_name)
+        self.assertEqual(len(mapped_rel['rule_ids']), 1)
+        self.assertTrue(rule_id in mapped_rel['rule_ids'])
```
bhearsum: These tests look good, but we need database-level tests as well. If we only had these, it could be possible to accidentally break this new functionality of getReleaseInfo for other callers.

bhearsum:
Functionally, this looks great. Just a couple of small improvements to be made, as noted above.

auslib/test/admin/views/test_releases.py
njirap: Hi bhearsum.. in the db tests I added rule_ids field in the getreleaseinfo tests..do those count as tests for the feature..?

bhearsum: Not quite...you added enough to those tests to make the existing ones pass, but there's nothing in test_db.py that makes sure this functionality specifically works. I know you've added similar tests to test_releases.py, but it's important to explicitly test the database layer to make sure that it works independently of the existing Releases views.

auslib/db.py
```diff
@@ -952,19 +952,13 @@ def getReleaseInfo(self, product=None, limit=None,
...
-                try:
-                    ref_dict[rel_name].append(rule_id)
-                except KeyError:
-                    ref_dict[rel_name] = [rule_id]
 
             for row in rows:
-                try:
-                    row['rule_ids'] = ref_dict[row['name']]
-                except KeyError:
+                refs = [ref for ref in ref_list if ref[0] == row['name']]
```
bhearsum: I don't think "ref for ref" is safe to do...those are two different variables that you're using the same name for. At the very least, it's confusing. You could use something like "r for ref in ref_list" instead.

njirap: both 'ref's refers to each item in ref_list during each iteration..

njirap: If we have "r for ref in ref_list if ref[0] == row['name']" then r will be undefined..

bhearsum: Oh, duh! Sorry, I parsed this wrong. You're totally right!

njirap:
Hi  bhearsum I need help troubleshooting why the two tests are failing.. 

https://public-artifacts.taskcluster.net/ZAXFxQ9XRsywFIpDGalwWw/0/public/logs/live_backing.log

auslib/db.py
```diff
@@ -946,7 +946,21 @@ def getReleaseInfo(self, product=None, limit=None,
...
         rows = self.select(where=where, columns=column, limit=limit, transaction=transaction)
+
+        if not nameOnly:
+            j = join(dbo.releases.t, dbo.rules.t, ((dbo.releases.name == dbo.rules.mapping) | (dbo.releases.name == dbo.rules.whitelist)))
+            ref_list = select([dbo.releases.name, dbo.rules.rule_id]).select_from(j).execute().fetchall()
+
+            for row in rows:
+                refs = [ref for ref in ref_list if ref[0] == row['name']]
+                ref_list = [ref for ref in ref_list if ref[0] != row['name']]
+                if len(refs) is not 0:
```
bhearsum: You should probably use "> 0" here. "is" tests for identity, and I'm not sure Python guarantees that 0 always has the same id().

bhearsum:
Hmm, it's hard to tell why that's failing...there's nothing obviously wrong. I think you may need to add some print statements to the tests or logging statements to the code to get some more detailed information.

auslib/test/test_db.py
```diff
@@ -1003,7 +1003,13 @@ def testWhitelistIncluded(self):
...
         rule_id = 6
 
-        releases = self.releases.getReleaseInfo()
+        with mock.patch('auslib.db.dbo') as mock_dbo:
+            mock_dbo.releases.t = self.db.releases.t
+            mock_dbo.rules.t = self.db.rules.t
+            mock_dbo.releases.name = self.releases.name
+            mock_dbo.rules.mapping = self.rules.mapping
+            mock_dbo.rules.whitelist = self.rules.whitelist
+            releases = self.releases.getReleaseInfo()
```
bhearsum: I think you may need to replace a few more things here. If you don't want to fuss with this any further, you can just remove these tests though. As I said on IRC, the real bug here is that db.py is using the dbo at *all*, which is not your fault.

PR #68, opened 2016-04-03
Bug 1074362 - balrog serves partial when it shouldn't when fromRelease doesn't exist

nthomas-mozilla:
Turns out _getFromRelease() can return None in two situations, and we're not handling the 'release missing from DB' case. The test is a little dirty, we could add something to the TestReleaseBlobBase class if you prefer.


bhearsum:
Yeah, it would be good to test the Blob class directly as well. The change looks fine otherwise though. Thanks!

bhearsum:
(looks like I assigned this wrong - oops)

mozbhearsum:
@nthomas-mozilla - do you want to get this merged and deal with additional tests later? It would be nice to get this footgun removed.

nthomas-mozilla:
Yes, we should merge and get the fix out there.

PR #69, opened 2016-04-05
Added script to run tests in docker

nurav:
Running `./run-tests.sh` will run the tests using tox. When https://github.com/docker/compose/issues/1896 is resolved, we can probably eliminate `run-tests.sh` and `docker-compose-test.yml` to make things cleaner.

Also, this adds tox to be installed in `Dockerfile.dev`.


Dockerfile.dev
```diff
@@ -16,6 +16,7 @@ WORKDIR /app
 # ... really speeds up building new containers
 COPY requirements.txt /app/
 RUN pip install -r requirements.txt
+RUN pip install tox
```
bhearsum: We recently enabled hash verification for everything in requirements.txt. Is there a way to do that for tox, as well?

nurav: I could have a `requirements-test.txt` with the hash, but it would have just tox. Do we absolutely need to have hash verification for tox, seeing that it will never be installed in production?

bhearsum:
It looks like the Docker image that's built doesn't have Python 2.6 installed, so that test fails. I'm not going to consider that a blocker here considering we'll be switching to 2.7 in the next month or so, though.

I also want to figure out have to make Taskcluster run tests through the new compose file (or at least Dockerfile.dev). I think I got it almost working with https://tools.taskcluster.net/task-inspector/#e_ABxuUDSuSCmQki_rPSXQ/, except that it made me realize that Dockerfile.dev needs to be running tox instead of run.sh, and we probably need tox.ini (and maybe other files) copied into the balrogtest image. Can you update this patch for that? Let me know if any of this is unclear.

nurav:
What we could do is build an image from Dockerfile.dev, upload it to Docker
Hub and use that to run the tests (instead of using python-test-runner).

This way, while testing locally, we could use our regular docker-compose
flow that rebuilds the image from Dockerfile.dev, catching changes, but on
taskcluster, we can use the prebuilt image and save rebuilding the image
every time.

On Wed, Apr 13, 2016 at 2:35 AM, Ben Hearsum <notifications@github.com>
wrote:

> It looks like the Docker image that's built doesn't have Python 2.6
> installed, so that test fails. I'm not going to consider that a blocker
> here considering we'll be switching to 2.7 in the next month or so, though.
>
> I also want to figure out have to make Taskcluster run tests through the
> new compose file (or at least Dockerfile.dev). I think I got it almost
> working with
> https://tools.taskcluster.net/task-inspector/#e_ABxuUDSuSCmQki_rPSXQ/,
> except that it made me realize that Dockerfile.dev needs to be running tox
> instead of run.sh, and we probably need tox.ini (and maybe other files)
> copied into the balrogtest image. Can you update this patch for that? Let
> me know if any of this is unclear.
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/mozilla/balrog/pull/69#issuecomment-209101720>
>



-- 
Regards,
Varun Joshi

bhearsum:
So just to be clear, this means that local developers would run "run-tests.sh", and CI would use the pushed image and just run "tox" directly?

nurav:
That's exactly what I meant.

run-tests.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+docker-compose -f docker-compose-test.yml run balrogtest
```
bhearsum: Does this exit non-zero if tests fail?

nurav: Yes, it does.

bhearsum:
After talking with @rail about this I realized the reason that I was uncomfortable building a pushing a dev image, is because it means that CI will fail for PRs that need a new dev image (because we'd only be building & pushing a new image on merges to master, not PRs). Rail helped come up with what we think is the best thing that minimizes the differences between CI & local dev, and minimizes pain for devs. Specifically:
* Kill Dockerfile.dev
* Switch docker-compose-dev.yml to the main Dockerfile
* Install requirements-test.txt as part of tox.ini (instead of building them image itself)

Doing this means that the workflow for local devs and CI will be "run-tests.sh". If it's not available already, it will build the Docker image (and rebuild it when necessary), and then run tests inside of it.

How does this sound to you, @nurav?

bhearsum:
I think this is almost ready to land. I did some more fiddling with Taskcluster today and discovered that docker-compose doesn't work with its docker-in-docker feature yet, but that should be fixed once https://github.com/taskcluster/dind-service/pull/3 lands in production.

I also had remove the volume mount, and copy requirements-test.txt and run-tests.sh into the the image instead. Volume mounting is not allowed with docker-in-docker (for security reasons, I think). Don't worry about updating this PR for that - I'll land this by hand and apply my fix on top instead.

Once the dind-service is updated with the fix I'll verify that the tests work, and then get this landed.

nurav:
I don't think we require docker-in-docker, can't we just use run-tests.sh as the entrypoint for TaskCluster? We'll have to somehow provide for docker-compose on the node though, I don't know if that's possible. In the worst case, we could simply pull the docker-compose binary as a part of the initialization for TaskCluster and use that in our scripts.

bhearsum:
As far as I know, there's no Taskcluster worker that supports Docker that doesn't require a prebuilt image. Unless I'm missing something, this means that we need to use dind.

nurav:
On checking the TaskCluster payload format, I agree we need to use dind, since we can't specify commands to run before the docker image is run.

PR #70, opened 2016-04-14
Bug 1233621 - Add validation for rule fields which accept operators

aksareen:


auslib/admin/views/forms.py
```diff
@@ -78,6 +80,34 @@ def coercer(value):
...
+        if field.data is None:
+            return
+        try:
+            op, operand = get_op(field.data)
+            log.debug('Got (%s, %s) from get_op', op, operand)
+        except TypeError:
+            # get_op field returns None if no operator or no match, can't be unpacked
+            raise ValidationError('Invalid input for %s' % field.label)
+
+        if check_version:
```
bhearsum: Have you tested this within the UI? I mentioned this in https://bugzilla.mozilla.org/show_bug.cgi?id=1021022#c8 as well...I'm a bit concerned that input from the actual UI may not end up as Python None. If this works fine within the UI, this is OK.

auslib/admin/views/forms.py
```diff
@@ -78,6 +80,34 @@ def coercer(value):
...
 
 
+def operator(check_version=False):
+    log.debug('starting in operator, check_version is %s' % check_version)
+
+    def _validator(form, field):
+        log.debug('starting in _validator: field.data is %s' % field.data)
+        log.debug('starting in _validator, check_version is %s' % check_version)
+        # empty input is fine
+        if field.data is None:
```
bhearsum: Have you tested this within the UI? I mentioned this in https://bugzilla.mozilla.org/show_bug.cgi?id=1021022#c8 as well...I'm a bit concerned that input from the actual UI may not end up as Python None. If this works fine within the UI, this is OK.

auslib/admin/views/forms.py
bhearsum: I mentioned this in the bug as well - I'd like to see the version check moved out to its own validator. Operators and versions are different things, and it's cleaner to have them implemented separately IMO.

auslib/admin/views/forms.py
aksareen: I wrote a test for it here.https://github.com/mozilla/balrog/pull/70/files#diff-3fd9d618fb33367f3022cd28992c87e3R75 , Now I'll manually check out via the UI too

auslib/admin/views/forms.py
aksareen: hmm I see. I'll try writing a new validator

auslib/admin/views/forms.py
bhearsum: Ah, great!

auslib/test/admin/views/test_rules.py
```diff
@@ -46,6 +47,47 @@ def testMissingFields(self):
...
+
+    def testValidationEmptyInput(self):
+        ret = self._post('/rules', data=dict(backgroundRate=42, mapping='d', priority=50,
+                         product='Firefox', update_type='minor', version='', buildID=''))
+        self.assertEquals(ret.status_code, 200, "Status Code: %d, Data: %s" % (ret.status_code, ret.data))
+        r = dbo.rules.t.select().where(dbo.rules.rule_id == ret.data).execute().fetchall()
+        self.assertEquals(len(r), 1)
+        self.assertEquals(r[0]['buildID'], None)
+        self.assertEquals(r[0]['version'], None)
+
```
bhearsum: Great job on these tests, they're very extensive!

auslib/admin/views/forms.py
```diff
@@ -78,6 +80,45 @@ def coercer(value):
...
+        try:
+            op, operand = get_op(field.data)
+            version = MozillaVersion(operand)
+        except ValueError:
+            raise ValidationError("Couldn't parse version for %s" % field.label)
+        except:
+            raise ValidationError('Invalid input for %s' % field.label)
+        # MozillaVersion doesn't error on empty strings
+        if not hasattr(version, 'version'):
+            raise ValidationError("Couldn't parse the version for %s" % field.label)
```
bhearsum: The error messages for both of these could be improved, this is what they look like now:
http://people.mozilla.org/~bhearsum/sattap/64b6797c.png

They should probably say something more precise than "invalid input" when the operator parsing fails, and it looks like field.label isn't the best thing to use here.

Once these are fixed, I'll get this merged.

bhearsum:
Looking good now! Thanks Ashish!

PR #71, opened 2016-04-15
bug 1241760 - Revamp rules view ui

njirap:
Split into columns(channels) and rows(products)


bhearsum:
Can you explain your motivation behind adding this to the backend? As far as I can tell, it returns the same information as /rules, but organized differently.

njirap:
I may have mis-named the response field 'rules'.. "/rules" returns a list of rules.. this one returns a list of products but not the product itself but all the channels associated with each product and then all the mappings associated with each channel associated with each product..

bhearsum:
Maybe we have different visions for how the UI will work...I'm having trouble seeing the need for this endpoint, though. If you have it, you'll still need to make requests to obtain all the other information about rules as the UI is constructed.

Is there a reason you can't use the existing /rules endpoint, and then organize the results by product+channel in Javascript on the client side?

PR #73, opened 2016-04-20
bug 1266392: Add link to Taskcluster build in version.json; create artifact with sha256 of created image

bhearsum:
@mostlygeek asked for both of these the other day. Combined, these let CloudOps do sha256 verification of the images that are pulled from Dockerhub.

@rail - can you look at this? I think you're probably the best person since you know Balrog+Taskcluster fairly well.


bhearsum:
You can see my test run in https://tools.taskcluster.net/task-inspector/#CbmgGc2ARC-XQUm9YGxsOA/0, which uses this code but comments out all the Dockerhub stuff.

PR #74, opened 2016-04-22
Enable http keepalive for the admin interface

mozbhearsum:
I discovered this was needed at least for dev while doing some local testing. I suspect we'll need it for prod as well, otherwise uwsgi will close sockets very quickly. What do you think, @mostlygeek?

### DIFF
--- a/uwsgi/admin.ini
+++ b/uwsgi/admin.ini
@@ -6,6 +6,9 @@ mount = /=/app/uwsgi/admin.wsgi
 # Rewrite PATH_INFO & SCRIPT_NAME to match mountpoint (so the app sees paths like /foo instead of /api/foo)
 manage-script-name = 1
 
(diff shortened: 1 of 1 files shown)


mostlygeek:
What is the reason we need to keep HTTP sockets open? 
Using keep-alives won't affect the server much. I'm R+

bhearsum:
The reason I added this was because I was having issues while working on a new, asyncio-based, client. The server kept closing the connection right away, and the client blew up. It's possible that aiohttp might not be handling non-keepalive connections correctly...so maybe this isn't strictly necessary. I admit I don't fully understand the pros and cons of using keepalive.

mozbhearsum:
Looks like this isn't going to be necessary, I can't reproduce the problems I was having before.

PR #76, opened 2016-04-29
Bug 1268350, add ReleaseBlobV5 to support promptWaitTime

nthomas-mozilla:
Adds promptWaitTime in ReleaseBlobV5, which we can use for Gecko 19.0 and higher. Tweaks the comment blocks on earlier blobs to define compat on them too (BTW, licenseURL will go away in bug 1267891). The schema is a copy of apprelease-v4.yml with promptWaitTime added as a number. For tests I've trimmed out the forbidden domain tests we carry around, those only really need to be done once (possibly separated out elsewhere).


bhearsum:
Did you create a new blob versions for this because of the Gecko version dependency change? I'm not sure we've ever explicitly decided when we must create new blob versions other than when the schema changes in an incompatible way (such as appv -> applicationVersion, or fileUrls). This doesn't seem like a bad policy, though, and perhaps allows us to do some sanity checking on not returning responses to clients that they don't understand.

nthomas-mozilla:
Originally I'd added promptWaitTime to ReleaseBlobV2...V4, but decided that set a trap for ourselves - we could accidentally use promptWaitTime with too-old Gecko. Given the watersheds we have at much higher versions this is pretty unlikely. The client may just ignore it too, I dunno. Let me know if you have a strong preference to go that way.

It's an interesting problem in general, given the blobs may change for client or server reasons. As well as checking at request-time, we could check in the blob validation function.

bhearsum:
I think this is fine, I was mostly curious about the thinking.

PR #78, opened 2016-05-04
Initialize UI submodule before building Docker image.

bhearsum:
Without this, we don't include the UI in the built Docker image...which means the UI doesn't work!


rail:
:shipit:

PR #79, opened 2016-05-04
Fix detailsUrl property in desupport schema.

bhearsum:
I tried to add a new desupport blob today and discovered this silly error....


nthomas-mozilla:
👍

PR #81, opened 2016-05-18
Bug 1266063 - send e-mail when a release changes from read only to read write

nurav:


auslib/db.py
```diff
@@ -1304,10 +1304,35 @@ def __repr__(self):
         return self.__str__()
 
 
-def make_change_notifier(relayhost, port, username, password, to_addr, from_addr):
+def send_email(relayhost, port, username, password, to_addr, from_addr, table, subj,
```
bhearsum: Nice work refactoring this to reduce duplication.

auslib/db.py
```diff
@@ -1329,32 +1354,34 @@ def bleet(table, type_, changed_by, query):
...
-
     return bleet
 
 
+def make_change_notifier_for_read_only(relayhost, port, username, password, to_addr, from_addr):
+    def bleet(table, type_, changed_by, query):
+        body = ["Changed by: %s" % changed_by]
+        where = [c for c in query._whereclause.get_children()]
+        row = table.select(where=where)[0]
+        if row['read_only'] != query.parameters['read_only']:
```
bhearsum: As you noted in IRC yesterday, we only want to send mail when we go from read_only -> read_write, so this needs updating. It's not a great idea to be running queries here either - you don't have access to the "transaction" object, so you won't see uncommitted changes. I *think* only need to check where.parameters to make this decision though?

auslib/db.py
```diff
@@ -1329,32 +1354,34 @@ def bleet(table, type_, changed_by, query):
...
+            body.append("Row(s) to be updated as follows:")
+            data = {}
+            data['name'] = UnquotedStr(repr(row['name']))
+            data['product'] = UnquotedStr(repr(row['product']))
+            data['read_only'] = UnquotedStr("%s ---> %s" %
+                                            (repr(row['read_only']),
+                                             repr(query.parameters['read_only'])))
+            body.append(UTF8PrettyPrinter().pformat(data))
+
+        subj = "%s to %s detected" % (type_, table.t.name)
```
bhearsum: This subject should be different than the existing change notification mail so that we can quickly distinguish between the two.

bhearsum:
We talked on IRC a bit about this. Varun pointed out that the existing e-mail notifiers do a select(), so we're not going to be blocking on that. Given that, and the most recent change that adjusts this to only notify when releases are going back to rw, this looks good to me!

PR #82, opened 2016-05-19
Look for HTTP_REMOTE_USER if REMOTE_USER isn't present.

bhearsum:
We discovered yesterday that the CloudOps nginx+uwsgi deployment doesn't quite work because REMOTE_USER isn't set. According to @mostlygeek, we should check for this instead.

@rail - r?


rail:
:sheep: it

mostlygeek:
I'm gonna start using :sheep it.

PR #84, opened 2016-05-20
bug 1274374: add support for /update/6 URLs

bhearsum:
This should be relatively straightforward, as it's mostly just a new column sprinkled around all the areas that touch rules and update queries. I chose to make the matching logic exactly like osVersion, because I expect that we'll eventually have multiple values in it. (Right now, it looks like it could be unknown, error, MMX, SSE, SSE2, or SSE3 based on my read of https://bug1271761.bmoattachments.org/attachment.cgi?id=8754952.)

I had a bit of trouble upgrading my local db for this, so I updated the init script to do for already initialized DBs. This will be a no-op if your database is up to date already.


auslib/db.py
```diff
@@ -713,17 +714,17 @@ def _buildIDMatchesRule(self, ruleBuildID, queryBuildID):
...
+           comma. Once split we do simple substring matching against the query
+           string. Unlike versions and channels, we assume these columns
+           contain a substring of what will be in the query, thus we don't
+           need to support globbing."""
+        if ruleString is None:
             return True
-        for os in ruleOsVersion.split(','):
-            if os in queryOsVersion:
+        for os in ruleString.split(','):
+            if os in queryString:
```
nthomas-mozilla: Nit, the os here looks like the last leftover from the renaming, so calling it something else would fit better.

auslib/test/web/test_client.py
```diff
@@ -42,7 +42,8 @@ def setUp(self):
         dbo.setDomainWhitelist(('a.com', 'boring.com'))
         self.client = app.test_client()
         self.view = ClientRequestView()
-        dbo.rules.t.insert().execute(backgroundRate=100, mapping='b', update_type='minor', product='b', data_version=1)
+        dbo.rules.t.insert().execute(backgroundRate=100, mapping='b', update_type='minor', product='b',
+                                     systemCapabilities="SSE2", data_version=1)
```
nthomas-mozilla: This new rule affects testVersion4Get() in this file, because it gets the new SSE2 rule but then doesn't match and the test fails. You may need a product to keep the tests independent, in which case the whitelist will need changing but you'll need to merge now that the per-product whitelist landed.

nthomas-mozilla:
f+ as I'd like to see the tests passing. Is the sample-data.sql change a recent dump from prod ?

auslib/test/web/test_client.py
bhearsum: Hm. Tests were passing locally...but you're clearly right about this.

bhearsum:
The sample data change is my manually appending a new NULL to each rules and rules_history row to make it work with the new column.

nthomas-mozilla:
Ah, OK. Review should be easy once the test issue is fixed up.

bhearsum:
OK, I've fixed the tests and renamed that variable. I _think_ this is ready to go.

nthomas-mozilla:
👍

PR #85, opened 2016-05-24
Unquote necessary url parts during routing

bhearsum:
I did some digging and AFAICT there's no way to make uwsgi unquote the path before handing them off to the app. This is different than Apache which seems to do that by default.

I was considering doing manual conversation of @, /, and : in the uwsgi config, but I thought this would be a more flexible approach.

The path unquoting should go away when #80 lands and removes URL-style permissions.

### DIFF
--- a/auslib/admin/base.py
+++ b/auslib/admin/base.py
@@ -3,8 +3,6 @@ import urllib
 from flask import Flask, request
 from flask_compress import Compress
 
-from werkzeug.routing import BaseConverter, PathConverter
-
 import logging
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


bhearsum:
From IRC, this isn't the right place to handle this. According to https://github.com/benoitc/gunicorn/pull/1211#issuecomment-188258878, it seems like uwsgi should be able to do this. I couldn't find a way or any other confirmation that the wsgi sever is the right place to do unquoting, so I filed https://github.com/unbit/uwsgi/issues/1263. If that doesn't work out, @catlee suggested that middleware would be a better place for this.

bhearsum:
I'm not sure why tc-gh isn't firing, but this latest version replaces the routing gunk with middleware to unquote the path. Tests all pass locally, and usernames/permissions aren't mangled when I add them locally.

catlee:
Looks fine to me. Is there a risk of double decoding causing issues if this is run under apache or other servers that do the path decoding for us?

bhearsum:
I just verified this with a local apache - I can add permissions to users with @ in their name and they show up correctly. I tested both an "admin" permission and an URL-style one. I'll do some additional verification in dev, just in case my local config was too different.

PR #86, opened 2016-05-26
bug 1274374: Use full string matches for system capabilities.

bhearsum:
As I was verifying the original patch that adds system capabilities in dev, I realized that in my head, substring matching would work fine beacuse we'd get an SSE2 request, not find that in the SSE deprecation rule, and move on. Of course, it turns out the substring matches goes the other way, and the SSE from the rule is a substring of SSE2 from the query....so there's no way just to match "SSE" wit hthe current code.

This PR fixes that.


bhearsum:
Sorry about the test failure - that was due to https://github.com/mozilla/balrog/commit/46afe49b04aece1224c0309444aced59d9be71a8, and since fixed.

nthomas-mozilla:
That'll be what my spidey sense was trying to warn me about then 😄   👍

bhearsum:
Thanks Nick. Serves me right for ignoring your intuition!

PR #87, opened 2016-05-27
Bug 1163160 - Corrected history count in UI

njirap:


njirap:
Hi @bhearsum it is my opinion the new sample data is causing the failed test

njirap:
I believe that same test also fails at your request pull request @bhearsum

bhearsum:
Yeah, that failed test in my fault...I'll fix it in master. Thanks for noticing that.

bhearsum:
While this works, it's a very inefficient solution - it means we'll retrieve all versions of an object from a _single_ request to the ui. For releases in particular, this could mean downloaded many megabytes of data (which would be an awful experience in the UI).

I think the backend already does exactly what we need it to -- requests to http://localhost:8080/api/rules/90/revisions return the first "page" of results, and the total number of rules. Based on that information the UI should be able to make enough pages to hold all of the results (instead of just enough for the first page of results). It will also need to make additional requests to the backend to load additional results. Eg: I can get just the 11th-20th history items with a request like:
http://localhost:8080/api/rules/90/revisions?limit=10&page=2

Does that make sense? What do you think?

mozbhearsum:
Unfortunately, I had to back this out because it broke pagination on the non-history views (ie: http://localhost:8080/rules). I found two problems:
1) The pagination controls disappeared.
2) When I managed to get them re-added (by adjusting the ng-show), the items didn't change as I changed pages, probably due to the dropping of the startFrom filter.

If there's a quick fix for it, that would be great. In the longer term, it might be worth looking at separating the history and non-history templates/controllers, as they are diverging a lot now.

PR #91, opened 2016-06-03
Added support for querying changes with data_version and value via getChange

nurav:
This is required for getting historical versions of blobs from the data_version.

### DIFF
--- a/auslib/db.py
+++ b/auslib/db.py
@@ -530,20 +530,31 @@ class History(AUSTable):
         """ Returns the unique change that matches the give change_id or
             combination of data_version and values for the specified columns.
             column_values is a dict that contains the column names that are
-            versioned and their values."""
-        if change_id is None:
-            where = [self.data_version == data_version]
-            column_names = {col.name: col for col in self.table.columns if col.name in self.base_primary_key}
+            versioned and their values.
+            Ignores non primary key attributes specified in column_values."""
+        # if change_id is not None, we use it to get the change, ignoring
+        # data_version and column_values
+        by_change_id = False if change_id is None else True
+        # column_names lists all primary keys as string keys with the column
+        # objects as values
+        column_names = {col.name: col for col in self.table.columns if col.name in self.base_primary_key}
+
+        if not by_change_id:
+            # we check if the entire primary key is present in column_values,
+            # since there might be multiple rows that match an incomplete
+            # primary key
             for col in column_names.keys():
-                if col in column_values.keys():
-                    where.append(column_names[col] == column_values[col])
-                else:
+                if col not in column_values.keys():
                     raise ValueError("Entire primary key not present")
-            if self.baseTable.versioned:
-                changes = self.select(where=where,
-                                      transaction=transaction)
-            else:
+            # data_version can only be queried for versioned tables
+            if not self.baseTable.versioned:
                 raise ValueError("data_version queried for non-versioned table")
+
+            where = [self.data_version == data_version]
+            for col in column_names.keys():
+                where.append(column_names[col] == column_values[col])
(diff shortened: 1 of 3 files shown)


auslib/db.py
```diff
@@ -526,9 +526,26 @@ def forUpdate(self, rowData, changed_by):
         row['timestamp'] = self.getTimestamp()
         return self._insertStatement(**row)
 
-    def getChange(self, change_id, transaction=None):
-        """ Returns the unique change that matches the give change_id """
-        changes = self.select(where=[self.change_id == change_id], transaction=transaction)
+    def getChange(self, change_id, key=None, value=None, data_version=None, transaction=None):
```
bhearsum: change_id should be an optional parameter now; callers shouldn't be forced to pass it when they're quering by data_version.

auslib/db.py
```diff
@@ -526,9 +526,26 @@ def forUpdate(self, rowData, changed_by):
...
+        """ Returns the unique change that matches the give change_id or
+            combination of data_version and primary key."""
+
+        if change_id is None:
+            for col in self.table.columns:
+                if col.name == key:
+                    pk = col
+            if pk is None:
+                self.debug.log("Invalid primary key name")
+                return None
```
bhearsum: This won't handle multipart keys, such as those on the permissions table. I'm also not a fan of passing key and value separately...how about passing a dict that contains the primary key names+values?

Some additional verification would be good as well, such as ensuring the pk names you have match self.baseTable.primary_key.

auslib/db.py
```diff
@@ -526,9 +526,26 @@ def forUpdate(self, rowData, changed_by):
...
+            if self.baseTable.versioned:
+                changes = self.select(where=[self.data_version == data_version,
+                                             pk == value],
+                                      transaction=transaction)
+            else:
+                self.debug.log("data_version queried for non-versioned table")
```
bhearsum: It's good that you're checking for this! I think it should raise an Exception instead of returning None though. The caller has definitely made an error if they try to do this.

bhearsum:
The logic here looks sound, just needs a bit of work to support multipart keys and improve the interface.

auslib/test/test_db.py
```diff
@@ -566,6 +606,36 @@ def testMultiplePrimaryHistoryDeleteRollback(self):
...
+            self.assertEquals(ret, None)
+
+    def testHistoryGetChangeWithDataVersionWithNonPrimaryKeyColumn(self):
+        with mock.patch('time.time') as t:
+            t.return_value = 1.0
+
+            self.test.insert(changed_by='george', id1=4, id2=5, foo=0)
+            self.assertRaises(ValueError, self.test.history.getChange, data_version=1,
+                              column_values={'id1': 4, 'foo': 4})
+
```
bhearsum: Great work on the test coverage!

auslib/db.py
```diff
@@ -526,9 +526,26 @@ def forUpdate(self, rowData, changed_by):
...
+                    where.append(column_names[col] == column_values[col])
+                else:
+                    raise ValueError("Entire primary key not present")
+            if self.baseTable.versioned:
+                changes = self.select(where=where,
+                                      transaction=transaction)
+            else:
+                raise ValueError("data_version queried for non-versioned table")
+        else:
+            changes = self.select(where=[self.change_id == change_id], transaction=transaction)
```
bhearsum: Does this mean that you can only retrieve history by change_id for non-versioned tables? That's probably OK for now, but please add a comment to make note of it.

I would like to see the arg checking happen earlier though. Eg: before the "if change_id" - it's a bit tidier this way. Should be ready to merge with that fixed.

PR #92, opened 2016-06-09
Add upgrade-db action to run script

bhearsum:
This is primarily for CloudOps' benefit - @mostlygeek requested a wrapper that would make trivial for them to do db upgrades.

I tested it locally by going back to db version 12, tweaking docker-compose to run this command instead of admin-dev, and then "docker-compose up" -- it upgraded the db to version 13 as expected.

How does this look to you, @rail?


uwsgi/run.sh
```diff
@@ -6,6 +6,12 @@ elif [ $1 == "admin" ]; then
    exec uwsgi --ini /app/uwsgi/admin.ini --python-autoreload 1
 elif [ $1 == "admin-dev" ]; then
    exec uwsgi --ini /app/uwsgi/admin.dev.ini --ini /app/uwsgi/admin.ini --python-autoreload 1
+elif [ $1 == "upgrade-db" ]; then
+    if [ -z ${DBURI} ]; then
```
rail: I think the variable should be quoted, I trust our old scripts :) http://hg.mozilla.org/build/tools/file/tip/release/updates/verify.sh#l148

PR #93, opened 2016-06-20
bug 1223872: Single locale release updates should not race each other

nurav:
Uses dictdiffer to merge the two versions of the blob that contain different locales, handling conflicts by raising an OutdatedDataError.


tox.ini
```diff
@@ -20,6 +20,7 @@ deps =
     pytest
     pytest-capturelog
     pytest-cov
+    dictdiffer
```
bhearsum: Is this necessary? I think it will be installed along with the other things in requirements.txt.

auslib/admin/views/releases.py
```diff
@@ -248,7 +248,7 @@ def _put(self, release, changed_by, transaction):
                 msg = "Couldn't update release: %s" % e
                 self.log.warning("Bad input: %s", msg)
                 return Response(status=400, response=json.dumps({"data": e.args}))
-            data_version += 1
+            data_version = dbo.releases.getReleases(name=release, transaction=transaction)[0]['data_version']
```
bhearsum: Can you provide some background on why this is changing?

auslib/db.py
```diff
@@ -15,6 +15,9 @@
 import migrate.versioning.schema
 import migrate.versioning.api
 
+from dictdiffer import merge, patch
```
bhearsum: Nit: merge and patch are not terribly descriptive, so it would be part to just import dictdiffer and call them as dictdiffer.merge/patch.

auslib/db.py
```diff
@@ -1085,7 +1088,38 @@ def updateRelease(self, name, changed_by, old_data_version, product=None, read_o
...
             what['data'] = blob.getJSON()
-        self.update(where=[self.name == name], what=what, changed_by=changed_by, old_data_version=old_data_version, transaction=transaction)
+        try:
+            self.update(where=[self.name == name], what=what, changed_by=changed_by, old_data_version=old_data_version, transaction=transaction)
+        except OutdatedDataError as e:
+            if blob is not None and read_only is None:
+                ancestor_blob = createBlob(self.history.getChange(data_version=old_data_version,
+                                                                  column_values={'name': name},
+                                                                  transaction=transaction)
+                                                       .get('data'))
```
bhearsum: Style nit: Assigning the getChange result and then passing it to createBlob would probably be much cleaner. The inline call to .get() after a multiple line invocation is tough to read.

auslib/db.py
```diff
@@ -1085,7 +1088,38 @@ def updateRelease(self, name, changed_by, old_data_version, product=None, read_o
             if self.containsForbiddenDomain(blob, product):
                 raise ValueError("Release blob contains forbidden domain.")
             what['data'] = blob.getJSON()
-        self.update(where=[self.name == name], what=what, changed_by=changed_by, old_data_version=old_data_version, transaction=transaction)
+        try:
+            self.update(where=[self.name == name], what=what, changed_by=changed_by, old_data_version=old_data_version, transaction=transaction)
+        except OutdatedDataError as e:
```
bhearsum: Some debugging logging would be helpful here, particularly around all the possible exit points of this block.

auslib/db.py
```diff
@@ -1085,7 +1088,38 @@ def updateRelease(self, name, changed_by, old_data_version, product=None, read_o
...
+                # if we have no historical information about the ancestor blob
+                if ancestor_blob is None:
+                    raise
+                tip_release = self.getReleases(name=name,
+                                               transaction=transaction)[0]
+                tip_blob = tip_release.get('data')
+                m = merge.Merger(ancestor_blob, tip_blob, blob, {})
+                try:
+                    m.run()
+                    unified_blob = patch(m.unified_patches, ancestor_blob)
```
bhearsum: I'm a bit confused about why Merger and patch are both needed. I suspect it's a quirk of the dictdiffer interface, but since it would be good to have a comment explaining what each of these do.

auslib/db.py
```diff
@@ -1085,7 +1088,38 @@ def updateRelease(self, name, changed_by, old_data_version, product=None, read_o
...
+                    what['data'] = createBlob(unified_blob).getJSON()
```
bhearsum: Are you running the unified blob through createBlob() in order to validate it? It struck me as a bit strange to pass in JSON only to have it converted back afterwards.

auslib/db.py
```diff
@@ -1085,7 +1088,38 @@ def updateRelease(self, name, changed_by, old_data_version, product=None, read_o
...
+                    # more than that of the latest blob
+                    tip_data_version = tip_release['data_version']
+                    self.update(where=[self.name == name], what=what, changed_by=changed_by,
+                                old_data_version=tip_data_version, transaction=transaction)
+                    # cache will have a data_version of one plus the tip
+                    # data_version
+                    new_data_version = tip_data_version + 1
+                    cache.put("blob", name, {"data_version": new_data_version, "blob": blob})
+                    cache.put("blob_version", name, new_data_version)
+                    return
```
bhearsum: I think the new_data_version and cache updating should be unified with the existing logic for that in the function. It looks like the only difference is how new_data_version is calculated? Can you give that a try?

bhearsum:
This looks pretty good. As noted on IRC, it will be very good to have some tests directly against Releases.updateRelease, and there's some small nits above - but it largely looks good. I was hoping to test this against automation before we merged it, but I can't think of a way to do that. This means that after it merges to master, there's a chance that some new bugs may come up once automation is validating it.

requirements.txt
```diff
@@ -40,3 +40,5 @@ Werkzeug==0.11.2 \
 #    --hash=sha256:eb4f98994b40a8e2edce289e811c9a38880ae53eb9ff6f031d0e243a69b0fcbf --hash=sha256:b6a93e909b31e3e14659584ff7329864275905ddc1b8653861bdfe1e597250f4
 wtforms==2.1 \
 #    --hash=sha256:ffdf10bd1fa565b8233380cb77a304cd36fd55c73023e91d4b803c96bc11d46f
+dictdiffer==0.5.0.post1 \
+#    --hash=sha256:271f82d68f20dff25cb6e72d5f4153d3522767240cb2e6962e9ea3a3a8481f23
```
bhearsum: Looks like this is the hash of the .whl. Can you add the .tar.gz hash as well?

auslib/db.py
nurav: The unified blob is not JSON but a python dict; using createBlob both validates it and enables me to convert it to JSON.

mozbhearsum: Ah, OK. A comment about this would be worthwhile.

mozbhearsum:
I'm going to merge this now so we can see how it performs against dev over the weekend. We currently see plenty of 400s in logs such as https://public-artifacts.taskcluster.net/VXQ3KJiKQrau_ecf1n6NTQ/0/public/logs/live_backing.log (the funsize balrog submitter job). With this merged, we should stop getting those entirely.

mozbhearsum:
Small correction: need to look at nightly repack jobs (like http://archive.mozilla.org/pub/firefox/nightly/2016/06/2016-06-24-03-02-12-mozilla-central-l10n/mozilla-central-win64-l10n-nightly-6-unknown-bm73-build1-build13.txt.gz), not funsize jobs. The latter don't submit to stage.

PR #94, opened 2016-06-23
Update sample data to include SystemAddon rules/releases and some release history

mozbhearsum:
Two things here:
1) I promised the Morgoth folks a dump with SystemAddon rules/releases in them
2) @njirap noted that it would be helpful to have some releases history while working on #87.

This patch adds both, and compresses the sample data (it shot up to ~100MB uncompressed with the releases history). We've gone from ~7MB uncompressed to ~18MB compressed. I can probably trim down releases_history a bit if that's too big - what I've included is full history for a Beta, which is nearly 500 versions.

### DIFF
--- a/auslib/test/test_db.py
+++ b/auslib/test/test_db.py
@@ -1,3 +1,4 @@
+from bz2 import BZ2File
 import logging
 import mock
 import os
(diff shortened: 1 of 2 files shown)


mozbhearsum:
What do you think, @nthomas-mozilla?

njirap:
@mozbhearsum do you have the sample data view..? I don't even after deleting and redoing the docker containers

rels

njirap:
@mozbhearsum its good now.. I got rid of the cache.

nthomas-mozilla:
Sorry, I missed this one. 18MB seems OK to me, not too onerous. Less than the ui/node_modules/ anyway!

nthomas-mozilla:
BTW, do you have a script for extracting data from prod ? Might be a good candidate for adding to the repo.

mozbhearsum:
I don't have a script yet, but I'm planning to write something in the near future to extract the "live" data that we talked about recently. That is, the entire rules+permissions tables, as well as anything that rules point at. I'll probably end up including some history in that as well. I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1284481 for this.

In the meantime, I just run mysqldump with a table and some conditions specified, and selectively add that to sample-data.sql.

PR #95, opened 2016-06-28
Added blob for SystemAddons and fixed response for no updates

nurav:


auslib/blobs/schemas/systemaddons.yml
```diff
@@ -0,0 +1,71 @@
+---
+  title: GMP Blob
```
mozbhearsum: Please fix the title here.

auslib/blobs/schemas/systemaddons.yml
```diff
@@ -0,0 +1,71 @@
...
+  type: object
+  required:
+    - name
+    - schema_version
+    - hashFunction
+    - vendors
```
mozbhearsum: Now that this is just a SystemAddons blob, we can rename this to something better - perhaps "addons"? Please update the references in the descriptions below as well.

auslib/blobs/schemas/systemaddons.yml
```diff
@@ -0,0 +1,71 @@
...
+    name:
+      type: string
+      description: Human readable identifier
+      minLength: 1
+      maxLength: 100
+    schema_version:
+      type: number
+      description: Blob identifier number
+      enum:
+        - 1000
```
mozbhearsum: This should be 5000.

auslib/blobs/schemas/systemaddons.yml
```diff
@@ -0,0 +1,71 @@
...
+    hashFunction:
+      type: string
+      description: Hash function of files pointed at (returned in responses generated by this Blob)
+    uninstall:
+      type: boolean
+      description: If this is set true, all system addons are disabled
+      default: false
+    vendors:
```
mozbhearsum: I think it would be good to make sure "uninstall" and "vendors" can't be set at the same time. It's not obvious what should happen if both are present, so it's best to make sure that can't happen.

auslib/test/web/test_client.py
```diff
@@ -642,6 +690,30 @@ def testGetWithResponseProductsWithOneRule(self):
...
+        ret = self.client.get('/update/4/systemaddons/1.0/1/z/p/a/b/c/d/1/update.xml')
+        self.assertEqual(ret.status_code, 200)
+        self.assertEqual(ret.mimetype, 'text/xml')
+        returned = minidom.parseString(ret.data)
+        expected = minidom.parseString("""<?xml version="1.0"?>
+<updates>
+</updates>
+""")
+        self.assertEqual(returned.toxml(), expected.toxml())
+
```
mozbhearsum: These tests are a good start, but please also add some tests in auslib/test/blobs so we can test them directly.

auslib/web/views/client.py
```diff
@@ -75,22 +75,27 @@ def get(self, **url):
...
+                                             response_blob['response_update_type'],
+                                             app.config["WHITELISTED_DOMAINS"],
+                                             app.config["SPECIAL_FORCE_HOSTS"]))
+            if hasHeaderIfInnerXMLIsEmpty or len(innerXML) != 0:
+                xml.append(response_blobs[0]['response_release']
+                           .getHeaderXML(response_blobs[0]['product_query'],
+                                         response_blobs[0]['response_update_type']))
+                xml.extend(innerXML)
+                # Sampling the footer from the first blob
+                xml.append(response_blobs[0]['response_release'].getFooterXML())
```
mozbhearsum: I think this would be cleaner if you let getHeaderXML/getFooterXML return nothing if uninstall is True instead of adding a new method to the blobs. You can get rid of the need to check for whether or addons exist as well if you use the JSONSchema to ensure that "uninstall" and "vendors" can't be present in the same blob.

What do you think? Is there a reason for this new method that I haven't considered?

mozbhearsum:
This is a good start, but has a few things that need to be addressed, as noted above.

auslib/blobs/schemas/systemaddons.yml
```diff
@@ -0,0 +1,72 @@
...
+      enum:
+        - 5000
+    hashFunction:
+      type: string
+      description: Hash function of files pointed at (returned in responses generated by this Blob)
+    oneOf:
+        uninstall:
+          type: boolean
+          description: If this is set true, all system addons are disabled
+          default: false
```
mozbhearsum: Does setting a default here make it impossible to have this value absent? If so, it may mean it's impossible to include any addons due to the oneOf requirement.

auslib/blobs/schemas/systemaddons.yml
```diff
@@ -0,0 +1,71 @@
+---
+  title: SystemAddons Blob
+  type: object
+  required:
+    - name
+    - schema_version
+    - hashFunction
```
mozbhearsum: For this blob, hashFunction is not actually required, because of the uninstall option. If it's possible to make it required only when addons is present, let's do that. Otherwise just remove it from here.

auslib/blobs/systemaddons.py
```diff
@@ -0,0 +1,72 @@
...
+            addonXML.append('        <addon id="%s" URL="%s" hashFunction="%s" hashValue="%s" size="%s" version="%s"/>' %
+                            (addon, url, self["hashFunction"], platformData["hashValue"],
+                             platformData["filesize"], addonInfo["version"]))
+
+        return addonXML
+
+    def getHeaderXML(self, updateQuery, update_type):
+        if self.get("uninstall", False):
+            return '    <addons>'
+        return None
```
mozbhearsum: This is not quite right. <addons> is required if we're uninstalling _or_ if there are actually addons present. Eg: as talked about in http://gecko.readthedocs.io/en/latest/toolkit/mozapps/extensions/addon-manager/SystemAddons.html#update-response or shown in responses like https://aus5.mozilla.org/update/3/SystemAddons/45.0.1/20160315153207/Linux_x86_64-gcc3/en-US/release/Linux%204.2.0-30-generic%20(GTK%203.16.7%2Clibpulse%206.0.0)/default/default/update.xml.

auslib/blobs/schemas/systemaddons.yml
```diff
@@ -18,9 +18,6 @@
       description: Blob identifier number
       enum:
         - 5000
-    hashFunction:
-      type: string
-      description: Hash function of files pointed at (returned in responses generated by this Blob)
```
mozbhearsum: Sorry - I think I was unclear about this. I meant to remove hashFunction from the "required" section, not from the schema altogether. Removing it here will break any System Addon blobs that actually do have addons, because hashFunction will not be allowed in the blob.

auslib/blobs/systemaddons.py
```diff
@@ -0,0 +1,92 @@
...
+
+    def shouldServeUpdate(self, updateQuery):
+        # SystemAddon updates should always be returned. It is the responsibility
+        # of the client to decide whether or not any action needs to be taken,
+        # similar to GMP
+        return True
+
+    # If there are are no updates, we have a special response for SystemAddons
+    # blobs. We return <updates></updates>, without the addons tags.
+    def noUpdates(self, updateQuery, whitelistedDomains):
```
mozbhearsum: I think this would be simpler to read and implement if you reversed it and wrote a "hasUpdates" function instead. It would mean that the conditions in the header/footer methods below would not need a double negative ("if self.hasUpdates(...)" instead of "if not self.noUpdates").

auslib/blobs/systemaddons.py
```diff
@@ -0,0 +1,92 @@
...
+        buildTarget = updateQuery["buildTarget"]
+        n_updates = 0
+        for addon in sorted(self.getAddonsForPlatform(buildTarget)):
```
mozbhearsum: Because this is a boolean method there's no need to sort or count the number of addons - just return True if you get into the body of the loop, and return False as the last statement of the method.

auslib/blobs/systemaddons.py
```diff
@@ -0,0 +1,92 @@
...
+            # Checking if the addon update is to be served
+            platformData = self.getPlatformData(addon, buildTarget)
+            url = platformData["fileUrl"]
+            if isForbiddenUrl(url, updateQuery["product"], whitelistedDomains):
```
mozbhearsum: I don't think there's any need to do the forbidden URL check here. It gets done during the actual construction of the response.

auslib/blobs/systemaddons.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def shouldServeUpdate(self, updateQuery):
+        # SystemAddon updates should always be returned. It is the responsibility
+        # of the client to decide whether or not any action needs to be taken,
+        # similar to GMP
+        return True
+
+    # If there are are no updates, we have a special response for SystemAddons
+    # blobs. We return <updates></updates>, without the addons tags.
+    def hasUpdates(self, updateQuery, whitelistedDomains):
```
mozbhearsum: This function is much more readable now, very nice.

mozbhearsum:
Noted on IRC as well, but for posterity:
If i take the SystemAddons-ff46-rel1 blob from the sample data, s/vendors/addons/g, s/1000/5000/, it tells me that additional properties are not allowed. And if I try to submit an uninstall blob, it tells me that uninstall is not allowed, and that addons is required.
I'm not sure oneOf is valid where you've put it come to think of it....it may only be valid as the value of a property, eg: maybe it needs to be at line 10.

mozbhearsum:
Looks good to me now. Let's :shipit:!

PR #96, opened 2016-06-30
Integrate retention policy with existing db management script

mozbhearsum:
I'd like to do a little bit more testing with a larger data set, but in my limited tests it's working well. The queries are based on the most recent stored procedures from https://bugzilla.mozilla.org/show_bug.cgi?id=1237991#c41. Because we intend to do this cleanup much more frequently (at least once an hour, probably more), @mostlygeek suggested that we just do plain DELETEs rather than complicating things with cursors, which seemed fine to me.

I'm not 100% sure this script is the right place for the queries, but it seemed even more awkward to put them in db.py - I'm curious what you think about this, @nthomas-mozilla . At one point I had them in a .sql file, but that made it much tougher to do the SELECTs for the dryruns. Any other comments are welcome too, of course.


scripts/manage-db.py
```diff
@@ -13,22 +14,74 @@
...
 from auslib.db import AUSDatabase
 
+def cleanup_releases(trans, nightly_age, dryrun=True):
+    # This and the subsequent queries use "%%%%%" because we end up going
+    # through two levels of Python string formatting. The first is here,
+    # and the second happens at a low level of SQLAlchemy when the transaction
+    # is being executed.
+    query = """
+LEFT OUTER JOIN releases_history USING (name)
+LEFT JOIN rules ON (name=mapping)
```
nthomas-mozilla: I'll preface all these comments by saying my SQL is a smidgen rusty so I'm not sure I 100% understand how this works, but tally ho chaps.

Is there any difference between LEFT JOIN and LEFT OUTER JOIN ? The docs indicate OUTER is optional in MySQL. The queries are a bit intensive to play around with on (r/o) prod, and there's not enough history in docker, so I couldn't experiment.

Speaking of expensive, I think the query need only consider the most recent row in releases_history for each release (the old procedure deletes from releases and releases_history, but you're only acting on releases here). I wonder if there's a condition we can use to exclude processing any older releases_history rows, and leave cleanup_releases_history() to handle those.

scripts/manage-db.py
```diff
@@ -13,22 +14,74 @@
...
+WHERE name LIKE '%%%%nightly%%%%'
+AND (rules.whitelist <> releases.name OR rules.whitelist IS NULL)
```
nthomas-mozilla: It looks like we're trying to leave alone any release which is referenced as an active whitelist, but the the table join only considers mappings. Maybe <tt>LEFT JOIN rules ON (name=mapping OR name=whitelist)</tt>. Fairly edgecase, would need a whitelist with nightly in the name.

scripts/manage-db.py
```diff
@@ -13,22 +14,74 @@
...
+AND (rules.mapping <> releases.name OR rules.mapping IS NULL)
+AND (timestamp<1000*UNIX_TIMESTAMP(NOW()-INTERVAL %s MONTH) OR change_id is NULL);
```
nthomas-mozilla: Nit, readability would be nicer if the <tt><</tt> is space padded.

scripts/manage-db.py
```diff
@@ -13,22 +14,74 @@
...
 if __name__ == "__main__":
     from optparse import OptionParser
-    usage = """%s --db dburi action\n""" % sys.argv[0]
+    usage = """%s --db dburi action [options]\n""" % sys.argv[0]
     usage += "Possible actions:\n"
     usage += "  create: Create all the tables required for a new Balrog database\n"
     usage += "  upgrade: Upgrade an existing balrog table to a newer version.\n"
-    usage += "  downgrade: Downgrade an existing balrog table to an older version."
+    usage += "  downgrade: Downgrade an existing balrog table to an older version.\n"
+    usage += "  cleanup: Cleanup old data from a database. Requires an extra arg of maximum age (in months) of releases. Anything older than this will be deleted."
```
nthomas-mozilla: Nit, change 'of releases' to something that indicates it only affects releases matching '*nightly*'. Bonus points for mentioning the history cleanup.

ui
```diff
@@ -1 +1 @@
-Subproject commit dab0d84d2471a6988f91718f0e8881e5099ca1a1
+Subproject commit 7a68f4e0a8564723efc2470610ef3160f73e589e
```
nthomas-mozilla: Accidental inclusion ?

nthomas-mozilla:
My first reaction was that manage-db.py was the wrong place for this, but I'm not so worried after looking more closely. There's not much boiler plate if you prefer to put it in a separate script in the same directory.

scripts/manage-db.py
```diff
@@ -13,22 +14,74 @@
...
+    query = """
+LEFT OUTER JOIN releases_history USING (name)
+LEFT JOIN rules ON (name=mapping)
+WHERE name LIKE '%%%%nightly%%%%'
+AND (rules.whitelist <> releases.name OR rules.whitelist IS NULL)
+AND (rules.mapping <> releases.name OR rules.mapping IS NULL)
+AND (timestamp<1000*UNIX_TIMESTAMP(NOW()-INTERVAL %s MONTH) OR change_id is NULL);
+""" % nightly_age
+    if dryrun:
+        todelete = trans.execute("SELECT name FROM releases" + query).fetchall()
```
mostlygeek: Would be good to add the boolean values from the WHERE to make debug easier: 

* `IF(rules.whitelist <> releases.name OR rules.whitelist IS NULL,1,0) as rule1`
* `IF(timestamp<1000*UNIX_TIMESTAMP(NOW()-INTERVAL %s MONTH,1,0) as rule2`
* `IF(change_id is null, 1, 0) as rule3`

scripts/manage-db.py
```diff
@@ -13,22 +14,74 @@
 
 from auslib.db import AUSDatabase
 
+def cleanup_releases(trans, nightly_age, dryrun=True):
+    # This and the subsequent queries use "%%%%%" because we end up going
+    # through two levels of Python string formatting. The first is here,
+    # and the second happens at a low level of SQLAlchemy when the transaction
+    # is being executed.
+    query = """
+LEFT OUTER JOIN releases_history USING (name)
```
mostlygeek: What does this do? The releases_history table isn't used in the WHERE conditions

scripts/manage-db.py
mozbhearsum: re: LEFT JOIN vs. LEFT OUTER JOIN - I don't think there's any difference either, I just didn't tweak those parts of the queries from https://bugzilla.mozilla.org/show_bug.cgi?id=1237991#c41.

It's a good point about only needing the most recent row - I wonder if it would be faster to do subquery with LIMIT 1 instead of a JOIN?

scripts/manage-db.py
mozbhearsum: It's a touch hidden, but we actually do check releases_history.change_id in the last line of this query, which is used to make sure we don't delete any releases that are too recent.

scripts/manage-db.py
mozbhearsum: Good catch. Probably need to do a second join or a subquery for this.

scripts/manage-db.py
mozbhearsum: Hm, TIL that IF exists in sql :). I'm not quite sure how this improves debugging though - can you expand?

ui
mozbhearsum: Yeah, whoops.

mozbhearsum:
re: using a separate script in the same dir - I think I'd rather keep it in manage-db.py than that. I don't really see any benefit to moving it alongside.

scripts/manage-db.py
mostlygeek: The example code isn't the best example. I like seeing the fields used in the WHERE condition to see how the sql engine matched things. Since there is an `AND ( a OR b )` condition, it would help to see how that part matched.

mozbhearsum: I think I'm still missing something here...I can see how getting the 1/0 result of each AND block is helpful in debugging, but I'm not sure where that would show up in normal or even debug output. Sorry for being dense :).

mozbhearsum:
The latest push fixes the formatting and should fix the whitelist matching (I need to test it more though). I also spent some time trying to improve the efficiency of the releases table cleanup, but couldn't come up with anything that run faster on the prod db than what we have already. The best I got was:
```
SELECT name FROM releases
LEFT JOIN rules rules_mapping ON name=rules_mapping.mapping)
LEFT JOIN rules rules_whitelist ON (name=rules_whitelist.whitelist)
WHERE (rules_mapping.mapping <> releases.name OR rules_mapping.mapping IS NULL)
AND (ISNULL((SELECT name FROM releases_history WHERE name=releases.name LIMIT 1)) OR (SELECT timestamp FROM releases_history WHERE name=releases.name ORDER BY change_id DESC LIMIT 1) < (1000 * UNIX_TIMESTAMP(NOW() - INTERVAL 6 MONTH)) );
```

```
This gets the same results, but runs significantly slower. EXPLAIN shows that it ends up in the releases_history table twice, so perhaps that's no surprise:
mysql> EXPLAIN SELECT name FROM releases LEFT JOIN rules rules_mapping ON (name=rules_mapping.mapping) LEFT JOIN rules rules_whitelist ON (name=rules_whitelist.whitelist) WHERE name LIKE '%%%%nightly%%%%' AND (rules_whitelist.whitelist <> releases.name OR rules_whitelist.whitelist IS NULL) AND (rules_mapping.mapping <> releases.name OR rules_mapping.mapping IS NULL) AND (ISNULL((SELECT name FROM releases_history WHERE name=releases.name LIMIT 1)) OR (SELECT timestamp FROM releases_history WHERE name=releases.name ORDER BY change_id DESC LIMIT 1) < (1000 * UNIX_TIMESTAMP(NOW() - INTERVAL 6 MONTH)) );
+----+--------------------+------------------+-------+---------------+---------+---------+--------------------------------+--------+----------------------------------------------------+
| id | select_type        | table            | type  | possible_keys | key     | key_len | ref                            | rows   | Extra                                              |
+----+--------------------+------------------+-------+---------------+---------+---------+--------------------------------+--------+----------------------------------------------------+
|  1 | PRIMARY            | releases         | index | NULL          | PRIMARY | 302     | NULL                           | 100205 | Using where; Using index                           |
|  1 | PRIMARY            | rules_mapping    | ALL   | NULL          | NULL    | NULL    | NULL                           |    118 | Using where; Using join buffer (Block Nested Loop) |
|  1 | PRIMARY            | rules_whitelist  | ALL   | NULL          | NULL    | NULL    | NULL                           |    118 | Using where; Using join buffer (Block Nested Loop) |
|  3 | DEPENDENT SUBQUERY | releases_history | index | name          | PRIMARY | 4       | NULL                           |   2104 | Using where                                        |
|  2 | DEPENDENT SUBQUERY | releases_history | ref   | name          | name    | 302     | aus4_mozilla_org.releases.name |  52990 | Using index                                        |
+----+--------------------+------------------+-------+---------------+---------+---------+--------------------------------+--------+----------------------------------------------------+

```

vs.

```
mysql> EXPLAIN SELECT name FROM releases LEFT JOIN releases_history USING (name) LEFT JOIN rules rules_mapping ON (name=rules_mapping.mapping) LEFT JOIN rules rules_whitelist ON (name=rules_whitelist.whitelist) WHERE name LIKE '%%%%nightly%%%%' AND (rules_whitelist.whitelist <> releases.name OR rules_whitelist.whitelist IS NULL) AND (rules_mapping.mapping <> releases.name OR rules_mapping.mapping IS NULL) AND (timestamp < (1000 * UNIX_TIMESTAMP(NOW() - INTERVAL 6 month)) OR change_id is NULL);+----+-------------+------------------+-------+---------------+---------+---------+--------------------------------+--------+----------------------------------------------------+
| id | select_type | table            | type  | possible_keys | key     | key_len | ref                            | rows   | Extra                                              |
+----+-------------+------------------+-------+---------------+---------+---------+--------------------------------+--------+----------------------------------------------------+
|  1 | SIMPLE      | releases         | index | NULL          | PRIMARY | 302     | NULL                           | 100205 | Using where; Using index                           |
|  1 | SIMPLE      | releases_history | ref   | name          | name    | 302     | aus4_mozilla_org.releases.name |  52990 | Using where                                        |
|  1 | SIMPLE      | rules_mapping    | ALL   | NULL          | NULL    | NULL    | NULL                           |    118 | Using where; Using join buffer (Block Nested Loop) |
|  1 | SIMPLE      | rules_whitelist  | ALL   | NULL          | NULL    | NULL    | NULL                           |    118 | Using where; Using join buffer (Block Nested Loop) |
+----+-------------+------------------+-------+---------------+---------+---------+--------------------------------+--------+----------------------------------------------------+

```

I'll probably have another look at that with fresh brains tomorrow before giving up.

mozbhearsum:
I think I've addressed everything except the IF() comments that @mostlygeek made. I also tested this more thoroughly locally to make sure that we don't clean-up anything that a mapping or whitelist points at (we don't).

mozbhearsum:
@mostlygeek - can you clarify the comments about the IF() statements? Other than that, this should be ready to go.

mostlygeek:
My suggestions shouldn't be blocking. Feel free to proceed without them.

uwsgi/run.sh
```diff
@@ -12,6 +12,16 @@ elif [ $1 == "upgrade-db" ]; then
...
+elif [ $1 == "cleanup-db" ]; then
+    if [ -z "${DBURI}" ]; then
+        echo "\${DBURI} must be set!"
+        exit 1
+    fi
+    if [ -z "${MAX_AGE}" ]; then
+        echo "\${MAX_AGE} must be set!"
+        exit 1
+    fi
+    exec python scripts/manage-db.py -d ${DBURI} cleanup-dryrun ${MAX_AGE}
```
nthomas-mozilla: Leftover from testing ?

scripts/manage-db.py
```diff
@@ -13,22 +14,77 @@
...
+    # and the second happens at a low level of SQLAlchemy when the transaction
+    # is being executed.
+    query = """
+LEFT JOIN releases_history USING (name)
+LEFT JOIN rules rules_mapping ON (name=rules_mapping.mapping)
+LEFT JOIN rules rules_whitelist ON (name=rules_whitelist.whitelist)
+WHERE name LIKE '%%%%nightly%%%%'
+AND (rules_whitelist.whitelist <> releases.name OR rules_whitelist.whitelist IS NULL)
+AND (rules_mapping.mapping <> releases.name OR rules_mapping.mapping IS NULL)
+AND (timestamp < (1000 * UNIX_TIMESTAMP(NOW() - INTERVAL %s MONTH)) OR change_id is NULL);
```
nthomas-mozilla: Something about this still jangles the warning bell at the back of my head, so either I'm not getting it or something is amiss. Perhaps this ...

IIRC the `change_id IS NULL` condition would only be true if there is no history, say if `cleanup_releases_history()` removed history older than 7 days. So `%nightly%` are removed from releases if we have history going back X months, or after 7 days + the cron period ? Seems a bit incongruous, and the timestamp comparison is expensive if it never catches anything.

If we just want 7 day lifetime for (unmapped) nightlies and their history then we could remove the history first, and avoid the timestamp check here. There may be an edgecase around restored database without history, or maybe a new db with direct inserts (but don't do that!).

Or we may want to keep the release for longer than its history, which would be useful on the odd occasion we don't get informed of watersheds soon enough. The query could select releases which are unmapped & unwhitelisted, and the name contains a date older than some threshold. This could be fairly cheap by avoiding a join to the releases_history table.

Also, I don't grok the condition `rules_whitelist.whitelist <> releases.name`, that seems impossible given the join. Is it just to allow `rules_whitelist.whitelist IS NULL` as a check ? Can we write it as 
 `... AND (rules_mapping.mapping IS NULL AND rules_whitelist.whitelist IS NULL) ...` ?

The correct response to all this may be 'Nick should go learn some more SQL', or 'stop optimising this thing that won't take very long to run on RDS'; I'm just trying to figure out what the query is doing. If you have confidence in it from experimenting with live data and seeing the output then do squash my brainos and we can move along.

uwsgi/run.sh
mozbhearsum: Yep. Whoops.

scripts/manage-db.py
mozbhearsum: Re: lifetime of unmapped releases. I agree that it's a bit confusing. The original idea was that we would only keep history for nightly releases for 7 or 14 days, but we'd keep the actual release for 6 months. When it was written, nobody realized that we'd actually remove the releases after 7 or 14 days as well, because we were using releases_history to determine age. Reading the query after reading your comment makes me realize just how misleading it is -- a newcomer would rightly come to the conclusion that we keep nightly releases for 6 months, even though that's not the case. Either of your suggestions (just looking for releases without history, or parsing the name) sound reasonable to me, we just need to decide which direction to go.

Re: potential edge cases around restoration...I'm not particularly concerned about them, as any restoration from backup should include history. If we're doing direct db inserts....I think we deserve what we get. As this new job is expected to run every 15-30min, a case like that would get noticed quite quickly as well. Do push on this if you feel it's important though.

How strongly do you feel about keeping actual releases for longer than their history? You identified the not-notified-soon-enough watershed, and I can't think of any other case we'd want them. For the watersheds, we could regenerate them if needed with something like https://github.com/mozilla/balrog/blob/master/scripts/generate-json.py (though admittedly it's a PITA, and we'd have to write new code to do it against S3.

Re: the rules conditions - you're right - the <> part should be unnecessary because of the JOIN. I just tried some SELECTs against prod, and this query returns the same results as the one in this PR:
```SELECT name, rules_mapping.mapping, rules_whitelist.whitelist FROM releases LEFT JOIN releases_history USING (name) LEFT JOIN rules rules_mapping ON (name=rules_mapping.mapping) LEFT JOIN rules rules_whitelist ON (name=rules_whitelist.whitelist) WHERE name LIKE '%%%%nightly%%%%' AND rules_whitelist.whitelist IS NULL AND rules_mapping.mapping IS NULL AND (change_id IS NULL OR timestamp < (1000 * UNIX_TIMESTAMP(NOW() - INTERVAL 6 month)));```


To summarize, I need to:
* Get rid of the unnecessary <> checks.
* Actually save releases for the time specified, or change the query to not look as if it's going to without doing it.

**Comment by mozbhearsum** [comment]
Alright, I think the final issues are fixed up. The latest push does name parsing to find the age of dated releases, removes the now-necessary <> checks, and I switched MAX_AGE to take days because @mostlygeek said it would help when doing our dryruns. Removing the JOIN to releases_history speeds things up incredibly. It returned instantly when I ran it by hand on the old WebOps DB.

File: scripts/manage-db.py [hunk 10]
```diff
@@ -13,22 +14,77 @@
...
+def cleanup_releases(trans, nightly_age, dryrun=True):
+    # This and the subsequent queries use "%%%%%" because we end up going
+    # through two levels of Python string formatting. The first is here,
+    # and the second happens at a low level of SQLAlchemy when the transaction
+    # is being executed.
+    query = """
+LEFT JOIN rules rules_mapping ON (name=rules_mapping.mapping)
+LEFT JOIN rules rules_whitelist ON (name=rules_whitelist.whitelist)
+WHERE name LIKE '%%%%nightly%%%%'
+AND name NOT LIKE '%%%%latest'
```
nthomas-mozilla: Nit - `%%%%latest%%%%` is used elsewhere but kinda I like `%%%%latest` better. Doubt it makes much real difference but lets pick one to use everywhere.

nthomas-mozilla:
lgtm to me, lets ship this thing. 👍

PR #97, opened 2016-07-06
First pass at rule caching.

mozbhearsum:
This is a first crack at doing full-table caching of rules (the second option listed at https://bugzilla.mozilla.org/show_bug.cgi?id=1111032#c0). There's no invalidation necessary here because this code only runs on the public nodes, and they cannot modify rules.

This patch introduces an extra delay of up to 30 seconds between making a rule change, and having it take effect. This is on top of the existing 60 second cache at zeus (webops) or nginx (cloudops).

I haven't done a lot of self review or testing yet, I wanted to get this up for quick feedback anyways though.

### DIFF
--- a/auslib/db.py
+++ b/auslib/db.py
@@ -787,32 +787,42 @@ class Rules(AUSTable):
         """Returns all of the rules that match the given update query.
            For cases where a particular updateQuery channel has no
            fallback, fallbackChannel should match the channel from the query."""
-        where = [
-            ((self.product == updateQuery['product']) | (self.product == null())) &
-            ((self.buildTarget == updateQuery['buildTarget']) | (self.buildTarget == null())) &
-            ((self.headerArchitecture == updateQuery['headerArchitecture']) | (self.headerArchitecture == null()))
-        ]
-        # Query version 2 doesn't have distribution information, and to keep
-        # us maximally flexible, we won't match any rules that have
-        # distribution update set.
-        if updateQuery['queryVersion'] == 2:
-            where.extend([(self.distribution == null()) & (self.distVersion == null())])
-        # Only query versions 3 and 4 have distribution information, so we
-        # need to consider it.
-        if updateQuery['queryVersion'] in (3, 4):
-            where.extend([
-                ((self.distribution == updateQuery['distribution']) | (self.distribution == null())) &
-                ((self.distVersion == updateQuery['distVersion']) | (self.distVersion == null()))
-            ])
-        if not updateQuery['force']:
-            where.append(self.backgroundRate > 0)
-        rules = self.select(where=where, transaction=transaction)
-        self.log.debug("where: %s" % where)
+
+        def getRawMatches():
+            where = [
+                ((self.product == updateQuery['product']) | (self.product == null())) &
+                ((self.buildTarget == updateQuery['buildTarget']) | (self.buildTarget == null())) &
+                ((self.headerArchitecture == updateQuery['headerArchitecture']) | (self.headerArchitecture == null()))
+            ]
+            # Query version 2 doesn't have distribution information, and to keep
+            # us maximally flexible, we won't match any rules that have
+            # distribution update set.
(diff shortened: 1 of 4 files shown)


auslib/db.py
```diff
@@ -787,32 +787,42 @@ def getRulesMatchingQuery(self, updateQuery, fallbackChannel, transaction=None):
...
+            if not updateQuery['force']:
+                where.append(self.backgroundRate > 0)
+
+            self.log.debug("where: %s" % where)
+            return self.select(where=where, transaction=transaction)
+
+        cache_key = "%s-%s-%s-%s-%s-%s" % \
+            (updateQuery["product"], updateQuery["buildTarget"], updateQuery["headerArchitecture"],
+             updateQuery.get("distribution"), updateQuery.get("distVersion"), updateQuery["force"])
+        rules = cache.get("rules", cache_key, getRawMatches)
```
nthomas-mozilla: Probably only a minor confusion while debugging, but buildTarget contains `-` characters. Could use `:` or some other delimiter in the key.

nthomas-mozilla:
f+. Nicely leveraged off the caching you added a while ago. The pedantic part of my brain wouldn't call it  full-table caching even though it's fairly broad. Unhelpfully it doesn't have an alternative suggestion (brain, you're fired! no wait, come back ... *drool*).

Are you planning any more tests ? Maybe changing the rules while they're cached and verifying they change after cache expiry, and the constituents of the key working as expected.

uwsgi/public.wsgi
```diff
@@ -37,6 +37,8 @@ cache.make_cache("blob", 500, 3600)
 cache.make_cache("blob_schema", 50, 24 * 60 * 60)
 cache.make_cache("blob_version", 500, 60)
 
+cache.make_cache("rules", 20, 30)
```
mostlygeek: does this mean store 20 entries with a 30 second TTL?

auslib/db.py
mozbhearsum: Good point, I'll change this up.

uwsgi/public.wsgi
mozbhearsum: Yeah. I think I might increased the size though. My fresh morning brain thinks there's probably more than 20 combinations of product+buildTarget, and it shouldn't take up too much memory anyways (especially not compared to the blob cache.

mozbhearsum:
Thanks for the quick feedback, folks.

@nthomas-mozilla - I agree that this isn't full table caching. My very very first crack (which never got committed) was going to be the entire table, so I probably just settled on that term in my head.

I'll add a few more tests as well. One case I thought of this morning was what to do when there's no rows returned, eg: in cases where we get silly requests with product literally set to %PRODUCT%. I'm thinking we may want to not cache those (or maybe we do, because it would be nice to save the round trip from the db for those cases anyways?).

auslib/db.py
mostlygeek: That cache key will take up a lot of RAM. Does the cache library hash it internally? If not, using something like FNV-1a hash first would be more space efficient.

uwsgi/public.wsgi
mostlygeek: Yah, cache more! :) Is the product+build target in the request URLs? I can do some cli log parsing to give you a good idea of how many unique combos there are.

auslib/db.py
mozbhearsum: We talked about this on IRC and decided it wouldn't be necessary to hash because the cache size is relatively small (500 items). I'm probably going to want to change the implementation of the rules cache later to just cache the entire table every 30s (rather than caching bits and pieces of it based on the query), which will make this even tinier (1 item cache!) anyways. If we don't end up doing that for some reason we can still look at hashing the keys later (doesn't hurt).

mozbhearsum:
For posterity, rail r+'ed on IRC:
11:25 < rail> bhearsum: looks straight forward to me
11:25 < rail> ship it now? :)

PR #98, opened 2016-07-13
Set cache control header for public app

mozbhearsum:
Not much to say about this one - it should enable a 60s cache that can be shared by requests from different users. We explicitly don't want to cache 500s, to minimize the TTL of transient errors, and I suspect there's no benefit to caching 404s either. I read over the other Cache-Control options as well, and I don't think any of them are of benefit to us (eg: we don't want a lengthier cache for proxies by setting s-max-age).

### DIFF
--- a/auslib/web/views/client.py
+++ b/auslib/web/views/client.py
@@ -10,11 +10,10 @@ import logging
 
 
 class ClientRequestView(MethodView):
-    responseHeaders = {
-        "Cache-Control": "public,max-age=60",
-    }
-
     def __init__(self, *args, **kwargs):
+        # By default, we want a cache that can be shared across requests from different users ("public")
+        # and a maximum age of 60 seconds, to keep our TTL low.
+        self.cacheControl = app.config.get("CACHE_CONTROL", "public,max-age=60")
         self.log = logging.getLogger(self.__class__.__name__)
         MethodView.__init__(self, *args, **kwargs)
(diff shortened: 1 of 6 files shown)


auslib/web/views/client.py
```diff
@@ -10,8 +10,10 @@
 
 
 class ClientRequestView(MethodView):
-
     def __init__(self, *args, **kwargs):
+        # By default, we want a cache that can be shared across requests from different users ("public")
+        # and a maximum age of 60 seconds, to keep our TTL low.
+        self.cacheControl = app.config.get("CACHE_CONTROL", "public,max-age=60")
```
rail: Do we care if the format of this header is somehow invalid?

mozbhearsum: I think it's probably OK to trust it. In production, Ops people would be overwriting it, and I think we can trust that they know what they're doing :). I'm not sure how we'd verify it either unless there's some library that exists. I suppose we could only allow specific parameters (like max-age value) to be passed, but I think it's better to allow this to be fully overridden.

rail: SGTM

mostlygeek:
R+

The `__heartbeat__` and `__lbheartbeat__` and `__version__` can have explicit `Cache-Control: no-cache` so openresty doesn't cache the responses from these.

PR #99, opened 2016-07-15
Add displayVersion to DesupportBlob.

mozbhearsum:
According to Robert, we want to use the last supported version in displayVersion, so we'll need per-channel versions of the existing desupport blobs (like the OS X and GTK deprecation ones). This shouldn't be a big deal.


rail:
Looks straight forward. :shipit:

PR #100, opened 2016-07-15
Bug 1284385 -  Release history no longer gets data for each revision of a release

njirap:


auslib/admin/views/releases.py
```diff
@@ -393,10 +393,28 @@ def get(self, release):
...
+        _mapping = {
+            'data_version': 'data_version',
+            'name': 'name',
+            'product': 'product',
+            'read_only': 'read_only',
+            '_different': '_different',
+            '_time_ago': '_time_ago',
+            'change_id': 'change_id',
+            'changed_by': 'changed_by',
+        }
```
mozbhearsum: What's the purpose of having this mapping when all the keys and values are exactly the same?

mozbhearsum:
It would be good to make sure we don't regress this by adding some test coverage. Perhaps testGetRevisions (https://github.com/mozilla/balrog/blob/master/auslib/test/admin/views/test_releases.py#L945) can be updated to ensure that "data" doesn't exist in the response?

auslib/admin/views/releases.py
mozbhearsum: It also looks like you're missing "timestamp" from here, which makes the releases history items show up with the wrong time in the UI.

njirap:
@mozbhearsum i think we forgot about adding the timestamp..

mozbhearsum:
Adding the timestamp where?

njirap:
You had a comment.."It also looks like you're missing "timestamp" from here, which makes the releases history items show up with the wrong time in the UI."

mozbhearsum:
Ah, right. I just finished merging this manually to master because of the conflicts. I added timestamp ina follow-up (https://github.com/mozilla/balrog/commit/a9ae3813a11db5190c21025fb6f8bc210f995b37).

PR #101, opened 2016-07-19
Compressed history tables

nurav:
Used raw SQL queries since sqlalchemy-migrate does not expose an interface for changing ROW_FORMAT.

### DIFF
--- a/auslib/db.py
+++ b/auslib/db.py
@@ -462,6 +462,7 @@ class History(AUSTable):
         self.table = Table('%s_history' % baseTable.t.name, metadata,
                            Column('change_id', Integer, primary_key=True, autoincrement=True),
                            Column('changed_by', String(100), nullable=False),
+                           mysql_row_format='COMPRESSED',
                            )
         # Timestamps are stored as an integer, but actually contain
         # precision down to the millisecond, achieved through
--- a/auslib/db.py
+++ b/auslib/db.py
@@ -167,9 +167,12 @@ class AUSTable(object):
     """
 
     def __init__(self, db, dialect, history=True, versioned=True, onInsert=None,
-                 onUpdate=None, onDelete=None):
+                 onUpdate=None, onDelete=None, compressHistory=False):
         self.db = db
-        self.t = self.table
+        if hasattr(self, 'table'):
+            self.t = self.table
+        else:
+            raise NotImplementedError
         # Enable versioning, if required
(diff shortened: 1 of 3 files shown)


auslib/db.py
```diff
@@ -461,6 +461,7 @@ def __init__(self, dialect, metadata, baseTable):
         self.table = Table('%s_history' % baseTable.t.name, metadata,
                            Column('change_id', Integer, primary_key=True, autoincrement=True),
                            Column('changed_by', String(100), nullable=False),
+                           mysql_row_format='COMPRESSED',
```
mozbhearsum: Does this code still work if the db is not mysql? We don't care about that as much as we did in the past, but it would be good to not knowingly bust support for other dbs.

auslib/migrate/versions/013_compress_history.py
```diff
@@ -0,0 +1,18 @@
+
+
+def upgrade(migrate_engine):
+    con = migrate_engine.connect()
+
+    if migrate_engine.name == 'mysql':
+        con.execute('ALTER TABLE releases_history ROW_FORMAT=COMPRESSED')
+        con.execute('ALTER TABLE permissions_history ROW_FORMAT=COMPRESSED')
+        con.execute('ALTER TABLE rules_history ROW_FORMAT=COMPRESSED')
```
mozbhearsum: I know it will make it a bit more of a pain to implement, but I think we should explicitly avoid compressing rules or permissions history. There's little to no benefit, so there's no point in making those tables slower.

mozbhearsum:
This looks fine to me now. I'm going to hold off merging it until we finish the current production push.

mozbhearsum:
@nurav - can you fix the merge conflicts here? I'd hoping to merge it in the next day or two.

nurav:
@mozbhearsum fixed.

nurav:
Hmph, had forgotten to commit changes in the tests. Fixed now.

mozbhearsum:
I've been thinking about this some more and I don't think it's worthwhile to pursue at this point. The RDS database is more CPU-bound than disk-bound, and anything we do that slows down submissions will likely raise the number of 400s we get. This might be something to look at in the future, but for now I think we're in the best place we can be.

Sorry for any sunk time on this @nurav, hopefully the research you did will be helpful in the future.

PR #104, opened 2016-07-28
API endpoints for scheduled changes

mozbhearsum:
(This is built on top of reviewed but unmerged patches, so https://github.com/mozbhearsum/balrog/compare/rule-changes...rule-changes-web?expand=1 is probably the best view for this)

I'm hoping this will be a bit more straightforward than the other scheduled changes PRs. It implements the endpoints needed to manage scheduled rule changes via the UI and agent. The guts of that are in auslib/admin/views/scheduled_changes.py, with small subclasses in auslib/admin/views/rules.py. As with other objects, we mostly just support CRUD operations.

I cleaned up some of the exception handling code in auslib/admin/views/base.py to make the responses more consistent (setting content-type should fix some bugs where errors may not be shown). I also fixed up some status code sillyness I found in the existing history endpoints (eg: returning 400 instead of 404).


auslib/admin/views/scheduled_changes.py
```diff
@@ -0,0 +1,214 @@
...
+        if request.args.get("all"):
+            rows = self.sc_table.select()
+        else:
+            rows = self.sc_table.select(where={"complete": False})
+        ret = {"count": len(rows), "scheduled_changes": []}
+        for row in rows:
+            r = {}
+            # TODO: maybe we should require API consumers to use actual column names instead of rewriting?
+            # that may mean we can't re-use rules forms, but it makes this less ugly...
+            # if there's a way to make callers use the column names and still use rules forms we should definitely do that
```
nthomas-mozilla: In some ways it's nicer for clients to use the set of columns we already have for rules, and hide anything yucky behind the curtain.

auslib/admin/views/scheduled_changes.py
```diff
@@ -0,0 +1,214 @@
...
+            self.sc_table.update({"sc_id": sc_id}, what, changed_by, form.sc_data_version.data, transaction)
+        except ValueError as e:
+            self.log.warning("Bad input: %s", e)
+            return Response(status=400, response=json.dumps({"exception": e.args}))
+
+        sc = self.sc_table.select(where={"sc_id": sc_id}, transaction=transaction, columns=["data_version"])[0]
+        new_data_version = sc['data_version']
+        response = make_response(json.dumps(dict(new_data_version=new_data_version)))
+        response.headers['Content-Type'] = 'application/json'
+        return response
```
nthomas-mozilla: Nit-ish: Do you recall why we construct a response like this for _post's, but use jsonify() for _get's ? Seems like the latter is a much more compact form when we don't want to change HTTP status code (here and in rules.py vs releases.py).

auslib/admin/views/scheduled_changes.py
```diff
@@ -0,0 +1,214 @@
...
+
+    def _delete(self, sc_id, transaction, changed_by):
+        where = {"sc_id": sc_id}
+        sc = self.sc_table.select(where, transaction, columns=["sc_id"])
+        if not sc:
+            return Response(status=404)
```
nthomas-mozilla: Missing a human friendly message for the API here ?

auslib/test/admin/views/test_rules.py
```diff
@@ -594,3 +623,335 @@ def testGetRules(self):
...
+        )
+        dbo.rules.scheduled_changes.history.t.insert().execute(
+            change_id=6, changed_by="bill", timestamp=5, sc_id=4
+        )
+        dbo.rules.scheduled_changes.history.t.insert().execute(
+            change_id=7, changed_by="bill", timestamp=6, sc_id=4, scheduled_by="bill", when=500, data_version=1, base_priority=80,
+            base_version="3.3", base_buildTarget="d", base_backgroundRate=0, base_mapping="c", base_update_type="minor", base_data_version=1
+        )
+        dbo.rules.scheduled_changes.history.t.insert().execute(
+            change_id=8, changed_by="bill", timestamp=6, sc_id=4, scheduled_by="bill", when=500, data_version=2, complete=True, base_rule_id=5,
```
nthomas-mozilla: Is this timestamp accidentally coincident with the previous change ?

auslib/test/admin/views/test_rules.py
```diff
@@ -594,3 +623,335 @@ def testGetRules(self):
...
+        ret = self._post("/scheduled_changes/rules", data=data, username="bob")
+        self.assertEquals(ret.status_code, 403, ret.data)
+
+    def testAddScheduledChangeNoPermissionsToMakeChange(self):
+        data = {
+            "when": 1234567, "priority": 120, "backgroundRate": 100, "product": "foo", "channel": "blah",
+            "update_type": "minor", "mapping": "a",
+        }
+        ret = self._post("/scheduled_changes/rules", data=data, username="mary")
+        self.assertEquals(ret.status_code, 403, ret.data)
```
nthomas-mozilla: IIRC 'mary' is allowed to enact changes but not change /rules.  Are you specifically testing this combination or could it be any user without /rules perm ? If the latter it would be clearer to use something else.

auslib/test/admin/views/test_rules.py
```diff
@@ -594,3 +623,335 @@ def testGetRules(self):
...
+    def testEnactScheduledChangeNoPermissions(self):
+        ret = self._post("/scheduled_changes/rules/2/enact", username="bob")
+        self.assertEquals(ret.status_code, 403, ret.data)
+
+    def testGetScheduledChangeHistoryRevisions(self):
+        ret = self._get("/scheduled_changes/rules/3/revisions")
+        self.assertEquals(ret.status_code, 200)
+        expected = {
+            "count": 2,
+            # TODO: should these have base_ on them ?
```
nthomas-mozilla: Presumably destined for the UI, so better to be consistent with what the human expects to see for rules.

nthomas-mozilla:
Looks good to me.

auslib/admin/views/scheduled_changes.py
mozbhearsum: Yeah, I think this is probably the best way. Making life simpler for clients is usually better...

auslib/admin/views/scheduled_changes.py
mozbhearsum: There's some cases (eg: https://github.com/mozilla/balrog/blob/cf68278bf46eddbfff895697ce0081a908d8ee9e/auslib/admin/views/releases.py#L491) where we need to set a different status code, and jsonify doesn't allow us to do that. In this case I don't think there's a reason not to just use jsonify...

mozbhearsum: In going through this I also realized that you can't add extras headers when using jsonify(), which we do in some places for X-Data-Version. I've gone through the entire admin app and made things consistent such that:
- We use jsonify wherever possible (json responses that return a 200)
- We use Response everywhere else. make_response is unnecessary AFAICT.

auslib/test/admin/views/test_rules.py
mozbhearsum: I'm specifically testing that combination. In production, the Balrog Agent will be granted the "scheduled_change" permission with "enact" as the only allowed action - which is very similar to what mary's been granted here. Maybe it would be clearer if the test granted her that specific action too...and probably more future-proof.

auslib/test/admin/views/test_rules.py
mozbhearsum: Yeah...I think I meant to take out this TODO, it's probably a note I left myself very early on.

mozbhearsum:
@nthomas-mozilla - I won't flag you for explicit review again, but feel free to have a look if you want. I'll need one more push still, to fix the merge conflicts, but I want to wait until the stuff that this depends on lands upstream.

nthomas-mozilla:
The last few changesets address everything I mentioned earlier. 👍

PR #106, opened 2016-08-03
Bug 1291649, add new download host for Widevine

nthomas-mozilla:


Callek:
I'm happy to :+1: this in deference of actually :shipit: -ing it. But if ben is back before this ships anyway might as well continue to get his eye

mozbhearsum:
r+

PR #107, opened 2016-08-08
Work around python:2.7 docker image issue

mozbhearsum:
tl;dr - the current python:2.7 image removes pcre libraries, which breaks uwsgi routing support. https://github.com/docker-library/python/issues/132#issuecomment-235691116 has additional background.


Callek:
:shipit: from me, to avoid breaking balrog in prod for our next deploy. @rail can look when he's able if still desired.

rail:
lgtm

PR #108, opened 2016-08-08
Added product-wise admin permissions

nurav:


mozbhearsum:
This looks reasonable, but can you add some tests to Rules and Releases tables to verify that the permissions work correctly from that level as well?

mozbhearsum:
When this lands https://wiki.mozilla.org/Balrog#Permissions will need an update.

auslib/test/admin/views/test_releases.py
```diff
@@ -417,6 +417,12 @@ def testDeleteWithoutPermissionForAction(self):
         ret = self._delete("/releases/b", username="bob", qs=dict(data_version=1))
         self.assertStatusCode(ret, 403)
 
+    def testDeleteWithProductAdminPermission(self):
+        ret = self._delete("/releases/a", username="billy", qs=dict(data_version=1))
+        self.assertStatusCode(ret, 200)
+        ret = self._delete("/releases/b", username="billy", qs=dict(data_version=1))
+        self.assertStatusCode(ret, 403)
```
mozbhearsum: You've done well covering all of the important cases, but tests like this should be split in two - one for the positive case, one for the negative. Others below need to be split up as well.

auslib/test/admin/views/test_rules.py
```diff
@@ -38,6 +38,25 @@ def testNewRulePostJSON(self):
...
         self.assertEquals(r[0]['data_version'], 1)
 
+    def testNewRuleWithProductAdminPermission(self):
+        data = json.dumps(dict(
+            backgroundRate=31, mapping="a", priority=33, product="Firefox",
+            update_type="minor", channel="nightly"
+        ))
+        ret = self._post("/rules", data=data, headers={"Content-Type":
+                                                       "application/json"},
+                         username="billy")
```
mozbhearsum: You shouldn't need to set Content-Type here nor dump to json, those should both be handled by ViewTest._post.

auslib/db.py
```diff
@@ -1399,11 +1399,19 @@ def getOptions(self, username, permission, transaction=None):
             else:
                 return {}
         else:
+            print permission
```
mozbhearsum: Probably not necessary now :)

mozbhearsum:
Functionality-wise, this looks great. I tried it out locally by giving balrogadmin admin permissions to Firefox, which let it do anything to Firefox rules and releases, but nothing to other releases, nor to rules with other products (or no product) specified.

Once you fix up the minor comments and merge conflicts we can get this landed.

PR #110, opened 2016-08-11
Add batch delete for releases and releases_history tables

mostlygeek:
- The `cleanup-db` will now run a shell script which batch deletes for a
  defined length of time
- The delete logic was rewritten to allow for batch deletes of records

### DIFF
--- a/scripts/manage-db.py
+++ b/scripts/manage-db.py
@@ -31,40 +31,58 @@ AND (STR_TO_DATE(RIGHT(name, 14), "%%%%Y%%%%m%%%%d%%%%H%%%%i%%%%S") < NOW() - IN
 """ % nightly_age
     if dryrun:
         todelete = trans.execute("SELECT name FROM releases" + query).fetchall()
+        print "Releases rows to be deleted:"
         if todelete:
-            print "Releases rows to be deleted:"
             print "\n".join(itertools.chain(*todelete))
+        else:
+            print "  - None"
     else:
         trans.execute("DELETE releases FROM releases" + query)
 
 
 def cleanup_releases_history(trans, dryrun=True):
-    query = """
-WHERE name LIKE '%%%%latest'
-AND timestamp<1000*UNIX_TIMESTAMP(NOW()-INTERVAL 14 DAY);
-"""
-    if dryrun:
-        todelete = trans.execute("SELECT name, change_id FROM releases_history" + query).fetchall()
-        if todelete:
-            print "Releases history rows to be deleted:"
-            for key, group in itertools.groupby(todelete, lambda x: x[0]):
-                print "%s: %s history rows" % (key, len(list(group)))
-    else:
-        trans.execute("DELETE releases_history FROM releases_history" + query)
+    num_to_delete = 100
+    queries = dict(
+        dated="""
+            SELECT name, change_id
+            FROM releases_history
+            WHERE name LIKE '%%%%latest'
+              AND timestamp<1000*UNIX_TIMESTAMP(NOW()-INTERVAL 14 DAY)
+            ORDER BY change_id
+            LIMIT %d """ % num_to_delete,
+        nightly="""
+            SELECT name, change_id
+            FROM releases_history
+            WHERE name LIKE '%%%%nightly%%%%'
+              AND name NOT LIKE '%%%%latest'
+              AND timestamp<1000*UNIX_TIMESTAMP(NOW()-INTERVAL 7 DAY)
+            ORDER BY change_id
+            LIMIT %d """ % num_to_delete
+    )
 
-    query = """
-WHERE name NOT LIKE '%%%%latest' AND name LIKE '%%%%nightly%%%%'
-AND timestamp<1000*UNIX_TIMESTAMP(NOW()-INTERVAL 7 DAY);
-"""
-    if dryrun:
-        todelete = trans.execute("SELECT name, change_id FROM releases_history" + query).fetchall()
-        if todelete:
-            print "Releases history rows to be deleted:"
-            for key, group in itertools.groupby(todelete, lambda x: x[0]):
-                print "%s: %s history rows" % (key, len(list(group)))
-    else:
-        trans.execute("DELETE releases_history FROM releases_history" + query)
+    total_deleted = 0
 
+    for name, query in queries.items():
+        if dryrun:
+            todelete = trans.execute(query).fetchall()
+            if todelete:
+                print "releases_history (%s) rows to be deleted:" % name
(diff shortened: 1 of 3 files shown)


mostlygeek:
@mozbhearsum r? 

This is my first attempt at it. I have it running against the test data and it seems to be alright. Usage will work like this: 

```
docker run -it\
  --env-file=/etc/dockerflow/balrog-admin.txt \
  --e MAX_AGE=180 \
  -e DELETE_RUN_TIME=21600 \
  hub.prod.mozaws.net/mozilla/balrog:master-2016-08-08-14-27 cleanup-db
```

Output looks like: 
```
Deleting...20 rows removed >>> 58 seconds of runtime remaining
Deleting...20 rows removed >>> 51 seconds of runtime remaining
...
Deleting...20 rows removed >>> 13 seconds of runtime remaining
```

I'm running it on the testing database in batches of 200 records. It currently deletes 100 nightly and 100 latest records. It will short circuit and stop if 0 records were deleted. Basically that means it has caught up. :)

uwsgi/run.sh
```diff
@@ -21,7 +21,14 @@ elif [ $1 == "cleanup-db" ]; then
...
-    exec python scripts/manage-db.py -d ${DBURI} cleanup ${MAX_AGE}
+    if [ -z "${DELETE_RUN_TIME}" ]; then
+        echo "\${DELETE_RUN_TIME} must be set!"
+        exit 1
+    fi
+
+    exec scripts/run-batch-deletes.sh $DBURI $MAX_AGE $DELETE_RUN_TIME
+
+    #exec python scripts/manage-db.py -d ${DBURI} cleanup ${MAX_AGE}
 elif [ $1 == "cleanup-db-dryrun" ]; then
```
mozbhearsum: If we're going to keep the dryrun command, it should be updated to use the new script as well. Otherwise it's just going to be confusing.

scripts/manage-db.py
```diff
@@ -31,40 +31,75 @@ def cleanup_releases(trans, nightly_age, dryrun=True):
...
+            print "Releases history (latest) rows to be deleted:"
             for key, group in itertools.groupby(todelete, lambda x: x[0]):
-                print "%s: %s history rows" % (key, len(list(group)))
+                print "  - %s: %s history rows" % (key, len(list(group)))
     else:
-        trans.execute("DELETE releases_history FROM releases_history" + query)
+        del_query = """
+            DELETE R.*
+            FROM releases_history R
+            WHERE R.change_id IN (   -- use a subquery to work around mysql limitation
```
mozbhearsum: Clever usage of sql comments to make inline comments here, nice!

scripts/manage-db.py
```diff
@@ -31,40 +31,75 @@ def cleanup_releases(trans, nightly_age, dryrun=True):
...
+                SELECT X.change_id   -- of select'ing from the same table as a DML
+                FROM ( """ + find_records_latest + ") X)"
+
+        results = trans.execute(del_query)
+        if results:
+            print "Deleted %s 'latest' records" % results.rowcount
```
mozbhearsum: It would be good to print the distinct names that were deleting history for. On the off chance we bugger this up in the future, or otherwise are trying to figure out howe some history disappeared, we'd be able to see the names in the output.

scripts/manage-db.py
```diff
@@ -31,40 +31,75 @@ def cleanup_releases(trans, nightly_age, dryrun=True):
...
-WHERE name LIKE '%%%%latest'
-AND timestamp<1000*UNIX_TIMESTAMP(NOW()-INTERVAL 14 DAY);
-"""
+    num_to_delete = 100
+    find_records_latest = """
+        SELECT name, change_id
+        FROM releases_history
+        WHERE name LIKE '%%%%latest'
+          AND timestamp<1000*UNIX_TIMESTAMP(NOW()-INTERVAL 14 DAY)
+        ORDER BY change_id
```
mozbhearsum: Is this ORDER BY there for a reason other than making sure we always delete from oldest -> newest? If it's expensive, we could forego it.

scripts/manage-db.py
```diff
@@ -31,40 +31,75 @@ def cleanup_releases(trans, nightly_age, dryrun=True):
...
+        LIMIT %d """ % num_to_delete
+
+    find_records_nightly = """
```
mozbhearsum: nit: find_records_dated would be a better name for this. Releases ending in -latest are still considered nightly releases.

scripts/manage-db.py
```diff
@@ -31,40 +31,75 @@ def cleanup_releases(trans, nightly_age, dryrun=True):
...
+        if results:
+            print "Deleted %s 'latest' records" % results.rowcount
+            total_deleted += results.rowcount
 
-    query = """
-WHERE name NOT LIKE '%%%%latest' AND name LIKE '%%%%nightly%%%%'
-AND timestamp<1000*UNIX_TIMESTAMP(NOW()-INTERVAL 7 DAY);
-"""
     if dryrun:
-        todelete = trans.execute("SELECT name, change_id FROM releases_history" + query).fetchall()
```
mozbhearsum: It looks like the only difference between this and the above section is which query gets used, and one of the print statements. I don't think the difference in the print statements is necessary, so you could combine this in a loop like "for query in (find_records_latest, find_records_nightly)" if you like. I don't feel strongly about this, just a thought.

scripts/manage-db.py
mostlygeek: It's so we have more deterministic results for replication. I'm not 100% sure it is needed. Though it doesn't hurt performance at all.

mozbhearsum: OK, cool. Let's keep it then!

uwsgi/run.sh
mostlygeek: I changed the design enough that the dryrun is likely to confuse. Maybe renaming it to smoketest? Are we bikeshedding?

mozbhearsum: It's just weird that "cleanup-db" and "cleanup-db-dryrun" do different things. If the latter exists, it should run through the same code paths (minus the DELETEs). I'm fine with just removing it though.

mostlygeek:
@mozbhearsum you [said](https://github.com/mozilla/balrog/pull/110#discussion_r74590204): 

> It would be good to print the distinct names that were deleting history for. On the off chance we bugger this up in the future, or otherwise are trying to figure out howe some history disappeared, we'd be able to see the names in the output.

I think this might be turtles all the way down. Then we have to remember to keep/purge logs after a while too.

mozbhearsum:
Oh, the cron logs won't be available anywhere? If that's the case...meh.

PR #112, opened 2016-08-26
Bug 1289178 - add update or insert to __heartbeat__ check for admin

JohanLorenzo:
~~WIP~~

### DIFF
--- a/auslib/db.py
+++ b/auslib/db.py
@@ -1772,7 +1772,7 @@ class Dockerflow(AUSTable):
             what['watchdog'] += 1
 
             if not dryrun:
-                super(Dockerflow, self).update(where=where, what=what, changed_by=changed_by,   old_data_version=old_data_version, transaction=transaction)
+                super(Dockerflow, self).update(where=where, what=what, changed_by=changed_by, old_data_version=old_data_version, transaction=transaction)
         except IndexError:
             super(Dockerflow, self).insert(changed_by=changed_by, transaction=transaction, watchdog=1)
 
--- a/auslib/db.py
+++ b/auslib/db.py
@@ -1707,52 +1707,39 @@ class Permissions(AUSTable):
         return ret
 
     def getOptions(self, username, permission, transaction=None):
-        self.log.debug('init %s %s %s', username, permission, transaction)
-        ret = self.select(columns=[self.options], transaction=transaction)
-        self.log.debug('ret %s', ret)
+        ret = self.select(columns=[self.options], where=[self.username == username, self.permission == permission], transaction=transaction)
         if ret:
             if ret[0]['options']:
                 return json.loads(ret[0]['options'])
             else:
                 return {}
         else:
-            self.log.debug('raise bitch!')
             raise ValueError('Permission "%s" doesn\'t exist' % permission)
 
     def hasPermission(self, username, thing, action, product=None, transaction=None):
         # Supporting product-wise admin permissions. If there are no options
         # with admin, we assume that the user has admin access over all
         # products.
-        self.log.debug('init %s %s %s %s %s', username, thing, action, product, transaction)
         if self.select(where=[self.username == username, self.permission == 'admin'], transaction=transaction):
-            self.log.debug('in first if')
             options = self.getOptions(username, 'admin', transaction=transaction)
-            self.log.debug('options %o', options)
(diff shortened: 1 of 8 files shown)


JohanLorenzo:
@mozbhearsum The PR doesn't work at the moment. I'd like your opinion on this error:
`balrogadmin_1  | AttributeError: 'AUSDatabase' object has no attribute 'dockerflow'`

Should I explicitly mention dockerflow in AUSDatabase? Or is there just something wrong with the DB init script?

mozbhearsum:
That error is because you haven't referenced the Dockerflow table in AUSDatabase. It needs to be manually created, like the others at https://github.com/mozilla/balrog/blob/master/auslib/db.py#L1877. You can also add a shortcut property like https://github.com/mozilla/balrog/blob/master/auslib/db.py#L1939 if you like, though I don't think it's necessary for this table.

auslib/db.py
```diff
@@ -1745,6 +1745,23 @@ def hasPermission(self, username, thing, action, product=None, transaction=None)
...
 
+class Dockerflow(AUSTable):
+    def getWatchdogValue(self, transaction=None):
+        return self.select(transaction=transaction)
+
+    def incrementWatchdogValue(self, changed_by, transaction=None, dryrun=False):
+        if not self.db.hasPermission(changed_by, "dockerflow", "modify", transaction=transaction):
+            raise PermissionDeniedError("%s is not allowed to modify dockerflow" % changed_by)
+
+        watchdog = self.getWatchdogValue()
```
mozbhearsum: It's totally fine to keep it, but it's not strictly necessary to have a getter here. AUSTable subclasses are allowed to invoke self.select directly.

mozbhearsum: Please add tests for these new methods, too!

auslib/dockerflow.py
```diff
@@ -29,7 +32,7 @@ def heartbeat():
         depend on services like the database to also ensure they are healthy."""
         # Counting the rules should be a trivial enough operation that it won't
         # cause notable load, but will verify that the database works.
-        dbo.rules.countRules()
+        dbo.dockerflow.incrementWatchdogValue(changed_by=DOCKERFLOW_DB_USER)
```
mozbhearsum: One thing I just realized here is that we can't use incrementWatchdogValue on the public app (because it correctly doesn't have write permissions). We should continue using countRules() there.

auslib/db.py
```diff
@@ -1745,6 +1745,23 @@ def hasPermission(self, username, thing, action, product=None, transaction=None)
         return True
 
 
+class Dockerflow(AUSTable):
+    def getWatchdogValue(self, transaction=None):
+        return self.select(transaction=transaction)
+
+    def incrementWatchdogValue(self, changed_by, transaction=None, dryrun=False):
+        if not self.db.hasPermission(changed_by, "dockerflow", "modify", transaction=transaction):
+            raise PermissionDeniedError("%s is not allowed to modify dockerflow" % changed_by)
```
mozbhearsum: Per IRC discussions, let's get rid of this. Because there's no reading of args from the HTTP request, there shouldn't be any opportunity for an injection attack.

auslib/dockerflow.py
JohanLorenzo: Okay, will do

auslib/db.py
JohanLorenzo: Done

auslib/db.py
```diff
@@ -1743,6 +1743,27 @@ def hasPermission(self, username, thing, action, product=None, transaction=None)
...
+        self.table = Table('dockerflow', metadata, Column('watchdog', Integer))
+        AUSTable.__init__(self, db, dialect, history=False)
+
+    def getDockerflowEntry(self, transaction=None):
+        return self.select(transaction=transaction)[0]
+
+    def incrementWatchdogValue(self, changed_by, transaction=None, dryrun=False):
+        try:
+            what = self.getDockerflowEntry()
+            old_data_version = what['data_version']
```
mozbhearsum: Hm, I just realized that versioning is enabled on this table. You can just flip this off (pass versioned=False to the AUSTable constructor) and you don't need to worry about data_version handling.

mozbhearsum:
Other than the data_version comment, this looks good. r=me with that fixed.

PR #113, opened 2016-08-27
[rule_history_api] Changed mapping keys for id and rate

njirap:
Rate changes and rule ids were not being shown on rule
history because the fields were named differently i.e.
'id' and 'background_rate' instead of what the ui uses
i.e. 'rule_id' and 'backgroundRate'

### DIFF
--- a/auslib/admin/views/rules.py
+++ b/auslib/admin/views/rules.py
@@ -182,13 +182,13 @@ class RuleHistoryAPIView(HistoryAdminView):
         _rules = []
         _mapping = {
             # return : db name
-            'id': 'rule_id',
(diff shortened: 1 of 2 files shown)


mozbhearsum:
This looks good, but could you update the test for this method to make sure we don't regress it in the future? https://github.com/mozilla/balrog/blob/18e9f8632f4d9d71cdab81cac490794b30e0ed02/auslib/test/admin/views/test_rules.py#L439 is only checking the length of the data returned from /rules/:id/revisions, not the content.

njirap:
@mozbhearsum tests added..

mozbhearsum:
Thanks, looks great!

PR #114, opened 2016-08-30
Remove unused files now that Dockerflow is in place

JohanLorenzo:
TLDR the diff of 384 files and over 19,800 deletions; this removes entirely the following files:

* `vendor/*`
* `{admin,web}.uwsgi` at the root folder (the one in `uwsgi/` are still in place)
* `scripts/{admin,balrog-server}.py`. I see some other scripts which are not declared elsewhere in the project, but I guess there's here to be manually called.

Tested manually with `docker-compose up`

r? @mozbhearsum


JohanLorenzo:
Side note: I haven't included the changes related to balrog ui, we discussed on IRC. I'm afraid they'd remain unseen in that huge amount of changes.

mozbhearsum:
Thanks for doing this, cleanup is great!

PR #115, opened 2016-08-31
Bug 1299213 - Clean up ui/* now that no submodule is used

JohanLorenzo:
We don't rely on the pre-built dist folder in balrog-ui anymore.

Should land before mozilla/balrog-ui#35


uwsgi/run.sh
```diff
@@ -1,10 +1,18 @@
...
+   cd -
+}
+
 if [ $1 == "public" ]; then
    exec uwsgi --ini /app/uwsgi/public.ini --python-autoreload 1
 elif [ $1 == "admin" ]; then
+   build_front_end
    exec uwsgi --ini /app/uwsgi/admin.ini --python-autoreload 1
 elif [ $1 == "admin-dev" ]; then
+   build_front_end
```
mozbhearsum: Is there a reason you're building the frontend as part of the ENTRYPOINT instead of doing it during the build?

JohanLorenzo: I initially put the build step in `Dockerfile`. It turned out it broke the dev environment because the volumes binding occurs after building. There may be another build script that I missed. In this case, I'd be happy to move it there 😃

scripts/test-entrypoint.sh
```diff
@@ -1,7 +1,24 @@
...
+  npm test
+}
+
+type_of_tests="$1"
+
+if [[ $type_of_tests == "backend" ]]; then
+  shift
+  run_back_end_tests $@
+elif [[ $type_of_tests == "frontend" ]]; then
+  shift
```
JohanLorenzo: Nit: `shift` not needed here

mozbhearsum:
Looks great! :shipit:

PR #117, opened 2016-09-01
Use different default value for permission options.

mozbhearsum:
This is for https://bugzilla.mozilla.org/show_bug.cgi?id=1297765, where it was discovered that I broke creating permissions without options. The problem comes down to the fact that "options" now comes in as a kwarg in https://github.com/mozilla/balrog/blob/a09fd8ebd1f9df402966682a634eb3c729a3c72a/auslib/db.py#L1641, and if it's present in there, we try to insert it. The old code checked "if options", which meant that even passing {} as options would not get into that block.

I _think_ the new behaviour of trying to insert it if present makes sense, so I've adjusted the web layer to be a better consumer of the new db layer.

### DIFF
--- a/auslib/admin/views/forms.py
+++ b/auslib/admin/views/forms.py
@@ -31,6 +31,10 @@ class DisableableTextInput(TextInput):
 class JSONStringField(StringField):
     """StringField that parses incoming data as JSON."""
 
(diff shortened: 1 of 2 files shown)


rail:
🚢 it

PR #118, opened 2016-09-07
Update sample data to include new-style GMP rules and releases.

mozbhearsum:
@Callek, can you sanity check this? The best way is probably to rm -rf .cache locally, and rebuild with the new sample data. It's hard to inspect this by hand.


mozbhearsum:
Callek offered to review this after the fact. It's just sample data, so very low risk, and I'd like to get it merged in to allow GMP folks to get themselves oriented.

PR #119, opened 2016-09-08
Don't allow changes in the past to be scheduled

mozbhearsum:
This turned out to be pretty simple, just a couple of notable things:
* I had to adjust most of the new tests to simulate millisecond-precision timestamps instead of second-precision ones. This is annoying, but makes sense because all "when" timestamps we receive (either coming out of the database, or from the API) are assumed to be to-the-millisecond.
* I changed enactChange to call AUSTable.update directly instead of using ScheduledChangeTable.update. The latter has two value-adds: base table column renaming, and condition validation - but neither of them are relevant to marking changes as complete.

If you want to test this with the UI you can apply that PR and comment out the if block here: https://github.com/mozilla/balrog-ui/pull/30/files#diff-65e9079076aadc61faf96f00687e1364R36

### DIFF
--- a/auslib/test/admin/views/test_rules.py
+++ b/auslib/test/admin/views/test_rules.py
@@ -796,9 +796,8 @@ class TestRuleScheduledChanges(ViewTest):
         }
         self.assertEquals(db_data, expected)
 
-    @mock.patch("time.time")
-    def testAddScheduledChangeNewRule(self, time):
-        time.return_value = 300
+    @mock.patch("time.time", mock.MagicMock(return_value=300))
+    def testAddScheduledChangeNewRule(self):
         data = {
             "when": 1234567, "priority": 120, "backgroundRate": 100, "product": "blah", "channel": "blah",
             "update_type": "minor", "mapping": "a",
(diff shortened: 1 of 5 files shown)


auslib/test/admin/views/test_rules.py
```diff
@@ -795,7 +796,9 @@ def testAddScheduledChangeExistingRule(self):
         }
         self.assertEquals(db_data, expected)
 
-    def testAddScheduledChangeNewRule(self):
+    @mock.patch("time.time")
+    def testAddScheduledChangeNewRule(self, time):
+        time.return_value = 300
```
nthomas-mozilla: Was curious if this could be done in a one-liner, and it turns out yes:
  `@mock.patch("time.time", mock.MagicMock(return_value=300))`
and then the function definition needn't be modified.

auslib/test/admin/views/test_rules.py
```diff
@@ -859,20 +880,24 @@ def testUpdateScheduledChange(self):
...
         ret = self._post("/scheduled_changes/rules/2", username="bob", data=data)
         self.assertEquals(ret.status_code, 403, ret.data)
 
-    def testUpdateRuleWithMergeError(self):
+    @mock.patch("time.time")
+    def testUpdateRuleWithMergeError(self, time):
+        time.return_value = 300
         data = {"mapping": "a", "data_version": 1}
         ret = self._post("/rules/1", data=data)
         self.assertEquals(ret.status_code, 400, ret.data)
```
nthomas-mozilla: This test doesn't involve submitting a scheduled change so I'm not sure why time is being mocked. I see you've done that widely but not everywhere, so perhaps you had a reason to ?

nthomas-mozilla:
r+

auslib/test/admin/views/test_rules.py
mozbhearsum: Oh, cool. That's much nicer - I'll change these up!

PR #120, opened 2016-09-09
Bug 1299203 - Stop using a submodule for UI

JohanLorenzo:


mozbhearsum:
Nice detective work on the issues running the UI tests! This looks pretty good. There's a few small things to address, but they can be done in a follow-up if you prefer:
- Get rid of unnecessary hidden files in ui/ - at least .taskcluster.yml and .gitignore. Possibly .editorconfig and .npmignore as well?
- Run the UI tests in their own Task in CI. This is good for parallelism, but more importantly it makes it more obvious where problems are. Bonus points if run-tests.sh still runs them outside of CI. Maybe run-tests.sh can check $@ to see which tests to run?

JohanLorenzo:
Sounds like good improvements. I'd be in favor of a follow up, just for the sake of having a small patch to review.

PR #121, opened 2016-09-12
Bug 1041584 - move balrog whitelist checks out of AUSDatabase

nurav:


auslib/blobs/base.py
```diff
@@ -123,3 +124,40 @@ def getFooterXML(self, updateQuery, update_type, whitelistedDomains, specialForc
...
         raise NotImplementedError()
+
+    def containsForbiddenDomain(self, product):
+        """Returns True if the blob contains any file URLs that contain a
+           domain that we're not allowed to serve updates to."""
+        # Check the top level URLs, if the exist.
+        f = open('testing', 'a')
+        f.write(str(app.config.get("WHITELISTED_DOMAINS")))
+        f.close()
+        for c in self.get('fileUrls', {}).values():
```
nthomas-mozilla: This is an validator which works app releases only. Could we keep bug 1165164 and bug 1293352 in mind and put it in apprelease.py instead ? We could subclass, and just make the GMP/SystemAddon copies no-ops for now.

nurav: Yes, Ben altered me to this. I'm working on it!

mozbhearsum:
From IRC, so I don't forget:
```
16:53 <~bhearsum> you're setting up app.config in setUp, and then explicitly referencing it in the individual tests, to pass it along to the blobs?
16:54 < vjoshi> i'm setting up app.config in setUp primarily so that the tests in test_releases don't fail 
16:55 < vjoshi> but they fail all the same :(
16:55 <~bhearsum> test_releases.py? or test_apprelease.py?
16:55 < vjoshi> the tests in test_apprelease check containsForbiddenDomain specifically, so i thought it made sense to pass the whiltelist explicitly
16:56 < vjoshi> the tests in apprelease pass, but the ones in test_releases fail
16:57 <~bhearsum> oh, i see
16:57 < vjoshi> since the whitelist config item somehow can't be accesses
16:57 < vjoshi> *accessed
16:59 <~bhearsum> vjoshi: i think https://github.com/mozilla/balrog/pull/121/files#diff-c98aa223a3239bab49a28a78e373f7bb is causing your issue
16:59 <~bhearsum> that's the class that the test_releases.py tests inherit from, and you've removed the domain whitelist setup
16:59 <~bhearsum> and haven't replaced it with anything
17:00 < vjoshi> i've removed setDomainWhitelist everywhere
17:00 <~bhearsum> yeah, and you're calling containsForbiddenDomain with calls like https://github.com/mozilla/balrog/pull/121/files#diff-7c2c8a646f34fec41c773940931bf694L1413
17:00 < vjoshi> since i thought that the aim of the bug was to replace it's usages with app.config["WHITELISTED_DOMAINS"]
17:00 <~bhearsum> they look at app.config
17:00 < vjoshi> yeah
17:01 <~bhearsum> so, if you were to add app.config["WHITELISTED_DOMAINS"] in auslib/test/admin/views/base.py, it would probably fix the failures
17:01 <~bhearsum> however...we've been trying to avoid having the database layer know anything about the web layer
17:02 < vjoshi> but https://github.com/mozilla/balrog/pull/121/files#diff-c98aa223a3239bab49a28a78e373f7bbL21 does that
'''

File: auslib/blobs/base.py [hunk 2]
```diff
@@ -123,3 +123,6 @@ def getFooterXML(self, updateQuery, update_type, whitelistedDomains, specialForc
 
     def getInnerXML(self, updateQuery, update_type, whitelistedDomains, specialForceHosts):
         raise NotImplementedError()
+
+    def containsForbiddenDomain(self, product, whitelistedDomains):
+        return False
```
mozbhearsum: I think it would be best to raise NotImplementedError here. Defaulting to False makes it easier for us to introduce bugs that would allow bad domains through in the future.

mozbhearsum: I also wonder if we should be calling this from .validate() instead of requiring callers to make a separate call for it. It's another thing that would make it harder for to make mistakes later. What do you think about that?

auslib/test/blobs/test_gmp.py
```diff
@@ -206,3 +206,51 @@ def testGMPWithForbiddenDomain(self):
         self.assertEqual(returned_header.strip(), expected_header.strip())
         self.assertItemsEqual(returned, expected)
         self.assertEqual(returned_footer.strip(), expected_footer.strip())
+
+    def testContainsForbiddenDomain(self):
```
mozbhearsum: Thank you for adding new tests here!

mozbhearsum:
This looks really good overall. All of the logic in the blobs looks good, and will enhance the security for GMP and SystemAddon blobs!

Regarding the test failures and setDomainWhitelist -- we need to find a solution that allows the database layer to have knowledge of the whitelist without looking at app.config. I think keeping setDomainWhitelist is a fine way to do this. It wasn't that we were trying to get rid of here - it was the blob-specific logic in Releases.containsForbiddenDomain.

If you can think of another way to give the Releases class access to the domain whitelist without db.py importing app, I'm totally open to it!

The reasoning behind this is that the db.py is the inner-most layer of the application, and the web layer is above it. It's fine for outer layers to call into more inner layers, but not the opposite. http://rhodesmill.org/brandon/slides/2014-07-pyohio/clean-architecture/ might be worth a read, if you're interested in the architecture we're aiming for.

auslib/blobs/base.py
```diff
@@ -123,3 +123,11 @@ def getFooterXML(self, updateQuery, update_type, whitelistedDomains, specialForc
 
     def getInnerXML(self, updateQuery, update_type, whitelistedDomains, specialForceHosts):
         raise NotImplementedError()
+
+    def containsForbiddenDomain(self, product, whitelistedDomains):
+        raise NotImplementedError()
```
mozbhearsum: I think you'll need to implement this method on all of the other blob types too, namely: SuperBlob, WhitelistBlob, and DesupportBlob. Otherwise I don't think we'll be able to use them anymore :).

mozbhearsum:
This all looks good now. Looking at this in its final form makes me think we should add "product" as a required property of all Blobs, which would get rid of the need to pass it along to .validate(). That's definitely out of scope here though - I'll file a new bug for that.

mozbhearsum:
I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1303106 for the product follow-up.

PR #122, opened 2016-09-12
apprelease.py: Update the DesupportBlob

meetmangukiya:
Make detailsURL dynamic

Bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1270827


meetmangukiya:
Let me know if I need to amend my commit.

auslib/blobs/apprelease.py
```diff
@@ -700,7 +700,7 @@ def getHeaderXML(self, updateQuery, update_type, whitelistedDomains, specialForc
 
     def getInnerXML(self, updateQuery, update_type, whitelistedDomains, specialForceHosts):
         xml = []
-        xml.append('    <update type="%s" unsupported="true" detailsURL="%s" displayVersion="%s">' % (update_type, self['detailsUrl'], self["displayVersion"]))
+        xml.append('    <update type="%s" unsupported="true" detailsURL="%s" displayVersion="%s">' % (update_type, self['detailsUrl'].replace("%LOCALE%", updateQuery['locale']).replace("%VERSION%", updateQuery['version']), self["displayVersion"]))
```
mozbhearsum: This line is quite long - can you split it up? It might be best to assign the substituted url to a temporary variable, and use that in this substitution.

mozbhearsum:
Pretty close now!

In addition to the line length above, you'll need to update the tests as well. There's one that's failing (full log is at https://tools.taskcluster.net/task-inspector/#Ag1D0gCNSb6z_PHxKZyGCg), probably because updateQuery is set-up wrong in it:
```
_______________________ TestDesupportBlob.testDesupport ________________________
[gw1] linux2 -- Python 2.7.12 /app/.tox/py27/bin/python2.7
self = <auslib.test.blobs.test_apprelease.TestDesupportBlob testMethod=testDesupport>

    def testDesupport(self):
        updateQuery = []
        returned_header = self.blob.getHeaderXML(updateQuery, "minor", self.whitelistedDomains, self.specialForceHosts)
>       returned = self.blob.getInnerXML(updateQuery, "minor", self.whitelistedDomains, self.specialForceHosts)

auslib/test/blobs/test_apprelease.py:1631: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = {'name': 'd1', 'detailsUrl': 'http://moo.com/cow', 'schema_version': 50, 'displayVersion': '50.0'}
updateQuery = [], update_type = 'minor', whitelistedDomains = {'a.com': ('a',)}
specialForceHosts = ['http://a.com']

    def getInnerXML(self, updateQuery, update_type, whitelistedDomains, specialForceHosts):
        xml = []
>       xml.append('    <update type="%s" unsupported="true" detailsURL="%s" displayVersion="%s">' % (update_type, self['detailsUrl'].replace("%LOCALE%", updateQuery['locale']).replace("%VERSION%", updateQuery['version']), self["displayVersion"]))
E       TypeError: list indices must be integers, not str
```

And you should write a new test that verifies that substitution actually works.

auslib/blobs/apprelease.py
meetmangukiya: This is done, in current commit, does the variable name seem alright :sweat_smile: ?

meetmangukiya:
Why'd it return `TypeError: list indices must be integers, not str` :confused: ? We didn't try to access any element or wherer we specified an index? 

Might need some help with the tests :see_no_evil:

mozbhearsum:
I think the problem with the existing test is that the fake "updateQuery" that's created at https://github.com/mozilla/balrog/blob/ecdf83e079af1f91ee5520c12a434f73db001191/auslib/test/blobs/test_apprelease.py#L1629 is incorrect. The "updateQuery" is supposed to be a dict, not a list. You can see an example of a properly created updateQuery in some other tests like https://github.com/mozilla/balrog/blob/ecdf83e079af1f91ee5520c12a434f73db001191/auslib/test/blobs/test_apprelease.py#L175. Doing something like that in testDesupport will probably fix that test up.

meetmangukiya:
I'm trying to run the tests, but I was not able to, I'll check it as soon as I can run tests.

meetmangukiya:
Tests passed locally.

mozbhearsum:
There's still a failing test actually, but it's being hidden by something else. Can you pull from the latest upstream and try again?

You also need to add a new test that makes sure that substitution works. Let me know if you need a hand with that.

meetmangukiya:
Do we have something like a CI, that'll test my code. If there is then, it'd have been better ?

mozbhearsum:
Yeah, CI will test it. The problem is that the version of your code that this PR is based on has an issue that prevents some failures from being reported. If you merge the latest upstream and push it, you should see a failure.

meetmangukiya:
Just did.

meetmangukiya:
Says not satisfied :/

meetmangukiya:
https://tools.taskcluster.net/task-inspector/#Go-E4dbQRd2wPXz4J8IyUw/0 Says 99% coverage and 2 misses. What does misses mean ?

mozbhearsum:
It looks like https://github.com/mozilla/balrog/pull/122/commits/e4d330bdf87ab07fe8ac1d36008f46096a7f2659 fixed up the test, and provided the coverage that we need.

The "satisified" part can be ignored...it's "state" that matters. You'll see something like https://tools.taskcluster.net/task-graph-inspector/#GIC7OUS_QiKOBJF_m2Qm4Q/ or https://tools.taskcluster.net/task-inspector/#VRDh0FW-Q3u_vnlyn2WOLQ/ when you get a failure.

mozbhearsum:
The "missing" parts are talking about lines of code that don't get run as part of the tests. This pull request hasn't lowered our code coverage percentage, so there's nothing to worry about there.

I'm going to get this merged now!

meetmangukiya:
Don't we need a commit amend? The commits aren't well :see_no_evil:

mozbhearsum:
No need - we use squash merging, so I'll take care of improving the commit message.

meetmangukiya:
OK, so I think I made it through the `good-first-bug` :smile: Yay!

mozbhearsum:
You did! 👍 . There may be a follow-up for %OS% later, but I need to sort out something else before that's possible.

PR #123, opened 2016-09-13
Don't allow multiple scheduled changes to the same PK

mozbhearsum:
I'm not sure if this is what we want in the medium or long term, but I think it's important to fix the current footgun quickly. This PR disallows mulitple scheduled changes to the same rule.

I also switched all of the mock calls in this file to the new one line way of mocking return values that @nthomas-mozilla discovered, since they were nearby. You can look at the second and third commits by themselves if you want to get rid of the noise that that caused.


nthomas-mozilla:
lgtm

PR #124, opened 2016-09-13
Bug 1278543 (history field view doesn't work on integer columns)

NinadBhat:
Solves Bug 1278543 and adds test case.


mozbhearsum:
You mentioned on IRC that you thought https://github.com/mozilla/balrog/blob/5d75609ed75b7a349b0ca5ba5faf6cd8940af775/auslib/admin/views/base.py#L92 might need updating as well. Is there a reason you decided not to do that?

The code and test that are here looks good, though.

PR #126, opened 2016-09-16
Bug 1112628 - disallow deletion of releases that have a rule.mapping or rule.whitelist pointing at them

NinadBhat:
The test case I have currently written should fail as
`self.releases.delete({"name": "d"}, changed_by="bill", old_data_version=1)`
should throw an error.

### DIFF
--- a/auslib/db.py
+++ b/auslib/db.py
@@ -9,7 +9,7 @@ import sys
 import time
 
 from sqlalchemy import Table, Column, Integer, Text, String, MetaData, \
-    create_engine, select, BigInteger, Boolean, join
+    create_engine, select, BigInteger, Boolean, join, ForeignKey
 from sqlalchemy.exc import SQLAlchemyError
 from sqlalchemy.sql.expression import null
 import sqlalchemy.types
@@ -1005,7 +1005,7 @@ class Rules(AUSTable):
                            Column('rule_id', Integer, primary_key=True, autoincrement=True),
                            Column('alias', String(50), unique=True),
                            Column('priority', Integer),
-                           Column('mapping', String(100)),
+                           Column('mapping', String(100), ForeignKey('releases.name', ondelete="SET NULL"), nullable=False),
                            Column('backgroundRate', Integer),
                            Column('update_type', String(15), nullable=False),
                            Column('product', String(15)),
@@ -1021,6 +1021,7 @@ class Rules(AUSTable):
                            Column('headerArchitecture', String(10)),
                            Column('comment', String(500)),
                            Column('whitelist', String(100)),
+
                            )
         AUSTable.__init__(self, db, dialect, scheduled_changes=True)
 
@@ -1591,7 +1592,6 @@ class Releases(AUSTable):
             self._proceedIfNotReadOnly(toDelete["name"], transaction=transaction)
             if not self.db.hasPermission(changed_by, "release", "delete", toDelete["product"], transaction):
                 raise PermissionDeniedError("%s is not allowed to delete releases for product %s" % (changed_by, toDelete["product"]))
-
         if not dryrun:
             super(Releases, self).delete(where=where, changed_by=changed_by, old_data_version=old_data_version, transaction=transaction)
             for name in names:
--- a/auslib/db.py
+++ b/auslib/db.py
@@ -9,7 +9,7 @@ import sys
 import time
 
 from sqlalchemy import Table, Column, Integer, Text, String, MetaData, \
-    create_engine, select, BigInteger, Boolean, join, ForeignKey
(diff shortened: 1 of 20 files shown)


NinadBhat:
I figured It has something to do with
http://docs.sqlalchemy.org/en/latest/dialects/sqlite.html#foreign-key-support

auslib/test/test_db.py
```diff
@@ -1524,6 +1524,14 @@ def testDeleteRelease(self):
         release = self.releases.t.select().where(self.releases.name == 'a').execute().fetchall()
         self.assertEquals(release, [])
 
+    def testDeleteWithRuleMapping(self):
+        self.releases.t.insert().execute(name='d', product='d', data=json.dumps(dict(name="d", schema_version=1, hashFunction="sha512")),
+                                         data_version=1)
+        self.rules.t.insert().execute(rule_id=1, priority=100, version='3.5', buildTarget='d', backgroundRate=100, mapping='d', update_type='z', data_version=1)
+        self.releases.delete({"name": "d"}, changed_by="bill", old_data_version=1)
+        release = self.releases.t.select().where(self.releases.name == 'd').execute().fetchall()
+        self.assertEquals(release, [])
```
mozbhearsum: I think this test is backwards. If Foreign Keys are being enforced, self.releases.delete should raise an Exception rather than letting the Release be deleted, right?

auslib/db.py
```diff
@@ -1547,12 +1548,15 @@ def localeExists(self, name, platform, locale, transaction=None):
 
     def delete(self, where, changed_by, old_data_version, transaction=None, dryrun=False):
         names = []
+        engine = super(Releases, self).getEngine()
+        connection = engine.connect()
+        connection.execute("PRAGMA foreign_keys=ON")
+        connection.close()
```
mozbhearsum: I think this is in the wrong place. This PRAGMA statement is sqlite-specific, and should only be used in the tests. If possible, please put it in the setUp() method of TestReleases. If it's not possible to put it there, it should probably go in https://github.com/mozilla/balrog/blob/7bae002dd0f178b37e1e8d73f77eafca0b3e79bd/auslib/db.py#L1912, and only run when the dialect is "sqlite".

auslib/db.py
```diff
@@ -1020,6 +1020,7 @@ def __init__(self, db, metadata, dialect):
                            Column('headerArchitecture', String(10)),
                            Column('comment', String(500)),
                            Column('whitelist', String(100)),
```
mozbhearsum: Whitelist also maps to Releases, so you should add a foreign key for it too. Please add a test as well.

mozbhearsum:
One other thing: you will need to add a migration script to https://github.com/mozilla/balrog/tree/master/auslib/migrate/versions so that we can upgrade existing databases to add the foreign key. It can be a bit tricky, so let me know if you need a hand.

NinadBhat:
@bhearsum 
https://github.com/mozilla/balrog/blob/master/auslib/test/test_db.py#L1856
fails giving error
https://pastebin.mozilla.org/8910806

mozbhearsum:
Hm. It's hard to say for sure, but the problem with testDeleteReleaseClobbersCache may be caused by the fact that TestBlobCaching doesn't initialize the Rules table. You've got other test failures (eg: testDeleteReadOnlyRelease) that are failing for similar reasons.

You'll need to update the data some of the tests are initialized with to make it possible for these tests to Delete releases. Give that a try, and then we can see if testDeleteReleaseClobbersCache is still failing.

auslib/db.py
```diff
@@ -1004,7 +1004,7 @@ def __init__(self, db, metadata, dialect):
                            Column('rule_id', Integer, primary_key=True, autoincrement=True),
                            Column('alias', String(50), unique=True),
                            Column('priority', Integer),
-                           Column('mapping', String(100)),
+                           Column('mapping', String(100), nullable=True),
```
mozbhearsum: What's the reasoning behind making mapping nullable?

auslib/test/test_db.py
```diff
@@ -1524,6 +1524,20 @@ def testDeleteRelease(self):
         release = self.releases.t.select().where(self.releases.name == 'a').execute().fetchall()
         self.assertEquals(release, [])
 
+    def testDeleteWithRuleMapping(self):
```
mozbhearsum: Yay, tests! Good work on these.

auslib/db.py
NinadBhat: Actually `ForeignKey('releases.name', ondelete="SET NULL"), nullable=False `
was supposed to return an error when Parent is deleted. As it did not work, will removing ForeignKey I must have set nullable to True. It isn't  needed I will remove it.

PR #127, opened 2016-09-20
WIP: apprelease: support OS substitution in desupport blob

meetmangukiya:
WIP

### DIFF
--- a/auslib/test/blobs/test_apprelease.py
+++ b/auslib/test/blobs/test_apprelease.py
@@ -1644,20 +1644,20 @@ class TestDesupportBlob(unittest.TestCase):
 {
     "name": "d1",
     "schema_version": 50,
-    "detailsUrl": "http://moo.com/%LOCALE%/cow/%VERSION%/",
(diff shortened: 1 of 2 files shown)


auslib/test/blobs/test_apprelease.py
```diff
@@ -1644,20 +1644,20 @@ def setUp(self):
...
     "schema_version": 50,
-    "detailsUrl": "http://moo.com/%LOCALE%/cow/%VERSION%/",
+    "detailsUrl": "http://moo.com/%LOCALE%/cow/%VERSION%/%OS%",
     "displayVersion": "50.0"
 }
 """)
 
     def testDesupport(self):
-        updateQuery = {"locale": "<locale>", "version": "<version>"}
+        updateQuery = {"locale": "<locale>", "version": "<version>", "buildTarget": "<OS_Something>"}
```
mozbhearsum: It would be better to use a real build target here. "Darwin_x86_64-gcc3-u-i386-x86_64" is probably the best one to use, since it is the most complex.

meetmangukiya:
I was of the view that `balrog` had 100% coverage requirement.

PR #128, opened 2016-09-20
Switch to 2.7-slim docker image.

mozbhearsum:
Pretty simple! I had to explicitly install libfontconfig1 because phantomjs depends on it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,10 +2,7 @@ FROM python:2.7-slim
 
 MAINTAINER bhearsum@mozilla.com
 
-# Some versions of the python:2.7 Docker image remove libpcre3, which
-# uwsgi needs for routing support to be enabled.
-# We may be able to remove this after https://github.com/docker-library/python/pull/137
-# is fixed.
+# Some versions of the python:2.7 Docker image remove libpcre3, which uwsgi needs for routing support to be enabled.
 # Node and npm are to build the frontend. nodejs-legacy is needed by this version of npm.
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,13 +1,18 @@
-FROM python:2.7
+FROM python:2.7-slim
 
 MAINTAINER bhearsum@mozilla.com
 
-# Some versions of the python:2.7 Docker image remove this library, which
+# Some versions of the python:2.7 Docker image remove libpcre3, which
 # uwsgi needs for routing support to be enabled.
 # We may be able to remove this after https://github.com/docker-library/python/pull/137
```
JohanLorenzo: The docker [PR was merged](https://github.com/docker-library/python/commit/2f0d83db398e760a14a831370d54c926d3542959). The [latest published version](https://github.com/docker-library/python/blob/855b85c8309e925814dfa97d61310080dcd08db6/2.7/Dockerfile) of the Dockerfile have the changes in it. I don't see anything related to libpcre3, though. I wonder if we can get rid of this comment, then.

mozbhearsum: Good point - I will adjust the comment. I think I'll leave the explicit install though...it never hurts to be explicit.

mozbhearsum:
The failure above is a transient network issue. I retriggered the failed tests in https://tools.taskcluster.net/task-inspector/#SlMkXurATo-h03-b9nsgMA/, and it succeeded.

PR #130, opened 2016-09-20
Bug 1296685 remove platform version

NinadBhat:
https://bugzilla.mozilla.org/show_bug.cgi?id=1296685


auslib/blobs/apprelease.py
```diff
@@ -705,6 +705,25 @@ def __init__(self, **kwargs):
             self['schema_version'] = 5
 
 
+class ReleaseBlobV6(ReleaseBlobBase, NewStyleVersionsMixin, MultipleUpdatesXMLMixin, UnifiedFileUrlsMixin):
+    """
+    """
```
mozbhearsum: Please add a comment with what's changed between V5 and V6. It makes things much easier for humans!

auslib/blobs/schemas/apprelease-v6.yml
```diff
@@ -0,0 +1,210 @@
+---
```
mozbhearsum: You'll need to add support for "product" here, as I just merged https://github.com/mozilla/balrog/pull/129.

auslib/test/blobs/test_apprelease.py
```diff
@@ -1629,6 +1629,123 @@ def testSchema5OptionalAttributes(self):
...
+        }
+    }
+}
+""")
+
+    def testIsValid(self):
+        # Raises on error
+        self.blobH2.validate('h', self.whitelistedDomains)
+
+    def testSchema5OptionalAttributes(self):
```
mozbhearsum: Please update this test name to reflect the new schema version.

auslib/test/blobs/test_apprelease.py
```diff
@@ -1629,6 +1629,123 @@ def testSchema5OptionalAttributes(self):
         self.assertEqual(returned_footer.strip(), expected_footer.strip())
 
 
+class TestSchema6Blob(unittest.TestCase):
```
mozbhearsum: Since the only change between schema 5 and schema 6 is the removal of attributes, it would be good to have a test that ensures a schema 6 blob fails validation if one of those fields is present.

PR #132, opened 2016-09-27
Bug 1284481 active data

NinadBhat:


scripts/manage-db.py
```diff
@@ -84,13 +85,42 @@ def cleanup_releases_history(trans, dryrun=True):
...
+def extract_active_data(URL, loc="."):
+    """
+    Function was added in to enhance testing in Balrog's stage environment.
+
+    :param URL: Database. eg : mysql://balrogadmin:balrogadmin@balrogdb/balrog
+    :param loc: location where sqldump file must be created
+    :return: Stores sqldump data in dump.sql file in the specified location, if not specified stores in current directory
+            If dump.sql file already exists it will override that file and not append it.
+    """
+    data = re.split(':|/|@', URL)
```
mozbhearsum: I know it's worked in your own testing, but there's a lot of edge cases when parsing URLs. Please use https://docs.python.org/2/library/urlparse.html instead.

scripts/manage-db.py
```diff
@@ -84,13 +85,42 @@ def cleanup_releases_history(trans, dryrun=True):
 
     print "Total Deleted: %d" % total_deleted
 
+
+def extract_active_data(URL, loc="."):
```
mozbhearsum: I see that "loc" is a directory that "dump.sql" will be written too. It would be better to not hardcode the filename, and let this be the filename to write to.

scripts/manage-db.py
```diff
@@ -84,13 +85,42 @@ def cleanup_releases_history(trans, dryrun=True):
...
+    data = re.split(':|/|@', URL)
+    user = data[3]
+    password = data[4]
+    host = data[5]
+    database = data[6]
+
+    popen('mysqldump -h %s -u %s -p%s --single-transaction %s rules rules_history rules_scheduled_changes \
+           rules_scheduled_changes_history permissions permissions_history migrate_version  > %s/dump.sql' % (host, user, password, database, loc,))
+    popen('mysqldump -h %s -u %s -p%s --single-transaction %s releases \
+           --where="exists (select * from rules where releases.name = rules.mapping)" >> %s/dump.sql ' % (host, user, password, database, loc,))
```
mozbhearsum: This query is a good start, but don't forget that you need to look for releases that are referenced by rules.whilelist, rules_scheduled_changes.mapping, and rules_scheduled_changes.whitelist as well.

scripts/manage-db.py
```diff
@@ -84,13 +85,42 @@ def cleanup_releases_history(trans, dryrun=True):
...
+
+    popen('mysqldump -h %s -u %s -p%s --single-transaction %s releases_history \
+    --where="releases_history.name=\'Firefox-mozilla-central-nightly-latest\' ">> %s/dump.sql' % (host, user, password, database, loc,))
+
+    popen('mysqldump -h %s -u %s -p%s --single-transaction %s releases_history --where="exists (select * from rules where \
+    releases_history.name = rules.mapping AND rules.alias LIKE \'firefox-release\' ) " >> %s/dump.sql' % (host, user, password, database, loc,))
```
mozbhearsum: Since we know the exact string we're matching, it's better (and more performant) to use rules.alias="firefox-release" instead of LIKE here.

scripts/manage-db.py
```diff
@@ -84,13 +85,42 @@ def cleanup_releases_history(trans, dryrun=True):
...
+
+
 if __name__ == "__main__":
     from optparse import OptionParser
     usage = """%s --db dburi action [options]\n""" % sys.argv[0]
     usage += "Possible actions:\n"
     usage += "  create: Create all the tables required for a new Balrog database\n"
     usage += "  upgrade: Upgrade an existing balrog table to a newer version.\n"
     usage += "  downgrade: Downgrade an existing balrog table to an older version.\n"
+    usage += "  extract: Extracts active data for testing "
```
mozbhearsum: Please add a note to this docstring that indicates an output file can be specified.

scripts/manage-db.py
```diff
@@ -2,7 +2,8 @@
 
 import itertools
 import logging
-from os import path
+from os import path, popen
+from sqlalchemy.engine.url import make_url
```
mozbhearsum: Nice find - this is even better than urlparse.

scripts/manage-db.py
```diff
@@ -84,13 +85,57 @@ def cleanup_releases_history(trans, dryrun=True):
...
+    :return: Stores sqldump data in  the specified location, if not specified stores in current directory in file dump.sql
+            If  file already exists it will override that file and not append it.
+    """
+    url = make_url(URL)
+    user = url.username
+    password = url.password
+    host = url.host
+    database = url.database
+
+    popen('mysqldump -h %s -u %s -p%s --single-transaction %s rules rules_history rules_scheduled_changes \
```
mozbhearsum: I just noticed that we added another table recently. Can you add "dockerflow" to this list as well?

scripts/manage-db.py
```diff
@@ -84,13 +85,57 @@ def cleanup_releases_history(trans, dryrun=True):
...
+           --where="exists (select * from rules where releases.name = rules.whitelist  )" \
+               >> %s ' % (host, user, password, database, loc,))
+
+    popen('mysqldump -h %s -u %s -p%s --single-transaction %s releases \
+          --where="exists (select * from rules_scheduled_changes where releases.name = rules_scheduled_changes.base_mapping  )"\
+            >> %s' % (host, user, password, database, loc,))
+
+    popen('mysqldump -h %s -u %s -p%s --single-transaction %s releases \
+          --where="exists (select * from rules_scheduled_changes where releases.name = rules_scheduled_changes.base_whitelist )"\
+                >> %s' % (host, user, password, database, loc,))
```
mozbhearsum: I think we're going to run into some issues with duplicate rows if these are run in separate queries. It's possible that the same releases.name row will match multiple queries (in particular, the 1st and the 3rd). Can you combine this into one larger query to avoid that problem?

NinadBhat: @mozbhearsum By 1st and 3rd you mean 1st and 3rd quires to  releases right?

mozbhearsum: That's right, though I'd prefer that all queries to "releases" are done in a single mysqldump command.

NinadBhat: @mozbhearsum  done

mozbhearsum:
I just finished testing this against the production database. It largely looks good, but there's a few small things that need to be addressed still:
* By default, mysqldump locks tables. We should disable this in case any writes come in while the script is running (--skip-lock-tables).
* Firefox-mozilla-central-nightly-latest had a lot more rows in releases_history than I anticipated. Can you limit that query to 1000 rows? It's okay to leave the other releases_history query without a limit.
* The second releases_history query looks like it's trying to select everything from releases_history, because EXISTS (select ....) is true for all rows. I think you need to use something like: --where "name = (SELECT mappping from rules WHERE rules.mapping='firefox-release')" instead.
* By default, mysqldump adds "DROP TABLE" and "CREATE TABLE" statements before the INSERTs. This means that the second mysqldump for releases_history will end up dropping everything from the first. Can you adjust that to either get rid of those in the second query, or merge both queries into one?

Great work so far, this is going to be really helpful to development and staging once completed!

mozbhearsum:
This is 99% fine - but it still has the second "CREATE TABLE releases_history" statement. It looks like you've granted me access to edit your branch, so it looks like I can make that tiny fix myself, and then merge.

Note: the test failures are not legitimate, but caused by https://bugzilla.mozilla.org/show_bug.cgi?id=1305994.

PR #133, opened 2016-09-28
Remove whitelist checks from desupport blobs.

mozbhearsum:
While testing https://bugzilla.mozilla.org/show_bug.cgi?id=1270827 I realized that we're requiring the detailsUrl of DesupportBlob to be in the domain whitelist. This is probably not desired, as detailsUrl doesn't deliver binaries, and adding domains like support.mozilla.org to it would mean adding them to "Firefox"'s whitelist, which could open up an attack vector.

This is my bad, I failed to catch this while reviewing https://bugzilla.mozilla.org/show_bug.cgi?id=1041584. I'll try to get a fix in today's push, as the current state of things means that we're not delivering any responses to desupported users at the moment.

@nurav - can you sanity check this?

### DIFF
--- a/auslib/blobs/apprelease.py
+++ b/auslib/blobs/apprelease.py
@@ -769,6 +769,6 @@ class DesupportBlob(Blob):
         return '</update>'
 
     def containsForbiddenDomain(self, product, whitelistedDomains):
-        if isForbiddenUrl(self.get('detailsUrl', None), product, whitelistedDomains):
-            return True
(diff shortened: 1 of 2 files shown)


mozbhearsum:
Test failures are because of https://bugzilla.mozilla.org/show_bug.cgi?id=1305994. I retrigerred the failed job in https://tools.taskcluster.net/task-inspector/#Hs87wfi2RWux9SoPz0H5Fg/0, and it succeeded.

PR #134, opened 2016-09-29
Add .gitattributes file

F3real:


.gitattributes
```diff
@@ -0,0 +1,5 @@
+# Set the default behavior, in case people don't have core.autocrlf set.
+* text=auto
+
+# Declare files that will always have LF line endings on checkout.
+*.sh text eol=lf
```
mozbhearsum: So, "* text=auto" means that all files will have lf line endings when checked in, but it may convert the endings to crlf upon checkout - that seems fine.

And "*.sh text eol=lf" means that all .sh files will be _not_ be converted to crlf upon checkout, and this is because bash doesn't like crlf line endings after the directory has been volume mounted into the image?

F3real: Yes, that's right.

mozbhearsum:
This seems fine, but I'm hoping to give it a try myself on a Windows 10 machine before merging. I should be able to do that in the next few days or so.

PR #135, opened 2016-09-30
Added script to add "product" to blobs currently in the database

nurav:


scripts/add-product-to-blob.py
```diff
@@ -0,0 +1,58 @@
...
+                      action="store", help="Username for the Admin interface")
+    parser.add_option("-p", "--password", dest="password", type="string", default=False,
+                      action="store", help="Password for the Admin interface")
+    options, args = parser.parse_args()
+
+    if not options.url or not options.username:
+        print "url and auth is required"
+        sys.exit(1)
+
+    url = 'http://' + options.url
```
mozbhearsum: This won't work for production, because we use https there. Probably best just to have this arg be protocol+host+port.

scripts/add-product-to-blob.py
```diff
@@ -0,0 +1,58 @@
...
+    releases = data.json()['releases']
+
+    unsuccessful_releases = []
+    for release in releases:
+        try:
+            client = requests.Session()
+            data = client.get(url + '/api/releases/' + release['name'])
+            blob = data.json()
+            csrf_token = data.headers['X-CSRF-Token']
+            if 'product' not in blob.keys():
```
mozbhearsum: Having this condition was a good idea, because it makes this safe and easy to rerun. Nice job!

scripts/add-product-to-blob.py
```diff
@@ -0,0 +1,58 @@
...
+    parser.add_option("-p", "--password", dest="password", type="string", default=False,
+                      action="store", help="Password for the Admin interface")
+    options, args = parser.parse_args()
+
+    if not options.url or not options.username:
+        print "url and auth is required"
+        sys.exit(1)
+
+    url = 'http://' + options.url
+    data = requests.get(url + '/api/releases')
```
mozbhearsum: Auth needs to be passed for the GETs as well, it's not set automatically for prod.

scripts/add-product-to-blob.py
```diff
@@ -0,0 +1,58 @@
...
+                print 'changing %s' % release['name']
+                blob['product'] = release['product']
+                data = dict(data=json.dumps(blob), product=release['product'],
+                            data_version=release['data_version'])
+
+                data['csrf_token'] = csrf_token
+
+                r = client.post(url + '/api/releases/' + release['name'],
+                            json=data, auth=(options.username,
+                                             options.password))
```
mozbhearsum: You should probably call r.raise_for_status() here - otherwise 400s don't cause errors.

mozbhearsum:
This looks fine now. I won't be merging, like I said, but I'll close this after I run it against prod.

mozbhearsum:
This worked well for updating the majority of the blobs! Some failed to update due to non-whitelisted domains, but we'll deal with them elsewhere. Closing this now.

PR #136, opened 2016-10-01
Fix entrypoints on Windows

F3real:


mozbhearsum:
Explicit is usually better, but I'm surprised this makes a difference (entrypoint is run in the container, so it shouldn't be affected by anything on the host). Can you explain what difference this makes?

F3real:
I managed to found similar issue on docker github. As it turns out this way script isn't being run directly, it runs using bash, which has the executable permission enabled already.
[Issue](https://github.com/docker/compose/issues/2301#issuecomment-154450785)

mozbhearsum:
Ah, I see. So this sortof works around the issue by avoiding the need to have an execute bit on run.sh. Very clever, nice work finding that!

PR #138, opened 2016-10-03
(apprelease.py) Raise value error of channel not in 'fileUrls'

njirap:
Fixes bug #1170919

### DIFF
--- a/auslib/blobs/apprelease.py
+++ b/auslib/blobs/apprelease.py
@@ -587,6 +587,8 @@ class UnifiedFileUrlsMixin(object):
             # 1) Its exact specified channel.
             # 2) Its fallback channel.
             # 3) In the catch-all "channel" ("*").
+            # When a channel is present in 'fileUrls' then we don't fall back to the
+            # 'catch-all' block if a product is missing from the channel block
             channels = [
                 updateQuery['channel'],
                 getFallbackChannel(updateQuery['channel']),
@@ -594,8 +596,9 @@ class UnifiedFileUrlsMixin(object):
             ]
             url = None
             for c in channels:
-                url = self.get("fileUrls", {}).get(c, {}).get(patchKey, {}).get(from_)
-                if url:
+                config_block = self.get("fileUrls", {}).get(c, {})
+                if config_block:
+                    url = config_block.get(patchKey, {}).get(from_)
                     break
 
             # If we still can't find a fileUrl, we cannot fulfill this request.
--- a/auslib/test/blobs/test_apprelease.py
+++ b/auslib/test/blobs/test_apprelease.py
@@ -3,6 +3,8 @@ try:
 except ImportError:
     # so you're in python <2.7
     from ordereddict import OrderedDict
+
+import mock
 import unittest
 
 from auslib.global_state import dbo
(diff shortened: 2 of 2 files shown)


njirap:
hi @mozbhearsum is this fail test error something I can fix from my end?

screen shot 2016-10-03 at 7 13 03 pm

mozbhearsum:
Yeah, that's an error that we've been getting intermittenly. Try rerunning, it should fix it. #131 will make it less likely to occur, too, which I expect to merge this week.

auslib/blobs/apprelease.py
```diff
@@ -594,8 +594,9 @@ def _getUrl(self, updateQuery, patchKey, patch, specialForceHosts):
             ]
```
mozbhearsum: Can you update the comment at line 585 to talk about this new behaviour?

njirap:
hi @mozbhearsum kindly advise how to rerun...clicking "Retriever task doesn't seem to get the job done"
screen shot 2016-10-04 at 7 46 45 pm

auslib/test/blobs/test_apprelease.py
```diff
@@ -1507,6 +1507,38 @@ def testForbiddenDomainAndAllowedDomain(self):
...
+            "buildTarget": "p", "locale": "l", "channel": "c1",
+            "osVersion": "a", "distribution": "a", "distVersion": "a",
+            "force": 0
+        }
+        patchKey = "partials"
+        patch = {"from": "h1"}
+        specialForceHosts = ["http://a.com"]
+
+        # overwrite the fileUrls section not to include the 'h1' product
+        self.blobH2['fileUrls']['c1']['partials'] = {'h2': 'http://a.com/h2-partial.mar'}
```
njirap: yes that makes sense...

auslib/blobs/apprelease.py
njirap: ok..

auslib/test/blobs/test_apprelease.py
```diff
@@ -1812,6 +1876,38 @@ def testCheckFailForUnsuportedAttributes(self):
...
+
+        with self.assertRaises(ValueError):
+            self.blobH2._getUrl(updateQuery, patchKey, patch, specialForceHosts)
+
```
njirap: @mozbhearsum I have been wondering if it would be better to test the `UnifiedFileUrlsMixin` instead having the same test for all the classes that make use of it......So just testing the `UnifiedFileUrlsMixin` as it's own independent unit(from where the term `unit tests` comes from)

mozbhearsum: That's a very good idea!

mozbhearsum:
Re: retriggering the tests. That button doesn't work for Taskcluster-Github (https://bugzilla.mozilla.org/show_bug.cgi?id=1295131). "Edit and Recreate" does, but I'm not sure if you'd have enough permissions to do that, but I've done it here: https://tools.taskcluster.net/task-inspector/#SbLehhFsQCWszylDZRTnvA/.

FYI: the problom causing all the failures is tracked in https://bugzilla.mozilla.org/show_bug.cgi?id=1305994.

nthomas-mozilla:
Ben, I think this looks good. Ideally we'd stop submitting betas into the '*' block for RCs, since it's confusing and I'm 99% sure unused. From memory we've tried to be explicit everywhere, so there shouldn't be any other cases we're relying on the bug.

auslib/test/blobs/test_apprelease.py
```diff
@@ -1853,3 +1856,67 @@ def testBrokenDesupport(self):
...
+        def side_effects(key, *args):
+            return values[key]
+
+        self.mixin_instance = UnifiedFileUrlsMixin()
+        self.mixin_instance.getPlatformData = mock.Mock(side_effect=side_effects)
+        self.mixin_instance.get = mock.Mock(side_effect=side_effects)
+        self.mixin_instance.log = mock.MagicMock()
+        self.mixin_instance.log.debug = mock.MagicMock()
+
```
mozbhearsum: Nice usage of mock to make this testable.

mozbhearsum:
I'm not sure why Taskcluster didn't fire on this, but I ran the tests locally instead, and they passed.

PR #140, opened 2016-10-06
Add a horrible hack to greatly reduce CPU usage of balrogui container.

mozbhearsum:
I'm pretty sure the root cause of this problem is as described in https://github.com/gruntjs/grunt-contrib-watch/issues/429#issuecomment-216011305 and https://github.com/docker/for-mac/issues/82#issuecomment-247563198 (basically: monitoring a bunch of files for changes inefficiently), but @JohanLorenzo and I couldn't come up with a proper fix.

Please forgive me for this horrible hack.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -72,8 +72,7 @@ balrogui:
     - WEB_PORT=8080
   entrypoint:
     - /bin/bash
-    - -c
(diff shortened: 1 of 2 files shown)


mozbhearsum:
@nthomas-mozilla - could you give this a shot, since you're the one that first noticed the CPU usage problem?

nthomas-mozilla:
This works for me too, yay!

PR #141, opened 2016-10-07
Copy in requirements files early again

mozbhearsum:
@JohanLorenzo - I seem to recall that we added the COPY ./ /app/ because we couldn't volume mount in run-tests.sh. However, I noticed that we end up reinstalling dependencies a lot by not doing the requirements files first. This should have the same pratcial effect of of the COPY ./, except be more cache friendly.

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -22,9 +22,6 @@ WORKDIR /app
 # The general app requirements and packages required to run Tox are installed into the system.
 # We copy them in early to avoid re-installing them unless absolutely necessary.
 COPY requirements.txt requirements-tox.txt /app/
-# install the requirements into the container first
-# these rarely change and is more cache friendly
-# ... really speeds up building new containers
 RUN pip install -r requirements.txt
 RUN pip install -r requirements-tox.txt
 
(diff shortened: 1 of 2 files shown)


Dockerfile.dev
```diff
@@ -19,16 +19,19 @@ RUN apt-get -q update \
...
+COPY requirements.txt requirements-tox.txt /app/
 RUN pip install -r requirements.txt
 RUN pip install -r requirements-tox.txt
 
+COPY aus-data-snapshots/ /app/aus-data-snapshots/
+COPY auslib/ /app/auslib/
+COPY ui/ /app/ui/
+COPY scripts/ /app/scripts/
+COPY uwsgi/ /app/uwsgi/
+COPY MANIFEST.in requirements-test.txt setup.py tox.ini version.json version.txt /app/
```
JohanLorenzo: Out of curiosity, do we have to include `MANIFEST.in` and `version.json`? IIRC, the docker image isn't built out of the dev Dockerfile.

JohanLorenzo:
And it makes us closer to the original `Dockerfile` 😃  LGTM. I wonder if some files could be taken out, but that'd be a very slight improvement.

Dockerfile.dev
mozbhearsum: Maybe not version.json, but MANIFEST.in is needed to make setup.py operations work, which tox does. Without it, some files aren't copied to the tempdir where the build happens.

PR #142, opened 2016-10-11
fix issues with change notifier in production

mozbhearsum:
This should fix the two issues we found in https://bugzilla.mozilla.org/show_bug.cgi?id=1304082:
1) Production uses a local MTA, so we need to be able to disable tls.
2) We sometimes (or maybe always) get IndexErrors when the read only change notifier fires. I haven't been able to reproduce this, so I just added a guard against it with a comment for now.

I also noticed that the recent Dockerfile changes caused us to start including .tox/ in coverage reports, so I fixed that.

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -30,7 +30,7 @@ COPY auslib/ /app/auslib/
 COPY ui/ /app/ui/
 COPY scripts/ /app/scripts/
 COPY uwsgi/ /app/uwsgi/
-COPY MANIFEST.in requirements-test.txt setup.py tox.ini version.json version.txt /app/
+COPY .coveragerc MANIFEST.in requirements-test.txt setup.py tox.ini version.json version.txt /app/
 
 # Using /bin/bash as the entrypoint works around some volume mount issues on Windows
 # where volume-mounted files do not have execute bits set.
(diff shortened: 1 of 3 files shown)


Dockerfile.dev
```diff
@@ -30,7 +30,7 @@ COPY auslib/ /app/auslib/
 COPY ui/ /app/ui/
 COPY scripts/ /app/scripts/
 COPY uwsgi/ /app/uwsgi/
-COPY MANIFEST.in requirements-test.txt setup.py tox.ini version.json version.txt /app/
+COPY .coveragerc MANIFEST.in requirements-test.txt setup.py tox.ini version.json version.txt /app/
```
rail: Not sure if this is related, but LGTM.

uwsgi/admin.wsgi
```diff
@@ -43,13 +43,17 @@ cache.make_cache("blob_schema", 50, 24 * 60 * 60)
 
 dbo.setDb(os.environ["DBURI"])
 if os.environ.get("NOTIFY_TO_ADDR"):
+    tls = False
```
rail: I would probably use `use_tls` or `tls_enabled`; `tls` sounds more like an object name to me, not a boolean.

PR #144, opened 2016-10-11
Reinstated Sentry logging

nurav:


mozbhearsum:
This looks like it will be fine, I'm waiting to get access to our Sentry server to verify before merging.

auslib/util/sentry.py
```diff
@@ -0,0 +1,13 @@
+from raven.processors import Processor
+
+
+class SanitizeHeadersProcessor(Processor):
```
mozbhearsum: It looks like this is the default now, based on https://github.com/getsentry/raven-python/blob/80490488fb5c62d927ac789313f971e02ee395e2/raven/conf/defaults.py - let's get rid of this!

mozbhearsum:
This looks good now. For posterity, the "failure" is the intermittent issue with Debian mirrors in the frontend tests. Nothing to worry about.

PR #145, opened 2016-10-12
Bug 1282898 allow multifile updates

NinadBhat:
https://bugzilla.mozilla.org/show_bug.cgi?id=1282898

### DIFF
--- a/auslib/test/blobs/test_superblob.py
+++ b/auslib/test/blobs/test_superblob.py
@@ -8,6 +8,8 @@ class TestSchema1Blob(unittest.TestCase, MemoryDatabaseMixin):
     maxDiff = 2000
 
     def setUp(self):
+        self.specialForceHosts = ('http://a.com',)
+        self.whitelistedDomains = {'a.com': ('b', 'c', 'e', 'b2g', 'response-a', 'response-b', 's')}
         self.superblob = SuperBlob()
         self.superblob.loadJSON("""
 {
@@ -23,3 +25,19 @@ class TestSchema1Blob(unittest.TestCase, MemoryDatabaseMixin):
     def testGetResponseProducts(self):
         products = self.superblob.getResponseProducts()
         self.assertEqual(products, ['c', 'd'])
+
+    def testXML(self):
+        updateQuery = {
+            "product": "gg", "version": "3", "buildID": "1",
+            "buildTarget": "p", "locale": "l", "channel": "a",
+            "osVersion": "a", "distribution": "a", "distVersion": "a",
+            "force": 0
+        }
+        headerXML = self.superblob.getHeaderXML(updateQuery, "minor", self.whitelistedDomains,
+                                                self.specialForceHosts)
+        footerXML = self.superblob.getFooterXML(updateQuery, "minor", self.whitelistedDomains,
+                                                self.specialForceHosts)
+        expected_header = '    <addons>'
+        expected_footer = '    </addons>'
+        self.assertEqual(headerXML, expected_header)
+        self.assertEqual(footerXML, expected_footer)
--- a/auslib/test/web/test_client.py
+++ b/auslib/test/web/test_client.py
@@ -442,6 +442,80 @@ class ClientTestBase(ClientTestCommon):
     }
 }
 """)
+        dbo.rules.t.insert().execute(priority=200, backgroundRate=100,
+                                     mapping='superblobaddon-with-mupltiple-response-bob', update_type='minor',
+                                     product='superblobaddon-with-mupltiple-response-bo',
+                                     data_version=1)
+        dbo.rules.t.insert().execute(priority=200, backgroundRate=100,
+                                     mapping='superblobaddon-with-one-response-bob', update_type='minor',
+                                     product='superblobaddon-with-one-response-bob',
+                                     data_version=1)
+        dbo.rules.t.insert().execute(priority=190, backgroundRate=100,
+                                     mapping='responseblob-a', update_type='minor',
+                                     product='responseblob-a',
+                                     data_version=1)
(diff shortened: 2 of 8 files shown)


NinadBhat:
@mozbhearsum  tests fail because at line https://github.com/mozilla/balrog/blob/master/auslib/test/web/test_client.py#L291
superblob points at response-a and response-b which themselves are apprelease blobs
shouldn't they be superblob be pointing at gmp blobs?

NinadBhat:
test failing as 
at https://github.com/mozilla/balrog/blob/master/auslib/test/web/test_client.py#L291
superblob points at response-a, response-b which are supposed to be gmp blob
but are apprelease blob

auslib/web/views/client.py
```diff
@@ -70,6 +72,17 @@ def get(self, **url):
                     response_blobs.append({'product_query': product_query,
                                            'response_release': response_release,
                                            'response_update_type': response_update_type})
+            elif response_blob_names:
+                for blob_name in response_blob_names:
+                    product_query = query.copy()
+                    product_query["product"] = blob_name
```
NinadBhat: @mozbhearsum 
https://github.com/mozilla/balrog/blob/master/auslib/web/views/client.py#L99
will fail if we do not update the query.

auslib/blobs/schemas/superblob.yml
```diff
@@ -1,31 +1,67 @@
 ---
```
mozbhearsum: Pretty good work on this schema. There's an issue on line 8 with indentation, but after I fixed that it worked fine. I'm surprised that the tests didn't pick up the error with the schema...

auslib/test/blobs/test_superblob.py
```diff
@@ -19,7 +21,39 @@ def setUp(self):
     ]
 }
 """)
+        self.superblob_addon = SuperBlob()
+        self.superblob_addon.loadJSON("""
+{
+    "name": "fake",
```
mozbhearsum: Best to use a different name for this to avoid confusion.

auslib/test/blobs/test_superblob.py
```diff
@@ -19,7 +21,39 @@ def setUp(self):
...
+    "schema_version": 6000,
```
mozbhearsum: Please update this schema number to be correct - there is no blob with schema 6000!

auslib/test/web/test_client.py
```diff
@@ -442,6 +442,95 @@ def setUp(self):
     }
 }
 """)
+        dbo.rules.t.insert().execute(priority=200, backgroundRate=100,
+                                     mapping='superblobaddon-with-multiple-response-blob', update_type='minor',
+                                     product='superblobaddon-with-multiple-response-blob',
+                                     data_version=1)
+        dbo.rules.t.insert().execute(priority=200, backgroundRate=100,
+                                     mapping='superblobaddon-with-one-response-blob', update_type='minor',
+                                     product='superblobaddon-with-one-response-blob',
```
mozbhearsum: I can see why you were confused about blob name and product now, since we often use the same value in these tests. It might help to have a look at http://localhost:8080/releases once you get your Docker images working again. You'll notice that there are only a few different product names, but many blobs in each product.

You're just following the current style here, so need to change anything in these tests.

auslib/test/web/test_client.py
```diff
@@ -639,30 +728,30 @@ def testGetWithResponseProducts(self):
         ret = self.client.get('/update/4/gmp/1.0/1/p/l/a/a/a/a/1/update.xml')
         self.assertUpdateEqual(ret, """<?xml version="1.0"?>
 <updates>
-    <update type="minor" version="None" extensionVersion="2.5" buildID="25">
+    <addons>
```
mozbhearsum: Thanks for updating these tests, they're less confusing now!

auslib/web/views/client.py
```diff
@@ -57,8 +58,10 @@ def get(self, **url):
         if release:
             response_products = release.getResponseProducts()
             response_blobs = []
+            response_blob_names = release.getResponseBlobs()
+            self.log.debug("got blob names: %s", response_blob_names)
```
mozbhearsum: Please remove this before merging.

auslib/web/views/client.py
```diff
@@ -70,40 +73,55 @@ def get(self, **url):
...
-                xml.append(headerXML)
+            # Extracting Header
+            # In case of superblob Extracting Header form parent release
+            innerHeaderXML = release.getInnerHeaderXML(query,
+                                                       update_type,
+                                                       app.config["WHITELISTED_DOMAINS"],
+                                                       app.config["SPECIAL_FORCE_HOSTS"])
+            # Extracting Footer
+            # In case of superblob Extracting Header form parent release
+            innerFooterXML = release.getInnerFooterXML(query,
```
NinadBhat: @mozbhearsum 
Used intermediate variable to check if None is returned ( None is returned in case of systemaddons blob having 'uninstall' as False https://github.com/mozilla/balrog/blob/master/auslib/blobs/systemaddons.py#L83 )

mozbhearsum: Can you change those functions to return an empty string instead of None? I don't think there's any value in having them return None.

NinadBhat: I actually tried it before but it ends up appending a newline.

mozbhearsum:
Functionally, this looks fine now. Just fix up noted issue with getInner{Header,Footer} and I can get this merged!

auslib/web/views/client.py
NinadBhat: K I will update the tests then

mozbhearsum:
You did excellent work here @NinadBhat. This was a large bug with a lot of change and churn, and you ended up with a great patch in the end - nice work!

PR #147, opened 2016-10-14
Use volume mount when running tests in local dev, but not in Taskcluster

mozbhearsum:
We tried to enable a volume mount for local tests in #131, but got foiled by Taskcluster not supporting it. I realized today that we can still do this as long as we have a way to disable it for Taskcluster.

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -10,6 +10,8 @@ tasks:
     payload:
       maxRunTime: 1200
       image: "rail/python-test-runner"
+      env:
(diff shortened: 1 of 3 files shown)


mozbhearsum:
Looks like this works! Locally, the volume mount is used when I run tests. In Taskcluster, it isn't. @JohanLorenzo - how does this look to you?

JohanLorenzo:
LGTM. As the logic in `*run-tests.sh` is similar and becoming more complex, I wonder if we shouldn't port this script into a python one. I won't block on that, though.

PR #149, opened 2016-10-18
Bug 1305994: Stop using httpredir.debian.org

rail:


sources.list.jessie
```diff
@@ -0,0 +1,4 @@
+deb http://ftp.debian.org/debian jessie main
+deb http://ftp.debian.org/debian jessie-updates main
```
mozbhearsum: I don't think ftp.debian.org is a good choice...it resolves to a single server in the Netherlands :(.

mozbhearsum:
I'm changing my mind on this after discovering the hacky runtime mirror selection won't work for us. Rail, do you mind updating this to move sources.list.jessie out of the root of the repo (maybe into scripts?) and then I'll merge this?

mozbhearsum:
Dustin also suggested that we could use a Python script to do the mirror selection in a less hacky, and python-slim compatible, way.

mozbhearsum:
Going to close this again - let's use Dustin's mirror selection suggestion instead.

PR #151, opened 2016-10-18
Make Scheduled Changes conditions configurable

mozbhearsum:
The main change here is reworking condition_groups to make it possible to track enabled conditions. We now have condition_groups, all_condition_args, and enabled_condition_groups. The latter two are just subsets of the former, and only exist for convenience.

I've added the "dryrun" kwarg to AUSTable's insert/update/delete methods. The original Scheduled Changes work made this part of the interface, but those tests did not catch that AUSTable didn't support them, because the TestTable used overrode them and added it. These new tests don't do that, so I caught the fact that AUSTable didn't support them.

Other notable things:
- I'm only adding columns for enabled conditions. This might get a bit awkward if the enabled conditions on a specific table change over time (we'd need to do extra schema changes & may lose data). After stepping back and looking at this again I'm starting to think we should just always add all condition columns...the downside being that it may be confusing if you look at the raw table.
- I'm not 100% happy with the interface for changing conditions (passing conditions in scheduled_changes_kwargs via the base table) but I haven't been able to come up with something better. We can't pass a pre-configurde ScheduledChangesTable instance to the base table, because it requires an instance of the base table as an arg. I'm curious if anyone else has ideas about this.


auslib/db.py
```diff
@@ -185,7 +185,7 @@ class AUSTable(object):
     """
 
     def __init__(self, db, dialect, history=True, versioned=True, scheduled_changes=False,
-                 onInsert=None, onUpdate=None, onDelete=None):
+                 scheduled_changes_kwargs={}, onInsert=None, onUpdate=None, onDelete=None):
```
mozbhearsum: Huh. I know there's some risk involved in using mutable values because they may end up being shared, but I didn't realize they could create global variables...

auslib/db.py
```diff
@@ -735,25 +746,41 @@ class ScheduledChangeTable(AUSTable):
...
-    )
+    condition_groups = {
+        "time": ("when",),
+        "uptake": ("telemetry_product", "telemetry_channel", "telemetry_uptake"),
+    }
+    # It's also useful just to know what all the valid condition args are sometimes.
+    # Note that this does not take into account conditions that may be disabled
+    # for some tables. enabled_condition_groups, which can only be constructed at
+    # instantiation time, tracks these.
+    all_condition_args = itertools.chain(*condition_groups.values())
```
mozbhearsum: Hm, you're right. I forgot to kill this after the final refactor.

JohanLorenzo:
Following up on last Friday's discussion. Here's [a small schema](https://docs.google.com/drawings/d/1TCevEYiqV95m1EB-TIUo-UOSWLnwWkRkuzW0Ovr542s/edit?usp=sharing) of the proposed tables.
In this propose, we add another table, in charge of knowing the trigger conditions.

A rule is then defined by 3 tables:
* The current state of the rule: `rules`. No changes needed to that one
* The past and upcoming changes: `rules_scheduled_changes`. It contains the values to put in `rules`, and who's the (latest) author.
* The conditions which trigger the values from `rules_scheduled_changes` to be written in `rules`. Let's call this table `rules_scheduled_changes_conditions` (open to a better name :) ). This table only the expected state. As of now, we support 2 kinds of states: the one given by telemetry and the other, time based. If new conditions come up, we'll add new columns to this table.

# Workflows proposal

From a back-end point of view:

* [Add a new scheduled change](https://www.websequencediagrams.com/?lz=dGl0bGUgQWRkIGEgbmV3IHNjaGVkdWxlZCBjaGFuZ2UKClMACghSdWxlc0FQSVZpZXctPgAJDlNlcnZpY2U6IGFkZF9uZXcocnVsZV9pZCwgcnVsZXNfdmFsdWVzLCBjb25kaXRpb25zLCBhdXRob3IpAFIPAEMHAFEQVGFibGU6IGluc2VydABZBQBLCgBBCG5vdGUgcmlnaHQgb2YgACsVYWxzbyBzYXZlcyB0aW1lc3RhbXAAgVMPAGUFAIFHGQCCFwlfAIFbBwCCERAAgSoWQwCBbwkAgTsOAD4RAIIXDCk&s=rose)
* [Update a condition](https://www.websequencediagrams.com/?lz=dGl0bGUgVXBkYXRlIGNvbmRpdGlvbnMKClNjaGVkdWxlZFJ1bGVzQVBJVmlldy0-AAkOU2VydmljZTogdQA8BV8ANwoocwA4CF9ydWxlX2lkLABUCywgYXV0aG9yKQBYDwBJBwBYD0MAgQ4JVGFibABlCQBFHgA1JgBEDl9sYXN0XwCBFQYAgS0UAIEtCA&s=rose)
* [Update a scheduled target value](https://www.websequencediagrams.com/?lz=dGl0bGUgVXBkYXRlIHNjaGVkdWxlZCB0YXJnZXQgdmFsdWVzCgpTABEIUnVsZXNBUElWaWV3LT4ACQ5TZXJ2aWNlOiB1AEkFKABFCV9ydWxlX2lkLCBydWxlc18AUAYsIGF1dGhvcikATw8AQAcAThBUYWJsACkz&s=rose)

Please tell me if something looks clumsy, or if some areas aren't too clear.

auslib/db.py
```diff
@@ -735,25 +746,36 @@ class ScheduledChangeTable(AUSTable):
...
-    )
+    condition_groups = {
+        "time": ("when",),
+        "uptake": ("telemetry_product", "telemetry_channel", "telemetry_uptake"),
+    }
+
+    def __init__(self, db, dialect, metadata, baseTable, conditions=("time", "uptake"), history=True):
+        if not conditions:
+            raise ValueError("No conditions enabled, cannot initialize ScheduledChangesTable for %s", baseTable.t.name)
+        if set(conditions).difference(self.condition_groups.keys()):
```
mozbhearsum: Good idea, will do.

auslib/test/test_db.py
```diff
@@ -1002,6 +995,132 @@ def testMergeUpdateWithConflict(self):
...
+    def testValidateConditionsJustWhen(self):
+        self.sc_table._validateConditions({"when": 12345678})
+
+    def testValidateConditionsTelemetryRaisesError(self):
+        conditions = {
+            "telemetry_product": "Firefox",
+            "telemetry_channel": "nightly",
+            "telemetry_uptake": "200000",
+        }
+        self.assertRaisesRegexp(ValueError, "uptake condition is disabled", self.sc_table._validateConditions, conditions)
```
mozbhearsum: Not in the new ConfigurableConditions tests, but TestScheduledChangesTable has a couple: https://github.com/mozilla/balrog/blob/a300741332da939c1ba7eefa95e8c1cd147b1515/auslib/test/test_db.py#L701 -- does that cover it, or do you think we also need to test when _only_ telemetry is enabled?

mozbhearsum:
Johan and I talked a bit more about this proposal on IRC. Two things of note/clarification:
- Web APIs will not be affected by moving conditions to their own Table, that complexity will be encapsulated.
- His proposal has us creating a new Service layer between the db and the web layers. I'm going to forego that complexity for now, and put that logic in the database layer instead. I think a Service layer is worth looking at though, as we have a mix of traditional Service stuff in the db and web layers at the moment. He's going to file a bug on that.

I'm going to take a stab at moving conditions to their own table. This may end up in its own pull request. Thanks folks!

auslib/test/test_db.py
nthomas-mozilla: That should cover it, thanks.

mozbhearsum:
I've just finished merging the conditions tables and other recently merged-to-master work into this branch. AFAIK, the only changes since last time you reviewed this were to unhork merge conflicts.

While doing my own review of this I started to wonder whether we should just disable uptake conditions for Rules for now, since the Agent doesn't support them anyways. I didn't do that, partly because schema changes are annoying, and partly because I'm still hopeful we'll get support in the future. I'm open to doing that, though.

auslib/db.py
```diff
@@ -871,7 +887,7 @@ def _splitColumns(self, columns):
         base_columns = {}
         condition_columns = {}
         for k in columns:
-            if k in itertools.chain(*self.conditions.condition_groups):
+            if k in itertools.chain(*self.conditions.condition_groups.values()):
```
JohanLorenzo: Nit: `k` seems a bit odd if we take the values of a condition group. How about something like `condition_type`?

auslib/db.py
JohanLorenzo: The current code looks alright from a global var point of view (even though like you said, `scheduled_changes_kwargs` will remain shared).

As far as I know, a global var happens with this kind of scenario:
```py
def my_func(foo={}):    # Side note: meh, it's not a pure function
    foo['bar'] = 'baz'
    return foo
 
a = my_func()
a['baz'] = 'real baz'

b = my_func()    # b == {'baz': 'real baz', 'bar': 'baz'}
```

auslib/db.py
mozbhearsum: Ah, I see what you mean re: global vars now - it's the same issue we see if we have a dict as the default arg in a constructor, and then modify it in the constructor - it will be modified for all consumers. Since this is essentially a static variable, this is probably OK for now. We will need to start copying it if AUSTable starts modifying it at some point.

Good catch on the naming here, I didn't think very hard when switching this!

mozbhearsum:
@nthomas-mozilla - this should be ready for final review now, at your convenience.

mozbhearsum:
Thanks guys!

PR #152, opened 2016-10-18
Peak into blob and read name during adding/updating releases

F3real:


ui/app/js/controllers/release_new_controller.js
```diff
@@ -18,6 +18,28 @@ function($scope, $http, $modalInstance, CSRF, Releases, releases) {
...
+    if (typeof file !== 'undefined') {
+      // should work
+      reader.readAsText(file);
+    }
+
+  };
+
+  $scope.changeName = function () {
+    //wait for actual file to be loaded
+    setTimeout( $scope.fillName , 500);
```
F3real: When I was calling function directly I was getting file undefined since it wasn't loaded yet.

ui/app/js/controllers/release_new_controller.js
```diff
@@ -18,6 +18,28 @@ function($scope, $http, $modalInstance, CSRF, Releases, releases) {
...
   $scope.saving = false;
 
+  $scope.fillName = function () {
+    var file = $scope.dataFile;
+
+    var reader = new FileReader();
+    reader.onloadend = function(evt) {
+      var blob = evt.target.result;
+      $scope.$apply( function() {
+         $scope.release.name = JSON.parse(blob).name;
```
F3real: I thought that name field is mandatory in blob. I'll change it to set it to empty if it's not present.

ui/app/templates/release_modal.html
```diff
@@ -4,9 +4,16 @@ <h3 class="modal-title" ng-show="!is_edit">Add Release</h3>
 </div>
 <div class="modal-body">
   <form role="form">
+    <div class="form-group" ng-class="{'has-error': errors.data}">
+      <label for="id_data">Data</label>
+      <input type="file" id="id_data" file-model="dataFile" onchange="angular.element(this).scope().changeName(this)">
```
F3real: Ok, I'll try it. This approach is from some example I dig out somewhere.

ui/app/js/controllers/release_new_controller.js
mozbhearsum: It is, but when a user selects the JSON file locally, there's no guarantee that they've remembered to include "name" in it. The best thing to do here is probably set an error for this field.

ui/app/js/controllers/release_new_controller.js
mozbhearsum: I think this might go away if you don't go through angular.element to call it...see if that helps. We do this sort of thing in lots of places, eg: when addRelease is called on releases.html

mozbhearsum: Turns out that Angular doesn't support onchange events to <input type="file">, per https://github.com/angular/angular.js/issues/1375, so this indirection is necessary.

Can you use 0 as the timeout instead of 500 though? The latter causes an unnecessary delay.

F3real: Yeah, I've seen that angular doesn't support it that's why I did it like this (it's kinda hacky but it works). I'll try to lower delay but it can't be zero since file won't get cached/loaded or whatever and we will just get file undefined. I can try lowering delay and see how far can I reduce it. It will probably vary a lot depending of internet speed of user, so I picked 500 ms as a safe value initially.

mozbhearsum: Hm, I thought the file was getting loaded in fillName - I see the parsing happening after an loadend event fires? I used 0 locally in my own testing, and it worked fine.

mozbhearsum:
Looks good now. Thanks @F3real!

PR #153, opened 2016-10-19
Reduce docker container from 776.9MB to 246.6MB (uncompressed)

mostlygeek:
Each command in a Dockerfile creates a new layer. To keep the final
docker image small build dependencies must be removed in the same
Dockerfile command. Removing data in separate commands does not make
the final image smaller.

This commit uses the above technique for adding and building with pip+gcc
and node+npm dependencies. It also adds a few extra cleanup commands for
files in /root/cache/.pip, /root/.npm and /tmp/phantomjs.

@bhearsum  r? please

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@ MAINTAINER bhearsum@mozilla.com
 # Node and npm are to build the frontend. nodejs-legacy is needed by this version of npm. These will get removed after building.
 # libmysqlclient-dev is required to use SQLAlchemy with MySQL, which we do in production.
 RUN apt-get -q update \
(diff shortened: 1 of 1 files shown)


mostlygeek:
@bhearsum the taskcluster failure seems to have blown up pulling down debian deps and I don't have the scope to retrigger.

mozbhearsum:
Yeah, that failure is known-intermittent, we're looking for a fix in https://bugzilla.mozilla.org/show_bug.cgi?id=1305994. I hit one issue locally, but I think it was a one-off race condition within "npm install". This looks fine to me, let's do it!

PR #154, opened 2016-10-19
Add extract command to run.sh

mozbhearsum:
This command was added to manage-db.py in #139, but in order to run it in production we need to make it accessible through the default entrypoint (run.sh). I tested this with and without OUTPUT_FILE set, and it correctly uses the default (dump.sql) when it is unset.

@JohanLorenzo - can you have a look?


mozbhearsum:
CI failures are caused by https://bugzilla.mozilla.org/show_bug.cgi?id=1305994, as usual. I retriggered the failed task in https://tools.taskcluster.net/task-inspector/#OEih07XhRsaOjLiRLZuf8g/

PR #155, opened 2016-10-20
Add change notifier for scheduled rule changes.

mozbhearsum:
Pretty straightforward, the hardest part was the tests!


mozbhearsum:
Test failure is the usual debian junk. Retriggered (and passing) version is here: https://tools.taskcluster.net/task-inspector/#KSHUUSk0TT2tSCHDk8Cc9Q/0.

auslib/db.py
```diff
@@ -1914,6 +1914,9 @@ def setupChangeMonitors(self, relayhost, port, username, password, to_addr, from
         self.rules.onInsert = bleeter
         self.rules.onUpdate = bleeter
         self.rules.onDelete = bleeter
+        self.rules.scheduled_changes.onInsert = bleeter
+        self.rules.scheduled_changes.onUpdate = bleeter
+        self.rules.scheduled_changes.onDelete = bleeter
```
nthomas-mozilla: Almost ready for a couple of loops here ? eg for table in (self.rules, ...):  for attr ('onInsert', ...)  setattr(table, attr, bleeter)

auslib/test/test_db.py
```diff
@@ -2752,20 +2756,53 @@ def testOnUpdate(self):
         def doit():
             self.db.rules.update({"rule_id": 2}, {"product": "blah"}, "bob", 1)
         mock_conn = self._runTest(doit)
+        # Updating a Rule causes its Scheduled Change to be updated as well, so we have to check both of those calls.
+        mock_conn.sendmail.assert_any_call("fake@from.com", "fake@to.com", PartialString("UPDATE"))
+        mock_conn.sendmail.assert_any_call("fake@from.com", "fake@to.com", PartialString("Row(s) to be updated as follows:"))
+        mock_conn.sendmail.assert_any_call("fake@from.com", "fake@to.com", PartialString("'product': None ---> 'blah'"))
+        mock_conn.sendmail.assert_any_call("fake@from.com", "fake@to.com", PartialString("'channel': u'release' (unchanged)"))
         mock_conn.sendmail.assert_called_with("fake@from.com", "fake@to.com", PartialString("UPDATE"))
```
nthomas-mozilla: Where we are modifying two tables it would be good to check for the table names, eg "UPDATE to rules" for the first one.

auslib/db.py
mozbhearsum: That's a good idea, let's do it!

mozbhearsum:
The latest failure is not the Debian issue (for once), but rather an issue when installing node dependencies. Retrigger is here, and passes: https://g233e3qaaaavqajnzrdk64mhcgtwkkbrv6pkjvcd2lfrt6ji.taskcluster-worker.net:32772/log/lL73mQd0ThS_fN0Txe979g

PR #156, opened 2016-10-20
Don't dump permissions or permissions history.

mozbhearsum:


nthomas-mozilla:
Looks good so far. Do we need to add an admin account so that the sample data is useful in dev ? And maybe something for the agent.

mozbhearsum:
The "balrogadmin" account gets added separately by https://github.com/mozilla/balrog/blob/master/scripts/initdb_and_run.sh. At the moment, the Agent uses that as well. Because of the uwsgi magic we use to automatically set REMOTE_USER (https://github.com/mozilla/balrog/blob/master/uwsgi/admin.dev.ini), we're forced into all local admin api access using the same account.

It's probably a good idea to try to fix this, in particular so that we catch permissions issues with the Agent in local dev. I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1313110 for this.

nthomas-mozilla:
Okie dokie, lets ship it.

PR #159, opened 2016-10-21
Migrate balrog.submitter.api to balrogclient

srfraser:
Moved from hg.m.o/build/tools to github
Basic test added
Python 3.5 compatibility

Minimal refactoring for the moment, to reduce the number of things to track during the module's location change.


python-lib/tox.ini
```diff
@@ -0,0 +1,51 @@
...
+[base]
+deps =
+    nose==1.3.0
+    rednose==0.4.1
+    coverage==3.6
+    pep8==1.4.3
+    jinja2==2.6
+    mock==1.0.1
+    requests
+    requests_mock
```
mozbhearsum: Why no pinning on these?

python-lib/tox.ini
```diff
@@ -0,0 +1,51 @@
...
+
+[testenv]
+
+commands =
+    coverage erase
+    coverage run --branch --source {toxinidir}/balrogclient {envbindir}/nosetests -v --with-xunit --rednose --force-color {toxinidir}/balrogclient {posargs}
+
+
+[testenv:py27]
+basepython = python2.7
```
mozbhearsum: Yay, 2.7 and 3.5! We may want to run against 2.6 as well, if we still use that in production anywhere...I'm trying to find out if that's the case or not.

python-lib/setup.py
```diff
@@ -0,0 +1,19 @@
+#! /usr/bin/env python
+
+from setuptools import setup, find_packages
+
+setup(
+    name="balrogclient",
+    version="0.0.1",
+    description="Mozilla RelEng Toolkit",
```
mozbhearsum: This should probably be "Balrog Admin API Client" or something like that.

python-lib/tox.ini
srfraser: Fixed, update coming.

python-lib/setup.py
srfraser: Fixed, update coming.

PR #160, opened 2016-10-24
Bug 1282891 Fallback mapping 

NinadBhat:
https://bugzilla.mozilla.org/show_bug.cgi?id=1282891


ui/app/templates/rule_modal.html
```diff
@@ -153,6 +153,15 @@ <h3 class="modal-title" ng-show="!is_edit && is_duplicate">Duplicate Existing Ru
         </div>
       </div>
     </div>
+    <div class="row">
```
mozbhearsum: This field also needs to autocomplete the mapping names, like the Mapping field does. Please avoid making an extra network request for this if possible (ie: mapping and fallbackMapping should be able to share the same data).

ui/app/templates/rule_modal.html
```diff
@@ -138,14 +141,21 @@ <h3 class="modal-title" ng-show="!is_edit && is_duplicate">Duplicate Existing Ru
...
-      <div class="col-md-6">
+      <div class="col-md-4">
         <div class="form-group" ng-class="{'has-error': errors.headerArchitecture}">
           <label for="id_headerArchitecture">Header Architecture</label>
           <input type="text" class="form-control" id="id_headerArchitecture" ng-model="rule.headerArchitecture">
           <p class="help-block" ng-show="errors.headerArchitecture">{{ errors.headerArchitecture.join(', ') }}</p>
         </div>
       </div>
-      <div class="col-md-6">
+      <div class="col-md-4">
```
mozbhearsum: I think I misspoke in my previous comment when I said "between". This has caused a three-column layout in the last row of the modal dialog, which is pretty ugly. Can you adjust this to put headerArchitecture and alias on the same row, and put comment on the final row by itself? Bonus points if comment is two columns wide.

PR #161, opened 2016-10-24
[client_view] Non-substituted url variables return empty response

njirap:
We get tens of thousands of requests per day that don't substitute
the URL variables. A route has been added that quickly returns an
empty response for these. There's no reason query the rules table,
 we'll never find anything useful in there.

Fixes bug #1282569

### DIFF
--- a/auslib/test/web/test_client.py
+++ b/auslib/test/web/test_client.py
@@ -865,6 +865,24 @@ class ClientTest(ClientTestBase):
         ret = self.client.get('/update/3/product_that_should_not_be_updated/1.0/1/p/l/a/a/a/a/update.xml')
         self.assertUpdatesAreEmpty(ret)
 
+    def testNonSubstitutedUrlVariablesReturn404(self):
+        request1 = '/update/1/%PRODUCT%/%VERSION%/%BUILD_ID%/%BUILD_TARGET%/%LOCALE%/%CHANNEL%/update.xml'
+        request2 = '/update/2/%PRODUCT%/%VERSION%/%BUILD_ID%/%BUILD_TARGET%/%LOCALE%/%CHANNEL%/%OS_VERSION%/update.xml'
+        request3 = '/update/3/%PRODUCT%/%VERSION%/%BUILD_ID%/%BUILD_TARGET%/%LOCALE%/%CHANNEL%/%OS_VERSION%/%DISTRIBUTION%/%DISTRIBUTION_VERSION%/' \
+            'update.xml'
+        request4 = '/update/4/%PRODUCT%/%VERSION%/%BUILD_ID%/%BUILD_TARGET%/%LOCALE%/%CHANNEL%/%OS_VERSION%/%DISTRIBUTION%/%DISTRIBUTION_VERSION%/' \
+            '%MOZ_VERSION%/update.xml'
+        request5 = '/update/5/%PRODUCT%/%VERSION%/%BUILD_ID%/%BUILD_TARGET%/%LOCALE%/%CHANNEL%/%OS_VERSION%/%DISTRIBUTION%/%DISTRIBUTION_VERSION%/' \
+            '%IMEI%/update.xml'
+        request6 = '/update/6/%PRODUCT%/%VERSION%/%BUILD_ID%/%BUILD_TARGET%/%LOCALE%/%CHANNEL%/%OS_VERSION%/%SYSTEM_CAPABILITIES%/%DISTRIBUTION%/' \
+            '%DISTRIBUTION_VERSION%/update.xml'
+
+        with mock.patch('auslib.web.views.client.ClientRequestView') as mock_cr_view:
+            for request in [request1, request2, request3, request4, request5, request6]:
+                ret = self.client.get(request)
+                self.assertEqual(ret.status_code, 404)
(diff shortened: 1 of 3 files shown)


auslib/web/base.py
```diff
@@ -105,3 +105,8 @@ def robots():
     view_func=ClientRequestView.as_view('clientrequest_esrnightly'),
     defaults={'queryVersion': 3},
 )
+app.add_url_rule(
+    '/update/4///////////update.xml',
+    view_func=ClientRequestView.as_view('clientrequest4_unsubstituted'),
+    defaults={'queryVersion': 4},
+)
```
mozbhearsum: This isn't like the URLs described in the bug. The ones we need to handle are:
/update/1/%PRODUCT%/%VERSION%/%BUILD_ID%/%BUILD_TARGET%/%LOCALE%/%CHANNEL%/update.xml
/update/2/%PRODUCT%/%VERSION%/%BUILD_ID%/%BUILD_TARGET%/%LOCALE%/%CHANNEL%/%OS_VERSION%/update.xml
/update/3/%PRODUCT%/%VERSION%/%BUILD_ID%/%BUILD_TARGET%/%LOCALE%/%CHANNEL%/%OS_VERSION%/%DISTRIBUTION%/%DISTRIBUTION_VERSION%/update.xml
/update/4/%PRODUCT%/%VERSION%/%BUILD_ID%/%BUILD_TARGET%/%LOCALE%/%CHANNEL%/%OS_VERSION%/%DISTRIBUTION%/%DISTRIBUTION_VERSION%/%MOZ_VERSION%/update.xml
/update/6/%PRODUCT%/%VERSION%/%BUILD_ID%/%BUILD_TARGET%/%LOCALE%/%CHANNEL%/%OS_VERSION%/%SYSTEM_CAPABILITIES%/%DISTRIBUTION%/%DISTRIBUTION_VERSION%/update.xml

And they should get routed differently - there's no point in putting them through ClientRequestView. You can probably just add the routes with @app.route (like we do for robots.txt, and then return a 404 from there. This avoids running _any_ database code at all, which should speed these queries up.

njirap:
@mozbhearsum changes made..

auslib/web/views/client.py
```diff
@@ -51,9 +51,13 @@ def getQueryFromURL(self, url):
...
-        query = self.getQueryFromURL(url)
-        self.log.debug("Got query: %s", query)
-        release, update_type = AUS.evaluateRules(query)
+        try:
+            query = self.getQueryFromURL(url)
+            self.log.debug("Got query: %s", query)
+            release, update_type = AUS.evaluateRules(query)
+        except KeyError:
+            release = None
+
```
mozbhearsum: Is there a reason you left this in? I don't think it's necessary...

auslib/test/web/test_client.py
```diff
@@ -854,3 +870,19 @@ def testSentryRealError(self):
...
+        request4 = '/update/4/%PRODUCT%/%VERSION%/%BUILD_ID%/%BUILD_TARGET%/%LOCALE%/%CHANNEL%/%OS_VERSION%/%DISTRIBUTION%/%DISTRIBUTION_VERSION%/' \
+            '%MOZ_VERSION%/update.xml'
+        request5 = '/update/5/%PRODUCT%/%VERSION%/%BUILD_ID%/%BUILD_TARGET%/%LOCALE%/%CHANNEL%/%OS_VERSION%/%DISTRIBUTION%/%DISTRIBUTION_VERSION%/' \
+            '%IMEI%/update.xml'
+        request6 = '/update/6/%PRODUCT%/%VERSION%/%BUILD_ID%/%BUILD_TARGET%/%LOCALE%/%CHANNEL%/%OS_VERSION%/%SYSTEM_CAPABILITIES%/%DISTRIBUTION%/' \
+            '%DISTRIBUTION_VERSION%/update.xml'
+
+        for request in [request1, request2, request3, request4, request5, request6]:
+            ret = self.client.get(request)
+            self.assertUpdatesAreEmpty(ret)
```
mozbhearsum: I think it's better to do these tests without error handlers, actually. We currently eat 404s, so you can't check that this is working. AFAICT, the routes you added aren't working, but because you're testing for empty updates you don't know that. Try moving this to ClientTest and checking for 404 instead - that should uncover the issue.

njirap: hi @mozbhearsum might you be talking about something like L818 - L820?

auslib/web/views/client.py
njirap: not having it raises a "release referenced before assignment" Error

auslib/test/web/test_client.py
mozbhearsum: I'm sorry, I totally missed those! You're right - you've got enough test coverage.

auslib/web/views/client.py
mozbhearsum: Hm, I can't reproduce this....all of tests still pass even if I remove this block. Are you still getting an error with the latest code without it?

njirap: Actually no..this block was needed for the first version of the fix that I had..

auslib/test/web/test_client.py
mozbhearsum: Great idea!

PR #162, opened 2016-10-24
Gracefully handle missing addons section

mozbhearsum:
I'm pretty sure this is a bug that's existed since the SystemAddons blob was created - it's always supported "addons" not being specified, and always tried look up whether or not a platform has addons with "hasUpdates". We only uncovered it recently for two reasons: 1) We started using blobs with "uninstall" set instead of "addons", and 2) Sentry :)

This is an aggressively conservative patch to make sure we don't do any lookups on None.


auslib/blobs/systemaddons.py
```diff
@@ -12,22 +12,24 @@ def __init__(self, **kwargs):
...
             return "default"
         raise BadDataError("No platform '%s' or default in addon '%s'",
                            platform, addon)
 
     def getPlatformData(self, addon, platform):
         platform = self.getResolvedPlatform(addon, platform)
         try:
-            return self['addons'][addon]['platforms'][platform]
+            return self.get("addons", {}).get(addon, {}).get("platforms", {}).get(platform)
         except KeyError:
```
rail: I think you can remove the try/except wrapper with the change above.

PR #163, opened 2016-10-28
bug 1170797: "unhang" release upload screen if no file provided

njirap:


mozbhearsum:
This is needed in the release edit controller as well, looks fine otherwise though. I discovered a related bug (https://bugzilla.mozilla.org/show_bug.cgi?id=1314040) while verifying this that you may be interested in as well.

PR #165, opened 2016-10-31
Move Scheduled Changes conditions to their own table

mozbhearsum:
#151 has some background on this.

I think this has cleaned things up nicely. It's mostly involved moving code around, with some new glue to deal with extracting or modifying the new table in sync with the primary Scheduled Changes one. I tried to keep all of this in db.py, but I couldn't find a non-terrible way to grab conditions table history there. The best I could come up with was monkey-patching the History class. In any case, the web APIs have not changed at all.

The changes to ScheduledChangesTable are probably going to be hard to read as a diff, it may be easier if you look at the new file for that part instead. While doing this, I discovered that we weren't fully validating Scheduled Changes during updates, and ended up reworking all of the validation code that was in insert() into a new validate() method, which is also called from update(). The other changes to select/insert/update/delete are mostly around combining or separating the conditions columns from the others, and passing them along correctly.

I had to move the AUSTable dryrun support from #151 to this because ConditionsTable doesn't override those methods. This let me get rid of dryrun handling in ScheduledChangesTable, because AUSTable will now do the aborting instead.


auslib/test/admin/views/test_rules.py
```diff
@@ -991,16 +1016,22 @@ def testRevertScheduledChange(self):
...
             "base_backgroundRate": 100, "base_product": "ff", "base_mapping": "def", "base_update_type": "minor", "base_version": None,
             "base_buildTarget": None, "base_alias": None, "base_channel": None, "base_buildID": None, "base_locale": None, "base_osVersion": None,
             "base_distribution": None, 'base_fallbackMapping': None, "base_distVersion": None, "base_headerArchitecture": None, "base_comment": None,
-            "base_whitelist": None, "base_data_version": None, "base_systemCapabilities": None, "telemetry_product": None, "telemetry_channel": None,
-            "telemetry_uptake": None,
+            "base_whitelist": None, "base_data_version": None, "base_systemCapabilities": None
         }
-        self.assertEquals(got, expected)
+        self.assertEquals(db_data, expected)
+        self.assertEquals(dbo.rules.scheduled_changes.conditions.history.t.count().execute().first()[0], 9)
```
JohanLorenzo: I'm not sure. Why do we have 9 changes in the history table?

auslib/db.py
```diff
@@ -319,6 +319,10 @@ def insert(self, changed_by=None, transaction=None, **columns):
 
            @rtype: sqlalchemy.engine.base.ResultProxy
         """
+        if dryrun:
```
mozbhearsum: I'm not sure what you mean? My top comment was just saying that I copied this from #151, because its needed here.

I'm not sure I see the value in moving this into prepareInsert, but maybe I'm missing something?

auslib/db.py
```diff
@@ -854,80 +896,98 @@ def insert(self, changed_by, transaction=None, dryrun=False, **columns):
...
-            new_row.update(self._prefixColumns(what))
-            base_table_where = {pk: new_row["base_%s" % pk] for pk in self.base_primary_key}
-            if new_row.get("base_data_version"):
-                self.baseTable.update(base_table_where, new_row, changed_by, new_row["base_data_version"], transaction=transaction, dryrun=True)
-            else:
-                self.baseTable.insert(changed_by, transaction=transaction, dryrun=True, **new_row)
+            # Before validation, we need to create the new version of the
+            # Scheduled Change by combining the old one with the new data.
+            sc_columns, condition_columns = self._splitColumns(row)
+            base_columns = {}
```
mozbhearsum: For which part?

auslib/db.py
JohanLorenzo: Okay, that unrelated then :) 

I was thinking of making the early return of dry-run happen just before the actual transaction happens. Something like:

```
def insert(self, changed_by=None, transaction=None, dryrun=False, **columns):
    if self.history and not changed_by: 
         raise ValueError("changed_by must be passed for Tables that have history") 

    if dryrun:
         self.log.debug("In dryrun mode, no transaction executed...")
         return

      if transaction:
            return self._prepareInsert(transaction, changed_by, **columns)
      else:
           with AUSTransaction(self.getEngine()) as trans:
              return self._prepareInsert(trans, changed_by, **columns)
```

Would it provide some value?

JohanLorenzo: Okay, that unrelated then :) 

I was thinking of making the early return of dry-run happen just before the actual transaction happens. Something like:

```
def insert(self, changed_by=None, transaction=None, dryrun=False, **columns):
    if self.history and not changed_by: 
         raise ValueError("changed_by must be passed for Tables that have history") 

    if dryrun:
         self.log.debug("In dryrun mode, no transaction executed...")
         return

      if transaction:
            return self._prepareInsert(transaction, changed_by, **columns)
      else:
           with AUSTransaction(self.getEngine()) as trans:
              return self._prepareInsert(trans, changed_by, **columns)
```

Would it provide some value?

auslib/db.py
JohanLorenzo: For that line, to the 3 next ones. I was thinking of:

```python
base_columns = {col: value for col, value in sc_columns.items() if col.startswith("base_") and value is not None}
```

auslib/db.py
mozbhearsum: Oh, I see. I agree it's better to do this after the other checks - will do!

auslib/db.py
mozbhearsum: Hm, I did this, but then realized it might be better to combine the two existing loops into one - since the purpose of both is to get the new base_columns. The first loop (or dict comprehension, if we use that) is really only an intermediate state. What do you think of:
```
            base_columns = {}
            for col in sc_columns:
                if not col.startswith("base_"):
                    continue
                base_col = col.replace("base_", "")
                if base_col in base_what:
                    base_columns[base_col] = base_what[base_col]
                elif sc_columns.get(col):
                    base_columns[base_col] = sc_columns[col]
```

JohanLorenzo: LGTM!

mozbhearsum:
@JohanLorenzo - the latest push should address all of your comments. I discovered that I could remove a few more lines of code from .update() as well, so I did that (I had originally written code to pull out existing conditions before I overrode select(), but it's no longer necessary now that select() returns conditions).

auslib/admin/views/scheduled_changes.py
```diff
@@ -149,13 +146,27 @@ def get(self, sc_id):
...
+        # so we do it here instead.
         revisions = self.table.scheduled_changes.history.select(
             where=[self.table.scheduled_changes.history.sc_id == sc_id,
                    self.table.scheduled_changes.history.data_version != null()],
             limit=limit,
             offset=offset,
             order_by=[self.table.scheduled_changes.history.timestamp.asc()],
         )
+        # There's a big 'ol assumption here that the primary Scheduled Changes
+        # table and the conditions table always keep their data version in sync.
```
mozbhearsum: I guess there's two things that I'm concerned about. One is that I've done something wrong in this patch, and there's a case where they will get out of sync as part of the normal operations. The other is that some future code change may change things in such a way that normal operations would cause them to get out of sync. Raising an execption is actually a clever idea - if you do the data_version checking within the transaction, and raise on mismatch, the transaction will be aborted before it gets committed. It's still not a great user experience, but it's better than letting the db get into a bad state?

Manual db changes or some sort of crazy problem with transactions certainly could break things as well, but I'm not sure I'm terribly concerned about them. They're pretty close to "everything is on fire" territory for me, I guess.

auslib/admin/views/scheduled_changes.py
```diff
@@ -149,13 +146,27 @@ def get(self, sc_id):
...
+        # history in both and return the combined version.
+        # This is done by the database layer for non-history parts of Scheduled Changes, but
+        # that's not feasible for History due to the inheritance structure of the tables,
+        # so we do it here instead.
         revisions = self.table.scheduled_changes.history.select(
             where=[self.table.scheduled_changes.history.sc_id == sc_id,
                    self.table.scheduled_changes.history.data_version != null()],
             limit=limit,
             offset=offset,
             order_by=[self.table.scheduled_changes.history.timestamp.asc()],
```
mozbhearsum: Will do.

auslib/db.py
```diff
@@ -1444,47 +1506,48 @@ def update(self, where, what, changed_by, old_data_version, transaction=None, dr
...
+                        # to the original blob
+                        unified_blob = dictdiffer.patch(m.unified_patches, ancestor_blob)
+                        # converting the resultant dict into a blob and then
+                        # converting it to JSON
+                        what['data'] = createBlob(unified_blob).getJSON()
+                        # we want the data_version for the dictdiffer.merged blob to be one
+                        # more than that of the latest blob
+                        tip_data_version = tip_release['data_version']
+                        super(Releases, self).update(where={"name": name}, what=what, changed_by=changed_by, old_data_version=tip_data_version,
+                                                     transaction=transaction)
```
mozbhearsum: Ah, indeed. Good catch.

auslib/test/admin/views/test_rules.py
```diff
@@ -665,51 +665,66 @@ class TestRuleScheduledChanges(ViewTest):
 
     def setUp(self):
         super(TestRuleScheduledChanges, self).setUp()
+        # TODO: make sure history tables return conditions columns too
```
mozbhearsum: I think this is covered in testGetScheduledChangeHistoryRevisions actually? This was a reminder to myself mid-refactor that I forgot to remove, heh.

nthomas-mozilla: Oh, that's fine.

auslib/test/admin/views/test_rules.py
```diff
@@ -991,16 +1016,22 @@ def testRevertScheduledChange(self):
...
-            "base_whitelist": None, "base_data_version": None, "base_systemCapabilities": None, "telemetry_product": None, "telemetry_channel": None,
-            "telemetry_uptake": None,
+            "base_whitelist": None, "base_data_version": None, "base_systemCapabilities": None
         }
-        self.assertEquals(got, expected)
+        self.assertEquals(db_data, expected)
+        self.assertEquals(dbo.rules.scheduled_changes.conditions.history.t.count().execute().first()[0], 9)
+        cond = dbo.rules.scheduled_changes.conditions.t.select().where(dbo.rules.scheduled_changes.conditions.sc_id == 3).execute().fetchall()
+        self.assertEquals(len(cond), 1)
+        cond_expected = {"sc_id": 3, "data_version": 3, "when": 2000000, "telemetry_product": None, "telemetry_channel": None, "telemetry_uptake": None}
```
mozbhearsum: I ended up doing this.

PR #166, opened 2016-10-31
update balrog README and documentation .

NinadBhat:
https://bugzilla.mozilla.org/show_bug.cgi?id=1282088


mozbhearsum:
This is a very promising start. It's so cool to have standard looking Python docs for Balrog! I'm mostly reviewing this by building it locally and looking at the output, so I'll leave comments here instead of inline.
* You seem to have trimmed some content in various places. Eg: the Overview section is much smaller, nothing about the limitations/caveats for running tox, there's no link to an example public interface request, no list of permissions that are available, nothing about the Agent. Was this on purpose, or an oversight?
* I should've mentioned this up front, but anything that relates to our _specific_ infrastructure (basically, anything talking about deployment or domains that our production instances are hosted on) doesn't belong here. These are operational docs, and will either remain on wiki.m.o or move to an internal page.
* There's a lot of small grammar errors and typos. If you haven't already, please make an edit pass to fix whatever you find. Don't worry if it's not perfect - I'll do a final edit pass myself before merging.
* Did you add https://wiki.mozilla.org/Balrog/Blobs somewhere? I couldn't find it.
* It would be nice to have a full TOC on the left panel. Eg: I should be able to directly browse to the "Code Overview" even if I'm on the "Database Model" page.
* I think it's better to put everything from "Installing Balrog" in the "Contributing" section. The contents of that page are really about how to use Balrog for local development, not how to install it for production usage.
* Along with the column types, all of the rule columns should be listed on the "Database Model" page. The detailed descriptions of those columns can stay in "Rules Table", though.
* Information about Scheduled Changes needs to be added to the Rules section of "Database Model".
* I like the "Rules Table" page overall, but there's a couple of small improvements that would be good:
** It doesn't appear to be part of the main TOC right now - there's no "next" or "previous" section links.
** I don't think bullet points are the best way to organize things within each section. Descriptions, matching logic, and examples can simply be written into a paragraph per section instead.
* In the "Admin app" section:
** The first paragraph says "this page documents all of the available endpoints", but then links to another page for the endpoints. I think it's best if the endpoints go right on this page.
** Just kill the "UI Use Cases" section. I don't think they're useful anymore.
* Is it possible to generate autodocs for the code in addition to doing this? I'd want them to be off to the side somewhere, away from the User docs.

NinadBhat:
>  I don't think bullet points are the best way to organize things within each section. Descriptions, matching logic, and examples can simply be written into a paragraph per section instead.

Writing it as paragraph might make it harder to read. I think it is easier to read it if data in form of bulleted point. But will change it if still needed.

> Is it possible to generate autodocs for the code in addition to doing this? I'd want them to be off to the side somewhere, away from the User docs.

We can make autodocs in a different directory. But, should the autodoc also shown in readthedocs ?

NinadBhat:
**Things remaining**  :-

-  Still haven't decided what should be the order of pages.
-  Adding example of superblob

mozbhearsum:
* The entire "Infrastructure" section can go - everything in there is specific to our installation, and doesn't belong in general docs.
* The "Admin API" section is not properly organized - it appears that all endpoints are in the "Rules" subsection - even those for Releases, Permissions, or History.
* I'm unable to find the Blob and Agent docs anywhere. Did you forget to add links to them? I see this in the output:
```
/home/bhearsum/repos/balrog/docs/source/index.rst:14: WARNING: toctree contains reference to nonexisting document u'blobs'
/home/bhearsum/repos/balrog/docs/source/index.rst:14: WARNING: toctree contains reference to nonexisting document u'gmp_example'
/home/bhearsum/repos/balrog/docs/source/index.rst:14: WARNING: toctree contains reference to nonexisting document u'balrog_agent'
```
* I'd still prefer a more detailed TOC on the left - is that possible? Eg: all subsections should always be shown - not just the ones for the current section.
* Regarding the bullet points on "Rules Table" - my issue with them is that the page is mostly just whitespace, and it makes for a lot of unnecessary scrolling. Now that you've added a table back to the "Rules" subsection of the "Database Model" page, how do you feel about just putting the "Rules Table" content back into it, like we have it on the wiki?
* Re: autodocs - yes, they should go on RTD. Ideally, the autodocs and the docs you've added here will be part of the same site. Eg: the final section of this could be autodocs of the entire codebase.

mozbhearsum:
Content-wise, I think you're good. I'll have a final read and comparison vs. the wiki of the content tomorrow to be sure.

There's still a few remaining layout issues to take care of, then we can get this merged:
* The Admin API section is still not organized properly - all endpoints show up under the "Rules" header. Releases and Users should have their own section, and you can put the csrf_token and history views in an "Other" section.
* The example of a GMP blob doesn't deserve its own section. That should go in the "GMP/System Addons" part of the "Blobs" section.
* The autodocs link disappears when I navigate to another section. Eg: if I'm looking at "Blobs", I can't see "Developer Documentation" in the TOC. Is there a way to avoid duplication of the TOC so this isn't an issue in the future?
* What's the point in the "Subpackages" and  "Submodules" parts of the autodocs? They don't appear to have any content, or any subsections. They should probably be removed, and just have the subpackages/modules listed directly underneath their containing package.

NinadBhat:
> The example of a GMP blob doesn't deserve its own section. That should go in the "GMP/System Addons" part of the "Blobs" section.

As adding GMP blob example to  "GMP/System Addons"  makes the page longer. What I did was remove the GMP blob example from TOC but they can still look at it by clicking the link.

 >The autodocs link disappears when I navigate to another section. Eg: if I'm looking at "Blobs", I can't see "Developer Documentation" in the TOC. Is there a way to avoid duplication of the TOC so this isn't an issue in the future?

I don't seem to face the same issue as you.
Can you try
`make clean`
`make html`
to see if you still face the same issue and let me know.
And what do  you mean by **"duplication of the TOC"**

>What's the point in the "Subpackages" and "Submodules" parts of the autodocs? They don't appear to have any content, or any subsections. They should probably be removed, and just have the subpackages/modules listed directly underneath their containing package.

So when sphinx looks through auslib directory it includes all the .py files in "submodules" and
all the directories in "subpackages"  so will searching in auslib it includes "auslib.admin" as a package because  admin is a directory in auslib and "auslib.AUS" as modules because of AUS.py file in auslib

I have improved the formating if there is a problem let me know.

mozbhearsum:
Looks like you're right about the TOC - it looks fine after doing a fresh build. I had thought that maybe it was listed multiple places in the source...but I see that's not the case! The GMP blob looks fine now too.

PR #167, opened 2016-11-01
Add UserRoles table + apis

mozbhearsum:
Not much to say about this one, it's a simple table with a simple CRD API. During the proposal stage of multiple signoffs, a couple of people suggested that we may want to tie Roles and Permissions together directly at some point (ie: grant permissions to Roles, and let users inherit those instead of being granted Permissions). That's out of scope for now, but it's part of the reason I made this table owned by Permissions instead of living beside it.

### DIFF
--- a/auslib/admin/views/permissions.py
+++ b/auslib/admin/views/permissions.py
@@ -113,6 +113,9 @@ class UserRoleView(AdminView):
 
     @requirelogin
     def _put(self, username, role, changed_by, transaction):
+        # These requests are idempotent - if the user already has the desired role,
+        # no change needs to be made. Because of this there's also no reason to
+        # return an error.
         r = dbo.permissions.user_roles.select({"username": username, "role": role}, transaction=transaction)
         if r:
             return Response(status=200, response=json.dumps({"new_data_version": r[0]["data_version"]}))
--- a/auslib/admin/views/permissions.py
+++ b/auslib/admin/views/permissions.py
@@ -101,7 +101,7 @@ class UserRolesView(AdminView):
     """/users/:username/roles"""
 
     def get(self, username):
-        roles = dbo.permissions.user_roles.getRoles(username)
+        roles = dbo.permissions.getUserRoles(username)
         if roles:
             return jsonify({"roles": roles})
         else:
@@ -120,12 +120,12 @@ class UserRoleView(AdminView):
         if r:
             return Response(status=200, response=json.dumps({"new_data_version": r[0]["data_version"]}))
 
-        dbo.permissions.user_roles.insert(changed_by, transaction, username=username, role=role)
+        dbo.permissions.grantRole(username, role, changed_by, transaction)
         return Response(status=201, response=json.dumps({"new_data_version": 1}))
 
     @requirelogin
     def _delete(self, username, role, changed_by, transaction):
-        if role not in dbo.permissions.user_roles.getRoles(username):
+        if role not in dbo.permissions.getUserRoles(username):
             return Response(status=404)
 
         form = DbEditableForm(request.args)
@@ -133,6 +133,5 @@ class UserRoleView(AdminView):
             self.log.warning("Bad input: %s", form.errors)
             return Response(status=400, response=json.dumps(form.errors))
 
-        dbo.permissions.user_roles.delete(where={"username": username, "role": role}, changed_by=changed_by,
-                                          old_data_version=form.data_version.data, transaction=transaction)
+        dbo.permissions.revokeRole(username, role, changed_by=changed_by, old_data_version=form.data_version.data, transaction=transaction)
         return Response(status=200)
--- a/auslib/admin/views/permissions.py
+++ b/auslib/admin/views/permissions.py
@@ -4,7 +4,7 @@ from flask import request, Response, jsonify
 
 from auslib.global_state import dbo
 from auslib.admin.views.base import requirelogin, AdminView
-from auslib.admin.views.forms import NewPermissionForm, ExistingPermissionForm
+from auslib.admin.views.forms import NewPermissionForm, ExistingPermissionForm, DbEditableForm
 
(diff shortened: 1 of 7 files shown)


auslib/admin/views/permissions.py
```diff
@@ -95,3 +95,41 @@ def _delete(self, username, permission, changed_by, transaction):
...
+
+
+class UserRoleView(AdminView):
+    """/users/:username/roles/:role"""
+
+    @requirelogin
+    def _put(self, username, role, changed_by, transaction):
+        r = dbo.permissions.user_roles.select({"username": username, "role": role}, transaction=transaction)
+        if r:
+            return Response(status=200, response=json.dumps({"new_data_version": r[0]["data_version"]}))
```
nthomas-mozilla: Could you please add a comment to the effect of 'adding a role which already exists is treated as a no-op'.

auslib/admin/views/permissions.py
```diff
@@ -95,3 +95,41 @@ def _delete(self, username, permission, changed_by, transaction):
...
+
+        dbo.permissions.user_roles.insert(changed_by, transaction, username=username, role=role)
+        return Response(status=201, response=json.dumps({"new_data_version": 1}))
+
+    @requirelogin
+    def _delete(self, username, role, changed_by, transaction):
+        if role not in dbo.permissions.user_roles.getRoles(username):
+            return Response(status=404)
+
+        form = DbEditableForm(request.args)
```
mozbhearsum: Because username and role are part of the URL, there's no need to create a Form (they don't show up in request.args/data). You can see similar with Permissions, where "options" is the only thing that's part of the form. This code can't fire without them being set because of the routing!

auslib/db.py
```diff
@@ -1617,8 +1657,10 @@ def assertOptionsExist(self, permission, options):
                 raise ValueError('Unknown option "%s" for permission "%s"' % (opt, permission))
 
     def getAllUsers(self, transaction=None):
-        res = self.select(columns=[self.username], distinct=True, transaction=transaction)
-        return [r['username'] for r in res]
+        users = self.select(columns=[self.username], distinct=True, transaction=transaction)
+        users = set([u["username"] for u in users])
+        users.update(self.user_roles.getAllUsers(transaction))
```
mozbhearsum: Good call. There's no use case I can think of for having a Role without a Permission. Such a User wouldn't be able to do any harm (Signoffs are likely to require a "signoff" permission), but it's a weird state to be in. I'll look at this, and at cascading the deletion of a User's last permission.

auslib/admin/views/permissions.py
nthomas-mozilla: 😊

mozbhearsum:
I had to rework a little bit to add better checks between permissions & roles. Now, Permissions has methods to grant/revoke/get roles instead of callers acting directly on UserRoles. This is because we need to select from the Permissions table to see if the user exists before granting them a Role. bc89c19 takes care of that refactor, and the most recent commit implements the checks you requested.

auslib/test/test_db.py
```diff
@@ -2629,12 +2649,50 @@ def testGrantPermissionsUnknownOption(self):
...
     def testRevokePermission(self):
         self.permissions.delete({"username": "bob", "permission": "release"}, changed_by="bill", old_data_version=1)
         query = self.permissions.t.select().where(self.permissions.username == "bob")
         query = query.where(self.permissions.permission == "release")
         self.assertEquals(len(query.execute().fetchall()), 0)
 
+    def testRevokeRoleWithPermission(self):
+        self.permissions.revokeRole("bob", "releng", "bill", old_data_version=1)
+        got = self.user_roles.t.select().where(self.user_roles.username == "bob").execute().fetchall()
+        self.assertEquals(len(got), 1)
```
nthomas-mozilla: Could make this test stronger by testing for a value of 'dev'.

auslib/test/test_db.py
```diff
@@ -2629,12 +2649,50 @@ def testGrantPermissionsUnknownOption(self):
...
+
+    def testRevokeRoleWithoutPermission(self):
+        self.assertRaises(PermissionDeniedError, self.permissions.revokeRole, username="bob", role="releng", changed_by="kirk", old_data_version=1)
+
+    def testRevokingPermissionAlsoRevokeRolesRole(self):
```
nthomas-mozilla: Looks like a typo with the extra Role at the end there.

auslib/test/test_db.py
```diff
@@ -2706,6 +2764,33 @@ def testHasPermissionNotAllowedByAction(self):
...
+
+    def testGetUserRolesNonExistantUser(self):
+        got = self.permissions.getUserRoles("kirk")
+        self.assertEquals(got, [])
+
+    def testUpdateUserRole(self):
+        self.assertRaises(AttributeError, self.user_roles.update, {"username": "bob"}, {"role": "relman"}, "bill", 1)
+
+
+class TestUserRolesTable(unittest.TestCase, MemoryDatabaseMixin):
```
nthomas-mozilla: All the tests moved away from this class, so no longer required ?

auslib/test/test_db.py
mozbhearsum: Good point, will do.

auslib/test/test_db.py
mozbhearsum: Ah, yes. Forgot to get rid of this.

auslib/admin/views/permissions.py
```diff
@@ -95,3 +95,43 @@ def _delete(self, username, permission, changed_by, transaction):
...
+            return jsonify({"roles": roles})
+        else:
+            return Response(status=404, response="No roles found for user")
+
+
+class UserRoleView(AdminView):
+    """/users/:username/roles/:role"""
+
+    @requirelogin
+    def _put(self, username, role, changed_by, transaction):
```
JohanLorenzo: Only allowing to change roles via PUT (and not POST) is a great idea!

PR #168, opened 2016-11-01
Fix db creation when sample data doesn't match current schema version.

mozbhearsum:
I discovered this small bug after recreating my database today. The current sample data is from version 14, but version 16 of the schema landed last week. This means that my database ends up on version 14, but with code that requires version 16. "upgrade" is a no-op if the database is on the latest version, so this is safe to run even if the sample data is at version 16.

@rail - can you sanity check this?


rail:
:shipit:

PR #169, opened 2016-11-02
Test  template page for docs

NinadBhat:
What I have done is written docstrings in rst format.
Every change in docstring will be updated in html by running `make html`


mozbhearsum:
This is a bit better, since the TOC doesn't mention classes, but I still see them in the body of the docs. If I'm a user of Balrog (not a developer), this is confusing and unnecessary. It looks like there's no way to keep these docs right beside the code without making things worse for users, so I'd rather not do this. Thanks for investigating this option, but I think your initial work in #166 is going to be a better approach - let's continue talking there.

PR #174, opened 2016-11-11
Added shutoff rules

nurav:


auslib/AUS.py
```diff
@@ -55,51 +55,63 @@ def __init__(self):
...
-                    return blob, rule['update_type']
-
-                self.log.debug("No fallback releases. Request was dropped")
+        is_shutoff, shutoff_mapping = dbo.shutoffs.getShutoff(product=updateQuery.get('product'),
+                                                              channel=updateQuery.get('channel'))
+        if is_shutoff:
+            self.log.debug("Updates are shutoff for this query, returning mapped empty release %s", shutoff_mapping)
+            release = dbo.releases.getReleases(name=shutoff_mapping, limit=1)[0]
+            blob = release['data']
+            if not blob.shouldServeUpdate(updateQuery):
```
mozbhearsum: I think you've got this code more or less in the right place, but it might be easier to read if the shutoff check was its own method that ClientRequestView called. I'm also not sure if calling shouldServeUpdate is the right thing to do here...what would actually be returned to the client if shouldServeUpdate returns False?

auslib/blobs/superblob.py
```diff
@@ -33,13 +33,22 @@ def getInnerHeaderXML(self, updateQuery, update_type, whitelistedDomains, specia
...
-        if not self.get("products"):
+        if self.get("blobs"):
             revision = self['revision']
             # In case of systemaddons superblob
             return '    <addons revision=\"%i\">' % (revision)
-        else:
+        elif self.get("products"):
             # In case of GMP superblob
             return '    <addons>'
+        else:
```
mozbhearsum: Is this case even possible? I'm pretty sure either blobs or products must be defined in a SuperBlob.

auslib/db.py
```diff
@@ -1764,6 +1765,56 @@ def _putWatchdogValue(self, changed_by, value, where=None, transaction=None):
...
 
+class Shutoffs(AUSTable):
+    def __init__(self, db, metadata, dialect):
+        self.table = Table('shutoffs', metadata,
+                           Column('shutoff_id', Integer, primary_key=True,
+                                  autoincrement=True),
+                           Column('product', String(15), nullable=False),
+                           Column('channel', String(75)),
+                           Column('mapping', String(100)),
+                           UniqueConstraint('product', 'channel')
```
mozbhearsum: As discussed on IRC, using a composite key of product+channel is probably the best approach.

auslib/db.py
```diff
@@ -1764,6 +1765,56 @@ def _putWatchdogValue(self, changed_by, value, where=None, transaction=None):
...
+                           )
+        AUSTable.__init__(self, db, dialect, history=False, versioned=False)
+
+    def getShutoff(self, transaction=None, **columns):
```
mozbhearsum: I don't think there's any reason to catch extra kwargs here. Is there a reason you're doing this instead of having "product" and "channel" as keyword arguments?

If you were overriding select() you'd have to do this, but since you're writing a custom method there's no reason you need to match select()'s interface.

auslib/db.py
mozbhearsum: I'm also wondering if there's a use case for mapping to be NULL. If we can't think of one, it's probably best to set nullable=False for it.

auslib/db.py
```diff
@@ -1764,6 +1765,56 @@ def _putWatchdogValue(self, changed_by, value, where=None, transaction=None):
...
+        empty release"""
+        # Since we are able to shutoff entire products without specifying a
+        # channel, we first get all shutoffs for a product and check if any of
+        # them has 'channel' set to None
+        matched_shutoffs = self.select(where=[self.product == columns.get('product')],
+                                       transaction=transaction)
+        for shutoff in matched_shutoffs:
+            if shutoff.get('channel') is None or shutoff.get('channel') == columns.get('channel'):
+                return (True, shutoff.get('mapping'))
+        return (False, None)
```
mozbhearsum: If mapping ends up being NOT NULL, you can simplify this interface and just return the blob name, or None if there's no shutoff.

auslib/db.py
```diff
@@ -1764,6 +1765,56 @@ def _putWatchdogValue(self, changed_by, value, where=None, transaction=None):
...
+        if not self.db.hasPermission(changed_by, "shutoff", "unset", columns['product'], transaction=transaction):
+            raise PermissionDeniedError("%s can't unset the shutoff rule on product: %s" % (changed_by, columns['product']))
+
+        columns['channel'] = columns['channel'] or None
+        where = [self.product == columns.get('product'),
+                 self.channel == columns.get('channel')]
+        super(Shutoffs, self).delete(changed_by=changed_by, where=where, transaction=transaction)
+
+    def getOrderedShutoffs(self, transaction=None):
+        """Returns all of the shutoffs, sorted in ascending order"""
```
mozbhearsum: I don't see this method ever getting used - is there a reason it exists?

auslib/db.py
```diff
@@ -1764,6 +1765,56 @@ def _putWatchdogValue(self, changed_by, value, where=None, transaction=None):
...
+    def __init__(self, db, metadata, dialect):
+        self.table = Table('shutoffs', metadata,
+                           Column('shutoff_id', Integer, primary_key=True,
+                                  autoincrement=True),
+                           Column('product', String(15), nullable=False),
+                           Column('channel', String(75)),
+                           Column('mapping', String(100)),
+                           UniqueConstraint('product', 'channel')
+                           )
+        AUSTable.__init__(self, db, dialect, history=False, versioned=False)
```
mozbhearsum: Did we talk at all about whether or not history should be enabled for this table? My gut says it should...

I think leaving versioning off is probably OK, since operations are idempotent.

auslib/db.py
```diff
@@ -1764,6 +1765,56 @@ def _putWatchdogValue(self, changed_by, value, where=None, transaction=None):
...
+        # channel, we first get all shutoffs for a product and check if any of
+        # them has 'channel' set to None
+        matched_shutoffs = self.select(where=[self.product == columns.get('product')],
+                                       transaction=transaction)
+        for shutoff in matched_shutoffs:
+            if shutoff.get('channel') is None or shutoff.get('channel') == columns.get('channel'):
+                return (True, shutoff.get('mapping'))
+        return (False, None)
+
+    def insert(self, changed_by, transaction=None, **columns):
```
mozbhearsum: Please be sure match the full interface of the base class method on this and delete, including accepting dryrun.

auslib/migrate/versions/016_add_shutoffs.py
```diff
@@ -0,0 +1,22 @@
...
+)
+
+
+def upgrade(migrate_engine):
+    metadata.bind = migrate_engine
+    metadata.create_all()
+
+
+def downgrade(migrate_engine):
+    metadata.bind = migrate_engine
```
mozbhearsum: I don't think this downgrade method does what you want...it appears to be creating the table instead of dropping it!

auslib/test/test_db.py
```diff
@@ -2729,6 +2729,89 @@ def testInitAndIncrementValue(self):
         self.assertEqual(entry['watchdog'], 2)
 
 
+class TestShutoffs(unittest.TestCase, MemoryDatabaseMixin):
```
mozbhearsum: These tests will need a few updates after the schema changes. Please make sure you explicitly test cases where channel is "*".

auslib/web/views/client.py
```diff
@@ -57,37 +57,38 @@ def get(self, **url):
...
-                    if not response_release:
-                        continue
-
-                    response_blobs.append({'product_query': product_query,
-                                           'response_release': response_release,
-                                           'response_update_type': response_update_type})
-            elif response_blob_names:
-                for blob_name in response_blob_names:
-                    # if we have a SuperBlob of systemaddons, we process the response products and
+            if release.get('schema_version') == 4000:  # if we have a Superblob
```
mozbhearsum: What's the reasoning behind this special case? This goes against the data driven approach I've been trying to keep us on.

auslib/db.py
nurav: I wrote it this way to make it easier to add new fields. Will change it.

auslib/db.py
nurav: It's going to be used for the API view.

auslib/AUS.py
nurav: Since 'mapping' can map to any release, someone might accidentally set it to something that cannot be served. Do we need to safeguard against that?

mozbhearsum:
I'm going to close this for now because we're probably going to be making significant changes to the client response in the coming months. I think it will be better to wait until after that's done, or at least detailed, to continue this. This is definitely a good base to start from though!

PR #176, opened 2016-11-11
Bug 1316929 - temporarily whitelist stage bucket to allow testing of …

MihaiTabara:
…nightlies Tier-2. r=bhearsum

### DIFF
--- a/uwsgi/admin.wsgi
+++ b/uwsgi/admin.wsgi
@@ -15,6 +15,7 @@ DOMAIN_WHITELIST = {
     "clients2.googleusercontent.com": ("Widevine",),
     "redirector.gvt1.com": ("Widevine",),
     "ftp.mozilla.org": ("SystemAddons",),
+    "bucketlister-delivery.stage.mozaws.net": ("Fennec-tcmigration", "Firefox-tcmigration"),
 }
 
 # Logging needs to be set-up before importing the application to make sure that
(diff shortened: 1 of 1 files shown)


mozbhearsum:
@mostlygeek - this PR is a hack I suggested that allows us to whitelist a domain in a way that doesn't affect production products. It's really not great though - I'd much prefer that we only modify the whitelist for stage. Is there a way to detect at runtime if we're in stage or prod? I was thinking maybe we could check some environment variable that would have a value of "stage" or "prod" or something? Then we could make adjustments to the whitelist based on that value.

mostlygeek:
if the app accepted an ENV environment variable we can pass through "stage" or "prod".

mozbhearsum:
We can totally do that - do you have a preference for what it's called?

uwsgi/admin.wsgi
```diff
@@ -15,6 +15,7 @@ DOMAIN_WHITELIST = {
     "clients2.googleusercontent.com": ("Widevine",),
     "redirector.gvt1.com": ("Widevine",),
     "ftp.mozilla.org": ("SystemAddons",),
+    "bucketlister-delivery.stage.mozaws.net": ("Fennec-tcmigration", "Firefox-tcmigration"),
```
mozbhearsum: Since @mostlygeek has offered to add an env variable for us, we can do better! Instead of this, add this new domain in an 'if os.environ.get("XXXXXXX") == "stage"' block. Since it will only apply to stage, you don't need to use fake product names either.

MihaiTabara: thanks both for this!
@mostlygeek what should I call that env var?

PR #178, opened 2016-11-14
Bug 1316929 - whitelist nightlies stage buckets for public uwsgi. r=bhearsum, r=Callek

MihaiTabara:
@Callek r? here too, similar to https://github.com/mozilla/balrog/pull/176

### DIFF
--- a/uwsgi/public.wsgi
+++ b/uwsgi/public.wsgi
@@ -19,6 +19,9 @@ DOMAIN_WHITELIST = {
     "ftp.mozilla.org": ("SystemAddons",),
 }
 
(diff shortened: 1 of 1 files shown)


MihaiTabara:
This is no longer blocking as we've switched to production buckets that backfill archive.m.o which is already whitelisted in both staging and production - so we've ran past the filters in Balrog. Keeping this PR open as @bhearsum said he'd like to tackle this problem anyway.

mozbhearsum:
No need to keep this open - we've figured out the strategy for this (check an environment variable) that we can use next time. Thanks for helping guinea pig that!

PR #181, opened 2016-11-23
Implement SignoffsTable and web apis for it.

mozbhearsum:
There's not too much to say about this - it ended up being pretty boilerplate-y. Most of business logic is in SignoffsTable.insert/delete.

One thing I realized while working on this is that while the ScheduledChangesTable can be fully tested without being enabled for any primary tables, all of the tests I previously wrote for the Scheduled Changes API layer are specifically around Scheduled Rule Changes. Because of this, I had to enable Signoffs for Rules in order to test the web layer. I guess this is OK (it doesn't make as much sense to create a fake version of of Scheduled Changes in the web layer as it does in the database layer), but it felt a bit odd at first.


auslib/test/test_db.py
```diff
@@ -1269,6 +1283,61 @@ def testUpdateChangeToDisabledCondition(self):
...
+        self.assertEquals(got, [(1, "bob", "releng")])
+
+    def testSignoffWithoutPermission(self):
+        self.assertRaisesRegexp(PermissionDeniedError, "jim cannot signoff with role 'releng'",
+                                self.signoffs.insert, "jim", sc_id=1, username="jim", role="releng")
+
+    def testSignoffASecondTimeWithSameRole(self):
+        self.signoffs.insert("nancy", sc_id=1, username="nancy", role="relman")
+        got = self.signoffs.t.select().where(self.signoffs.sc_id == 1).where(self.signoffs.username == "nancy").execute().fetchall()
+        self.assertEquals(got, [(1, "nancy", "relman")])
```
nthomas-mozilla: Could also check no history rows were added.

auslib/db.py
```diff
@@ -1115,6 +1118,51 @@ def mergeUpdate(self, old_row, what, changed_by, transaction=None):
...
+        columns["username"] = changed_by
+        super(SignoffsTable, self).insert(changed_by=changed_by, transaction=transaction, dryrun=dryrun, **columns)
+
+    def update(self, where, what, changed_by=None, transaction=None, dryrun=False):
+        raise AttributeError("Signoffs cannot be modified (only granted and revoked)")
+
+    def delete(self, where, changed_by=None, transaction=None, dryrun=False):
+        for row in self.select(where, transaction):
+            if row["username"] != changed_by:
+                raise PermissionDeniedError("Cannot revoke a signoff made by another user")
```
nthomas-mozilla: Wondering if we might occasionally have situations like 'relman_1 signs off but goes incommunicado, relman_2 spots a problem and wants to stop the scheduled change'. Could you expand on why admin users cannot revoke a signoff ? We'd just delete the whole change ?

auslib/admin/views/base.py
```diff
@@ -42,7 +42,7 @@ def decorated(*args, **kwargs):
                 logging.warning(e)
                 return Response(status=400, response=json.dumps({"exception": msg}), mimetype="application/json")
             except PermissionDeniedError as e:
-                msg = "Permission denied to perform the request"
+                msg = "Permission denied to perform the request. {}".format(e.message)
                 logging.warning(msg)
                 logging.warning(e)
```
JohanLorenzo: Do we still need to log out the error, now that's a part of the message?

auslib/admin/views/scheduled_changes.py
```diff
@@ -118,6 +125,44 @@ def _post(self, sc_id, transaction, changed_by):
...
+class SignoffsView(AdminView):
+    """/scheduled_change/:namespace/:sc_id/signoffs"""
+
+    def __init__(self, namespace, table):
+        self.namespace = namespace
+        self.path = "/scheduled_changes/%s/:sc_id/signoffs" % namespace
+        self.signoffs_table = table.scheduled_changes.signoffs
+        super(SignoffsView, self).__init__()
+
+    def _post(self, sc_id, transaction, changed_by):
```
JohanLorenzo: If I understand correctly, both `_post` and `_delete` don't require the user to be logged in. What use case does that allow?

auslib/db.py
```diff
@@ -1115,6 +1118,51 @@ def mergeUpdate(self, old_row, what, changed_by, transaction=None):
...
+        existing_signoff = self.select({"sc_id": columns["sc_id"], "username": changed_by}, transaction)
+        if existing_signoff:
+            # It shouldn't be possible for there to be more than one signoff,
+            # so not iterating over this should be fine.
+            existing_signoff = existing_signoff[0]
+            if existing_signoff["role"] != columns["role"]:
+                raise PermissionDeniedError("Cannot signoff with a second role")
+            # Signoff already made under the same role, we don't need to do
+            # anything!
+            return True
```
JohanLorenzo: sqlalchemy n00b question: What does `return True` allows, compared to just simply `return`?

auslib/migrate/versions/018_add_scheduled_rule_changes_signoffs.py
```diff
@@ -0,0 +1,35 @@
...
+
+    rules_scheduled_changes_signoffs_history = Table(
+        "rules_scheduled_changes_signoffs_history", metadata,
+        Column("change_id", Integer, primary_key=True, autoincrement=True),
+        Column("changed_by", String(100), nullable=False),
+        Column("sc_id", Integer, nullable=False, autoincrement=False),
+        Column("username", String(100), nullable=False),
+        Column("role", String(50)),
+    )
+    if migrate_engine.name == "mysql":
```
JohanLorenzo: Another noob question: I wonder if we could extract that logic in another file under `auslib/migrate/versions/`

auslib/db.py
mozbhearsum: I didn't allow users to revoke other signoffs because I couldn't come up with a case where that would be useful, and it seemed like the safer default. The situation you describe definitely seems plausible though, and I think it's important that we have the ability to abort in those cases. With that in mind, admin users should clearly be able to revoke anyones signoff, and I could see a case for users to be able to revoke signoffs made by the same group (eg: relman_2 can revoke relman_1).

auslib/admin/views/base.py
mozbhearsum: Good point, we don't!

auslib/admin/views/scheduled_changes.py
mozbhearsum: This gets enforced in the subclasses, actually. I think there was a complicated technical reason I couldn't put it here when I originally wrote this code...

auslib/migrate/versions/018_add_scheduled_rule_changes_signoffs.py
mozbhearsum: Maybe...I'm a bit concerned about having these scripts depend on shared code because it may make it easier to break an older version. OTOH, we use this pattern *everywhere*.

auslib/db.py
```diff
@@ -1891,6 +1939,12 @@ def hasPermission(self, username, thing, action, product=None, transaction=None)
 
         return True
 
+    def hasRole(self, username, role, transaction=None):
+        if role in self.getUserRoles(username, transaction):
```
mozbhearsum: Much nicer, will do!

mozbhearsum:
The latest commits fix the nits, adjust the rules for revoking Signoffs, and update the now-in-repo docs. @nthomas-mozilla - how do you feel about the way revocation works now?

nthomas-mozilla:
lgtm, thanks.

PR #182, opened 2016-11-25
Bug 1310213 - enable scheduled deletions

NinadBhat:


auslib/db.py
```diff
@@ -975,13 +976,15 @@ def select(self, where=None, transaction=None, **kwargs):
             ret.append(row)
         return ret
 
-    def insert(self, changed_by, transaction=None, dryrun=False, **columns):
+    def insert(self, changed_by, transaction=None, dryrun=False, change_type=None, **columns):
```
mozbhearsum: Unfortunately, you'll have to look this up in "columns" instead of adding it its own argument. This method overrides a base class one, so the interface needs to be kept the same. You can see a similar example at https://github.com/mozilla/balrog/blob/master/auslib/db.py#L1501. This probably means you don't need to add it to base_columns yourself either.

auslib/db.py
```diff
@@ -834,6 +834,7 @@ def __init__(self, db, dialect, metadata, baseTable, conditions=("time", "uptake
                            Column("sc_id", Integer, primary_key=True, autoincrement=True),
                            Column("scheduled_by", String(100), nullable=False),
                            Column("complete", Boolean, default=False),
+                           Column("change_type", String(50)),
```
mozbhearsum: Best to make this nullable=False if we're making decisions on it.

auslib/test/admin/views/test_rules.py
```diff
@@ -678,22 +678,22 @@ def setUp(self):
         super(TestRuleScheduledChanges, self).setUp()
         dbo.rules.scheduled_changes.t.insert().execute(
             sc_id=1, scheduled_by="bill", data_version=1, base_rule_id=1, base_priority=100, base_version="3.5", base_buildTarget="d",
-            base_backgroundRate=100, base_mapping="b", base_update_type="minor", base_data_version=1,
+            base_backgroundRate=100, base_mapping="b", base_update_type="minor", base_data_version=1, change_type="update",
```
mozbhearsum: I know it's a pain in the butt, but please add change_type to all of the history tables as well - otherwise this test data is not quite representative of production.

auslib/test/test_db.py
```diff
@@ -1109,6 +1113,11 @@ def testEnactChangeExistingRow(self):
         self.assertEquals(history_row.data_version, 2)
         self.assertEquals(sc_row.complete, True)
 
+    def testEnactChangeForDeletingExistingRow(self):
```
mozbhearsum: You've got a good start on the tests here, but you need to add some to verify that things work correctly when inserting, updating, and deleting scheduled changes. I'm particularly concerned that it might currently be possible to insert a scheduled change that asks for a deletion, but with no primary key defined.

You also need to update testTablesHaveCorrectColumns to check for the new column.

auslib/admin/views/rules.py
```diff
@@ -298,14 +299,21 @@ def __init__(self):
...
     @requirelogin
     def _post(self, transaction, changed_by):
-        if request.json and request.json.get("data_version"):
-            form = ScheduledChangeExistingRuleForm()
-        else:
-            form = ScheduledChangeNewRuleForm()
+        if request.json.get("change_type") != "delete":
+            if request.json and request.json.get("change_type") == "update":
+                form = ScheduledChangeExistingRuleForm()
+            elif request.json.get("change_type") == "new":
```
mozbhearsum: Nit: "insert" would be better than "new", since we're naming the other two (update/delete) like database operations.

auslib/db.py
```diff
@@ -1074,7 +1083,12 @@ def enactChange(self, sc_id, enacted_by, transaction=None):
 
         # If the scheduled change had a data version, it means the row already
         # exists, and we need to use update() to enact it.
-        if what["data_version"]:
+        if change_type == "delete":
```
mozbhearsum: The _checkBaseTablePermissions block needs updating too: https://github.com/mozilla/balrog/blob/37b499a7d0c3d117186e3341eec7e2ea8f6e972d/auslib/db.py#L897.

auslib/migrate/versions/018_add_change_type.py
```diff
@@ -0,0 +1,56 @@
...
+        WHERE
+            base_data_version==NULL
+        """)
+    migrate_engine.execute("""
+        UPDATE "rules_scheduled_changes"
+        SET change_type = "update"
+        WHERE
+            base_data_version!=NULL
+        """)
+
```
mozbhearsum: How much testing have you done already this? It's very difficult to recover from any mistakes in migration scripts, so I want to be very confident this will work.

auslib/test/test_db.py
mozbhearsum: I see you've added a few new tests, but it looks like you're still missing one that verifies that creating a new scheduled change that deletes an existing row will work, unless I'm missing something?

auslib/test/test_db.py
```diff
@@ -674,6 +679,7 @@ def testAllTablesCreated(self):
 
     def testTablesHaveCorrectColumns(self):
         sc_columns = [c.name for c in self.sc_table.t.get_children()]
+        self.assertEquals(len(sc_columns), 9)
```
mozbhearsum: This is a great enhancement, thank you!

auslib/db.py
```diff
@@ -978,10 +979,17 @@ def select(self, where=None, transaction=None, **kwargs):
     def insert(self, changed_by, transaction=None, dryrun=False, **columns):
         base_columns, condition_columns = self._splitColumns(columns)
 
+        if base_columns["change_type"] == "delete":
```
mozbhearsum: Unfortunately, at this level you can't assume that "change_type" will be present in columns. It certainly should be in all existing uses of it, but we don't know what will happen in the future. Your best bet is to explicitly check for it before this, and raise a ValueError if it's not present. You can see something similar at https://github.com/mozilla/balrog/blob/37b499a7d0c3d117186e3341eec7e2ea8f6e972d/auslib/db.py#L1501.

auslib/migrate/versions/018_add_change_type.py
```diff
@@ -0,0 +1,56 @@
...
+        WHERE
+            base_mapping!=NULL AND
+            base_data_version==NULL
+        """)
+    migrate_engine.execute("""
+        UPDATE "rules_scheduled_changes_history"
+        SET change_type = "update"
+        WHERE
+            base_mapping!=NULL AND
+            base_data_version==NULL
```
mozbhearsum: Why is base_mapping part of the WHERE clause here? And why do both queries have the same conditions?

auslib/admin/views/rules.py
```diff
@@ -298,14 +299,21 @@ def __init__(self):
...
     def _post(self, transaction, changed_by):
-        if request.json and request.json.get("data_version"):
-            form = ScheduledChangeExistingRuleForm()
-        else:
-            form = ScheduledChangeNewRuleForm()
+        if request.json.get("change_type") != "delete":
+            if request.json and request.json.get("change_type") == "update":
+                form = ScheduledChangeExistingRuleForm()
+            elif request.json.get("change_type") == "insert":
+                form = ScheduledChangeNewRuleForm()
```
mozbhearsum: Careful with this block - if change_type is set to something other than "update" or "insert", this will raise an Exception. You should probably return a 400 in an else block here.

auslib/admin/views/forms.py
```diff
@@ -144,6 +144,7 @@ class ScheduledChangeForm(Form):
     telemetry_channel = NullableStringField("Telemetry Channel")
     telemetry_uptake = NullableStringField("Telemetry Uptake")
     when = IntegerField("When", validators=[Optional(), not_in_the_past()])
+    change_type = StringField("Change Type")
```
mozbhearsum: Can change_type be changed when modifying an existing Scheduled Change? If not, this should only be in the forms for new Scheduled Changes.

auslib/admin/views/forms.py
```diff
@@ -223,6 +224,14 @@ class ScheduledChangeExistingRuleForm(ScheduledChangeForm, EditRuleForm):
     rule_id = IntegerField('Rule ID', validators=[Required()])
 
 
+class ScheduledChangeDeleteRuleForm(ScheduledChangeForm):
+    """
+    ScheduledChangeDeletionForm includes all the PK columns and ScheduledChangeForm columns
+    """
```
mozbhearsum: This comment needs an update - data_version is neither a ScheduledChangeForm column nor a PK.

auslib/admin/views/rules.py
```diff
@@ -298,14 +299,21 @@ def __init__(self):
 
     @requirelogin
     def _post(self, transaction, changed_by):
-        if request.json and request.json.get("data_version"):
-            form = ScheduledChangeExistingRuleForm()
-        else:
-            form = ScheduledChangeNewRuleForm()
+        if request.json.get("change_type") != "delete":
```
mozbhearsum: I think change_type need to be a required parameter. The way this is written now, if it's not specified we assume it's a Scheduled Deletion, which seems like a bad assumption to make. You should probably adjust the Form to give it a Required() Validator. This should be easier if change_type is only present in the Forms for new Scheduled Changes.

And a small nit: It would be a bit easier to read this if you used == here, and put update/insert in the else block.

auslib/db.py
```diff
@@ -895,10 +896,15 @@ def _splitColumns(self, columns):
...
+        if "change_type" not in new_row:
+            raise ValueError("change_type needed to check Permission")
+
+        if new_row.get("change_type") == "update":
             self.baseTable.update(base_table_where, new_row, changed_by, new_row["data_version"], transaction=transaction, dryrun=True)
-        else:
+        elif new_row.get("change_type") == "insert":
             self.baseTable.insert(changed_by, transaction=transaction, dryrun=True, **new_row)
+        elif new_row.get("change_type") == "delete":
+            self.baseTable.delete(base_table_where, changed_by, new_row["data_version"], transaction=transaction, dryrun=True)
```
mozbhearsum: Best to be on the safe side and raise a ValueError in an else block here. If somehow we managed to get a Scheduled Change with an unknown change_type, we definitely don't want it to get into the database.

auslib/admin/views/forms.py
mozbhearsum: Also, did you try using a SelectField for this? There's only a few valid values, so it would be good to enforce that. See "update_type" for an example.

auslib/db.py
```diff
@@ -977,11 +984,20 @@ def select(self, where=None, transaction=None, **kwargs):
...
-
+        if "change_type" not in base_columns:
+            raise ValueError("Change type is required")
+
+        if base_columns["change_type"] == "delete":
+            for pk in self.base_primary_key:
+                if pk not in base_columns:
+                    raise ValueError("Missing primary key column %s. PK values needed for deletion" % (pk))
+                if base_columns[pk] is None:
+                    raise ValueError("%s value found to be None. PK value can not be None for deletion" % (pk))
```
mozbhearsum: I think these checks need to happen in update() as well - we wouldn't want an existing scheduled deletion to have its PK removed. If you can, moving these to validate() is probably the best thing to do.

auslib/db.py
```diff
@@ -1014,7 +1030,7 @@ def update(self, where, what, changed_by, old_data_version, transaction=None, dr
                     base_columns[base_col] = base_what[base_col]
                 elif sc_columns.get(col):
                     base_columns[base_col] = sc_columns[col]
-
+            base_columns["change_type"] = sc_columns["change_type"]
```
mozbhearsum: Is this necessary? I'm surprised that it doesn't happen as part of the if/elif block above.

I see you're doing it in update/delete/enactChange as well. Can you add a comment explaining why? It looks a bit odd to be copying one specific column after copying a bunch of others in a loop or comprehension.

auslib/test/test_db.py
mozbhearsum: Great work on the tests so far. Just one more to add still: something that tests mergeUpdate for a scheduled deletion.

auslib/migrate/versions/018_add_change_type.py
```diff
@@ -0,0 +1,50 @@
...
+        SET change_type = "insert"
+        WHERE base_data_version is NULL;
+        """)
+    migrate_engine.execute("""
+        UPDATE rules_scheduled_changes_history
+        SET change_type = "update"
+        WHERE base_data_version is not NULL;
+        """)
+    rules_scheduled_changes = Table("rules_scheduled_changes", metadata, autoload=True)
+    rules_scheduled_changes.c.base_update_type.alter(nullable=True)
```
mozbhearsum: Ah, that's right! Carry on.

NinadBhat:
Delete change UI working now.
The following problems
delete
It does not show product and channel
I tried to pass product and channel as hidden attributes
like https://github.com/mozilla/balrog/blob/master/ui/app/templates/rule_scheduled_change_modal.html#L219
but still got the same result
I think is is because in DB product column is null.

mozbhearsum:
I suspect the UI issue is because the UI is built based on the scheduled changes table. For deletes, that table has no value for anything except PK and data version...

mozbhearsum:
Something else I noticed in the UI there - it looks like there's two deletions scheduled for the same Rule? That's not something that should be possible...the second one would always fail.

mozbhearsum:
Small issue I just noticed with the UI: it shows the "View" link for mapping, even though mapping is null. Can you fix that up as well?

PR #184, opened 2016-12-01
Use custom column types for JSON columns

mozbhearsum:
While working on https://bugzilla.mozilla.org/show_bug.cgi?id=1310188 I discovered a bunch of test failures rooted in the releases.data and permissions.options columns. The problem boiled down to the fact that we retrieve the Scheduled Changes from the database when doing an update() or delete(), then pass the data to the base class to do a dry run of enacting the change. Because the rows are retrieved with ScheduledChanges.select(), the mirrored versions of releases.data and permissions.options end up as a string, _not_ a Blob/dict. This causes problems when the base classes try to validate the data, eg when .validate() is called here: https://github.com/mozilla/balrog/blob/a72bbc2289761ec20478665e16d21682a169d46e/auslib/db.py#L1558

This pull request fixes the issue by doing serialization/deserialization at the lowest possible level, using a custom column type as described here: http://docs.sqlalchemy.org/en/latest/core/custom_types.html. This might seem a bit heavy handed, but the only alternative I could come up with was monkey-patching self.scheduled_changes.select in Releases/Permissions.__init__, which seemed much worse.

Most of the patch is simply getting rid of now-unnecessary json.loads/dumps, and fixing some tests to compare against deserialized versions blobs/options instead of strings. A couple of other things of note:
- I had to get a bit more familiar with auslib/admin/views/history.py to make it cope properly. I ended up improving the release diffs to show data version + get rid of of unnecessary context at the same time.
- I made a small interface change to Permissions.getOptions() - it now returns None instead of {} if no options are found to simplify that code a bit. Its only consumer is hasPermission, which has been updated to cope.
- I was initially concerned that I might break caching, or make it less efficient. Manual tests + the cache tests (which verify hits/misses) have given me confidence that I didn't. AFAICT, the only thing that has changed is where deserialization happens, not whether it happens at all.
- I haven't touched the migration scripts because the underlying column type is still the same - the only change is in Python code that sits on top of it. It's a bit weird to have the types differ, but I didn't think it was useful to make a migration script that did nothing.

### DIFF
--- a/auslib/admin/views/base.py
+++ b/auslib/admin/views/base.py
@@ -105,6 +105,11 @@ class HistoryAdminView(AdminView):
                 # prep the value for being shown in revision_row.html
                 if value is None:
                     value = 'NULL'
+                elif isinstance(value, dict):
+                    try:
+                        value = json.dumps(value, indent=2, sort_keys=True)
+                    except ValueError:
+                        pass
                 elif isinstance(value, int):
                     value = unicode(str(value), 'utf8')
                 elif not isinstance(value, basestring):
--- a/auslib/db.py
+++ b/auslib/db.py
@@ -1376,18 +1376,11 @@ class Rules(AUSTable):
 
 
 class BlobColumn(sqlalchemy.types.TypeDecorator):
-    # TODO: use LONGTEXT for mysql
 
     impl = Text
 
     def process_bind_param(self, value, dialect):
-        from auslib.blobs.base import Blob
-        if isinstance(value, Blob):
-            return value.getJSON()
-        elif isinstance(value, dict):
-            return json.dumps(value)
-        else:
-            return value
+        return value.getJSON()
 
     def process_result_value(self, value, dialect):
         return createBlob(value)
@@ -1526,7 +1519,6 @@ class Releases(AUSTable):
         blob.validate(columns["product"], self.domainWhitelist)
         if columns["name"] != blob["name"]:
             raise ValueError("name in database (%s) does not match name in blob (%s)" % (columns["name"], blob["name"]))
-        columns["data"] = blob.getJSON()
 
         ret = super(Releases, self).insert(changed_by=changed_by, transaction=transaction, dryrun=dryrun, **columns)
         if not dryrun:
@@ -1576,7 +1568,6 @@ class Releases(AUSTable):
                 name = what.get("name", name)
                 if name != blob["name"]:
                     raise ValueError("name in database (%s) does not match name in blob (%s)" % (name, blob.get("name")))
-                what['data'] = blob.getJSON()
 
         for release in current_releases:
             name = current_release["name"]
@@ -1594,7 +1585,7 @@ class Releases(AUSTable):
                     if ancestor_change is None:
                         self.log.debug("history for data_version %s for release %s absent" % (old_data_version, name))
                         raise
-                    ancestor_blob = createBlob(ancestor_change.get('data'))
+                    ancestor_blob = ancestor_change.get('data')
                     tip_release = self.getReleases(name=name, transaction=transaction)[0]
                     tip_blob = tip_release.get('data')
                     m = dictdiffer.merge.Merger(ancestor_blob, tip_blob, blob, {})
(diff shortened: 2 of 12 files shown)


auslib/admin/views/base.py
```diff
@@ -105,6 +105,11 @@ def annotateRevisionDifferences(self, revisions):
                 # prep the value for being shown in revision_row.html
                 if value is None:
                     value = 'NULL'
+                elif isinstance(value, dict):
+                    try:
+                        value = json.dumps(value, indent=2, sort_keys=True)
```
nthomas-mozilla: Oh hell yeah. See also https://bugzilla.mozilla.org/show_bug.cgi?id=1304562.

auslib/db.py
```diff
@@ -83,6 +83,34 @@ class ChangeScheduledError(SQLAlchemyError):
     for data consistency reasons."""
 
 
+class JSONColumn(sqlalchemy.types.TypeDecorator):
```
nthomas-mozilla: Maybe some comments around these two classes. The functions they contain are opaque until you go look them up.

auslib/db.py
```diff
@@ -1880,14 +1890,15 @@ def hasPermission(self, username, thing, action, product=None, transaction=None)
...
+            # If a user has a permission that doesn't explicitly limit the type of
+            # actions they can perform, they are allowed to do any type of action.
+            if options.get("actions") and action not in options["actions"]:
+                return False
+            # Similarly, permissions without products specified grant that
+            # that permission without any limitation on the product.
+            if options.get("products") and product not in options["products"]:
+                return False
 
         return True
```
mozbhearsum: I like it. Defaulting to closed for permissions checking seems like the right thing to do, and getting rid of the try/except improves readability. Are you going to send that as a PR, or do you want me to include it here?

auslib/admin/views/base.py
mozbhearsum: That looks like a trivial ride along, let's do it :)

auslib/db.py
JohanLorenzo: Nice simplication! Thanks to it, I believe we can extract the conditions into function with more business-oriented names:
```py
def hasPermission(self, username, thing, action, product=None, transaction=None):
    perm = self.getPermission(username, "admin", transaction=transaction)
    if perm:
        return self._isProductAllowed(perm)    # Admins don't have actions.

    perm = self.getPermission(username, thing, transaction=transaction)
    if perm:    
        return all([self._isProductAllowed(perm), self._isActionAllowed(perm)])

    return False

def _isProductAllowed(permissions, product):
    return self._isFieldAbsentOrExplicitlyWhitelisted(permissions, 'products', product)

def _isActionAllowed(permissions, action):
    return self._isFieldAbsentOrExplicitlyWhitelisted(permissions, 'actions', action)

def _isFieldAbsentOrExplicitlyWhitelisted(permissions, field, field_value):
    # If the field is missing, we assume that the user has access over all elements of this field.
    options = perm["options"]
    return not options or not options.get(field) or field_value in options[field]
```

auslib/db.py
```diff
@@ -83,6 +83,43 @@ class ChangeScheduledError(SQLAlchemyError):
...
+            value = json.dumps(value)
+        return value
+
+    def process_result_value(self, value, dialect):
+        if value:
+            value = json.loads(value)
+        return value
+
+
+def BlobColumn(impl=Text):
```
mozbhearsum: Hmmm, I think this is more of a Column Factory than a decorator (at least, I'm not invoking it as a decorator further down). In any case, the reason is because impl in BlobColumn needs to be configurable at runtime based on the database dialect. JSONColumn is different because it uses Text as the impl regardless of dialect.

auslib/db.py
mozbhearsum: What do you think of this @nthomas-mozilla ? I'm happy to include it in my branch if we're all agreed on it.

auslib/db.py
```diff
@@ -1878,34 +1907,24 @@ def revokeRole(self, username, role, changed_by=None, old_data_version=None, tra
...
         for row in rows:
-            perm = row['permission']
-            opt = row['options']
-            ret[perm] = dict()
-            ret[perm]['data_version'] = row['data_version']
-            if opt:
-                ret[perm]['options'] = json.loads(opt)
-            else:
-                ret[perm]['options'] = None
+            ret[row["permission"]] = {
```
mozbhearsum: I tried rewriting this as:
`return {row["permission"]: {"options": row["options"], "data_version": row["data_version"]} for row in rows}`

...but I actually think the current version is more readable.

auslib/db.py
nthomas-mozilla: I find _isFieldAbsentOrExplicitlyWhitelisted() a little hard to parse, but that's OK if we test/doc it like mad. Otherwise looks good.

auslib/admin/views/base.py
nthomas-mozilla: Thanks!

PR #185, opened 2016-12-01
Bug 1282841 - can't set background rate to 0 for new rules

gilleycro:
Fix for mozilla bug  https://bugzilla.mozilla.org/show_bug.cgi?id=1282841

review: bhearsum

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -55,6 +55,12 @@ Tests can also run by using tox
     $ tox
 
 
+Documentation
+-------------
+
+Balrog's documentation is hosted at http://mozilla-balrog.readthedocs.io/en/latest/index.html
+
+
 Getting Involved
 ----------------
 If you like to get involved in the development of Balrog there're lots of areas where we could use some help.


auslib/admin/views/forms.py
```diff
@@ -168,7 +168,7 @@ class PartialReleaseForm(Form):
 
 
 class RuleForm(Form):
-    backgroundRate = IntegerField('Background Rate', validators=[Required(), NumberRange(0, 100)])
+    backgroundRate = IntegerField('Background Rate', validators=[validators.any_of(range(0,100,1))])
```
mozbhearsum: This looks it fixes the problem, nice work! I fiddled a bit locally to make sure I understood why it worked, and it looks like the part that fixes it is removing Required(). NumberRange() and any_of(range()) do the same thing as far as I can tell, except that NumberRange provides a better error message.

For a test, the easiest way to go about writing one for a change like this is to copy an existing one, and tweak it for the new case you're testing. For example, here is a test that ensures that a properly formatted request for a new rule works: https://github.com/mozilla/balrog/blob/a72bbc2289761ec20478665e16d21682a169d46e/auslib/test/admin/views/test_rules.py#L39

Try the following:
* Remove your changes to forms.py
* Copy that function to a new one, and rename it to something appropriate
* Change "backgroundRate" in the "data" dict to 0
* Run the tests, and make sure that your new test fails. This is a key step because it ensure that your test fails correctly.
* Now apply your fix to forms.py and run the tests again to see if it passes. If it does, great! If the new test (or other tests) fail, the output should help you debug your change to forms.py.

gilleycro: I made new pull request. I have tried to execute tests and confirmed that it doesn't work with old method, and that it works with new one.  I also noticed that I had some mistake in code so I fixed it. 

run-test.sh script had some failed stuff but I don't think they were related to my changes in code.

PR #186, opened 2016-12-03
bug 1170797: "unhang" release upload screen if no file provided

njirap:
Hi @mozbhearsum I am unable to replicate this not working in stage so I have moved stuff around hoping they will work in stage.


mozbhearsum:
I can't reproduce this anymore either...I think that #177 must've fixed this. Do you think there's still a good reason to move this code around?

njirap:
haha...yes I do...old code only checks if the product is not filled then returns the No Product error...For the No File error and No name errors the old code first tries to do stuff then later on discovers that they were not there....This moving around checks that everything else exists before any action can be taken on anything..

mozbhearsum:
Thanks Njira!

PR #187, opened 2016-12-04
Bug 1282841

gilleycro:
Modified, should work now.

run-test.sh script passed, had some errors but not related to my changes.

### DIFF
--- a/auslib/admin/views/forms.py
+++ b/auslib/admin/views/forms.py
@@ -168,7 +168,7 @@ class PartialReleaseForm(Form):
 
 
 class RuleForm(Form):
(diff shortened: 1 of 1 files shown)


gilleycro:
when I run docker-compose up and I try to open rules I get error and see this in response

API Proxying to `/api/rules` failed with: `Error: getaddrinfo ENOTFOUND`

gilleycro:
I cloned fresh copy so I can see did I change anything to make that error. Same things happens with freshly cloned repository.

mozbhearsum:
We'll continue working out of #185 for this!

PR #188, opened 2016-12-04
bug 1285975: product/channel filter not updated after adding/editting

njirap:


njirap:
- [ ] The dropdown needs to be updated when edits add new products or channels, too.

- [ ] Adding a channel-only rule adds a ",channel" entry to the drop down, but that disappears on a refresh. Nothing should be added to the drop down for these cases because we don't support filtering on channel-only.

- [ ] Adding a product-only rule adds two entries to the drop down. One for just the product (that's good), and one with "product,undefined", which we don't want.

- [ ] Sorting is a bit messed up in some cases. If I add a rule with product "AA", it ends up earlier in the list than "All rules".

njirap:
hi @mozbhearsum I was unable to "unify the two" but I got all issues solved.

PR #189, opened 2016-12-05
fixing background rate bug

gilleycro:


auslib/admin/views/forms.py
```diff
@@ -168,7 +168,7 @@ class PartialReleaseForm(Form):
 
 
 class RuleForm(Form):
-    backgroundRate = IntegerField('Background Rate', validators=[Required(), NumberRange(0, 100)])
+    backgroundRate = IntegerField('Background Rate', validators=[any_of(range(0,100,1))])
```
mozbhearsum: This looks it fixes the problem, nice work! I fiddled a bit locally to make sure I understood why it worked, and it looks like the part that fixes it is removing Required(). NumberRange() and any_of(range()) do the same thing as far as I can tell, except that NumberRange provides a better error message.

For a test, the easiest way to go about writing one for a change like this is to copy an existing one, and tweak it for the new case you're testing. For example, here is a test that ensures that a properly formatted request for a new rule works: https://github.com/mozilla/balrog/blob/a72bbc2289761ec20478665e16d21682a169d46e/auslib/test/admin/views/test_rules.py#L39

Try the following:
* Remove your changes to forms.py
* Copy that function to a new one, and rename it to something appropriate
* Change "backgroundRate" in the "data" dict to 0
* Run the tests, and make sure that your new test fails. This is a key step because it ensure that your test fails correctly.
* Now apply your fix to forms.py and run the tests again to see if it passes. If it does, great! If the new test (or other tests) fail, the output should help you debug your change to forms.py.

mozbhearsum:
@gilleycro - are you still interested in working on this?

gilleycro:
@mozbhearsum Yes I am, will resume now. Thank you

gilleycro:
This is the log when I run test locally:

https://pastebin.mozilla.org/8953478


On line 59. I have error but I did not touch that part, could you help me with that please ?

Thank you

auslib/admin/views/forms.py
```diff
@@ -1,13 +1,14 @@
 import simplejson as json
 
 from flask_wtf import Form
-from wtforms import StringField, IntegerField, SelectField, BooleanField
+from wtforms import StringField, IntegerField, SelectField, BooleanField, TextField
```
gilleycro: Well I had other idea in mind when I first tried to fix this so i guess I just forgot about it. I can remove it no problem.

auslib/admin/views/forms.py
```diff
@@ -168,7 +169,7 @@ class PartialReleaseForm(Form):
 
 
 class RuleForm(Form):
-    backgroundRate = IntegerField('Background Rate', validators=[Required(), NumberRange(0, 100)])
+    backgroundRate = IntegerField('Background Rate', validators=[any_of(range(0, 101, 1))])
```
mozbhearsum: Is there a reason you're using any_of() instead of NumberRange()? As I mentioned before, NumberRange seems to do the same thing, except it provides a better error message.

gilleycro: No really, I forgot about that TBH cause of error "phantom" errors I was getting from test. I will commit now version with NumberRange and then we can alter error message to your liking.

PR #190, opened 2016-12-09
bug 1138418: Sanity check for hashes

meetmangukiya:
Does this look like the right way of doing it ? Addressing to these comments :
> Zach and I exchanged some e-mail and it looks like he'll be taking a crack at this. I'll assign it to you to make sure no one else picks it up!
The best place to do this is probably in the individual blob classes that require a hash function (https://github.com/mozilla/balrog/blob/master/auslib/blobs/apprelease.py, https://github.com/mozilla/balrog/blob/master/auslib/blobs/gmp.py, and https://github.com/mozilla/balrog/blob/master/auslib/blobs/systemaddons.py). They will need to override the base class' validate method (but still call it to do the schema validation), and then walk through the blob data to ensure that all hashValues are the correct length for the given hashFunction

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/balrog/190)

### DIFF
--- a/auslib/blobs/apprelease.py
+++ b/auslib/blobs/apprelease.py
@@ -118,6 +118,16 @@ class ReleaseBlobBase(Blob):
         return '        <patch type="%s" URL="%s" hashFunction="%s" hashValue="%s" size="%s"/>' % \
             (patchType, url, self["hashFunction"], patch["hashValue"], patch["filesize"])
 
+    def validate(self, *args, **kwargs, ignore=False):
+        Blob.validate(self, *args, **kwargs)
+        required_length = Blob.hashes[self["hashFunction"].upper()]
+        if not ignore:
+            for platform in self["platforms"]:
+                for locale in platform["locales"]:
+                    for state in locale:
+                        if len(state["hashValue"]) != required_length:
+                            raise ValueError("Length of hash is not equal to that of {} function".format(self["hashFunction"]))
+
     def getInnerHeaderXML(self, updateQuery, update_type, whitelistedDomains, specialForceHosts):
         buildTarget = updateQuery["buildTarget"]
         locale = updateQuery["locale"]
--- a/auslib/blobs/gmp.py
+++ b/auslib/blobs/gmp.py
@@ -61,6 +61,15 @@ class GMPBlobV1(Blob):
 
         return vendorXML
 
+    def validate(self, *args, **kwargs, ignore=False):
+        Blob.validate(self, *args, **kwargs)
+        required_length = Blob.hashes[self["hashFunction"].upper()]
+        if not ignore:
+            for vendors in self["vendors"]:
+                for platform in vendors["platforms"]:
+                    if len(platform["hashValue"]) != required_length:
+                        raise ValueError("Length of hash is not equal to that of {} function".format(self["hashFunction"]))
+
     def containsForbiddenDomain(self, product, whitelistedDomains):
         """Returns True if the blob contains any file URLs that contain a
            domain that we're not allowed to serve updates to."""
--- a/auslib/blobs/base.py
+++ b/auslib/blobs/base.py
@@ -58,6 +58,7 @@ def createBlob(data):
 
 class Blob(dict):
     jsonschema = None
+    hashes = dict(SHA1=40, SHA256=64, SHA512=128, SHA384=96)
 
     def __init__(self, *args, **kwargs):
         super(Blob, self).__init__(self, *args, **kwargs)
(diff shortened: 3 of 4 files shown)


auslib/blobs/gmp.py
```diff
@@ -61,6 +61,15 @@ def getInnerXML(self, updateQuery, update_type, whitelistedDomains, specialForce
 
         return vendorXML
 
+    def validate(self, *args, **kwargs):
+        Blob.validate(self, *args, **kwargs)
+        hashes = dict(SHA1=40, SHA256=64, SHA512=128)
```
mozbhearsum: Two things here:
1) I just found out this week that we expect to use sha384 in the near future. Please add this, too.
2) Since this will be shared between multiple blob types, it would be good to move the hash length definitions and actual checking of them to a separate function.

auslib/blobs/gmp.py
```diff
@@ -61,6 +61,15 @@ def getInnerXML(self, updateQuery, update_type, whitelistedDomains, specialForce
...
+        required_length = hashes[self["hashFunction"]]
```
meetmangukiya: @mozbhearsum can you confirm if `self["hashFunction"]` will always be upper case, if not I should use `.upper()` for it to get NoSuchKey type of error.

auslib/blobs/gmp.py
meetmangukiya: > I just found out this week that we expect to use sha384 in the near future. Please add this, too.

[x] Added

auslib/blobs/apprelease.py
```diff
@@ -118,6 +118,16 @@ def _getSpecificPatchXML(self, patchKey, patchType, patch, updateQuery, whitelis
         return '        <patch type="%s" URL="%s" hashFunction="%s" hashValue="%s" size="%s"/>' % \
             (patchType, url, self["hashFunction"], patch["hashValue"], patch["filesize"])
 
+    def validate(self, *args, **kwargs, ignore=False):
+        Blob.validate(self, *args, **kwargs)
+        required_length = Blob.hashes[self["hashFunction"].upper()]
```
meetmangukiya: I figured `.upper` wouldn't do any bad anyways

auslib/blobs/apprelease.py
```diff
@@ -118,6 +118,16 @@ def _getSpecificPatchXML(self, patchKey, patchType, patch, updateQuery, whitelis
         return '        <patch type="%s" URL="%s" hashFunction="%s" hashValue="%s" size="%s"/>' % \
             (patchType, url, self["hashFunction"], patch["hashValue"], patch["filesize"])
 
+    def validate(self, *args, **kwargs, ignore=False):
```
meetmangukiya: Regarding the `ignore` parameter.. 
### Purpose
- To be used while testing, I noticed that the `hashValue` in most of the test cases is a single digit number and not a _dummy hash_, so during testing we will flip `ignore = True`

meetmangukiya:
As far as the tests are concerned, I'll be adding them soon.

auslib/blobs/gmp.py
mozbhearsum: I think I wasn't quite clear on what I meant here...I think that the part that validates the hash length can be factored out into its own function, eg:
def validate_hash_length(type_, hash):

...and then that can be called from classes like ReleaseBlobBase and GMPBlobV1.

Does that make sense?

auslib/blobs/gmp.py
mozbhearsum: We don't enforce that anywhere, so that's not a safe assumption to make. Using .upper() is a good idea.

auslib/blobs/gmp.py
meetmangukiya: Yeah, that does. I guess I just move out the code, make a function, make the call **here** itself, or should it be called somewhere else. Any particular place, where this function belongs? I'd suggest `Blob` itself :thinking: ?

mozbhearsum: I think the function for validating whether a hash is the correct length or not doesn't need to live in Blob - there's nothing inherently Blob specific about it. I would suggest putting it somewhere in https://github.com/mozilla/balrog/tree/master/auslib/util.

mozbhearsum:
Are you still looking at this, @meetmangukiya?

meetmangukiya:
@mozbhearsum I am not able to find time to complete this, maybe someone can work over this, he will just have to write the tests..

mozbhearsum:
I'm going to close this pull request for now - a new one will need to be opened when someone else writes the tests, anyways. Thank your for your work on this so far, @meetmangukiya.

PR #192, opened 2016-12-12
bug 1315369: pull latest balrogdb from hosted location when rebuildi…

sirMackk:
…ng local db

This is a patch for https://bugzilla.mozilla.org/show_bug.cgi?id=1315369. It modifies the `initdb_and_run.sh` script to attempt to use the newest available database dump when populating the local database.

At first, it compares the time stamp of the dump at the specified URL and the local dump. If the dump at the specified URL is newer, or if the local dump doesn't exist, the script downloads the new dump and uses it to populate the database. If there is no network connectivity, the script uses the local dump, which is the newest available dump in this situation. If there is no network connectivity _and_ no local dump, the script will use the dump that's part of the repository (`scripts/sample-data.sql.bz2`).

**Edit 12/16/2016**: It just hit me that maybe adding `wget` as a dependency to the Dockerfile isn't the best solution here and that most of this can be performed by a Python script. What do you think?

### DIFF
--- a/scripts/import-db.py
+++ b/scripts/import-db.py
@@ -1,0 +1,101 @@
+#!/usr/bin/env python
+
+from calendar import timegm
+from datetime import datetime
+from httplib import HTTPSConnection
+import logging
+import os
+from socket import gaierror
+import time
+from urllib2 import urlopen, HTTPError, URLError
+
+
+HOST = 'https://balrog-public-dump-prod.s3.amazonaws.com'
+PATH = '/dump.sql.txt'
+LOCAL_DB_PATH = os.getenv('LOCAL_DUMP', '/app/scripts/prod_db_dump.sql')
+TIMEOUT = 10
+
+logging.basicConfig(level=logging.DEBUG, format='%(asctime)s: %(message)s')
+
+
+def getRemoteDBModifiedTS():
+    """
+    Performs a HEAD request to get the Last-Modified date-time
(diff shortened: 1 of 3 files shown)


mozbhearsum:
I don't have a problem depending on wget, but I do think that this would be much easier to read and maintain in Python. Could you give that a shot?

sirMackk:
Yeah, I agree. I redid the PR to shift the heavy lifting over to Python. No 3rd party dependencies and definitely more readable than before.

scripts/import-db.py
```diff
@@ -0,0 +1,87 @@
...
+    """
+    Sets mtime on the local database dump file to the remote database dump
+    file's mtime. Sets atime to now.
+    """
+    now_ts = int(time.time())
+    os.utime(LOCAL_DB_PATH, (now_ts, prod_db_ts,))
+
+
+def setLocalDBPermissions():
+    os.chmod(LOCAL_DB_PATH, 0666)
```
mozbhearsum: It's unfortunate that we can't set the ownership correctly (boo Docker), but this is a good workaround - nice idea!

sirMackk:
Good point - I added more verbose logging messages and I changed the code to exit sooner if there's a problem getting the remote database. Btw, thanks for checking all the scenarios, it's always better when there's a second pair of eyes.

sirMackk:
Sweet! Thanks. Onto the next one :)

PR #194, opened 2016-12-20
fix for bug id 1281459 which propagate errors with status 500 to client

dpaks:


auslib/web/base.py
```diff
@@ -117,4 +113,4 @@ def unsubstituted_url_variables():
     '/update/3/<product>/<version>esrpre/<buildID>/<buildTarget>/<locale>/<channel>/<osVersion>/<distribution>/<distVersion>/update.xml',
     view_func=ClientRequestView.as_view('clientrequest_esrnightly'),
     defaults={'queryVersion': 3},
-)
+)
```
dpaks: flake8 complained. Hence.

auslib/web/base.py
```diff
@@ -41,9 +41,8 @@ def generic(error):
     a 200 response with no updates is returned, because that's what the client
     expects. See bugs 885173 and 1069454 for additional background."""
 
-    if not isinstance(error, BadDataError):
-        if sentry.client:
-            sentry.captureException()
+    if isinstance(error, BadDataError):
+        raise BadDataError
 
     # We don't want to eat exceptions from this special Dockerflow endpoint because
```
dpaks: irrelevant?

auslib/test/web/test_client.py
```diff
@@ -906,11 +906,12 @@ def testCacheControlIsNotSetFor404(self):
         ret = self.client.get('/whizzybang')
         self.assertEqual(ret.headers.get("Cache-Control"), None)
 
-    def testCacheControlIsNotSetFor500(self):
+    def testCacheControlIsSetFor500(self):
         with mock.patch('auslib.web.views.client.ClientRequestView.get') as m:
```
mozbhearsum: I don't think test has any value anymore. It's purpose was to check the response headers for Cache-Control, but we don't get a response with headers to check now. Removing it is the right thing to do.

auslib/test/web/test_client.py
```diff
@@ -919,22 +920,23 @@ def testEmptySnippetOn404(self):
     def testEmptySnippetOn500(self):
         with mock.patch('auslib.web.views.client.ClientRequestView.get') as m:
             m.side_effect = Exception('I break!')
-            ret = self.client.get('/update/4/b/1.0/1/p/l/a/a/a/a/1/update.xml')
-            self.assertUpdatesAreEmpty(ret)
+            with self.assertRaises(Exception) as exc:
+                self.client.get('/update/4/b/1.0/1/p/l/a/a/a/a/1/update.xml')
+                self.assertEqual('I break!', exc.exception.value())
```
mozbhearsum: I didn't know you could use this as a context manager - neat!

However, I don't think this assertEqual ever runs. It looks to me like nothing in the block will run after the Exception is raised. The example on https://docs.python.org/2/library/unittest.html does some verification of the exception outside the block, like this:
```
with self.assertRaises(SomeException) as cm:
    do_something()

the_exception = cm.exception
self.assertEqual(the_exception.error_code, 3)
```

auslib/test/web/test_client.py
```diff
@@ -919,22 +920,23 @@ def testEmptySnippetOn404(self):
...
 
     def testSentryBadDataError(self):
         with mock.patch("auslib.web.views.client.ClientRequestView.get") as m:
             m.side_effect = BadDataError("exterminate!")
-            with mock.patch("auslib.web.base.sentry") as sentry:
+            with mock.patch("auslib.web.base.sentry"), self.assertRaises(BadDataError) as exc:
                 self.client.get("/update/4/b/1.0/1/p/l/a/a/a/a/1/update.xml")
-                self.assertFalse(sentry.captureException.called)
+                self.assertEqual('exterminate!', exc.exception.value())
```
mozbhearsum: It's important to check sentry.captureException.called, because these tests' purpose is to verify that Sentry handling is correct. Just like the previous comment, you'll need to do that outside of the context manager, though.

auslib/web/base.py
```diff
@@ -41,9 +41,8 @@ def generic(error):
     a 200 response with no updates is returned, because that's what the client
     expects. See bugs 885173 and 1069454 for additional background."""
 
-    if not isinstance(error, BadDataError):
-        if sentry.client:
-            sentry.captureException()
+    if isinstance(error, BadDataError):
+        raise BadDataError
```
mozbhearsum: We still want to send BadDataErrors to Sentry, so please don't remove that. I think the only change needed in this function is re-raise the orginal error instead of only doing it sometimes.

auslib/test/web/test_client.py
dpaks: Addressed.

auslib/test/web/test_client.py
dpaks: Addressed.

auslib/web/base.py
dpaks: Addressed.

auslib/web/base.py
```diff
@@ -38,24 +38,13 @@ def fourohfour(error):
...
-    # functional. See https://github.com/mozilla-services/Dockerflow for additional
-    # details.
-    if request.path == "/__heartbeat__":
-        return error
-
-    log.debug('Hit exception, sending an empty response', exc_info=True)
-    response = make_response('<?xml version="1.0"?>\n<updates>\n</updates>')
-    response.mimetype = 'text/xml'
-    return response
+    raise Exception(error)
```
mozbhearsum: "error" is already an Exception, so it's better not to wrap it in a new one. Doing this ends up with specific exceptions (like BadDataError) showing up as Exception - so we lose granularity.

mozbhearsum: Thank you for that!

PR #196, opened 2016-12-20
Enable Scheduled Changes for Releases and Permissions

mozbhearsum:
This patch is largely boilerplate (eg: routes, views, forms), but there's a few notable things besides that:
- Splitting up Uptake and Time based fields into two forms (because Releases and Permissions don't have uptake-based scheduled changes enabled)
- Some changes in the ScheduledChange views to make it work with a field called "data".
- Add a None guard to BlobColumn methods, after a test found a way to hit an Exception. AFAICT there's no way to hit this exception through the API endpoints, so #184 should be fine in production.
- Adjusting ScheduledChanges.validate to stop guessing about what type of change things are, made possible by @NinadBhat's addition of "change_type".
- Fixing Releases.delete to check mapping/whitelist/fallbackMapping of all affected releases, instead of just the one passed. I uncovered this because ScheduledChangesTable passes "where" as SQLAlchemy clauses instead of by dict.

UI for this will be handled in https://bugzilla.mozilla.org/show_bug.cgi?id=1310214.


nthomas-mozilla:
I'm hoping to review before xmas but may run out of time. Would you mind resolving the conflicts in forms.py ?

mozbhearsum:
Whoops, I didn't notice that those cropped up. Will do.

auslib/test/admin/views/test_releases.py
```diff
@@ -1016,18 +1017,308 @@ def testReleasesPost(self):
...
+            "base_name": "ab", "base_data": {"name": "ab", "hashFunction": "sha256", "schema_version": 1}, "base_data_version": 1
+        }
+        self.assertEquals(db_data, expected)
+        cond = dbo.releases.scheduled_changes.conditions.t.select().where(dbo.releases.scheduled_changes.conditions.sc_id == 4).execute().fetchall()
+        self.assertEquals(len(cond), 1)
+        cond_expected = {"sc_id": 4, "data_version": 1, "when": 2300000000}
+        self.assertEquals(dict(cond[0]), cond_expected)
+
+    @mock.patch("time.time", mock.MagicMock(return_value=300))
+    def testAddScheduledChangeDeleteRelease(self):
```
nthomas-mozilla: No testEnactScheduledChangeDeleteRelease to go with this ?

auslib/test/test_db.py
```diff
@@ -1311,7 +1316,7 @@ def testValidateConditionsTelemetryRaisesError(self):
 
     @mock.patch("time.time", mock.MagicMock(return_value=200))
     def testInsertWithEnabledCondition(self):
-        what = {"fooid": 11, "foo": "i", "bar": "jjj", "data_version": 1, "when": 909000, "change_type": "insert"}
+        what = {"fooid": 11, "foo": "i", "bar": "jjj", "data_version": 1, "when": 909000, "change_type": "update"}
```
nthomas-mozilla: Looks like change_type is not verified when the db is read back, by design or just missed when change_type came along ?

auslib/admin/views/forms.py
```diff
@@ -155,6 +158,33 @@ class ExistingPermissionForm(DbEditableForm):
     options = JSONStringField(None, 'Options')
 
 
+class ScheduledChangeNewPermissionForm(ScheduledChangeTimeForm, NewPermissionForm):
+    permission = StringField('Permission', validators=[Length(0, 50), InputRequired()])
+    username = StringField('Username', validators=[Length(0, 100), InputRequired()])
+    change_type = SelectField("Change Type", choices=[('insert', 'insert'), ('update', 'update'), ('delete'), ('delete')])
```
mozbhearsum: Yeah, I struggled with what to do here. There's so many different combinations of required/not required/not allowed for PKs, non-PK base table columns, scheduled changes columns, data_version/sc_data_version that I opted to just be more explicit and verbose.

In the medium-to-long term, moving towards Swagger may help make this better. We should be able to separate "definition and validation of a field" from "required/not required/not allowed".

auslib/admin/views/forms.py
```diff
@@ -265,3 +295,41 @@ class ReadOnlyForm(Form):
     product = StringField('Product', validators=[InputRequired()])
     read_only = BooleanField('read_only')
     data_version = IntegerField('data_version', widget=HiddenInput())
+
+
+class ScheduledChangeNewReleaseForm(ScheduledChangeTimeForm):
+    name = StringField('Name', validators=[InputRequired()])
+    product = StringField('Product', validators=[InputRequired()])
+    data = JSONStringField({}, 'Data', validators=[InputRequired()], widget=FileInput())
+    data_version = IntegerField('data_version', widget=HiddenInput())
```
mozbhearsum: Good catch, we don't. This probably ends up never being used because it's optional.

auslib/admin/views/forms.py
```diff
@@ -265,3 +295,41 @@ class ReadOnlyForm(Form):
...
+    change_type = SelectField("Change Type", choices=[('insert', 'insert'), ('update', 'update'), ('delete'), ('delete')])
+
+
+class ScheduledChangeExistingReleaseForm(ScheduledChangeTimeForm):
+    name = StringField('Name', validators=[InputRequired()])
+    product = StringField('Product', validators=[Optional()])
+    data = JSONStringField({}, 'Data', validators=[Optional()], widget=FileInput())
+    data_version = IntegerField('data_version', widget=HiddenInput())
```
mozbhearsum: Great questions. This is something we didn't need worry about for Rule Scheduled Changes, I suppose, because the data is simpler and the PK is hidden and unchangeable.

The reason name is required here isn't to allow for changing it - it's required because we need to know which Release to update. Unlike when editing a Release directly, it's not included in the URL for Scheduled Changes, so we need to put it in the Form. I agree that changing the Product of an existing Release is unlikely (and we can always do that by deleting/adding a new one, if we need to). How about this for Releases:
* For Scheduled Changes that create new Releases:
** When creating, name, product, and data will be required, matching the db rules. Neither data_version nor sc_data_version will be allowed.
** When editing, name, product, and data will be optional. We update based on sc_id, so we don't need to know name to edit, but we should allow for other base table fields to be passed in case there's a typo fix or other need to change the name. Same thing for product & data. sc_data_version will be required, data_version will not be allowed.
* For Scheduled Changes that change existing Releases:
** When creating, name and data will be required. Product will not be allowed, per above about it being confusing to allow product to change. data_version will be required, sc_data_version will not be allowed.
** When editing, only data will be allowed. sc_data_version will be required, data_version will not be allowed.

I think Permissions may be similar, too, except we'd treat both permission and username like we treat name, and treat options like data. Spelled out more, this means:
* For Scheduled Changes that create new Permissions:
** When creating, permission, username, and options will be required. Neither data_version nor sc_data_version will be allowed.
** When editing, permission, username, and options will all be optionaltd. sc_data_version will be required, data_version will not be allowed.
* For Scheduled Changes that modify a Permission:
** When creating, permission, username, and options will be required. data_version will be required, sc_data_version will not be allowed.
** When editing, only options will be allowed. sc_data_version will be required, data_version will not be allowed.

auslib/db.py
```diff
@@ -1785,20 +1794,21 @@ def localeExists(self, name, platform, locale, transaction=None):
...
-        if mapping_count > 0 or whitelist_count > 0 or fallbackMapping_count > 0:
-            msg = "%s has rules pointing to it. Hence it cannot be deleted." % (self.name)
-            raise ValueError(msg)
         for toDelete in self.select(where=where, columns=[self.name, self.product], transaction=transaction):
             names.append(toDelete["name"])
             self._proceedIfNotReadOnly(toDelete["name"], transaction=transaction)
             if not self.db.hasPermission(changed_by, "release", "delete", toDelete["product"], transaction):
                 raise PermissionDeniedError("%s is not allowed to delete releases for product %s" % (changed_by, toDelete["product"]))
+
+        for name in names:
```
mozbhearsum: We always sort-of did at the database layer. Prior to Scheduled Changes, you could've called Releases.delete() to do it (because we didn't override it at that time). One of the side effects of the Scheduled Changes implementation was overriding methods such as this, which required changing their interface to match the base class, which in turn means we have methods like this that allow deleting anything any number of things at once.

I agree that this feels more dangerous when it's being down in this specific class. Perhaps it's best to bail out early in this method if "name" isn't in "where", or if the clause matches more than one row?

auslib/migrate/versions/020_add_release_and_permission_scheduled_changes.py
```diff
@@ -0,0 +1,161 @@
...
+        "releases_scheduled_changes_signoffs_history", metadata,
+        Column("change_id", Integer, primary_key=True, autoincrement=True),
+        Column("changed_by", String(100), nullable=False),
+        Column("sc_id", Integer, nullable=False, autoincrement=False),
+        Column("username", String(100), nullable=False),
+        Column("role", String(50)),
+    )
+
+    releases_scheduled_changes.append_column(Column("base_data", dataType))
+    releases_scheduled_changes_history.append_column(Column("base_data", dataType))
```
mozbhearsum: Good point, I'll put them with the rest of the Columns.

auslib/test/admin/views/test_permissions.py
```diff
@@ -169,6 +170,304 @@ def testPermissionDeleteWithoutPermission(self):
...
+        dbo.permissions.scheduled_changes.t.insert().execute(
+            sc_id=1, scheduled_by="bill", change_type="insert", data_version=1, base_permission="rule", base_username="janet",
+            base_options={"products": ["foo"]},
+        )
+        dbo.permissions.scheduled_changes.history.t.insert().execute(change_id=1, changed_by="bill", timestamp=20, sc_id=1)
+        dbo.permissions.scheduled_changes.history.t.insert().execute(
+            change_id=2, changed_by="bill", timestamp=21, sc_id=1, scheduled_by="bill", change_type="insert", data_version=1,
+            base_permission="rule", base_username="janet", base_options={"products": ["foo"]},
+        )
+        dbo.permissions.scheduled_changes.signoffs.t.insert().execute(sc_id=1, username="bill", role="releng")
```
mozbhearsum: I think I got a bit lazy with these and the Release tests, will fix.

auslib/test/test_db.py
mozbhearsum: Hmm...I'm not sure I understand what you mean by verified here, can you clarify?

mozbhearsum:
> One question I have is how the 'when' of a scheduled change will work with signoffs. Say someone adds a change, and maybe signs off for their group, then acquires the other signoffs. In the permissions case, does the 'when' default to the first submission time, so that the agent actions the change as soon as all signoffs are done ? Releases would have a genuine future date like rules, I'm guessing.

Your Permissions example is how I was thinking about it, yeah. For that specifically, maybe we'll even hide "when" in the UI and just set it to a minute in the future automatically or something. Depending on how https://bugzilla.mozilla.org/show_bug.cgi?id=1310207 gets implemented, we may just be able to disable the time condition for Permissions altogether, though.

Releases and Rules will generally have future dates, but don't forget that even things such as throttle changes will need signoff once we're all done here - which means that we may start wanting to schedule Rule and Release changes very near in the future. This creates an interesting edge case of what to do when signoffs only come in after the desired enacting time passes. Can we safely assume that it's still safe to enact if the final signoff doesn't come until minutes/hours/days after "when"? Perhaps we cancel those sorts of changes after X minutes.

auslib/test/admin/views/test_releases.py
```diff
@@ -1016,18 +1017,308 @@ def testReleasesPost(self):
...
+        dbo.releases.scheduled_changes.t.insert().execute(
+            sc_id=1, scheduled_by="bill", change_type="insert", data_version=1, base_name="m", base_product="m",
+            base_data=createBlob(dict(name="m", hashFunction="sha512", schema_version=1))
+        )
+        dbo.releases.scheduled_changes.history.t.insert().execute(change_id=1, changed_by="bill", timestamp=50, sc_id=1)
+        dbo.releases.scheduled_changes.history.t.insert().execute(
+            change_id=2, changed_by="bill", timestamp=51, sc_id=1, scheduled_by="bill", change_type="insert", data_version=1, base_name="m",
+            base_product="m", base_data=createBlob(dict(name="m", hashFunction="sha512", schema_version=1))
+        )
+        dbo.releases.scheduled_changes.signoffs.t.insert().execute(sc_id=1, username="bill", role="releng")
```
mozbhearsum: Yeah, I sortof struggled to decide what was the best balance here. We're basically just testing base class stuff from auslib/admin/views/scheduled_changes.py and history.py here, so it's a bit of a tough sell to add the same tests for each type of Scheduled Change. OTOH, adding these tests found bugs in the base classes with handling of JSON columns.

auslib/admin/views/forms.py
```diff
@@ -155,6 +158,33 @@ class ExistingPermissionForm(DbEditableForm):
...
+
+class ScheduledChangeDeletePermissionForm(ScheduledChangeTimeForm):
+    change_type = SelectField("Change Type", choices=[('insert', 'insert'), ('update', 'update'), ('delete', 'delete')])
+    permission = StringField('Permission', validators=[Length(0, 50), InputRequired()])
+    username = StringField('Username', validators=[Length(0, 100), InputRequired()])
+    data_version = IntegerField('data_version', validators=[InputRequired()], widget=HiddenInput())
+
+
+class EditScheduledChangeNewPermissionForm(ScheduledChangeTimeForm, NewPermissionForm):
+    sc_data_version = IntegerField('sc_data_version', validators=[InputRequired()], widget=HiddenInput())
```
mozbhearsum: Yeah, I see what you mean. The difference here is that the existing Permissions forms didn't have anything that wasn't desired is the Scheduled Changes versions, whereas the Releases ones did (eg: alias, "blob" instead of "data"). Perhaps it's less confusing to explicitly specify them rather than inheriting, though.

auslib/test/test_db.py
mozbhearsum: Oh, duh. You mean checking the value like we do for the other columns.

mozbhearsum:
@nthomas-mozilla - I think this addresses most of the things you pointed out. Writing the extra tests for Releases actually uncovered a bug in Releases.delete where checking mappings outside of the Transaction broke the active transaction (in sqlite, anyways), so I've fixed that as well. The other notable part of this is the reworking of forms.py - I've switched all of the Releases and Permissions Forms to be fully explicit - only inheriting from ScheduledChangeTimeForm. This should get rid of any confusion around what's required and not.

If there's other things you're concerned about or would like you changed based on my earlier responses just let me know.

auslib/test/admin/views/test_releases.py
```diff
@@ -1281,6 +1330,22 @@ def testEnactScheduledChangeNewRelease(self):
...
+        self.assertEquals(len(r), 1)
+        db_data = dict(r[0])
+        expected = {
+            "sc_id": 4, "complete": True, "data_version": 2, "scheduled_by": "bill", "change_type": "delete", "base_name": "ab", "base_product": None,
+            "base_read_only": False, "base_data": None, "base_data_version": 1,
+        }
+        self.assertEquals(db_data, expected)
+
+        base_row = dbo.releases.t.select().where(dbo.releases.name == "a").execute().fetchall()
+        self.assertEquals(len(base_row), 1)
```
nthomas-mozilla: Was expecting a check for the "ab" not existing here first, then you're checking that "a" didn't get taken out as collateral damage ?

mozbhearsum: This is simply just me not updating this after a copy/paste correctly, unfortunately. I don't think a collatoral damage check has much value (we don't generally check non-named rows in other tests).

PR #198, opened 2016-12-21
Correcting UI and backend for Delete Scheduled change update button

NinadBhat:
The backed error were caused by IntegerField and SelectField hence creating a new form.

### DIFF
--- a/auslib/admin/views/forms.py
+++ b/auslib/admin/views/forms.py
@@ -245,27 +245,8 @@ class EditScheduledChangeExistingRuleForm(ScheduledChangeForm, EditRuleForm):
     sc_data_version = IntegerField('sc_data_version', validators=[Required()], widget=HiddenInput())
 
 
-class EditScheduledChangeDeleteRuleForm(ScheduledChangeDeleteRuleForm):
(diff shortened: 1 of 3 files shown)


auslib/admin/views/forms.py
```diff
@@ -245,6 +245,29 @@ class EditScheduledChangeExistingRuleForm(ScheduledChangeForm, EditRuleForm):
     sc_data_version = IntegerField('sc_data_version', validators=[Required()], widget=HiddenInput())
 
 
+class EditScheduledChangeDeleteRuleForm(ScheduledChangeDeleteRuleForm):
+    sc_data_version = IntegerField('sc_data_version', validators=[Required()], widget=HiddenInput())
+    backgroundRate = NullableStringField('Background Rate', validators=[Optional()])
```
mozbhearsum: I don't think you need to include backgroundRate or any of the other Rule fields in this form - they aren't used for Scheduled Deletions (or even in the UI), right? I think the only needed (and valid) fields for updating a scheduled deletion of a rule are "when" and "sc_data_version"? Even change_type, rule_id, and data_version can't be changed?

PR #200, opened 2016-12-23
Changing Required to InputRequired

NinadBhat:
https://github.com/mozilla/balrog/blob/master/auslib/admin/views/forms.py#L249 should be InputRequired in place of required Required .
Currently sudo tox on master branch is failing .
@mozbhearsum: I saw that you made this change in your PR. So if not needed please close the PR.


mozbhearsum:
Whoops, thanks for this!

PR #201, opened 2016-12-28
bug 1325460: rule history ui broken by bug 1285975

mozbhearsum:
It looks like the issue here is that https://github.com/mozilla/balrog/commit/a23096f36fa474ac8df6c83d8f5abbea7c04f605#diff-71ee39a9a5979b46096d0bbc53a88574L8 changed the default value of $scope.pr_ch_options. This was the correct thing to do for the main rules page, but it broke the check that filterBySelect does to see if it should return all revisions or not.

This is a quick bustage fix because having broken history is really bad. In the medium-to-long term, we should probably rework the UI so that code like this is not shared.

### DIFF
--- a/agent/tox.ini
+++ b/agent/tox.ini
@@ -11,7 +11,9 @@ setenv =
 deps =
     asynctest
     flake8
-    coverage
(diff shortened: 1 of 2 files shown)


NinadBhat:
LGTM

mozbhearsum:
Thanks for the sanity check!

PR #205, opened 2017-01-04
UI for scheduled changes in Releases

NinadBhat:
https://bugzilla.mozilla.org/show_bug.cgi?id=1310214


mozbhearsum:
Can you update this against the latest version of #196? It looks like it's a week or so behind.

ui/app/js/controllers/release_scheduled_change_edit_controller.js
```diff
@@ -0,0 +1,144 @@
...
+
+  };
+
+  $scope.changeName = function () {
+    //wait for actual file to be loaded
+    setTimeout($scope.fillName, 0);
+  };
+
+  $scope.saveChanges = function () {
+    if (!$scope.sc.product.trim()) {
```
NinadBhat: I have created release_scheduled_change_delete_controller.js  for Scheduled Deletion

ui/app/js/controllers/release_scheduled_changes_controller.js
```diff
@@ -0,0 +1,164 @@
...
+      },
+    ];
+  } else {
+  $scope.ordering_options = [
+    {
+      text: "When",
+      value: "when"
+    },
+    {
+      text: "Product, Name",
```
mozbhearsum: Close! These should be separate items though. There's really no reason to have a two-part sort, but sorting on either exclusively is useful.

auslib/admin/views/scheduled_changes.py
```diff
@@ -82,7 +82,7 @@ def _post(self, sc_id, form, transaction, changed_by):
         for k, v in form._fields.iteritems():
             # sc_data_version is a "special" column, in that it's not part of the
             # primary data, and shouldn't be updatable by the user.
-            if k == "sc_data_version":
+            if k == "sc_data_version" or v.data is None:
```
mozbhearsum: Unfortunately, this has made it impossible to update values to None in the Rules Scheduled Changes :(. Eg: if I scheduled a change, then try to update that scheduled change to change "product" to an empty value, it accepts the update but doesn't make the change. I _think_ the right thing to do here is to make sure the UI doesn't set "data" in the request if no new file is being uploaded - the if block below this should take care of handling that correctly.

Sorry for misleading you by suggesting modifying this yesterday =/.

PR #206, opened 2017-01-04
UI for User Roles

njirap:
Hi @mozbhearsum 

This PR is still under construction. I have created it so that the reviewing can be on a rolling basis and not request for a PR review on a gigantic lumpsum.

Each commit will be the ui for one of the CRUD operations on the User roles.

This first commit is for Viewing a User's roles

### DIFF
--- a/auslib/admin/views/permissions.py
+++ b/auslib/admin/views/permissions.py
@@ -180,6 +180,14 @@ class UserRolesView(AdminView):
         return jsonify({"roles": roles})
 
 
+class AllRolesView(AdminView):
+    """/users/roles"""
+
+    def get(self):
+        roles_list = dbo.permissions.getAllRoles()
+        return jsonify({"roles_list": roles_list})
+
+
 class UserRoleView(AdminView):
     """/users/:username/roles/:role"""
 
--- a/auslib/admin/views/permissions.py
+++ b/auslib/admin/views/permissions.py
@@ -197,7 +197,8 @@ class UserRoleView(AdminView):
 
     @requirelogin
     def _delete(self, username, role, changed_by, transaction):
-        if role not in dbo.permissions.getUserRoles(username):
+        roles_list = [r['role'] for r in dbo.permissions.getUserRoles(username)]
+        if role not in roles_list:
             return Response(status=404)
 
         form = DbEditableForm(request.args)
--- a/auslib/admin/views/permissions.py
+++ b/auslib/admin/views/permissions.py
@@ -184,8 +184,8 @@ class AllRolesView(AdminView):
     """/users/roles"""
 
     def get(self):
-        roles_list = dbo.permissions.getAllRoles()
-        return jsonify({"roles_list": roles_list})
+        roles = dbo.permissions.getAllRoles()
+        return jsonify({"roles": roles})
 
 
 class UserRoleView(AdminView):
@@ -205,8 +205,8 @@ class UserRoleView(AdminView):
 
     @requirelogin
     def _delete(self, username, role, changed_by, transaction):
-        roles_list = [r['role'] for r in dbo.permissions.getUserRoles(username)]
-        if role not in roles_list:
+        roles = [r['role'] for r in dbo.permissions.getUserRoles(username)]
+        if role not in roles:
             return Response(status=404)
 
         form = DbEditableForm(request.args)
--- a/auslib/db.py
+++ b/auslib/db.py
@@ -2237,8 +2237,8 @@ class Permissions(AUSTable):
             raise ValueError('Permission "%s" doesn\'t exist' % permission)
 
     def getUserRoles(self, username, transaction=None):
-        res = self.user_roles.select(where=[self.user_roles.username == username], columns=[self.user_roles.role], distinct=True, transaction=transaction)
-        return [r["role"] for r in res]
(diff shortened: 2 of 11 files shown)


ui/app/js/controllers/user_edit_controller.js
```diff
@@ -31,6 +31,15 @@ function ($scope, $modalInstance, CSRF, Permissions, user, users) {
     $scope.loading = false;
   });
 
+  $scope.user.roles = [];
+  Permissions.getUserRoles(user.username)
+  .success(function(response) {
+    $scope.user.roles = response.roles;
+  })
+  .error(function(response) {
+    console.error(response);
```
mozbhearsum: We're trying to get away from silent errors - better to show an error in a modal dialog for this.

auslib/db.py
```diff
@@ -1970,8 +1970,8 @@ def getOptions(self, username, permission, transaction=None):
             raise ValueError('Permission "%s" doesn\'t exist' % permission)
 
     def getUserRoles(self, username, transaction=None):
-        res = self.user_roles.select(where=[self.user_roles.username == username], columns=[self.user_roles.role], distinct=True, transaction=transaction)
-        return [r["role"] for r in res]
+        res = self.user_roles.select(where=[self.user_roles.username == username], columns=[self.user_roles.role, self.user_roles.data_version], distinct=True, transaction=transaction)
+        return [{"role": r["role"], "data_version": r["data_version"]} for r in res]
```
njirap: hi @mozbhearsum I require the roles to return the `data_version` because this is required when revoking a role from the ui

mozbhearsum: Good catch!

njirap:
@mozbhearsum the latest commit ends my poking in the backend(hopefully)

auslib/test/admin/views/test_permissions.py
```diff
@@ -518,8 +518,15 @@ class TestUserRolesAPI_JSON(ViewTest):
     def testGetRoles(self):
         ret = self._get("/users/bill/roles")
         self.assertStatusCode(ret, 200)
-        got = set(json.loads(ret.data)["roles"])
-        self.assertEquals(got, set(["releng", "qa"]))
+        got = json.loads(ret.data)["roles"]
+        self.assertEquals(got, [{"role": "qa", "data_version": 1},
+                          {"role": "releng", "data_version": 1}])
+
+    def testGettAllRoles(self):
```
mozbhearsum: Nit: typo here - "Gett" vs "Get".

auslib/test/test_db.py
```diff
@@ -3256,7 +3256,7 @@ def testHasPermissionNotAllowedByProduct(self):
 
     def testGetUserRoles(self):
         got = self.permissions.getUserRoles("bob")
-        self.assertEquals(set(got), set(["releng", "dev"]))
+        self.assertEquals(got, [{'data_version': 1, 'role': u'dev'}, {'data_version': 1, 'role': u'releng'}])
```
mozbhearsum: Please continue using a set here - the order in which these are returned from the database is not guaranteed.

auslib/admin/views/permissions.py
```diff
@@ -180,6 +180,14 @@ def get(self, username):
         return jsonify({"roles": roles})
 
 
+class AllRolesView(AdminView):
+    """/users/roles"""
+
+    def get(self):
+        roles_list = dbo.permissions.getAllRoles()
+        return jsonify({"roles_list": roles_list})
```
mozbhearsum: Nit: I think just plain "roles" would be a better name here - the type of the value is obvious to consumers.

ui/app/js/controllers/user_edit_controller.js
mozbhearsum: Don't forget to address this, please!

auslib/test/test_db.py
njirap: I used `sorted` instead because a dictionary is not hashable

njirap:
@mozbhearsum this PR is no longer a work in progress..

auslib/test/test_db.py
mozbhearsum: 👍

ui/app/js/controllers/user_edit_controller.js
```diff
@@ -31,12 +31,96 @@ function ($scope, $modalInstance, CSRF, Permissions, user, users) {
...
     value.options = value.options_as_json;
   });
 
+  $scope.grantRole = function() {
+    $scope.saving = true;
+    CSRF.getToken()
+    .then(function(csrf_token) {
+      Permissions.grantRole($scope.user.username, $scope.role.role, $scope.role.data_version, csrf_token)
+      .success(function(response) {
+        $scope.user.roles.push($scope.role);
```
mozbhearsum: I think you need to parse the response for "new_data_version" and set it in the right place. Right now, if I try to revoke a role immediately after granting it, I get "data version is not a valid integer value".

mozbhearsum: You should also update $scope.roles_list if the new Role isn't already in it.

njirap:
@mozbhearsum changes made..

mozbhearsum:
Looks great now!

PR #208, opened 2017-01-09
Implement Required Signoffs tables

mozbhearsum:
This patch implements both of the Required Signoffs tables as well as checking that signoffs are made before enacting Scheduled Changes. The latter was supposed to happen in https://bugzilla.mozilla.org/show_bug.cgi?id=1310232, but I found it difficult to feel sure about the RequiredSignoffsTable implementation without doing verifications at the same time. Here's an overview of the changes:
* Create new tables needed to track Required Signoffs.
* Look for necessary signoffs before making changes to any row. This was very tricky because both enactChange and the web layer call insert/update/delete to make changes to a table. As things stood before this patch, there was no way for those methods to know whether or not signoffs have been made (because they do not have access to sc_id). I discussed this a bit with @JohanLorenzo, and he suggested passing in signoffs, which ended up working up pretty well. It lets insert/update/delete always look up whether or not signoffs are required, and only allow the change if the signoffs given are enough. I don't particularly like making the signature of these methods even larger, but I honestly can't come up with any other way to do it. Totally open to changing this if someone has a better idea!
* Looking up the signoffs required is necessarily factored out to subclasses, because only they know which Required Signoffs table to use, and how to determine how to find the products and/or channels that are affected by the change. I put all of the generic parts of that algorithm into verify_signoffs, to avoid repetition.
* Speaking of verify_signoffs, I'm not super happy about where it lives. It doesn't really make sense to attach to a class (though I suppose it could be a static method on RequiredSignoffsTable)...ideas welcome!
* I found and fixed a couple of unrelated bugs in db.py:
  * Looking up rules that map to a release in getReleaseInfo breaks when run outside of transaction in certain cases
  * Permissions checks when setting read_only sometimes don't set product correctly.
* Changes to Permissions to ensure we can't get into a state where we don't have enough users with a Role to signoff on a change.
* Tons of changes to tests to workaround failures caused by setting up Required Signoffs. I tried to keep this to a minimum to keep things readable.

I suspect this will need a few small fixes at least, but I'd appreciate some review to make sure I'm not missing anything major at this point!


auslib/db.py
```diff
@@ -341,13 +382,17 @@ def _prepareInsert(self, trans, changed_by, **columns):
         if self.onInsert:
             self.onInsert(self, "INSERT", changed_by, query, trans)
 
+        potential_signoffs = self.getPotentialSignoffs(None, columns, transaction=trans)
```
mozbhearsum: Yeah, that's not a bad idea. It also makes me wonder if it would be better to have it take a list of rows instead, since we just loop over them in getPotentialSignoffs implementations...I'm not sure there's a use case for treating old and new differently.

auslib/db.py
```diff
@@ -341,13 +382,17 @@ def _prepareInsert(self, trans, changed_by, **columns):
...
+        if potential_signoffs:
```
mozbhearsum: That's a good idea, will do.

auslib/db.py
```diff
@@ -1189,6 +1252,93 @@ def mergeUpdate(self, old_row, what, changed_by, transaction=None):
...
+            self.validate(rs, transaction=transaction)
+
+        if not self.db.hasPermission(changed_by, "required_signoff", "modify", transaction=transaction):
+            raise PermissionDeniedError("{} is not allowed to modify Required Signoffs.".format(changed_by))
+
+        return super(RequiredSignoffsTable, self).update(where=where, what=what, changed_by=changed_by, old_data_version=old_data_version,
+                                                         transaction=transaction, dryrun=dryrun, signoffs=signoffs)
+
+    def delete(self, where, changed_by=None, old_data_version=None, transaction=None, dryrun=False, signoffs=None):
+        if not self.db.hasPermission(changed_by, "required_signoff", "delete", transaction=transaction):
```
mozbhearsum: Do you mean a call to self.validate, or verify_signoffs?

auslib/test/test_db.py
```diff
@@ -2052,6 +2326,18 @@ def testUpdateReleaseChangeReadOnly(self):
     def testUpdateReleaseNoPermissionToSetReadOnly(self):
         self.assertRaises(PermissionDeniedError, self.releases.update, {"name": "c"}, {"read_only": True}, "bob", 1)
 
+    def testUpdateReleaseWithRuleMappingThatRequiresSignoff(self):
+        newBlob = ReleaseBlobV1(name="h", schema_version=1, hashFunction="sha256")
```
mozbhearsum: No good reason, it's pretty much just an arbitrary string. In some of these tests, I'm explicitly changing the current hashFunction because it seems silly to have a scheduled change (even in tests) that doesn't actually change anything.

auslib/db.py
JohanLorenzo: I mean `self.validate()`

auslib/db.py
mozbhearsum: Actually - can you expand on this a bit? What do you mean bypassing, exactly?

auslib/db.py
mozbhearsum: I don't think that would do anything useful. We always validate before making a change to the database, so we'd just be revalidating the latest database state before deleting it? Unless I'm missing something...

auslib/db.py
JohanLorenzo: Oh wait, I guess I understood this condition wrong. "No potential signoffs" means "sometimes a product (for instance) doesn't require any sign off", right? If that's the case, I initially thought there was a way to contact this layer without any signoff at all.

mozbhearsum: Ah, I think see the confusion..."potential signoffs" here actually ends up being "all required signoffs that match the affected product and/or channel". These end up getting reduced to one item per role. So the "potential" part here means that not all of the items returned by getPotentialSignoffs get used, not that we may bypass signoff entirely.

As an example, if Firefox:release requires signoff, and you try to change a Firefox:release Rule without passing any signoffs in (eg: like the web layer does when it calls Rules.update), SignoffRequiredError will be raised. When a similar call is made by enactChange, any signoffs made will be passed as well, and the change will be allowed if all the required ones have been made.

Perhaps "potential signoffs" is not the best language to use here.

Does that make sense? Did I understanding your confusion?

JohanLorenzo: You got it ;) How about `potential_required_signoffs`?

auslib/migrate/versions/021_add_required_signoffs.py
```diff
@@ -0,0 +1,187 @@
...
+    product_required_signoffs_scheduled_changes_conditions_history.append_column(Column("when", bigintType))
+
+    product_required_signoffs_scheduled_changes_signoffs = Table( # noqa
+        "product_req_signoffs_scheduled_changes_signoffs", metadata,
+        Column("sc_id", Integer, primary_key=True, autoincrement=False),
+        Column("username", String(100), primary_key=True),
+        Column("role", String(50), nullable=False),
+    )
+
+    product_required_signoffs_scheduled_changes_signoffs_history = Table( # noqa
```
nthomas-mozilla: This nearly inceptioned my head to a hard crash :1st_place_medal: Also wondered about product__req_signoffs__scheduled_changes__signoffs__history, or some other delimiter, to cope with the spaces in the names when concatenated.

auslib/test/test_db.py
```diff
@@ -1503,23 +1695,32 @@ def setUp(self):
         self.db = AUSDatabase(self.dburi)
         self.db.create()
         self.paths = self.db.rules
-        self.paths.t.insert().execute(rule_id=1, priority=100, version='3.5', buildTarget='d', backgroundRate=100, mapping='c', update_type='z', data_version=1)
-        self.paths.t.insert().execute(rule_id=2, priority=100, version='3.3', buildTarget='d', backgroundRate=100, mapping='b', update_type='z', data_version=1)
-        self.paths.t.insert().execute(rule_id=3, priority=100, version='3.5', buildTarget='a', backgroundRate=100, mapping='a', update_type='z', data_version=1)
+        self.paths.t.insert().execute(rule_id=1, priority=100, version='3.5', buildTarget='d', backgroundRate=100, mapping='c', update_type='z',
+                                      product="a", channel="a", data_version=1)
```
nthomas-mozilla: There are a lot of changes to tests adding a product and channel to rules. Do we have any tests left where they're not set ? Mainly channel I guess. Seems like we still support it.

docs/source/database.rst
```diff
@@ -219,11 +225,45 @@ Users may hold any number of Roles. Roles are used when signing off on Scheduled
...
 Roles and Permissions are not directly related - assigning a User a Role does not inherently grant them any Permissions.
 
+
+-----------------
+Required Signoffs
+-----------------
+
+Some types of changes to Balrog's database require more than one person to approve them before they can be done. The Required Signoffs tables specify how many signoffs are needed from different Roles for each type of change. For example, a change may required 3 signoffs from users that hold the "releng" Role as well as 1 signoff from a user that holds the "relman" role.
+
+Changes to Required Signoffs tables them generally require signoff as well. If you are adding, modifying, or removing signoff requirements for something that already has signoff requirements, you must obtain signoff to do so. For example, if a change requires 2 signoffs from users who hold the "releng" Role, and you want to also require signoff from 1 user who holds the "relman" Role, you must get signoff from 2 "releng" users first. The one exception to this is that if you are adding a new signoff requirement for something that doesn't require any signoff yet, you do not need any signoff to do so.
```
nthomas-mozilla: Nit, s/ them / /. Love that the docs are modified in sync with the code.

auslib/test/admin/views/test_permissions.py
```diff
@@ -32,60 +32,52 @@ def testPermissionGetMissing(self):
         self.assertEqual(ret.status_code, 404)
 
     def testPermissionPut(self):
-        ret = self._put('/users/bob/permissions/admin')
+        ret = self._put('/users/bob/permissions/admin', data=dict(options=json.dumps(dict(products=["a"]))))
```
mozbhearsum: They are. I'll fix up this coverage.

auslib/migrate/versions/021_add_required_signoffs.py
```diff
@@ -0,0 +1,187 @@
+from sqlalchemy import Table, Column, Integer, String, MetaData, \
+    BigInteger, Boolean
+
+
+def upgrade(migrate_engine):
```
mozbhearsum: The Balrog's appetite for tables in unquenchable!

auslib/migrate/versions/021_add_required_signoffs.py
mozbhearsum: Yeah, I'm not particularly happy with some of these names =\. We're actually bumping up against the default mysql table name length in some cases (which is why "required" became "req"). I'm not sure I'm a big fan of double underscore as a special delimeter, but I agree that a special delimeter would be helpful. I wonder if mysql supports emoji...

Is it just the database names that throw you? Eg: are references like dbo.productRequiredSignoffs.scheduled_changes.signoffs.history also confusing?

auslib/migrate/versions/021_add_required_signoffs.py
```diff
@@ -0,0 +1,187 @@
...
+    permissions_signoffs_scheduled_changes_signoffs_history.append_column(Column("timestamp", bigintType, nullable=False))
+
+    metadata.create_all()
+
+
+def downgrade(migrate_engine):
+    metadata = MetaData(bind=migrate_engine)
+
+    Table("product_req_signoffs", metadata, autoload=True).drop()
+    Table("product_req_signoffs_history", metadata, autoload=True).drop()
```
mozbhearsum: That's...an oversight. And it makes me wonder if there's a way we could have a test for downgrades. I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1332000 on this.

auslib/db.py
```diff
@@ -345,13 +381,17 @@ def _prepareInsert(self, trans, changed_by, **columns):
         if self.onInsert:
             self.onInsert(self, "INSERT", changed_by, query, trans)
 
+        potential_required_signoffs = self.getPotentialRequiredSignoffs([columns], transaction=trans)
+        if potential_required_signoffs:
+            verify_signoffs(potential_required_signoffs, signoffs)
```
mozbhearsum: I agree that this before the on* handlers. I'll make this change.

auslib/db.py
```diff
@@ -414,6 +457,10 @@ def _prepareDelete(self, trans, where, changed_by, old_data_version):
         if self.onDelete:
             self.onDelete(self, "DELETE", changed_by, query, trans)
 
+        potential_required_signoffs = self.getPotentialRequiredSignoffs([orig_row], transaction=trans)
+        if potential_required_signoffs:
+            verify_signoffs(potential_required_signoffs, signoffs)
```
mozbhearsum: I put these here for two reasons: 1) because it seemed a bit weird if AUSTable.insert/update/delete signatures were different than subclass methods of the same name. And 2) Because it seemed like the most obvious way to satisfy https://bugzilla.mozilla.org/show_bug.cgi?id=1310232.

I think putting these checks right beside permission checks is a good enough reason to override both of these, though, and it will make it more obvious which tables actual require signoffs. I'll give this a shot - hopefully the signatures differences don't cause any major problems.

auslib/db.py
```diff
@@ -1200,7 +1358,11 @@ def __init__(self, db, metadata, dialect, baseName):
         # update race, so they do not need to be versioned.
         super(SignoffsTable, self).__init__(db, dialect, versioned=False)
 
-    def insert(self, changed_by=None, transaction=None, dryrun=False, **columns):
+    def getPotentialRequiredSignoffs(self, affected_rows, transaction=None):
+        # Signoffs tables do not require signoffs
```
mozbhearsum: Probably not. I mainly did this to match the base class signature, but that may be going away anyways per above. This class will probably end up unchanged if the required signoff verification moves to subclasses.

auslib/test/test_db.py
```diff
@@ -1452,35 +1484,195 @@ def testRevokeOtherUsersSignoffWithoutPermission(self):
...
+        self.rs.scheduled_changes.signoffs.t.insert().execute(sc_id=3, username="bob", role="releng")
+        self.rs.scheduled_changes.signoffs.t.insert().execute(sc_id=3, username="janet", role="relman")
+        self.rs.scheduled_changes.signoffs.t.insert().execute(sc_id=3, username="nancy", role="relman")
+        self.rs.scheduled_changes.t.insert().execute(sc_id=4, scheduled_by="bob", complete=False, change_type="insert", data_version=1,
+                                                     base_product="foo", base_channel="bar", base_role="dev", base_signoffs_required=1)
+        self.rs.scheduled_changes.conditions.t.insert().execute(sc_id=4, when=300000, data_version=1)
+        self.rs.scheduled_changes.signoffs.t.insert().execute(sc_id=4, username="bob", role="releng")
+
+    def testInsertNewRequiredSignoff(self):
+        self.rs.insert(changed_by="bill", product="carrot", channel="celery", role="releng", signoffs_required=1)
```
mozbhearsum: Good call, I'll fix this.

auslib/test/test_db.py
mozbhearsum: I'll fix this new gap, too.

auslib/db.py
mozbhearsum: I didn't end up doing this because of the signoffs refactor, but I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1332412.

mozbhearsum:
I think I've fixed everything you identified. The most notable part is the reworking of where signoff checks happen - I've moved it into Rules, Releases, etc. This ended up being pretty easy. I took the opportunity to ensure that we do checks in the same order everywhere (object/data validation, permissions, then signoffs) for moar consistency. I _think_ it's cleaner than the previous version overall, but I'm curious what you think - I don't feel super strong about either. There's a couple of new ugly spots with this version:
- The duplication, of course. I ended up needing to duplicate the logic in update() that builds the new version of an object before looking for Signoffs as well.
- The extra "dryrun" check in all of the insert/update/delete methods, which was necessary avoid SignoffRequiredErrors when scheduleding changes for things that need Signoff. Until know, checking that has only been done at the AUSTable level. Checking it in the subclasses right before calling AUSTable.whatever doesn't actually change anything (we bail right away in AUSTable if it's True), but it could set a bad precedent. I considered passing it to verify_signoffs, but it didn't feel right.

Other notable changes:
- Fixed a few more cases where we run queries without a Transaction or Connection object (the signoffs refactor uncovered these).
- I added some direct tests of verify_signoffs.

@nthomas-mozilla - if you have time, I'm curious what you think of the signoffs rework.

auslib/migrate/versions/021_add_required_signoffs.py
nthomas-mozilla: The dotted camelCase is fine, and I guess we use that most of all, so lets not worry to much.

auslib/test/test_db.py
```diff
@@ -52,6 +53,93 @@ def getTempfile(self):
         return t
 
 
+class TestVerifySignoffs(unittest.TestCase):
```
nthomas-mozilla: Nice.

auslib/db.py
```diff
@@ -1188,6 +1223,114 @@ def mergeUpdate(self, old_row, what, changed_by, transaction=None):
...
+
+    def insert(self, changed_by, transaction=None, dryrun=False, signoffs=None, **columns):
+        self.validate(columns, transaction=transaction)
+
+        if not self.db.hasPermission(changed_by, "required_signoff", "create", transaction=transaction):
+            raise PermissionDeniedError("{} is not allowed to create new Required Signoffs.".format(changed_by))
+
+        if not dryrun:
+            potential_required_signoffs = self.getPotentialRequiredSignoffs([columns], transaction=transaction)
+            verify_signoffs(potential_required_signoffs, signoffs)
```
nthomas-mozilla: re the dryrun mode, which looks to be for ScheduledChangeTable.validate()'s call to _checkBaseTablePermissions(), I wondered it we could separate out the permission check to it's own method(s). Then insert() et al and _checkBaseTablePermissions() would both call them. That probably leads to more duplication of effort for updates though, so that's not very desirable. So I dunno. What precedent specifically are you worried about ?

auslib/db.py
```diff
@@ -1188,6 +1223,114 @@ def mergeUpdate(self, old_row, what, changed_by, transaction=None):
...
+
+        return super(RequiredSignoffsTable, self).update(where=where, what=what, changed_by=changed_by, old_data_version=old_data_version,
+                                                         transaction=transaction, dryrun=dryrun)
+
+    def delete(self, where, changed_by=None, old_data_version=None, transaction=None, dryrun=False, signoffs=None):
+        if not self.db.hasPermission(changed_by, "required_signoff", "delete", transaction=transaction):
+            raise PermissionDeniedError("{} is not allowed to remove Required Signoffs.".format(changed_by))
+
+        for rs in self.select(where=where, transaction=transaction):
+            if not dryrun:
```
nthomas-mozilla: Could avoid a db hit in the dryrun case by swapping these two lines over.

auslib/db.py
```diff
@@ -1490,6 +1668,11 @@ def delete(self, where, changed_by=None, old_data_version=None, transaction=None
         if not self.db.hasPermission(changed_by, "rule", "delete", product, transaction):
             raise PermissionDeniedError("%s is not allowed to delete rules for product %s" % (changed_by, product))
 
+        for current_rule in self.select(where=where, transaction=transaction):
+            if not dryrun:
```
nthomas-mozilla: Could avoid a db hit in the dryrun case by swapping these two lines over.

auslib/db.py
mozbhearsum: Specifically, I'm concerned that we may start having more and more conditionals in Rules, Releases, etc. based on dryrun or not. Eventually this could be problematic because the more differences between dryrun mode and real run mode, the fewer issues it's able to catch. I think it's probably OK not to block on this for now, but something to watch out for in the future.

Separating out the permissions checks might help, you're right. We'd have to decide how much we care about the data validation that's done as a side effect, though. Most of this is already checked at the web layer (eg, validating a blob), but things like checking whether or not a Release is mapped to before deletion wouldn't be caught at scheduling time.

PR #210, opened 2017-01-16
Initial commit of some contract tests

chartjes:
These tests validate responses by using a DTD and then validating the XML responses against it.

r? @bhearsum


api-tests/test_update_responses.py
```diff
@@ -0,0 +1,62 @@
...
+    size CDATA #REQUIRED
+    version CDATA #REQUIRED >
+    """
+    f = StringIO(dtd_info)
+    dtd = etree.DTD(f)
+
+    for url in addon_urls:
+        response = requests.get(url)
+        root = etree.XML(response.text)
+        assert dtd.validate(root)
```
mozbhearsum: It looks like dtd.validate() doesn't provide any useful output when it fails. Is there a way to have it tell you which part of the tree is broken?

api-tests/test_update_responses.py
```diff
@@ -0,0 +1,62 @@
...
+    <!ELEMENT addons (addon+) >
+    <!ELEMENT addon (#PCDATA)>
+    <!ATTLIST addon
+    id CDATA #REQUIRED
+    URL CDATA #REQUIRED
+    hashFunction CDATA #REQUIRED
+    hashValue CDATA #REQUIRED
+    size CDATA #REQUIRED
+    version CDATA #REQUIRED >
+    """
```
mozbhearsum: It looks like this dtd doesn't work when no System Addons are returned, eg: https://aus5.mozilla.org/update/3/SystemAddons/38.0/20160310153207/Darwin_x86_64-gcc3-u-i386-x86_64/en-GB/release/Darwin%2014.5.0/default/default/update.xml, which is a valid response.

api-tests/test_update_responses.py
```diff
@@ -0,0 +1,62 @@
...
+    dtd_info = """
+    <!ELEMENT updates (update) >
+    <!ELEMENT update (patch+) >
+    <!ATTLIST update
+    type CDATA #REQUIRED
+    displayVersion CDATA #REQUIRED
+    appVersion CDATA #REQUIRED
+    platformVersion CDATA #REQUIRED
+    buildID CDATA #REQUIRED
+    detailsURL CDATA #REQUIRED >
```
mozbhearsum: detailsURL is actually an optional parameter here, and there's numerous other optional parameters allowed:
- licenseURL
- everything listed in https://github.com/mozilla/balrog/blob/master/auslib/blobs/apprelease.py#L764

This is probably fine for now, but I should also note that this response changes over time as we add/remove support for different things. The required response generally tracks with the Firefox version number.

api-tests/test_update_responses.py
```diff
@@ -0,0 +1,62 @@
...
+    <!ELEMENT patch (#PCDATA)>
+    <!ATTLIST patch
+    type CDATA #REQUIRED
+    URL CDATA #REQUIRED
+    hashFunction CDATA #REQUIRED
+    hashValue CDATA #REQUIRED
+    size CDATA #REQUIRED >
+    """
```
mozbhearsum: It looks like this handles update URLs that don't have a partial, which is great.

api-tests/test_update_responses.py
```diff
@@ -0,0 +1,62 @@
+import requests
+from io import StringIO
+from lxml import etree
+
+
+def test_update_addon_responses():
+    addon_urls = [
+        'https://aus5.mozilla.org/update/3/SystemAddons/44.0/20160310153207/Darwin_x86_64-gcc3-u-i386-x86_64/en-GB/release/Darwin%2014.5.0/default/default/update.xml',
```
chartjes: The tests should be run using 'py.test  api-tests/' which you should be able to do via CI. I will do up some longer instructions.

We don't have to hardcode the server at all, can definitely be passed in. Our choices are via the CLI or in a configuration file grouped by environment and then we pass in the environment as a CLI switch.

api-tests/test_update_responses.py
chartjes: Ah, okay. I can definitely make those values optional so it does pass.

api-tests/test_update_responses.py
chartjes: There is a command that can throw any validation errors, I will find out the best way to display that on test failure.

api-tests/test_update_responses.py
chartjes: Thanks for that list of optional parameters

api-tests/test_update_responses.py
mozbhearsum: Ah, duh. I don't have a strong preference on which route we take here.

Eventually, we may want to support querying with Flask's test client (like we do for other CI tests: https://github.com/mozilla/balrog/blob/c6b8b919362ace0d819018cd602bafb5d9bfbdd5/auslib/test/web/test_client.py#L617), but that's not something we need to block on. There will probably be other changes needed when we do CI integration.

chartjes:
r? @bhearsum 

* Refactored to put those DTD's in an external file to avoid cluttering up the tests
* Added in some logic to display which validations failed on test failure
* Modified DTD's to account for more parameters

Let me know if I am missing other optional parameters

requirements-test.txt
```diff
@@ -20,3 +20,5 @@ pytest-cov==2.3.1 \
 #    --hash=sha256:fa0a212283cdf52e2eecc24dd6459bb7687cc29adb60cb84258fab73be8dda0f
 pytest-xdist==1.15.0 \
 #    --hash=sha256:6238395f8bd050f9288a3b10f34330edece80f4424cf2b4204d6e7d622f0f00b
+requests==2.12.4
+lxml==3.7.2
```
mozbhearsum: Thanks for pinning these, can you specify hashes as well, like we do for other packages?

api-tests/addon.dtd
```diff
@@ -0,0 +1,3 @@
+<!ELEMENT update (addons) >
```
mozbhearsum: Is this file used? I don't see any references to it. Let's kill it if not.

chartjes:
r? @bhearsum 

Removed that unneeded DTD file (added by mistake) and updated all the version requirements and associated hashes for the pip packages needed for testing.

requirements-test.txt
```diff
@@ -1,22 +1,38 @@
 certifi==2016.8.31 \
-#    --hash=sha256:f7708a42d86f29ccc7c8c4ff9d34a8d854d8d78eb2973d1f28406bb43d6b8919
```
mozbhearsum: Ack, I didn't realize things were in such a sorry state here. There's test failures with these changes, and it's really not your fault. Let's just revert this back to the old state, and I'll deal with fixing hashes somewhere else. I think I can push to your branch to fix this and save you the trouble...

mozbhearsum:
Thanks for doing this as a PR, @chartjes!

For posterity, I've filed https://bugzilla.mozilla.org/show_bug.cgi?id=1331992 on the hash checking bustage and https://bugzilla.mozilla.org/show_bug.cgi?id=1331993 to track integrating contract tests with CI.

PR #211, opened 2017-01-16
 Bug 1308625 - improve active data extraction script to grab referenc…

aksareen:
The extraction script  is modified to include partial release names and data as well in the sql dump. The idea behind the extraction of partial release from the json data fields : the references to partial releases is always either simple json object where each key denotes the partial release name (in case under fileUrls json) or is a list of releases where (such as in platforms) "from" points to the referred release value.


scripts/manage-db.py
```diff
@@ -130,22 +150,64 @@ def extract_active_data(url, dump_location='dump.sql'):
...
+            SELECT rules.fallbackMapping \
+            FROM rules \
+            WHERE rules.fallbackMapping IS NOT NULL \
+            UNION \
+            SELECT rules.whitelist \
+            FROM rules \
+            WHERE rules.whitelist IS NOT NULL"""
+
+    popen(_strip_multiple_spaces('''%s %s releases --where="releases.name IN (%s)" \
+        >> %s''' % (mysql_default_command, database, query_release_mapping, dump_location)))
```
mozbhearsum: Does this do the same thing that query from line 125 does? If so, please integrate the two. Also note that you need to look at rules_scheduled_changes's base_mapping, base_fallbackMapping, and base_whitelist as well.

scripts/manage-db.py
```diff
@@ -86,7 +87,26 @@ def cleanup_releases_history(trans, dryrun=True):
     print "Total Deleted: %d" % total_deleted
 
 
-def extract_active_data(url, dump_location='dump.sql'):
+def _extract_partials(json_object):
+    """Returns a generator that contains \
+    all value objects having key as 'partials'"""
```
mozbhearsum: Very clever usage of a recursive function here!

scripts/manage-db.py
```diff
@@ -130,22 +150,64 @@ def extract_active_data(url, dump_location='dump.sql'):
...
+    partial_release_names = []
+    for row in result:
+        row_data = None
+        try:
+            row_data = json.loads(row['data'])
+        except ValueError:
+            continue
+        partials_generator = _extract_partials(row_data)
+        for item in partials_generator:
+            if isinstance(item, dict):
```
mozbhearsum: It's a little bit weird to me that _extract_partials only does half the job of extracting the partial release names. I think this would read better if it returned a set of all partial names found in 'row_data'. This probably means moving this if and elif block into it, and updating "partial_release_names" with the result.

scripts/manage-db.py
```diff
@@ -130,22 +150,64 @@ def extract_active_data(url, dump_location='dump.sql'):
...
+
+    popen(_strip_multiple_spaces('''%s %s releases --where="releases.name IN (%s)" \
+        >> %s''' % (mysql_default_command, database, query_release_mapping, dump_location)))
+
+    query_release = """SELECT * \
+    FROM releases \
+    WHERE releases.name IN (%s)""" % query_release_mapping
+
+    result = trans.execute(query_release).fetchall()
+    partial_release_names = []
```
mozbhearsum: You're going to end up with a lot of duplicates, this will be much easier if you use a set() (https://docs.python.org/2/library/sets.html).

scripts/manage-db.py
```diff
@@ -86,7 +87,26 @@ def cleanup_releases_history(trans, dryrun=True):
     print "Total Deleted: %d" % total_deleted
 
 
-def extract_active_data(url, dump_location='dump.sql'):
+def _extract_partials(json_object):
+    """Returns a generator that contains \
+    all value objects having key as 'partials'"""
+    if isinstance(json_object, dict):
+        for key, value in json_object.iteritems():
+            if key == "partials":
```
mozbhearsum: Don't forget about old-style blobs, which have a singular "partial" block instead of a list of them. Eg: https://github.com/mozilla/balrog/blob/5d75609ed75b7a349b0ca5ba5faf6cd8940af775/aus-data-snapshots/3.6.12-3.6.13/Firefox-3.6.13-build3.json

scripts/manage-db.py
```diff
@@ -183,11 +245,12 @@ def _strip_multiple_spaces(string):
     elif action == 'downgrade':
         db.downgrade(options.version)
     elif action == 'extract':
-        if len(args) < 2:
-            extract_active_data(options.db)
-        else:
-            location = args[1]
-            extract_active_data(options.db, location)
+        with db.begin() as trans:
```
mozbhearsum: Good idea using a transaction, we'll probably prevent some edge cases (eg: releases existing one moment but not the next) this way.

scripts/manage-db.py
aksareen: I tried using set , however this error message kept proping up :  

mysqldump: Couldn't execute 'SELECT /*!40001 SQL_NO_CACHE */ * FROM `releases` WHERE releases.name IN (Firefox-mozilla-aurora-nightly-20160926004000, Firefox-45.0b3-build1, Firefox-48.0-build2, Thunderbird-38.4.0-build3, Firefox-mozilla-aurora-nightly-20160921004005, Thunderbird-42.0b2-build1, Firefox-49.0-build4, Firefox-oak-nightly-20160921233010, Firefox-45.0b1-build1, Firefox-mozilla-aurora-nightly-20160922004007, Firefox-mozilla-central-nightly-20160925030226, Firefox-43.0-build1, Firefox-45.0.1-build1, Firefox-43.0.4-build3, Firefox-mozilla-central-nightly-20160926030203, Firefox-50.0b1-build2, Thunderbird-45.1b1-build1, Firefox-43.0b8-build1, Firefox-44.0-build2, Firefox-45.0esr-build2, Firefox-maple-nightly-20150812100120, Firefox-49.0b2-build1, Firefox-45.1.1esr-build1, Thunderbird-38.2.0-build1, Firefox-38.0.1esr-build2, Firefox-45.3.0esr-build1, Firefox-38.3.0esr-build2, Firefox-46.0.1-build1, Firefox-49.0b10-build1, Firefox-45.0.2-build1, Firefox-48.0.2-build1, Firefox-42.0-build2, T

scripts/manage-db.py
aksareen: Checking for keyword "partial" too

scripts/manage-db.py
aksareen: query on line 125 dumps the entire release table if any rule mapping or whitelist column matches any value in the releases.name. I'll add the rules_scheduled_changes's base_mapping, base_fallbackMapping, and base_whitelist to that query and query on 169 and delete this dump statement on line 182.

scripts/manage-db.py
aksareen: Sounds good. Moving the block in next commit.

scripts/manage-db.py
mozbhearsum: I think something got cut off here, can you try pasting again?

scripts/manage-db.py
mozbhearsum: Are you sure this is true? I would expect our production dumps to have hundreds of releases in them if that were the case, and they don't as far as I can tell.

scripts/manage-db.py
aksareen: No that's the last line of the error output. Here's the complete error log : http://pastebin.com/r8dy4GFh . 

this error occurs when I add partial_rel = list(set(partial_rel))

mozbhearsum: Huh. Maybe try printing out the query before running it to see what the full version looks like? This is very weird, and would be good to get to the bottom of.

scripts/manage-db.py
aksareen: Yeah I think so. This [link](https://github.com/mozilla/balrog/pull/211/files#diff-5b9c31a834a09a47043b6c6d63cfbbf7R145) dumps the entire release table.

scripts/manage-db.py
aksareen: Thanks for the advice. Had made a mistake of using double quotes in place of single quote.

scripts/manage-db.py
```diff
@@ -86,13 +87,53 @@ def cleanup_releases_history(trans, dryrun=True):
...
+    for item in _extract_partials_generator(release_data):
+        if isinstance(item, dict):
+            map(partial_release_names.add, list(item.keys()))
+        elif isinstance(item, list):
+            for list_item in item:
+                if 'from' in list_item:
+                    partial_release_names.add(str(list_item['from']))
+        else:
+            pass
+    return partial_release_names
```
mozbhearsum: For posterity, we talked about this not working for certain types of releases on IRC: https://gist.github.com/mozbhearsum/2f9669007476a960fa4a51d50bc6e0c4

aksareen: The issue is that the code considers the possible scenarios that we had discussed forFirefox-mozilla-central-nightly-latest and Firefox-44.0-build3 . In older releases the schema changes a bit. Example : in Firefox-10.0.2-build1 , partial is present as a string. So either changes have to be made in order to account for each scenario possible into the script or to add a method in Blob class to return all releases.  I agree with @mozbhearsum  that adding a method is a better and more maintainable code choice. Working on it.

auslib/blobs/apprelease.py
```diff
@@ -306,6 +306,22 @@ def getApplicationVersion(self, platform, locale):
         may have been a pretty version for users to see"""
         return self.getExtv(platform, locale)
 
+    def getReferencedReleases(self):
+        """
+        :return: Returns set of names of partially referenced releases that the current
+        release references
+        """
+        referencedReleases = set()
+        if self.get('platforms'):
```
mozbhearsum: You can remove a lot of the nesting here by using statements like "for platform in self.get("platforms", {})".

auslib/blobs/apprelease.py
```diff
@@ -585,6 +617,23 @@ def createSnippets(self, updateQuery, update_type, whitelistedDomains, specialFo
...
+            for platform in self['platforms']:
+                if self['platforms'][platform].get('locales'):
+                    for locale in self['platforms'][platform]['locales']:
+                        if self['platforms'][platform]['locales'][locale].get('partials'):
+                            for partials in self['platforms'][platform]['locales'][locale]['partials']:
```
mozbhearsum: Nit: "partial" instead of "partials" as the variable name here - the loop only works on one at a time.

auslib/blobs/base.py
```diff
@@ -161,3 +161,10 @@ def getFooterXML(self):
         """
         footer = '</updates>'
         return footer
+
+    def getReferencedReleases(self):
+        """
+        :return: Returns set of names of partially referenced releases that the current
+        release references
+        """
+        return NotImplementedError()
```
mozbhearsum: As is, this will cause errors whenever we try to extract non-AppRelease blobs from the database. This should either return an empty set, or all other blobs should implement this and return an empty set.

auslib/test/blobs/test_apprelease.py
```diff
@@ -104,6 +104,143 @@ def setUp(self):
...
+                }
+            }
+        }
+        """)
+
+    def testGetPartialReleaseReferences_Happy_Case(self):
+        partial_releases = self.sampleReleaseBlob.getReferencedReleases()
+        self.assertTrue(4, len(partial_releases))
+        for item in (['samplePartial1', 'samplePartial2', 'samplePartial3', 'samplePartial4']):
+            self.assertTrue(item in partial_releases)
```
mozbhearsum: This test is correct, but the asserts could be simplified as:
self.assertTrue(partial_releases, ["samplePartial1", "samplePartial2", "samplePartial3", samplePartial4"])

auslib/test/blobs/test_apprelease.py
```diff
@@ -104,6 +104,143 @@ def setUp(self):
...
+            }
+        }
+        """
+        sample_release_blob = ReleaseBlobV1()
+        sample_release_blob.loadJSON(sample_release_JSON)
+        partial_releases = sample_release_blob.getReferencedReleases()
+        self.assertEquals(0, len(partial_releases))
+
+    def testGetPartialReleaseReferences_No_Partial_Release_Case(self):
+        sample_release_JSON = """
```
mozbhearsum: Awesome job on these tests, you're verifying a lot of edge cases!

scripts/manage-db.py
```diff
@@ -118,34 +120,61 @@ def extract_active_data(url, dump_location='dump.sql'):
...
+            rules_scheduled_changes.base_fallbackMapping \
+            )"""
+
+    result = trans.execute(query_release_mapping).fetchall()
+    partial_release_names = set()
+    for row in result:
+        try:
+            release_blob = createBlob(row['data'])
+            partial_release_names = partial_release_names.union(release_blob.getReferencedReleases())
+        except:
```
mozbhearsum: Bare excepts like this can be dangerous - you may eat unexpected exceptions sometimes. At the very least, you should be logging the traceback in a case like this, but I'm curious which errors you're actually expecting you might see?

scripts/manage-db.py
```diff
@@ -118,34 +120,61 @@ def extract_active_data(url, dump_location='dump.sql'):
...
+            rules_scheduled_changes.base_whitelist, \
+            rules_scheduled_changes.base_fallbackMapping \
+            )"""
+
+    result = trans.execute(query_release_mapping).fetchall()
+    partial_release_names = set()
+    for row in result:
+        try:
+            release_blob = createBlob(row['data'])
+            partial_release_names = partial_release_names.union(release_blob.getReferencedReleases())
```
mozbhearsum: I think this would read better if you used .update() instead of .union() - it will modify the set in place rather than return a new version.

scripts/manage-db.py
```diff
@@ -118,34 +120,61 @@ def extract_active_data(url, dump_location='dump.sql'):
...
+        except:
+            continue
+    if partial_release_names:
+        qry = ", ".join("'" + release_names + "'" for release_names in partial_release_names)
+        popen(_strip_multiple_spaces('%s %s releases --where="releases.name IN (%s)" \
+               >> %s' % (mysql_default_command, database, qry, dump_location)))
```
mozbhearsum: I think this will be okay, but there's a small possibility that we may need to batch this if the queries or shell commands end up too long. We'll see what happens when we run it against the production db.

mozbhearsum:
@mostlygeek - this PR contains the update to the extract script that I mentioned. Could we give it a try against production today or tomorrow?

PR #213, opened 2017-01-18
Bug 1304375 - no error in UI when release cannot be deleted

tieu:
The error message when deleting a release with rules pointing to it now shows a more comprehensive error. The response is now a 400 status when the error occurs.

Overview of the changes:

- Added new error response "HasRulesError"
- Modified UI Release Delete controller to display exceptions instead of blank message when errors occur.

### DIFF
--- a/auslib/admin/views/base.py
+++ b/auslib/admin/views/base.py
@@ -51,6 +51,10 @@ def handleGeneralExceptions(messages):
                 msg = "Permission denied to perform the request. {}".format(e.message)
                 logging.warning(msg)
                 return Response(status=403, response=json.dumps({"exception": msg}), mimetype="application/json")
+            except ValueError as e:
+                msg = "Error. {}".format(e.message)
+                logging.warning(msg)
+                return Response(status=403, response=json.dumps({"exception": msg}), mimetype="application/json")
         return decorated
     return wrap
 
--- a/auslib/admin/views/base.py
+++ b/auslib/admin/views/base.py
@@ -54,7 +54,7 @@ def handleGeneralExceptions(messages):
             except ValueError as e:
                 msg = "Error. {}".format(e.message)
                 logging.warning(msg)
-                return Response(status=403, response=json.dumps({"exception": msg}), mimetype="application/json")
+                return Response(status=400, response=json.dumps({"exception": msg}), mimetype="application/json")
         return decorated
     return wrap
 
--- a/auslib/test/admin/views/test_releases.py
+++ b/auslib/test/admin/views/test_releases.py
@@ -439,8 +439,8 @@ class TestReleasesAPI_JSON(ViewTest):
         self.assertStatusCode(ret, 403)
         
     def testDeleteWithRules(self):
-        ret = self._delete("/releases/d", qs=dict(name='ab', data_version=1))
-        self.assertStatusCode(ret, 403)
+        ret = self._delete("/releases/a", qs=dict(data_version=1))
+        self.assertStatusCode(ret, 400)
 
     def testLocalePut(self):
         data = json.dumps({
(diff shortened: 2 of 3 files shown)


auslib/test/admin/views/test_releases.py
```diff
@@ -341,13 +341,6 @@ def testReleasePostInvalidBlob(self):
         data = json.dumps(dict(uehont="uhetn", schema_version=1))
         ret = self._post("/releases/c", data=dict(data=data, product="c", data_version=1))
         self.assertStatusCode(ret, 400)
-        self.assertIn("Additional properties are not allowed", ret.data)
-
-    def testReleasePostWithSignoffRequired(self):
-        data = json.dumps(dict(bouncerProducts=dict(partial='foo'), name='a', hashFunction="sha512"))
-        ret = self._post("/releases/a", data=dict(data=data, product="a", data_version=1, schema_version=1))
-        self.assertStatusCode(ret, 400)
-        self.assertIn("This change requires signoff", ret.data)
```
mozbhearsum: Is there a reason you removed the this test, and the assert from the previous one?

tieu:
Thanks for the review, I will make the suggested changes now.

Overview of the new changes:

Removed error response "HasRulesError"
Added a new test case

auslib/admin/views/base.py
```diff
@@ -51,6 +51,10 @@ def decorated(*args, **kwargs):
                 msg = "Permission denied to perform the request. {}".format(e.message)
                 logging.warning(msg)
                 return Response(status=403, response=json.dumps({"exception": msg}), mimetype="application/json")
+            except ValueError as e:
+                msg = "Error. {}".format(e.message)
+                logging.warning(msg)
+                return Response(status=403, response=json.dumps({"exception": msg}), mimetype="application/json")
```
mozbhearsum: This should be a 400 actually - 403 is specifically for cases where somebody is not authorized to do something (eg, because of permissions).

auslib/test/admin/views/test_releases.py
```diff
@@ -437,6 +437,10 @@ def testDeleteReadOnlyRelease(self):
         dbo.releases.t.update(values=dict(read_only=True, data_version=2)).where(dbo.releases.name == "d").execute()
         ret = self._delete("/releases/d", username="bill", qs=dict(data_version=2))
         self.assertStatusCode(ret, 403)
+        
+    def testDeleteWithRules(self):
+        ret = self._delete("/releases/d", qs=dict(name='ab', data_version=1))
```
mozbhearsum: Two things here:
1) No need to include "name" in the query string - it's part of the URL.
2) As far as I can tell, "d" is not mapped to. This test is failing because it returns 200! You'll need to update this to choose one of the releases mentioned by the rules from https://github.com/mozilla/balrog/blob/master/auslib/test/admin/views/base.py#L82

mozbhearsum:
Hi @tieu, are you still looking at this?

aksareen:
Hey  @tieu, great job. 
Your checks are failing due to " auslib/test/admin/views/test_releases.py:440:1: W293 blank line contains whitespace ".
For future reference : try running "tox" or "./run-tests.sh backend" cli commands to see if the all PEP8 style checks and the tests succeed every time before pushing the changes.

ui/app/js/controllers/release_delete_controller.js
```diff
@@ -17,7 +17,7 @@ function ($scope, $modalInstance, CSRF, Releases, release, releases) {
       })
       .error(function(response) {
          if (typeof response === 'object') {
-          message = '';
+          message = response.exception;
```
aksareen: Hey @tieu, why are you setting default value to exception ?

aksareen: As discussed on IRC with tieu, the exception message in response is being assigned to the variable to be displayed on UI. Previously the message was set as blank, hence the UI wasn't displaying the error message.

mozbhearsum: Nice catch!

PR #215, opened 2017-01-21
1321036 verify column attributes

sirMackk:
This is a work in progress PR for https://bugzilla.mozilla.org/show_bug.cgi?id=1321036.

I've finished the tests that check out column attributes. In the process I've found that some columns in the models don't match up with the columns in the database, mainly when it comes to null values. We touched on this on IRC the other day and I thought I'd get the columns in questions in front of your eyes before changing the models. 

It looks like some of the "child" models are setting `nullable` to `False` - for example the `data_version` column, _but other_ "child" models have the same column's nullable attribute set to True. For example - `ScheduledChanges.data_version.nullable` differs depending on what the parent model is. 

You can run the tests on this branch (`./run-tests.sh backend auslib/test/test_db.py::TestDBModel`) (edit: or you can just go to line 3327 [in the test log](https://tools.taskcluster.net/task-group-inspector/#/ApFImzLyQgqBGcoZ-Q28sg/ApFImzLyQgqBGcoZ-Q28sg?_k=hodcw6) in Taskcluster) to get this list of differences:

 - "Property 'nullable' on 'releases_scheduled_changes_conditions.when' differs between model and migration: (model) True != (reflected) False",
 - "Property 'nullable' on 'releases_scheduled_changes_conditions.data_version' differs between model and migration: (model) False != (reflected) True",
 - "Property 'nullable' on 'product_req_signoffs_scheduled_changes_conditions.data_version' differs between model and migration: (model) False != (reflected) True",
 - "Property 'unique' on 'rules_scheduled_changes.base_alias' differs between model and migration: (model) True != (reflected) None",
 - "Property 'nullable' on 'rules_scheduled_changes.base_update_type' differs between model and migration: (model) False != (reflected) True",
 - "Property 'nullable' on 'rules_scheduled_changes.data_version' differs between model and migration: (model) False != (reflected) True",
 - "Property 'nullable' on 'permissions_scheduled_changes.base_permission' differs between model and migration: (model) True != (reflected) False",
 - "Property 'nullable' on 'permissions_scheduled_changes.base_username' differs between model and migration: (model) True != (reflected) False",
 - "Property 'nullable' on 'permissions_scheduled_changes.data_version' differs between model and migration: (model) False != (reflected) True",
 - "Property 'nullable' on 'permissions_req_signoffs_scheduled_changes_conditions.data_version' differs between model and migration: (model) False != (reflected) True",
 - "Property 'nullable' on 'permissions_scheduled_changes_conditions.when' differs between model and migration: (model) True != (reflected) False",
 - "Property 'nullable' on 'permissions_scheduled_changes_conditions.data_version' differs between model and migration: (model) False != (reflected) True",
 - "Property 'unique' on 'rules_history.alias' differs between model and migration: (model) True != (reflected) None",
 - "Property 'nullable' on 'rules_scheduled_changes_conditions.data_version' differs between model and migration: (model) False != (reflected) True",
 - "Property 'nullable' on 'product_req_signoffs_scheduled_changes.base_product' differs between model and migration: (model) True != (reflected) False",
 - "Property 'nullable' on 'product_req_signoffs_scheduled_changes.base_channel' differs between model and migration: (model) True != (reflected) False",
 - "Property 'nullable' on 'product_req_signoffs_scheduled_changes.base_role' differs between model and migration: (model) True != (reflected) False",
 - "Property 'nullable' on 'product_req_signoffs_scheduled_changes.base_signoffs_required' differs between model and migration: (model) False != (reflected) True",
 - "Property 'nullable' on 'user_roles.data_version' differs between model and migration: (model) False != (reflected) True",
 - "Property 'unique' on 'rules_scheduled_changes_history.base_alias' differs between model and migration: (model) True != (reflected) None",
 - "Property 'nullable' on 'permissions_req_signoffs_scheduled_changes.base_product' differs between model and migration: (model) True != (reflected) False",
 - "Property 'nullable' on 'permissions_req_signoffs_scheduled_changes.base_role' differs between model and migration: (model) True != (reflected) False",
 - "Property 'nullable' on 'permissions_req_signoffs_scheduled_changes.base_signoffs_required' differs between model and migration: (model) False != (reflected) True",
 - "Property 'nullable' on 'releases_scheduled_changes.base_name' differs between model and migration: (model) True != (reflected) False",
 - "Property 'nullable' on 'releases_scheduled_changes.base_product' differs between model and migration: (model) False != (reflected) True",
 - "Property 'nullable' on 'releases_scheduled_changes.base_data' differs between model and migration: (model) False != (reflected) True",
 - "Property 'nullable' on 'releases_scheduled_changes.data_version' differs between model and migration: (model) False != (reflected) True"

I'm still trying to find a system in how changes to these columns propagate from parent to child models and I was wondering if you have any advice? 

Thanks!

### DIFF
--- a/auslib/db.py
+++ b/auslib/db.py
@@ -644,6 +644,9 @@ class History(AUSTable):
                 newcol.primary_key = False
             else:
                 newcol.nullable = True
+                # Setting unique to None because SQLAlchemy marks column attribute as None
+                # unless they have been explicitely set to True or False.
+                newcol.unique = None
             self.table.append_column(newcol)
         AUSTable.__init__(self, db, dialect, history=False, versioned=False)
 
@@ -928,16 +931,26 @@ class ScheduledChangeTable(AUSTable):
             # See https://github.com/zzzeek/sqlalchemy/blob/rel_0_7/lib/sqlalchemy/schema.py#L781
             # for background.
             newcol.key = newcol.name = "base_%s" % col.name
-            # 2) Primary keys on the base table are normal columns here,
-            # and are allowed to be null (because we support adding new rows
-            # to tables with autoincrementing keys via scheduled changes).
-            # The base table's data version may also be null for the same reason.
-            if col.primary_key or newcol.name == "base_data_version":
+            # 2) Primary Key Integer Autoincrement columns from the baseTable become normal nullable
+            # columns in ScheduledChanges because we can schedule changes that insert into baseTable
+            # and the DB will handle inserting the correct value. However, nulls aren't allowed when
+            # we schedule updates or deletes -this is enforced in self.validate().
+            # For Primary Key columns that aren't Integer or Autoincrement but are nullable, we preserve
+            # this non-nullability because we need a value to insert into the baseTable when the
+            # scheduled change gets executed.
+            # Non-Primary Key columns from the baseTable become nullable and non-unique in ScheduledChanges
+            # because they aren't part of the ScheduledChanges business logic and become simple data storage.
+            if col.primary_key:
                 newcol.primary_key = False
+
+                # Only integer columns can be AUTOINCREMENT. The isinstance statement guards
+                # against false positives from SQLAlchemy.
+                if col.autoincrement and isinstance(col.type, Integer):
+                    newcol.nullable = True
+            else:
+                newcol.unique = None
                 newcol.nullable = True
-            # Notable because of its abscence, other columns retain their
-            # nullability because whether or not we're adding a new row or
-            # modifying an existing one, those NOT NULL columns are required.
+
             self.table.append_column(newcol)
 
         super(ScheduledChangeTable, self).__init__(db, dialect, history=history, versioned=True)
--- a/auslib/db.py
+++ b/auslib/db.py
@@ -2285,7 +2285,7 @@ class Permissions(AUSTable):
 
 class Dockerflow(AUSTable):
     def __init__(self, db, metadata, dialect):
-        self.table = Table('dockerflow', metadata, Column('watchdog', Integer))
+        self.table = Table('dockerflow', metadata, Column('watchdog', Integer, nullable=False))
         AUSTable.__init__(self, db, dialect, history=False, versioned=False)
 
     def getDockerflowEntry(self, transaction=None):
--- a/auslib/test/test_db.py
+++ b/auslib/test/test_db.py
@@ -4,6 +4,7 @@ import os
 import sys
 from tempfile import mkstemp
 import unittest
+import re
 
 from sqlalchemy import create_engine, MetaData, Table, Column, Integer, select, String
 from sqlalchemy.engine.reflection import Inspector
@@ -16,6 +17,7 @@ from auslib.db import AUSDatabase, AUSTable, AlreadySetupError, \
     ReadOnlyError, PermissionDeniedError, ChangeScheduledError, \
     MismatchedDataVersionError, SignoffsTable, SignoffRequiredError, \
     verify_signoffs
+import auslib.db
 from auslib.blobs.base import BlobValidationError, createBlob
 from auslib.blobs.apprelease import ReleaseBlobV1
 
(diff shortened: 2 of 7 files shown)


mozbhearsum:
Wow, there's a lot more differences than I expected here! It looks like there's a few different categories of issues, so let's group them together.

All of the data_version columns with errors are actually correct in the model. It looks like we screwed up the migration scripts for these. You'll need to create a new migration script to fix them all:
permissions_req_signoffs_scheduled_changes_conditions.data_version
permissions_scheduled_changes.data_version
permissions_scheduled_changes_conditions.data_version
product_req_signoffs_scheduled_changes_conditions.data_version
releases_scheduled_changes.data_version
releases_scheduled_changes_conditions.data_version
rules_scheduled_changes.data_version
rules_scheduled_changes_conditions.data_version
user_roles.data_version

Similarly, the "when" columns are correct in the model. It looks like I mistakenly set these to nullable=False when creating them:
permissions_scheduled_changes_conditions.when
releases_scheduled_changes_conditions.when

These columns are non-PK fields coming from a baseTable in ScheduledChangesTable. They were non-nullable in the baseTable, but they should be nullable in ScheduledChangesTable. There's a comment at https://github.com/mozilla/balrog/blob/870a1dc821f32a331d4542f22b2af93a6d8782a3/auslib/db.py#L938 about this, but that become invalid with https://bugzilla.mozilla.org/show_bug.cgi?id=1310213 -- all non PK columns coming from the baseTable should now be nullable in their ScheduledChanges version.
permissions_req_signoffs_scheduled_changes.base_signoffs_required
product_req_signoffs_scheduled_changes.base_signoffs_required
releases_scheduled_changes.base_data
releases_scheduled_changes.base_product
rules_scheduled_changes.base_update_type

It looks like History and ScheduledChangesTable need to override the "unique" property. In both cases, any columns coming from baseTable are safe to set unique to False on.
rules_history.alias
rules_scheduled_changes.base_alias
rules_scheduled_changes_history.base_alias

Finally, these columns are all correct in the migration scripts as well, and are all ScheduledChanges columns that were PK fields on their associated baseTable.
permissions_req_signoffs_scheduled_changes.base_product
permissions_req_signoffs_scheduled_changes.base_role
permissions_scheduled_changes.base_permission
permissions_scheduled_changes.base_username
product_req_signoffs_scheduled_changes.base_channel
product_req_signoffs_scheduled_changes.base_product
product_req_signoffs_scheduled_changes.base_role
releases_scheduled_changes.base_name

They all end up nullable in the model because of https://github.com/mozilla/balrog/blob/870a1dc821f32a331d4542f22b2af93a6d8782a3/auslib/db.py#L931. That code was originally written just for the Rules table (where the only PK column is autoincrement), and the logic didn't end up transferring well to these tables. The best thing to do here is to inspect the base table more deeply. Base table PK columns which are autoincrement should be nullable in the Scheduled Changes version of the column. Base table PK columns which are not autoincrement should not be nullable.

auslib/db.py
```diff
@@ -935,6 +935,11 @@ def __init__(self, db, dialect, metadata, baseTable, conditions=("time", "uptake
             if col.primary_key or newcol.name == "base_data_version":
                 newcol.primary_key = False
                 newcol.nullable = True
+            # elif newcol.name in ('base_update_type', 'base_product', 'base_data', 'data_version'):
+                # newcol.nullable = True
+
+            # if newcol.name in ('base_name', 'base_permission', 'base_username'):
+                # newcol.nullable = False
```
mozbhearsum: You shouldn't need to be hardcoding column names from specific baseTables here. My previous comment has some pointers on how to find these programatically instead.

auslib/test/test_db.py
```diff
@@ -3919,11 +3915,110 @@ def testAllTablesExist(self):
...
+                          'timetuple', )
+        cls.property_err_msg = ("Property '{property}' on '{table_name}.{column}' differs between model "
+                                "and migration: (model) {model_prop} != (reflected) {reflected_prop}")
+
+    def setUp(self):
+        NamedFileDatabaseMixin.setUp(self)
+        self.db = AUSDatabase(self.dburi)
+        self.db.metadata.create_all()
+        self.aus_db_tables = [self.db.rulesTable, self.db.releasesTable,
+                              self.db.permissionsTable, self.db.dockerflowTable]
```
mozbhearsum: It looks like this doesn't end up being used?

auslib/test/test_db.py
```diff
@@ -3919,11 +3915,110 @@ def testAllTablesExist(self):
...
+        """
+        Check to see if a column is unique using a Sqlite-specific query.
+        """
+        col_engine = col_obj.table.metadata.bind
+        table_name = col_obj.table.name
+        res = col_engine.execute("SELECT sql FROM sqlite_master WHERE name = :table_name", table_name=table_name)
+        res = res.fetchone()[0]
+
+        if re.search(r'(?:CONSTRAINT (\w+) +)?UNIQUE *\({0}\)'.format(col_obj.name), res) is None:
+            return None
```
mozbhearsum: I wonder if False is better here. All of the failures on nullability show "True != False", while uniqueness failures show "True != None".

auslib/test/test_db.py
```diff
@@ -3919,11 +3915,110 @@ def testAllTablesExist(self):
...
+                if db_py_col.unique != ref_uniq:
+                    failures.append(
+                        self.property_err_msg.format(
+                            property="unique",
+                            table_name=table_model_instance.name,
+                            column=col_name,
+                            model_prop=db_py_col.unique,
+                            reflected_prop=ref_uniq))
+
+        self.assertEqual(failures, [], str(failures))
```
mozbhearsum: I like that we have all of the errors in one assert, but I wonder if there's a way to improve the formatting such that it includes newlines - it's tough to read right now. Not a blocker, just nice to have if it's easy.

auslib/test/test_db.py
sirMackk: I went with None here because if a property hasn't been set on a column, then `some_column.unique` returns None, and setting unique=False in sqlalchemy (afaik) doesn't affect the database. This, in turn, means that reflecting a column from a database will never set unique to False.

It _is_ a bit convoluted - do you think I should add a comment to make the "why" behind this more explicit?

mozbhearsum: Ah, I see. I guess the difference with nullable is that it defaults to True instead of None? Using None here sounds like the right thing in that case, and an updated comment would help.

sirMackk:
Hey @mozbhearsum ! I've added some more commentary to answer why certain things happen and fixed a few misfiring tests.

I'd love to get feedback on removing the `testInsertWithNonNullableColumn` test. I removed it because the test doesn't make sense after updating the logic in ScheduledChanges that's responsible for base tables. 

The only way to get a null insertion error is if one of the columns was PK but not autoincrement (this would raise a ValueError in ScheduledChangesTable.validate, line ~1020 in db.py). Adding a test like this adds requires changes in the setup that break 20-something tests. What do you think would be a better solution here - is it better to set up an isolated test for this and avoid changing other tests _or_ set this test up but fix the other tests?

Thanks!

mozbhearsum:
I agree - testInsertWithNonNullableColumn only made sense when we weren't supporting scheduled deletions - removing it makes sense.

Setting up an isolated test for the non-autoincrement PK is fine as well. I think we already have one test that does that: https://github.com/mozilla/balrog/blob/b5b736220ae504b9cd00aa0e5f5b91c3bc02bb84/auslib/test/test_db.py#L885 - perhaps we should have a separate class entirely for tests that depend on a baseTable like that - that's not a blocker though, a new test in the existing class is also fine.

auslib/db.py
```diff
@@ -644,6 +644,7 @@ def __init__(self, db, dialect, metadata, baseTable):
                 newcol.primary_key = False
             else:
                 newcol.nullable = True
+                newcol.unique = None
```
mozbhearsum: An explanation of why None is used here instead of False would be good - it's not obvious.

auslib/db.py
```diff
@@ -928,16 +929,20 @@ def __init__(self, db, dialect, metadata, baseTable, conditions=("time", "uptake
...
             # for background.
             newcol.key = newcol.name = "base_%s" % col.name
-            # 2) Primary keys on the base table are normal columns here,
-            # and are allowed to be null (because we support adding new rows
-            # to tables with autoincrementing keys via scheduled changes).
-            # The base table's data version may also be null for the same reason.
-            if col.primary_key or newcol.name == "base_data_version":
+            # 2) All non PK columns from baseTable should be nullable. Base table PK columns
+            # which are autoincrement should be nullable in the Scheduled Changes version of
+            # the column. Base table PK columns which are not autoincrement should not be nullable.
```
mozbhearsum: This is OK, but it would be better to answer the question of "why" instead of just saying what the code does. Eg, something that talks about scheduled deletions having a NULL value even for columns which are non-nullable in the baseTable would answer that question for the non-PK columns.

auslib/test/test_db.py
```diff
@@ -20,6 +21,12 @@
 from auslib.blobs.apprelease import ReleaseBlobV1
 
 
+def setUpModule():
+    # This is meant to silence the debug information coming from SQLAlchemy-Migrate since
+    # AUSDatabase provides decent debugging logs by itself.
+    logging.getLogger('migrate').setLevel(logging.CRITICAL)
```
mozbhearsum: This is awesome - it really reduces the spamminess of failures. If you don't mind, could you add it to the other test files that set up databases, too?

auslib/test/test_db.py
```diff
@@ -3816,7 +3817,7 @@ def doit():
     def testOnInsertRuleSC(self):
         def doit():
             self.db.rules.scheduled_changes.insert("bob", when=2000000000000000, product="foo", channel="bar", backgroundRate=100, priority=50,
-                                                   update_type="minor", change_type="insert")
+                                                   update_type="minor", change_type="insert", base_rule_id=1)
```
mozbhearsum: Is this a necessary change? This should work without specifying rule_id, because it is autoincrement. I think base_rule_id ends up getting ignored, maybe, because insert wants columns specified without base_.

auslib/db.py
sirMackk: Thanks for this - why aren't non-integer PK columns allowed to be null? I'm looking at some recent pull requests to find out, but it would be super helpful if you could point me in the right direction or give me an idea of what I'm looking for. Thanks again!

auslib/test/test_db.py
sirMackk: Done and done :+1:

auslib/db.py
mozbhearsum: A Scheduled Change needs to know when baseTable row it will be operating on, or inserting.

For baseTable PK columns which are autoincrement integers, this will be determined automatically when the Scheduled Change insert is enacted. Technically, we shouldn't allow NULLs for these columns if change_type is update or delete (otherwise we don't know which baseTable row to update or delete) - this gets enforced in validate(): https://github.com/mozilla/balrog/blob/897c39144caafaa41c53bfdd95c70f71e66b4b03/auslib/db.py#L1017.

For baseTable PK columns which aren't autoincrement, or aren't integers, we *always* need a value, even for inserts because there's no way to create it at enact time.

Does that help?

sirMackk: Yes, it's a huge help! Let me run an updated version through you to see if I got this correct:

Primary Key Integer Autoincrement columns from the baseTable become normal nullable
columns because we can schedule changes that insert into baseTable and the DB will handle
inserting the correct value. However, nulls aren't allowed when we schedule updates or deletes - this is enforced in self.validate().
For Primary Key columns that aren't Integer or Autoincrement but are nullable, we preserve
this non-nullability because we need a value to insert into the baseTable when the
scheduled change gets executed.
Non-Primary Key columns from the baseTable become nullable and non-unique because a
scheduled deletion will have null values in all non-PK even if they are non-nullable or
unique in the baseTable.

mozbhearsum: Sorry - missed this comment until just now. Just a couple of small clarifications:
> For Primary Key columns that aren't Integer or Autoincrement but are nullable

When you say "but are nullable" are you talking about the base table or the scheduled changes table? It wouldn't be possible to make it nullable in the base table, because it woul be part of the key.

> Non-Primary Key columns from the baseTable become nullable and non-unique because a
> scheduled deletion will have null values in all non-PK even if they are non-nullable or
> unique in the baseTable.

100% right on the nullable part. When it comes to uniqueness, it's a bit more subtle. Take a case like like rules.alias. In the Rules table, we allow alias to be used in place of rule_id in some places, so uniqueness is necessary. A field like alias is used in the business logic of Rules, and that business logic requires things like uniqueness to be in place. In rules_scheduled_changes, things are different. The base table columns are _not_ really part of the business logic of a Scheduled Changes table. We need to store values in them, but the real job of Scheduled Changes is putting in place new versions of a base table row upon certain conditions being met. To do that job, Scheduled Changes needs to keep track of conditions and signoffs, but the base table columns are more or less just simple data storage.

That's probably more info than you need for the comment update, but hopefully it's helpful background.

sirMackk: Hey @mozbhearsum, thank you \_immensely\_ for explaining the context behind this. I updated the comment to factor all of this in, even re-used some of your wording because it fit in perfectly. The comment is a bit larger than the previous one, but I think it reflects the new, slightly more complex, logic in more detail.

PR #217, opened 2017-01-23
Bug 1257298 - support a list of values in rules' version field

aksareen:
https://bugzilla.mozilla.org/show_bug.cgi?id=1257298

### DIFF
--- a/auslib/admin/views/forms.py
+++ b/auslib/admin/views/forms.py
@@ -1,4 +1,5 @@
 import simplejson as json
+import re
 
 from flask_wtf import Form
 from wtforms import StringField, IntegerField, SelectField, BooleanField
@@ -103,6 +104,14 @@ def operator_validator():
     return _validator
 
 
+def _has_op(pattern):
+    # only alphanumeric characters means no operator in version.
+    if re.match('\w+', pattern):
+        return False
+    else:
+        return True
+
+
 def version_validator():
     log.debug('starting in version_validator for version')
 
@@ -111,16 +120,27 @@ def version_validator():
         # empty input
         if field.data is None:
             return
-        try:
-            op, operand = get_op(field.data)
-            version = MozillaVersion(operand)
-        except ValueError:
-            raise ValidationError("ValueError. Couldn't parse version for %s. Invalid '%s' input value" % (field.name, field.name))
-        except:
-            raise ValidationError('Invalid input for %s . No Operator or Match found.' % field.name)
-        # MozillaVersion doesn't error on empty strings
-        if not hasattr(version, 'version'):
-            raise ValidationError("Couldn't parse the version for %s. No attribute 'version' was detected." % field.name)
+        rulesVersionList = field.data.split(",")
+        has_more_than_one_version = 1 < len(rulesVersionList)
+
+        for rule_version in rulesVersionList:
+            try:
+                op, operand = get_op(rule_version)
+                version = MozillaVersion(operand)
+            except ValueError:
+                raise ValidationError("ValueError. Couldn't parse version for %s. Invalid '%s' input value"
+                                      % (field.name, field.name))
+            except:
+                raise ValidationError('Invalid input for %s . No Operator or Match found.' % field.name)
+            # MozillaVersion doesn't error on empty strings
+            if not hasattr(version, 'version'):
+                raise ValidationError("Couldn't parse the version for %s. No attribute 'version' was detected."
+                                      % field.name)
+
+            # If more than one version number present , then all versions must be alphanumeric only.
+            if has_more_than_one_version and _has_op(rule_version):
+                raise ValidationError('Invalid input for %s . All versions must have alphanumeric characters only.'
+                                      % field.name)
 
     return _validator
 
--- a/auslib/admin/views/forms.py
+++ b/auslib/admin/views/forms.py
@@ -1,5 +1,4 @@
 import simplejson as json
-import re
 import operator
 
 from flask_wtf import Form
@@ -105,14 +104,6 @@ def operator_validator():
     return _validator
 
 
-def _has_op(pattern):
-    # only alphanumeric characters means no operator in version.
-    if re.match('\w+', pattern):
-        return False
-    else:
-        return True
-
-
 def version_validator():
     log.debug('starting in version_validator for version')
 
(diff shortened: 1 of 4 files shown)


auslib/admin/views/forms.py
```diff
@@ -111,16 +120,27 @@ def _validator(form, field):
...
-            version = MozillaVersion(operand)
-        except ValueError:
-            raise ValidationError("ValueError. Couldn't parse version for %s. Invalid '%s' input value" % (field.name, field.name))
-        except:
-            raise ValidationError('Invalid input for %s . No Operator or Match found.' % field.name)
-        # MozillaVersion doesn't error on empty strings
-        if not hasattr(version, 'version'):
-            raise ValidationError("Couldn't parse the version for %s. No attribute 'version' was detected." % field.name)
+        rulesVersionList = field.data.split(",")
+        has_more_than_one_version = 1 < len(rulesVersionList)
```
mozbhearsum: This block looks like it will work, but I think it would be more readable if you returned early if there's only one version, and only used the loop if there's more than one.

Some tests in https://github.com/mozilla/balrog/blob/master/auslib/test/admin/views/test_rules.py are necessary too. Even if it's perfect now, we need to make sure we don't accidentally break it later!

auslib/admin/views/forms.py
```diff
@@ -111,16 +120,27 @@ def _validator(form, field):
...
+                                      % (field.name, field.name))
+            except:
+                raise ValidationError('Invalid input for %s . No Operator or Match found.' % field.name)
+            # MozillaVersion doesn't error on empty strings
+            if not hasattr(version, 'version'):
+                raise ValidationError("Couldn't parse the version for %s. No attribute 'version' was detected."
+                                      % field.name)
+
+            # If more than one version number present , then all versions must be alphanumeric only.
+            if has_more_than_one_version and _has_op(rule_version):
```
mozbhearsum: I don't think _has_op needs to exist here - just check if "op" is operator.eq or not. It's probably best to do this at the start, too - the other validation doesn't matter if there's an operator present.

auslib/admin/views/forms.py
```diff
@@ -111,16 +120,27 @@ def _validator(form, field):
...
+                raise ValidationError('Invalid input for %s . All versions must have alphanumeric characters only.'
```
mozbhearsum: Nit on this message: What you've tested is whether or not there's an operator present, so that's what the error message should talk about. Eg, something like: "Operators are not allowed when providing a list of values".

auslib/test/test_db.py
```diff
@@ -2228,6 +2230,33 @@ def testGetRulesMatchingQueryVersionComparison(self):
...
+                     buildTarget='', buildID='', locale='', osVersion='',
+                     distribution='', distVersion='', headerArchitecture='',
+                     force=False, queryVersion=3,
+                     ),
+                fallbackChannel=''
+            )
+            rules = self._stripNullColumns(rules)
+            self.assertEquals(rules, expected)
+
+        # version number isn't present in the list of version numbers.
```
mozbhearsum: Very good job on these tests - you're covering most the cases. Small nit here: this test of the negative case should be a separate test. Generally, you should only be calling the code under test once per test method (the loop above is OK because you are testing the same expected result, just with different input values).

One other case to test too: ensuring that a partial version match doesn't match a longer version. Eg: if the version is set to "30.0.1,30.0.2,30.0.2", passing a version of "30.0" to getRulesMatchingQuery shouldn't come up with a match.

auslib/db.py
```diff
@@ -1465,7 +1465,13 @@ def _versionMatchesRule(self, ruleVersion, queryVersion):
         self.log.debug('ruleVersion: %s, queryVersion: %s', ruleVersion, queryVersion)
         if ruleVersion is None:
             return True
-        return version_compare(queryVersion, ruleVersion)
+        rulesVersionList = ruleVersion.split(",")
+        if 1 == len(rulesVersionList):
```
mozbhearsum: Minor nit: 1 should be on the right hand side of this comparison.

auslib/db.py
```diff
@@ -1465,7 +1465,13 @@ def _versionMatchesRule(self, ruleVersion, queryVersion):
...
+            return version_compare(queryVersion, rulesVersionList[0])
+        versionMatchesRule = False
+        for rule in rulesVersionList:
+            versionMatchesRule = versionMatchesRule or version_compare(queryVersion, rule)
```
mozbhearsum: I don't think there's a need for versionMatchesRule here - simply returning True if version_compare() returns True is fine. If you already found a match, there's no point in testing the remaining values!

auslib/admin/views/forms.py
```diff
@@ -103,6 +105,14 @@ def _validator(form, field):
     return _validator
 
 
+def _has_op(pattern):
```
mozbhearsum: This is dead now, so you can remove it :).

auslib/db.py
```diff
@@ -1465,7 +1465,13 @@ def _versionMatchesRule(self, ruleVersion, queryVersion):
...
         if ruleVersion is None:
             return True
-        return version_compare(queryVersion, ruleVersion)
+        rulesVersionList = ruleVersion.split(",")
+        versionMatchesRule = False
+        for rule in rulesVersionList:
+            versionMatchesRule = version_compare(queryVersion, rule)
+            if versionMatchesRule:
+                break
+        return versionMatchesRule
```
mozbhearsum: This is an improvement, but it can still be simplified by getting rid of the break and just returning from within the loop.

auslib/test/admin/views/test_rules.py
```diff
@@ -112,8 +112,17 @@ def testBuildIDValidation(self):
...
+            ret = self._post('/rules', data=dict(backgroundRate=42, mapping='d', priority=50,
+                                                 product='Firefox', channel="nightly",
+                                                 update_type='minor', version=validVersionList))
+            self.assertEquals(ret.status_code, 200,
+                              "Status Code: %d, Data: %s, Input: %s" % (ret.status_code, ret.data, validVersionList))
+
     def testVersionValidationBogusInput(self):
-        for bogus in ('<= 4.0', ' <=4.0', '<>4.0', '<=-4.0', '=4.0', '> 4.0', '>= 4.0', ' >=4.0', ' 4.0 '):
+        for bogus in ('<= 4.0', ' <=4.0', '<>4.0', '<=-4.0', '=4.0', '> 4.0', '>= 4.0', ' >=4.0', ' 4.0 ',
+                      '<=4.0,3.0', '>=4.0,3.0', '4.0,<3.0', '4.0,>3.0', '=4.0,3.0', '=4.0,=3.0', '4.0,=3.0'):
```
mozbhearsum: Nice enhancement to this test.

auslib/admin/views/forms.py
```diff
@@ -111,17 +121,28 @@ def _validator(form, field):
...
+                raise ValidationError("ValueError. Couldn't parse version for %s. Invalid '%s' input value"
+                                      % (field.name, field.name))
+            except:
+                raise ValidationError('Invalid input for %s . No Operator or Match found.' % field.name)
+            # MozillaVersion doesn't error on empty strings
+            if not hasattr(version, 'version'):
+                raise ValidationError("Couldn't parse the version for %s. No attribute 'version' was detected."
+                                      % field.name)
+        if has_relational_op:
+            raise ValidationError('Invalid input for %s .Relational Operators are not allowed'
```
mozbhearsum: This would be more readable if you got rid of has_relational_op and raised this inside the if condition on line 130. Raising it so far away from the if condition makes it seem like other things may modify the value as well.

PR #218, opened 2017-01-24
Web APIs for Required Signoffs

mozbhearsum:
Because we have two nearly identical objects, I took a similar approach to Scheduled Changes and used a base class to implement the logic for these. Other than that, I think much of this work is pretty boilerplate-y. I'm not terribly happy with all of the duplication in the Forms, but each has its own subtle differences, and it gets to be a mess when trying to use shared base classes. Swagger may help here, because it should let us more easily separate field definition, fields-for-a-given-endpoint, and optional vs required.

I realize these are a little bit difficult to review without UI, but I'm hoping to get initial feedback while I continue working on the UI.


mozbhearsum:
I've been using the latest version of this with my "rs-ui" branch, and the APIs have worked as expected so far (though the UI is not using all of them just yet).

mozbhearsum:
I'll probably be pushing an update or two to this. During UI development I discovered a couple more endpoints that would be helpful.

mozbhearsum:
Apologies for the pushes after requesting for review - I added the other changes to the API (beyond Required Signoffs CRUD) that I discovered were necessary while working on the UI. Speaking of which, it's pretty far along, and may help in reviewing this. Running my "rs-ui" branch will let you use it at http://localhost:8080/required_signoffs. The Edit/Delete buttons may not work yet, but the other functionality should.

auslib/admin/views/scheduled_changes.py
```diff
@@ -26,18 +26,26 @@ def get(self):
...
+                        base_row[k] = v
+                    r[k] = v
+
+            for signoff in self.sc_table.signoffs.select({"sc_id": row["sc_id"]}):
+                r["signoffs"][signoff["username"]] = signoff["role"]
+
+            # No point in retrieving this for completed scheduled changes...
+            if not row["complete"]:
+                for rs in self.table.getPotentialRequiredSignoffs([base_row]):
+                    r["required_signoffs"][rs["role"]] = max(r["required_signoffs"].get(rs["role"], 0), rs["signoffs_required"])
```
JohanLorenzo: Bikeshed: I don't know if that helps in understanding the logic, but I'd split this line into 2:
``` py
number_of_required_signoff_for_current_role = max(r["required_signoffs"].get(rs["role"], 0), rs["signoffs_required"])
r["required_signoffs"][rs["role"]] = number_of_required_signoff_for_current_role
```

What do you think?

auslib/test/admin/views/base.py
```diff
@@ -53,10 +53,27 @@ def setUp(self):
         dbo.permissions.user_roles.t.insert().execute(username="bill", role="releng", data_version=1)
         dbo.permissions.user_roles.t.insert().execute(username="bill", role="qa", data_version=1)
         dbo.permissions.user_roles.t.insert().execute(username="bob", role="relman", data_version=1)
+        dbo.permissions.user_roles.t.insert().execute(username="julie", role="releng", data_version=1)
```
JohanLorenzo: I'm not sure to understand what test case Julie helps testing. (I have nothing against Julies 😉 , but I only see her added to `testUsers`)

auslib/test/admin/views/test_required_signoffs.py
```diff
@@ -0,0 +1,923 @@
...
+            .where(dbo.productRequiredSignoffs.scheduled_changes.signoffs.sc_id == 2).execute().fetchall()
+        self.assertEquals(len(r), 2)
+        self.assertEquals(dict(r[0]), {"sc_id": 2, "username": "bill", "role": "releng"})
+        self.assertEquals(dict(r[1]), {"sc_id": 2, "username": "bob", "role": "relman"})
+        r = dbo.productRequiredSignoffs.scheduled_changes.signoffs.history.t.select()\
+            .where(dbo.productRequiredSignoffs.scheduled_changes.signoffs.history.sc_id == 2).execute().fetchall()
+        self.assertEquals(len(r), 4)
+        self.assertEquals(dict(r[0]), {"change_id": 3, "changed_by": "bill", "timestamp": 60, "sc_id": 2, "username": "bill", "role": None})
+        self.assertEquals(dict(r[1]), {"change_id": 4, "changed_by": "bill", "timestamp": 61, "sc_id": 2, "username": "bill", "role": "releng"})
+        self.assertEquals(dict(r[2]), {"change_id": 10, "changed_by": "bob", "timestamp": 99999, "sc_id": 2, "username": "bob", "role": None})
```
JohanLorenzo: Is it the behavior that we talked about yesterday? The one where we're force to stack insertions in the database?

auslib/admin/base.py
```diff
@@ -77,6 +91,7 @@ def add_security_headers(response):
 # hosted at "/api", which is stripped away by the web server before we see
 # these requests.
 app.add_url_rule("/csrf_token", view_func=CSRFView.as_view("csrf"))
+app.add_url_rule("/current_user", view_func=CurrentUserView.as_view("current_user"))
```
mozbhearsum: /users/:id actually doesn't exist (though maybe it should...). I do like the idea of having this endpoint somewhere in the /users namespace, though I'm not sure I like having a magic string there. I think I'll give this a try and see how it looks/feels.

auslib/admin/views/forms.py
```diff
@@ -374,3 +374,101 @@ class EditScheduledChangeExistingReleaseForm(ScheduledChangeTimeForm):
     data = JSONStringField(None, 'Data', validators=[Optional()], widget=FileInput())
     data_version = IntegerField('data_version', widget=HiddenInput())
     sc_data_version = IntegerField('sc_data_version', validators=[InputRequired()], widget=HiddenInput())
+
+
+class ProductRequiredSignoffForm(Form):
+    product = StringField('Product', validators=[Length(0, 15), InputRequired()])
```
mozbhearsum: Yeah, I hear you. I'm sure this is easy if we were to switch to Django or some other full stack ORM...but that's not going to happen. We might be able to reference database tables here instead of copying the value over - I'm not sure how easy it is to get a reference to things like column length. Sounds like something that would be good to address as part of https://bugzilla.mozilla.org/show_bug.cgi?id=1336452, unless you think it's pressing.

auslib/admin/views/forms.py
```diff
@@ -374,3 +374,101 @@ class EditScheduledChangeExistingReleaseForm(ScheduledChangeTimeForm):
...
+    channel = StringField('Channel', validators=[Length(0, 75), InputRequired()])
+    role = StringField('Role', validators=[InputRequired()])
+
+
+class ScheduledChangeNewProductRequiredSignoffForm(ScheduledChangeTimeForm):
+    product = StringField('Product', validators=[Length(0, 15), InputRequired()])
+    channel = StringField('Channel', validators=[Length(0, 75), InputRequired()])
+    role = StringField('Role', validators=[InputRequired()])
+    signoffs_required = IntegerField("Signoffs Required", validators=[InputRequired()])
+    change_type = SelectField("Change Type", choices=[('insert', 'insert'), ('update', 'update'), ('delete'), ('delete')])
```
mozbhearsum: Oops. To be honest, this is just copy/paste. I think the two element tuple defines the value + label for each <option>...

auslib/admin/views/required_signoffs.py
```diff
@@ -0,0 +1,272 @@
...
+    def _post(self, sc_id, transaction, changed_by):
+        return super(EnactProductRequiredSignoffScheduledChangeView, self)._post(sc_id, transaction, changed_by)
+
+
+class ProductRequiredSignoffScheduledChangeSignoffsView(SignoffsView):
+    def __init__(self):
+        super(ProductRequiredSignoffScheduledChangeSignoffsView, self).__init__("product_req_signoffs", dbo.productRequiredSignoffs)
+
+    @requirelogin
+    def _post(self, sc_id, transaction, changed_by):
```
mozbhearsum: I think you're probably right that I could move requirelogin to the base class - I'll give it a try.

auslib/admin/views/required_signoffs.py
```diff
@@ -0,0 +1,272 @@
...
+        change_type = request.json.get("change_type")
+
+        if change_type == "update":
+            form = ScheduledChangeExistingPermissionsRequiredSignoffForm()
+        elif change_type == "insert":
+            form = ScheduledChangeNewPermissionsRequiredSignoffForm()
+        elif change_type == "delete":
+            form = ScheduledChangeDeletePermissionsRequiredSignoffForm()
+        else:
+            return Response(status=400, response="Invalid or missing change_type")
```
mozbhearsum: Yes!

auslib/admin/views/scheduled_changes.py
```diff
@@ -26,18 +26,26 @@ def get(self):
             rows = self.sc_table.select(where={"complete": False})
         ret = {"count": len(rows), "scheduled_changes": []}
         for row in rows:
-            # TODO: Probably need to return signoffs still required after
-            # that's been implemented. That + existing signoffs may end up
-            # in the same data structure.
-            r = {"signoffs": {}}
-            for signoff in self.sc_table.signoffs.select({"sc_id": row["sc_id"]}):
-                r["signoffs"][signoff["username"]] = signoff["role"]
+            r = {"signoffs": {}, "required_signoffs": {}}
```
mozbhearsum: Will do.

auslib/admin/views/scheduled_changes.py
mozbhearsum: I'd like to shorten the variable name, but I agree that it improves readability.

auslib/test/admin/views/test_required_signoffs.py
mozbhearsum: Hmm, I don't think so. This test just validates that a Signoff of an already scheduled change works. There's no tests that fully simulate the "stacking". There's really nothing special about it on the backend (things that require signoff must be scheduled, just like with Rules and Releases). The thing that makes it a bit different with Required Signoffs is that the UI allows you to do multiple things at once - so one frontend operation may result in multiple requests to the backend.

auslib/test/admin/views/base.py
mozbhearsum: She's not used directly, but we need RelEng to have more than one user for the Required Signoffs tests to work. I'll add a comment to that effect.

mozbhearsum:
@JohanLorenzo - I think I've fixed up everything you pointed out.

auslib/admin/views/permissions.py
```diff
@@ -26,6 +26,21 @@ def get(self):
...
 
+class CurrentUserView(AdminView):
+    """Returns all of the details about the logged in user. The UI needs this
+    method to know things about the current user because it does not have
+    access to REMOTE_USER, so it cannot query directly by name."""
+
+    def get(self, username):
+        if username == "current":
+            username = request.environ.get('REMOTE_USER', request.environ.get("HTTP_REMOTE_USER"))
+        permissions = dbo.permissions.getUserPermissions(username)
```
mozbhearsum: This would be to try to help minimize the possibility of privilege escalations? Eg: someone gets in as ffxbld, and then uses this endpoint to figure out which account to target to gain admin access?

auslib/admin/base.py
```diff
@@ -114,3 +134,22 @@ def add_security_headers(response):
...
                  view_func=ReleaseScheduledChangeHistoryView.as_view("scheduled_change_releases_history"))
+app.add_url_rule("/scheduled_changes/required_signoffs/product", view_func=ProductRequiredSignoffsScheduledChangesView.as_view("scheduled_changes_product_rs"))
+app.add_url_rule("/scheduled_changes/required_signoffs/product/<int:sc_id>",
+                 view_func=ProductRequiredSignoffScheduledChangeView.as_view("scheduled_change_product_rs"))
+app.add_url_rule("/scheduled_changes/required_signoffs/product/<int:sc_id>/enact",
+                 view_func=EnactProductRequiredSignoffScheduledChangeView.as_view("enact_scheduled_change_product_rs"))
+app.add_url_rule("/scheduled_changes/required_signoffs/product/<int:sc_id>/signoffs",
+                 view_func=ProductRequiredSignoffScheduledChangeSignoffsView.as_view("scheduled_change_product_rs_signoffs"))
+app.add_url_rule("/scheduled_changes/required_signoffs/product/<int:sc_id>/revisions",
+                 view_func=ProductRequiredSignoffScheduledChangeHistoryView.as_view("scheduled_change_product_rs_history"))
```
mozbhearsum: Good catch - we do not. The APIs to support this are probably trivial to write though. If it's not pressing, might make a good contributor bug.

auslib/test/admin/views/base.py
```diff
@@ -121,14 +138,14 @@ def _getHttpRemoteUserAuth(self, username):
     def _getAuth(self, username):
         return {'REMOTE_USER': username}
 
-    def _get(self, url, qs={}):
+    def _get(self, url, qs={}, username="bill"):
```
mozbhearsum: Good point, will do!

auslib/test/admin/views/test_permissions.py
```diff
@@ -250,17 +319,17 @@ def testGetScheduledChanges(self):
...
                     "sc_id": 1, "when": 10000000, "scheduled_by": "bill", "change_type": "insert", "complete": False, "sc_data_version": 1,
                     "permission": "rule", "username": "janet", "options": {"products": ["foo"]}, "data_version": None,
-                    "signoffs": {"bill": "releng"},
+                    "signoffs": {"bill": "releng"}, "required_signoffs": {},
                 },
                 {
                     "sc_id": 2, "when": 20000000, "scheduled_by": "bill", "change_type": "update", "complete": False, "sc_data_version": 1,
                     "permission": "release_locale", "username": "ashanti", "options": None, "data_version": 1,
-                    "signoffs": {"bill": "releng", "mary": "relman"},
+                    "signoffs": {"bill": "releng", "mary": "relman"}, "required_signoffs": {"releng": 1, "relman": 1},
```
mozbhearsum: That's right. No product specified means they've got access to all products.

auslib/test/admin/views/test_required_signoffs.py
```diff
@@ -0,0 +1,923 @@
...
+        cond = dbo.productRequiredSignoffs.scheduled_changes.conditions.t.select()\
+            .where(dbo.productRequiredSignoffs.scheduled_changes.conditions.sc_id == 5).execute().fetchall()
+        self.assertEquals(len(cond), 1)
+        cond_expected = {"sc_id": 5, "data_version": 1, "when": 400000000}
+        self.assertEquals(dict(cond[0]), cond_expected)
+
+    @mock.patch("time.time", mock.MagicMock(return_value=300))
+    def testUpdateScheduledChangeExistingRequiredSignoff(self):
+        data = {
+            "signoffs_required": 1, "data_version": 1, "sc_data_version": 1, "when": 200000000,
```
mozbhearsum: Yes! This was implemented in https://bugzilla.mozilla.org/show_bug.cgi?id=1333095.

auslib/admin/views/permissions.py
nthomas-mozilla: Yes, that kind of thing. I was mentally comparing to other systems where I'm not an admin and can't go snooping around.

auslib/admin/base.py
nthomas-mozilla: It fits into the 'add an audit trail' kick I've been on for a while, doesn't need to block here.

auslib/admin/views/permissions.py
mozbhearsum: I think that makes a lot of sense, let's do it!

mozbhearsum: Implementing this in the most robust way (at the database layer) requires an interface change to AUSTable.select. I'm not sure we should scope creep this PR to do that, so I did a quick implementation in this endpoint, and filed https://bugzilla.mozilla.org/show_bug.cgi?id=1340167 to move it to the database layer. I'd like to get it fixed soon, but the status quo allows this already, so it doesn't seem extremely pressing.

auslib/admin/base.py
mozbhearsum: https://bugzilla.mozilla.org/show_bug.cgi?id=1340170 and https://bugzilla.mozilla.org/show_bug.cgi?id=1340171 for this.

auslib/test/admin/views/test_permissions.py
mozbhearsum: ...and because they've got access to all products, we must take all required signoffs into consideration. I have a feeling we're going to want to change this a little bit in the future. Eg: if end up with a "seamonkey" role, and I don't think we should require their signoff to a a new full fledged admin user.

auslib/test/admin/views/test_releases.py
```diff
@@ -1033,6 +1033,8 @@ class TestReleasesScheduledChanges(ViewTest):
 
     def setUp(self):
         super(TestReleasesScheduledChanges, self).setUp()
+        # TODO: Although sc_id has a signoff, it's actually not required. We should have a test
+        # that actually tests a scheduled change that requires a signoff
```
mozbhearsum: Was this meant for test_required_signoffs.py? I just realized there's no follow-up to this in test_releases.py.

auslib/test/admin/views/test_required_signoffs.py
```diff
@@ -0,0 +1,923 @@
...
+        self.assertEquals(got["required_signoffs"], expected)
+
+
+class TestProductRequiredSignoffsScheduledChanges(ViewTest):
+    maxDiff = 10000
+
+    def setUp(self):
+        super(TestProductRequiredSignoffsScheduledChanges, self).setUp()
+        dbo.productRequiredSignoffs.scheduled_changes.t.insert().execute(
+            sc_id=1, scheduled_by="bill", change_type="insert", data_version=1, base_product="fake", base_channel="c", base_role="releng",
```
mozbhearsum: Good point, we will not. I'll improve this to be more realistic.

auslib/test/admin/views/test_required_signoffs.py
```diff
@@ -0,0 +1,923 @@
...
+                                                         .where(dbo.productRequiredSignoffs.product == "fake")\
+                                                         .where(dbo.productRequiredSignoffs.role == "releng")\
+                                                         .execute().fetchall()[0]
+        base_expected = {
+            "product": "fake", "channel": "a", "role": "releng", "signoffs_required": 2, "data_version": 2
+        }
+        self.assertEquals(dict(base_row), base_expected)
+
+    def testEnactScheduledChangeNewRequiredSignoff(self):
+        ret = self._post("/scheduled_changes/required_signoffs/product/1/enact")
```
mozbhearsum: Fixing this.

auslib/test/admin/views/test_releases.py
mozbhearsum: Probably not, since this TODO is in test_releases.py....

mozbhearsum:
Alright, I think I've fixed up all of the issues you've both identified. Any reason I can't go ahead and merge this now, @JohanLorenzo and @nthomas-mozilla?

JohanLorenzo:
I don't see any 😃

nthomas-mozilla:
Go for it

PR #219, opened 2017-01-26
Configure logging in auslib.web component

tdz:
Logging in auslib.web currently doesn't handle missing 'requestid'
keys gracefully, which results in a KeyError and logs of stack traces.
Switching the logger to BalrogLogger fixes this problem.

### DIFF
--- a/auslib/web/base.py
+++ b/auslib/web/base.py
@@ -1,3 +1,6 @@
+from auslib.log import configure_logging
+configure_logging()
+
 import logging
 log = logging.getLogger(__name__)
 
--- a/uwsgi/public.wsgi
+++ b/uwsgi/public.wsgi
@@ -2,7 +2,6 @@ import logging
 import os
 
 from auslib.log import configure_logging
-from auslib.web.base import sentry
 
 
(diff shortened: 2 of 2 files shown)


tdz:
Without this patch, I see error output like:

balrogpub_1    | Traceback (most recent call last):
balrogpub_1    |   File "/usr/local/lib/python2.7/logging/__init__.py", line 861, in emit
balrogpub_1    |     msg = self.format(record)
balrogpub_1    |   File "/usr/local/lib/python2.7/logging/__init__.py", line 734, in format
balrogpub_1    |     return fmt.format(record)
balrogpub_1    |   File "/usr/local/lib/python2.7/logging/__init__.py", line 469, in format
balrogpub_1    |     s = self._fmt % record.__dict__
balrogpub_1    | KeyError: 'requestid'
balrogpub_1    | Logged from file AUS.py, line 71

mozbhearsum:
This sounds like https://bugzilla.mozilla.org/show_bug.cgi?id=1187735.

It looks like the problem is that we import auslib.web.base before we configure logging in https://github.com/mozilla/balrog/blob/master/uwsgi/public.wsgi#L5. It looks like adjusting the import order in this file fixes it, and lets us continue to configure logging in the entry point, which is good.

Thanks for doing the hard work here!

PR #220, opened 2017-01-26
Improve text on button for creating new objects with a Scheduled Change.

mozbhearsum:
@MihaiTabara - does this text make the button's purpose seem more obvious to you?


MihaiTabara:
@mozbhearsum Yes, thanks for tweaking this up! "Create" is more intuitive than "add" 👍

PR #225, opened 2017-01-27
Ui for scheduled permission

NinadBhat:


ui/app/js/controllers/permission_scheduled_change_new_controller.js
```diff
@@ -0,0 +1,241 @@
...
+        p.options_as_json = JSON.stringify(p['options']);
+      }
+    });
+    $scope.sc.permissions = permissions;
+    // console.log('$scope.sc.permissions');
+    // console.dbg($scope.sc.permissions);
+    $scope.loading = false;
+  });
+  $scope.sc.roles = [];
+  Permissions.getUserRoles(sc.username)
```
mozbhearsum: I don't think you need to worry about getting Roles here - they're not used at all, right? Same thing below when you're getting the list of all of the Roles.

ui/app/js/controllers/permission_scheduled_changes_controller.js
```diff
@@ -0,0 +1,204 @@
+angular.module("app").controller("PermissionScheduledChangesController",
+function($scope, $routeParams, $location, $timeout, Permissions, Rules, Search, $modal, $route, Releases) {
+
+  $scope.loading = true;
+  $scope.failed = false;
+
+  $scope.sc_id = parseInt($routeParams.sc_id, 10);
+
+  function loadPage(newPage) {
+    Permissions.getScheduledChangeHistory($scope.sc_id, $scope.pageSize, newPage)
```
mozbhearsum: Don't retrieve the history if you're not going to show it :)

NinadBhat:
@mozbhearsum PR read for review now

ui/app/templates/new_user_scheduled_change_modal.html
```diff
@@ -0,0 +1,51 @@
...
+
+    </div>
+  </div>
+
+  <div loader ng-show="loading"></div>
+
+</div>
+<div class="modal-footer">
+  <div ng-show="saving" small-loader></div>
+  <button type="submit" class="btn btn-primary" ng-click="addScheduledUser()">Save Changes</button>
```
mozbhearsum: I see you're calling this in ng-click and the ng-submit of the form - I think only one of them is necessary - can you double check?

mozbhearsum: I see you're calling this in ng-click and the ng-submit of the form - I think only one of them is necessary - can you double check?

ui/app/js/controllers/permission_scheduled_change_new_controller.js
```diff
@@ -0,0 +1,216 @@
...
+  });
+
+
+  $scope.saving = false;
+
+
+
+
+  $scope.addScheduledPermission = function() {
+  if($scope.permission.permission) {
```
mozbhearsum: I see that this controller is used by different modals, and that you've got addScheduledPermission for dealing with the modal that opens from the Permissions page, and addScheduledUser for dealing with the modal that opens from the Scheduled Changes page. There's a lot of duplication between them, and they both have the same job (scheduled a change for a new permission) - is there a way to combine them? If not, I think you should use a different controller for the modal that opens from the Scheduled Changes page - it's confusing to have two different handlers that both add a new scheduled change, but from different data sources.

ui/app/js/controllers/permission_scheduled_changes_controller.js
```diff
@@ -0,0 +1,179 @@
...
+   if ($scope.sc_id) {
+    $scope.ordering_options = [
+      {
+        text: "Data Version",
+        value: "-data_version"
+      },
+    ];
+  } else {
+    $scope.ordering_options = [
+
```
mozbhearsum: Since there's only one ordering option, it's better to just leave this out and hide the dropdown.

ui/app/js/controllers/permission_scheduled_changes_controller.js
```diff
@@ -0,0 +1,179 @@
...
+  $scope.$watchCollection('filters.search', function(value) {
+    $location.hash(value);
+    Search.noticeSearchChange(
+      value,
+      ['username']
+    );
+  });
+
+  $scope.filterBySearch = function(item) {
```
mozbhearsum: Is this filterBySearch stuff actually getting used, or is it just inherited from Releases? If it's not used, please remove it.

ui/app/js/controllers/permission_scheduled_changes_controller.js
```diff
@@ -0,0 +1,179 @@
...
+    if ($scope.state_str.value === "complete" && sc.complete) {
+      return true;
+    }
+    else if ($scope.state_str.value === "active" && !sc.complete) {
+      return true;
+    }
+    return false;
+  };
+
+  $scope.formatMoment = function(when) {
```
mozbhearsum: This doesn't look like it is used either.

ui/app/templates/permission_scheduled_changes.html
```diff
@@ -0,0 +1,90 @@
+<div loader ng-show="loading"></div>
+
+<h2>
+  <span ng-show="!scheduled_changes_permissions_count">Scheduled Permissions Changes</span>
+  <span ng-show="scheduled_changes_permissions_count">History for Scheduled Permissions Change</span>
```
mozbhearsum: Is this used? I don't see references to scheduled_changes_permissions_count in the controller.

PR #227, opened 2017-01-30
Revert unwanted configure_logging() call from #219.

mozbhearsum:
Accidentally included the unwanted configure_logging call from #219 - need to revert it!


mozbhearsum:
This ended up happening as part of #228.

PR #229, opened 2017-01-30
Bug 1333095 - reset signoffs when a scheduled change is updated

aksareen:


auslib/test/admin/views/test_releases.py
```diff
@@ -1239,6 +1241,35 @@ def testUpdateScheduledChangeExistingRelease(self):
         self.assertEquals(dict(cond[0]), cond_expected)
 
     @mock.patch("time.time", mock.MagicMock(return_value=300))
+    def testUpdateScheduledChangeExistingReleaseResetSignOffs(self):
+        data = {
+            "data": '{"name": "ab", "hashFunction": "sha512", "extv": "3.0", "schema_version": 1}', "name": "ab",
```
mozbhearsum: While adding the "data" column to a scheduled deletion works, it's not really something that's done in real usage. If you're going to continue using sc_id 4 for this test, I'd suggest updating "when" instead. Or you can use a different sc_id for it - up to you!

auslib/db.py
```diff
@@ -485,7 +485,7 @@ def delete(self, where, changed_by=None, old_data_version=None, transaction=None
            @rtype: sqlalchemy.engine.base.ResultProxy
         """
         # If "where" is key/value pairs, we need to convert it to SQLAlchemy
-        # clauses before porceeding.
+        # clauses before proceeding.
```
mozbhearsum: Thanks for fixing these typos!

auslib/admin/views/scheduled_changes.py
```diff
@@ -94,13 +94,18 @@ def _post(self, sc_id, form, transaction, changed_by):
...
                 what[k] = v.data
 
+        where = {"sc_id": sc_id}
         try:
-            self.sc_table.update({"sc_id": sc_id}, what, changed_by, form.sc_data_version.data, transaction)
+            self.sc_table.update(where, what, changed_by, form.sc_data_version.data, transaction)
+            signOffs = self.sc_table.signoffs.select(where=where, transaction=transaction, columns=["sc_id", "username"])
+            for signOff in signOffs:
+                self.sc_table.signoffs.delete(where={"sc_id": sc_id, "username": signOff["username"]},
+                                              changed_by=changed_by, transaction=transaction)
```
mozbhearsum: This logic looks fine, but it belongs in ScheduledChangesTable.update. If we put it here, and other code comes along that calls into ScheduledChangesTable.update, we'd have to duplicate this logic (or forget, and introduce a bug).

aksareen: That seems fair.

auslib/db.py
```diff
@@ -1140,6 +1140,13 @@ def update(self, where, what, changed_by, old_data_version, transaction=None, dr
             if not self._dataVersionsAreSynced(sc_id, transaction):
                 raise MismatchedDataVersionError("Conditions data version is out of sync with main table for sc_id %s" % sc_id)
 
+        if type(where) == dict and where.get("sc_id", None):
```
mozbhearsum: Hmmm, I'm not sure this is going to work, especially for calls that use a list of SQLAlchemy expressions instead of a dict. I think you can probably make use of the "affected_ids" that are calculated above.

PR #230, opened 2017-01-31
Bug 1278539 - History refactor

ex-Dev:
-Make the FieldView and DiffView subclassable and available only in release contexts;
-Refactoring the HistoryViews to be needed less reimplementations for each table.

Just opening the PR to validate the ReleaseHistoryView, RuleHistoryAPIView and ScheduledChangeHistoryView refactory approach.
So far, only the get method was refactored.

### DIFF
--- a/auslib/admin/views/base.py
+++ b/auslib/admin/views/base.py
@@ -1,11 +1,9 @@
 import json
-import time
 
 from flask import request, Response
 from flask.views import MethodView
 
 from auslib.global_state import dbo
-from auslib.util.timesince import timesince
 from auslib.db import OutdatedDataError, PermissionDeniedError, UpdateMergeError, ChangeScheduledError, \
     SignoffRequiredError
 import logging
@@ -78,64 +76,3 @@ class AdminView(MethodView):
         self.log.debug("processing DELETE request to %s" % request.path)
         with dbo.begin() as trans:
             return self._delete(*args, transaction=trans, **kwargs)
-
-
-class HistoryAdminView(AdminView):
-
-    history_keys = ('timestamp', 'change_id', 'data_version', 'changed_by')
-
-    def getAllRevisionKeys(self, revisions, primary_keys):
-        try:
-            all_keys = sorted([
-                x for x in revisions[0].keys()
-                if x not in self.history_keys and x not in primary_keys
-            ])
-        except IndexError:
-            all_keys = None
-        return all_keys
-
-    def annotateRevisionDifferences(self, revisions):
-        _prev = {}
-        for i, rev in enumerate(revisions):
-            different = []
-            for key, value in rev.items():
-                if key in self.history_keys:
-                    continue
-                if key not in _prev:
-                    _prev[key] = value
-                else:
-                    prev = _prev[key]
-                    if prev != value:
-                        different.append(key)
-                # prep the value for being shown in revision_row.html
-                if value is None:
-                    value = 'NULL'
-                elif isinstance(value, dict):
-                    try:
-                        value = json.dumps(value, indent=2, sort_keys=True)
-                    except ValueError:
-                        pass
-                elif isinstance(value, int):
-                    value = unicode(str(value), 'utf8')
-                elif not isinstance(value, basestring):
-                    value = unicode(value, 'utf8')
-                rev[key] = value
-
-            rev['_different'] = different
-            rev['_time_ago'] = getTimeAgo(rev['timestamp'])
-
-
-def getTimeAgo(timestamp):
-    # keeping this here amongst the view code because the use of the
-    # timesince() function is specific to the view
-    now, then = int(time.time()), int(timestamp / 1000.0)
-    time_ago = timesince(
-        then,
-        now,
-        afterword='ago',
-        minute_granularity=True,
-        max_no_sections=2
-    )
-    if not time_ago:
-        time_ago = 'seconds ago'
-    return time_ago
--- a/auslib/admin/views/history.py
+++ b/auslib/admin/views/history.py
@@ -1,27 +1,17 @@
-import difflib
 import json
-
-from sqlalchemy.sql.expression import null
 
 from flask import Response
 
-from auslib.global_state import dbo
 from auslib.admin.views.base import AdminView
 
 
 class FieldView(AdminView):
-    """/view/:type/:id/:field"""
+    """/view/:id/:field"""
+    def __init__(self, table, *args, **kwargs):
+        self.table = table
 
-    def get_value(self, type_, change_id, field):
-        tables = {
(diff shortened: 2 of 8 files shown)


auslib/admin/views/base.py
```diff
@@ -63,7 +63,6 @@ def __init__(self, *args, **kwargs):
 
     @handleGeneralExceptions("POST")
     def post(self, *args, **kwargs):
-        self.log.debug("processing POST request to %s" % request.path)
```
ex-Dev: I will turn back with this, was a little mistake.

auslib/admin/views/history.py
```diff
@@ -1,27 +1,17 @@
...
-
-from auslib.global_state import dbo
-from auslib.admin.views.base import AdminView
+from flask import Response, jsonify, request
+from sqlalchemy import and_
+from auslib.admin.views.base import AdminView, HistoryAdminView
 
 
 class FieldView(AdminView):
-    """/view/:type/:id/:field"""
```
mozbhearsum: I see that FieldView is only ever used by ReleaseDiffView, and it makes me wonder if we should just kill it (and move the small bit of necessary logic into ReleaseDiffView. What do you think?

auslib/admin/views/releases.py
```diff
@@ -348,40 +352,14 @@ def _put(self, release, changed_by, transaction):
...
-        revisions = table.select(
-            where=[
-                table.name == release['name'],
-                table.data_version != null()
-            ],
-            limit=limit,
-            offset=offset,
-            order_by=[table.timestamp.desc()],
-        )
+    def __process_revisions(self, revisions):
```
mozbhearsum: Nit: don't use __ - that's usually reserved for language-level stuff. A single _ will suffice, and is sometimes used to try to signal that things are private methods.

auslib/admin/views/history.py
```diff
@@ -54,43 +44,54 @@ def get(self, type_, change_id, field):
...
-            value.splitlines(),
-            fromfile="Data Version {}".format(prev_data_version),
-            tofile="Data Version {}".format(data_version),
-            lineterm=""
-        )
-
-        return Response('\n'.join(result), content_type='text/plain')
+class HistoryViewParameters(object):
+    def __init__(self,
+                 get_object_callback,
```
mozbhearsum: I'm not necessarily opposed to it, but I'm curious why you're using callbacks when some of them end up being methods on the subclass. Did you consider calling them directly instead?

ex-Dev: Yes, I thought to do this too, but then I used callbacks to get a more contextualized code.
Other people could see "_get_filters" and think: -"why is this not used?".

auslib/admin/views/history.py
ex-Dev: I think there is no problem, once the FieldView is used by releases table only.

auslib/admin/views/history.py
mozbhearsum: That's a great point - it's much explicit with the callbacks. Let's stick with that!

ex-Dev:
Hello @mozbhearsum ,
This is ready to review!

auslib/admin/views/history.py
```diff
@@ -1,96 +1,108 @@
...
-            value.splitlines(),
-            fromfile="Data Version {}".format(prev_data_version),
-            tofile="Data Version {}".format(data_version),
-            lineterm=""
-        )
-
-        return Response('\n'.join(result), content_type='text/plain')
+from flask import Response, jsonify, request
+from sqlalchemy import and_
+from auslib.admin.views.base import HistoryAdminView
```
mozbhearsum: What do you think about moving HistoryAdminView into this file? Does it make sense to factor it at all? It seems odd to me that it lives in auslib/admin/views/base.py, while the rest of the History stuff is here. I definitely think there's some more refactoring to do here, because it seems that annotateRevisionDifferences is the only consumer of getTimeAgo, and I think that getAllRevisionKeys is actually unused?

auslib/admin/views/history.py
```diff
@@ -1,96 +1,108 @@
...
+                 response_key='revisions'):
+        self.get_object = get_object_callback
+        self.filters = history_filters_callback
+        self.process_revisions = process_revisions_callback
+        self.order_by = revisions_order_by
+        self.not_found_msg = obj_not_found_msg
+        self.response_key = response_key
+
+
+class HistoryViewPostParameters(object):
```
mozbhearsum: I think it would be better to get rid of these and put the necessary parameters into get_revisions and revert_revision. Putting them here adds a layer of indirection that makes it more difficult to read the code. Some docstrings similar to what we have in AUSTable (to describe the parameters) would be greatly appreciated, too.

auslib/admin/views/history.py
```diff
@@ -1,96 +1,108 @@
...
+            order_by=history_parameters.order_by)
+
+        _revisions = history_parameters.process_revisions(revisions)
+
+        ret = {}
+        ret[history_parameters.response_key] = _revisions
+        ret['count'] = total_count
+        return jsonify(ret)
+
+    def revert_version(self, history_parameters, transaction):
```
mozbhearsum: Nit: "revert_to_revision" would be a better name for this.

auslib/admin/views/history.py
ex-Dev: I agree, makes all sense. All the history things in the same context.
I had not visit the HistoryAdminView class, I just kept inheritance from it.

auslib/admin/views/history.py
```diff
@@ -1,96 +1,150 @@
...
-        elif value is None:
-            value = 'NULL'
-        elif isinstance(value, int):
-            value = unicode(str(value), 'utf8')
-        else:
-            value = unicode(value, 'utf8')
-        return value
-
-    def get(self, type_, change_id, field):
+    def getAllRevisionKeys(self, revisions, primary_keys):
```
mozbhearsum: I still think this method is no longer used? If so, please remove it.

ex-Dev: I'm just starting the review. 
I had a free time at  work and I started, now I'm going home, haha.
I still have to put the comments on history.py, check the not used code and squash the commits.

mozbhearsum: Oh, sorry! I got too excited :). I'll wait until you declare it ready again.

ex-Dev: NP :)
Indeed I wanted to have more time to contribute. The Mozilla culture is much more cool that any company that I worked in Brazil.

ex-Dev:
Hey @mozbhearsum,
This PR is ready to review, again!

ex-Dev:
Great!
Let me know if you have any problems in production.

PR #233, opened 2017-01-31
Bug 1257298 : Altering rules.version field's length from 10 to 75

aksareen:


mozbhearsum:
This looks like it should be fine. I think that failure is related to @sirMackk's work in #215. Matt - can you have a look at the error log here - it almost looks like the "alias" column's unique property is somehow getting affected by this upgrade script?

> auslib/test/test_db.py:4063: in assert_attributes_for_tables
>     self.assertEqual(failures, [], 'Column properties different between models and migrations:\n' + '\n'.join(failures))
> E   AssertionError: Column properties different between models and migrations:
> E   Property 'unique' on 'rules.alias' differs between model and migration: (model) True != (migration) None

sirMackk:
I've looked into this a bit and it seems that the new migration is _somehow_ changing the unique property on `alias`. I basically dumped the migrated db from the test and inspected it using sqlite3. I don't know if the culprit is sqlite in some way or sqlalchemy-migrate - I'll be looking into this further today and tomorrow. Thanks for catching this!

mozbhearsum:
Thanks @sirMackk - that's quite strange. I wonder if this is an sqlite-only thing. If it is, and we can't find a way to fix it, we might have to stop comparing this attribute =\.

sirMackk:
I'm willing to bet it's a sqlite-only thing here - I bumped sqlalchemy-migrate to the newest version (0.10.0) and then I bumped sqlalchemy's version until things worked and I found that the test passes fine when sqlalchemy is at 1.0.0. Anything before that and migration 023 removes the unique constraint - you can actually see the CREATE TABLE statements for each migration if you add

    import logging
    logging.getLogger('sqlalchemy.engine').setLevel(logging.INFO)

... at the beginning of testColumnAttributesAreSameAsDb test.

I won't get a chance to test this against a mysql database until Monday :(, so if this is a blocker, I'd just disable the unique test, at least for now.

mozbhearsum:
@aksareen - is the above enough for you to move forward? I think you may need to comment out the testing of unique at https://github.com/mozilla/balrog/blob/fac5aee1003d7a13810cf2b7fbdd7127921637da/auslib/test/test_db.py#L4052, remove "unique" from the Column in db.py, or possibly both.

We'll probably want to revisit that when we finally get sqlalchemy updated, I've made a note in https://bugzilla.mozilla.org/show_bug.cgi?id=1303424 about it.

PR #236, opened 2017-02-03
Bug 1325441 - don't allow completed scheduled changes to be modified

aksareen:


auslib/admin/views/scheduled_changes.py
```diff
@@ -91,10 +91,14 @@ def _post(self, sc_id, form, transaction, changed_by):
...
             # will have already converted it to None, so we can treat it the
             # same as a modification here.
-            if (request.json and k in request.json):
+            if request.json and k in request.json:
                 what[k] = v.data
 
         where = {"sc_id": sc_id}
+        completed_scheduled_change = self.sc_table.select(where=where, transaction=transaction, columns=["complete"])
+        if completed_scheduled_change and completed_scheduled_change[0].get("complete"):
+            return Response(status=403, response="Scheduled change already completed. Cannot modify now.")
```
mozbhearsum: Similar to #229, this is something that should be enforced at the database level, to make sure that other things that call into it (that may exist in the future) don't have to reimplement this.

auslib/test/admin/views/test_permissions.py
```diff
@@ -406,6 +406,14 @@ def testUpdateScheduledChangeExistingPermissionResetSignOffs(self):
         self.assertEquals(len(rows), 0)
 
     @mock.patch("time.time", mock.MagicMock(return_value=300))
+    def testUpdateCompletedScheduledChangeExistingPermission(self):
```
mozbhearsum: Do keep these tests though! It's good to test both the web layer and the database layer.

auslib/admin/views/scheduled_changes.py
```diff
@@ -91,12 +92,14 @@ def _post(self, sc_id, form, transaction, changed_by):
...
             # same as a modification here.
-            if (request.json and k in request.json):
+            if request.json and k in request.json:
                 what[k] = v.data
 
         where = {"sc_id": sc_id}
         try:
             self.sc_table.update(where, what, changed_by, form.sc_data_version.data, transaction)
+        except PermissionDeniedError as e:
+            return Response(status=403, response=str(e))
```
mozbhearsum: No need to be catching this here - it's already caught by the general exception handler: https://github.com/mozilla/balrog/blob/df8c0af96753702989e48d60d1715e3a2aefc748/auslib/admin/views/base.py#L50

auslib/db.py
```diff
@@ -1008,6 +1012,23 @@ def _dataVersionsAreSynced(self, sc_id, transaction):
...
+
+        @param where: A list of SQLAlchemy clauses, or a key/value pair of columns and values.
+        @type where: list of clauses or key/value pairs.
+
+        @param transaction: A transaction object to add the update statement (and history changes) to.
+                           If provided, you must commit the transaction yourself. If None, they will
+                           be added to a locally-scoped transaction and committed.
+        @param change: a string denoting the type of change. Eg: Delete/Update.
+        """
+        completed_scheduled_change = self.select(where=where, transaction=transaction, columns=["complete"])
```
mozbhearsum: This method isn't really doing much, and raising the exception here makes things a bit harder to read. I think you should either be raising the exception in the caller, or integrating this with .validate().

auslib/db.py
```diff
@@ -1008,6 +1012,23 @@ def _dataVersionsAreSynced(self, sc_id, transaction):
...
+        if completed_scheduled_change and completed_scheduled_change[0].get("complete"):
+            raise PermissionDeniedError("Scheduled change already completed. Cannot {} now.".format(change))
```
mozbhearsum: A plain ValueError is better here too. PermissionDeniedError is reserved for when someone does not have the necessary ACLs to make a change.

auslib/admin/views/scheduled_changes.py
aksareen: so should we also stop catching valueError too ? : https://github.com/aksareen/balrog/blob/1f06024cfbcf96fe77febf56eeb1124c2ab58e0c/auslib/admin/views/scheduled_changes.py#L103

auslib/db.py
aksareen: ValueError will throw BadRequest 400 response. Since we're denying the user permission to modify the completed change, shouldn't a 403 be more suitable here ?

auslib/db.py
aksareen: Raising in caller seems to be better choice as we want to check if any further modification operations (update/delete) are being performed on an already completed change.

auslib/admin/views/scheduled_changes.py
mozbhearsum: Yes, that's a good idea!

auslib/db.py
mozbhearsum: Yup, sounds good.

auslib/db.py
```diff
@@ -1101,6 +1105,13 @@ def update(self, where, what, changed_by, old_data_version, transaction=None, dr
         # We need to check each Scheduled Change that would be affected by this
         # to ensure the new row will be valid.
         for row in self.select(where=where, transaction=transaction):
+            # verify whether the scheduled change has already been completed or not. If completed,
+            # then cannot modify the scheduled change anymore.
+            completed_scheduled_change = self.select(where={"sc_id": row["sc_id"]},
+                                                     transaction=transaction, columns=["complete"])
```
mozbhearsum: Are you sure you need to do this select? It looks like you should already have the complete row in "row".

aksareen: hehe, forgot to change this. Updating the PR.

PR #239, opened 2017-02-06
Bug 1301040 Filter state persists after switching pages

wasifhyder:
Filter state is now stored in localStorage under the key 'pr_ch_filter'. When page is switched, the state is persisted, and when we revert back to the page, we reload the last known filter from localStorage. Its "All rules" by default


aksareen:
Hi @wasifhyder , nit: specify the bug id when posting a new PR , makes it easier to look it up at bugzilla.

mozbhearsum:
@wasifhyder - can you accept the collaboration invite I sent? That we can get tests run on this PR.

wasifhyder:
@mozbhearsum - I accepted the request. I've also addressed the failing tests, and made sure they pass now.

PR #240, opened 2017-02-07
Bug 1337353: Set HSTS header in balrog responses

jvehent:


mozbhearsum:
Thanks for the PR, @jvehent. This should end up in production tomorrow.

PR #243, opened 2017-02-08
Bug 1337642 - fallbackMapping sidesteps checks on version and buildID increasing 

NinadBhat:


mozbhearsum:
Thank you! It was awesome seeing this filed and fixed in less than 24 hours!

PR #245, opened 2017-02-08
Bug 1333874:  mergeUpdate is broken when a base table row updates a v…

sirMackk:
…alue to the same value in the scheduled change

Bug: [1333874](https://bugzilla.mozilla.org/show_bug.cgi?id=1333874)

I changed the condition to check the base object against the update and I added a test to check against this case. I'll test this out manually today or tomorrow so we can get this merged soon :).

### DIFF
--- a/auslib/db.py
+++ b/auslib/db.py
@@ -1215,6 +1215,10 @@ class ScheduledChangeTable(AUSTable):
         (meaning: the scheduled change and the new version of the row modify
         the same columns), an UpdateMergeError is raised."""
 
+        # Filter the update to only include fields that are different than
+        # what's in the base (old_row).
+        what = {k: v for k, v in what.items() if v != old_row.get(k)}
+
         # pyflakes thinks this should be "is False", but that's not how SQLAlchemy
         # works, so we need to shut it up.
         # http://stackoverflow.com/questions/18998010/flake8-complains-on-boolean-comparison-in-filter-clause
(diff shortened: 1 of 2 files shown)


mozbhearsum:
I gave this a quick try and it looks like mergeUpdate overrides whatever the Scheduled Change is supposed to be changing. Eg: if I schedule a change to a Rule's channel, and then update its product, the channel in the scheduled change is overridden.

sirMackk:
Cool, thanks for giving it a whirl. I think this happens because an update consists of all the columns. 

What would be the ideal outcome for this scenario ie. given a Scheduled Change to a channel, should the user be allowed to update the product (this would update the product on the Scheduled Change _but not_ the channel)? Or should the user be blocked from making this update, because it conflicts with the Scheduled Change? 

I think that implementing the latter would require the code to isolate the updated column and only update that specific column - am I thinking of this correctly?

mozbhearsum:
> What would be the ideal outcome for this scenario ie. given a Scheduled Change to a channel, should the user be allowed to update the product (this would update the product on the Scheduled Change but not the channel)? Or should the user be blocked from making this update, because it conflicts with the Scheduled Change?

Taking Rules as an example...technically, the API supports only partial updates of them. The UI ends up providing all of the columns, because they are all shown to the user. I don't think that this should impact how we detect conflicts. This is essentially a 3-way merge where the base is current version of the Rule (the one in the database _before_ the update is committed), one head is the Scheduled Change, and the other head is the proposed update to the Rule - that is, the current version of the Rule with the proposed changes applied to it.

A conflict occurs when the two heads are both modifying the same field. So in your scenario above, if the only thing that differs between the current version and the Scheduled Change is the channel, and the only thing that differs between the current version and the proposed update is the product, there should not be a conflict. We should allow the change to the current version of the Rule, and update the Scheduled Change with the new product.

If the proposed update were to be changing the channel, that *would* be a conflict, and raise an UpdateMergeError.

> I think that implementing the latter would require the code to isolate the updated column and only update that specific column - am I thinking of this correctly?

I think so. It looks like mergeUpdate currently assumes that "what" only contains columns that have changed. This is probably true for most of the current tests, but as you've noticed, the UI always sends all of the columns through.

sirMackk:
Thanks, with that extra context I got what was wrong with my earlier approach. I added a filter to get rid of all the fields that aren't getting updated and this takes care of the issue you bumped into while keeping the rest of the function more or less unchanged.

I gave it a go manually to test for the ideal case, the issue you bumped into, and the "usual" error (trying to update the same column that scheduled change is updating with a different value) - things worked out correctly. Let me know what you think!

PR #247, opened 2017-02-09
Bump Sphinx version to fix docs.

mozbhearsum:


docs/source/conf.py
```diff
@@ -28,7 +28,7 @@
 
 # If your documentation needs a minimal Sphinx version, state it here.
 #
-# needs_sphinx = '1.0'
```
JohanLorenzo: Bikeshed: Do we need to keep this line in the file? Isn't having the old version only in VCS good enough?

mozbhearsum: Clarified on IRC - the new version of this line is uncommented!

PR #249, opened 2017-02-12
Bug 1304555:Remove Product/Channel filter from Rules history pages

tpingili:
Product/Channel filter is shown in all pages where rule_id is not available and ng-if evaluates to true. rule_id is available in the history page, hence ng-if evaluates to false and the filter is removed from the DOM.

### DIFF
--- a/ui/app/templates/rule_modal.html
+++ b/ui/app/templates/rule_modal.html
@@ -47,14 +47,14 @@
       <div class="col-md-6">
         <div class="form-group" ng-class="{'has-error': errors.backgroundRate}">
           <label for="id_rate">Rate</label>
-          <input type="number" class="form-control" id="id_rate" ng-model="rule.backgroundRate">
+          <input type="text" class="form-control" id="id_rate" ng-model="rule.backgroundRate">
           <p class="help-block" ng-show="errors.backgroundRate">{{ errors.backgroundRate.join(', ') }}</p>
         </div>
       </div>
       <div class="col-md-6">
         <div class="form-group" ng-class="{'has-error': errors.priority}">
           <label for="id_priority">Priority</label>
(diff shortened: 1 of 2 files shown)


NinadBhat:
@tpingili Thanks for you work. I will try to review the PR tonight.
Advice :- Please create different PR's for different bugs from next time.

tpingili:
Thanks Ninad. Unintentionally, the second PR got tacked onto the first one. I'm learning how to create different PRs for my consecutive commits.

NinadBhat:
@tpingili np. You can create different PR's by working on different branches. 
Your code for Bug 1304555 seems right.
As for  Bug 1301051 changing to "text" does the trick, but I am not sure if that is the right method, @mozbhearsum can you have a look and see if this is what we need.

@tpingili : let me know if you need help finding new bugs. bhearsum will be able to review this tomorrow.

mozbhearsum:
Sorry for the push to your branch - I had to do it to get the CI tests to run! Please feel free to join us in IRC (irc://irc.mozilla.org/#balrog) if you'd like some help with git branches.

mozbhearsum:
I was a bit concerned about accessibility problems if we change the inputs to type="text", but I asked some folks that know that space, and that should be fine.

You'll need to split up your patches into separate PRs before I can merge, though.

NinadBhat:
@tpingili : I think the same error might also be caused at 
https://github.com/mozilla/balrog/blob/master/ui/app/templates/rule_scheduled_change_modal.html#L107
can you kindly make the changes there too.

mozbhearsum:
I'd really love to get these things merged. Do you need a hand separating out the patches, @tpingili?

mozbhearsum:
This was split into #259 and #260.

PR #252, opened 2017-02-13
Bug 1310303 - teach the Balrog agent how to determine whether or not scheduled changes meet signoff requirements 

nurav:


agent/balrogagent/cmd.py
```diff
@@ -43,7 +43,21 @@
...
+                    for user, role in signoff_list.items():
+                        if obtained_signoffs.get(role) is not None:
+                            obtained_signoffs[role] += 1
+                        else:
+                            obtained_signoffs[role] = 1
+                    for role, number in required_signoffs.items():
+                        if required_signoffs.get(role, 0) > obtained_signoffs.get(role, 0):
+                            logging.debug("required signoffs not obtained, aborting")
+                            ready = False
+                    # If we have all required signoffs, go ahead
```
mozbhearsum: This logic looks fine, but it belongs in a separate function. I also wonder if it would be better to just re-use https://github.com/mozilla/balrog/blob/ac29e19c2d3e4f29af74296d03c908f0ff32320b/auslib/db.py#L132 ?

nurav: It seems we can't reuse the function verbatim since the format of the arguments is different.

agent/balrogagent/cmd.py
```diff
@@ -43,7 +44,14 @@
                     else:
                         logging.debug("Unknown change type!")
 
-                    # If it *is* ready, enact it!
+                    # If it *is* ready, check if all the required signoffs have
+                    # been obtained
```
mozbhearsum: This comment needs an update - "if it *is ready" doesn't make sense when that's not checked until ~8 lines later.

agent/balrogagent/cmd.py
```diff
@@ -43,7 +44,14 @@
...
+                    required_signoffs = change.get('required_signoffs') or {}
+                    signoffs = change.get('signoffs') or {}
+                    if not verify_signoffs(required_signoffs, signoffs):
```
mozbhearsum: Can you add a log message here? Right now we have output like:
balrogagent_1  | 2017-02-23 14:37:19,906 - DEBUG - root.time_is_ready#22: Comparing time for change 1 (now: 1487860639.9067075, scheduled time: 1487860530.912)
balrogagent_1  | 2017-02-23 14:37:19,907 - DEBUG - root.run_agent#61: Change 1 is not ready

It logs the details of the time (which make it look it the change is ready), but then it says "not ready".

PR #253, opened 2017-02-13
UI for Required Signoffs management

mozbhearsum:
This is my attempt at a unified view for Required Signoffs, Scheduled Changes, and Signoffs. I was originally going to implement this as simple CRUD like we have for other parts of the UI, but after starting down that path I thought it was going to be much nicer to be able to see and manipulate all Required Signoffs as a group, rather than needing to flip back and forth through current/pending. I'd love to know what you think of that, and if there's things you think would make it better. I _think_ the UI is pretty self explanatory, but let me know if anything is unclear.

Random notes:
- All of the Signoff/Revocation UI and controllers should be fully re-usable for Rules, Releases, and Permissions.
- In order to make Scheduled Changes get enacted, I included the one line needed to finish bug 1310217.
- I've tried to simplify ui/app/js/controllers/required_signoffs_modal_base_controller.js as best I can, but I realize it's still pretty tough to read. Sorry :(.
- I'm not entirely happy with having a new top level nav item for Required Signoffs. I considered separating the UI into Product and Permissions, and putting the Permissions one in the Permissions drop down...but I couldn't figure out where Product Required Signoffs would go.

I should mention that there's a few bugs I'm aware of:
- Whenever we make multiple changes in one dialog, we send multiple requests to the backend in parallel. This means that even if one fails, the others might succeed. Ideally, we'd revert everything if anything fails, but I'm not sure this is entirely fixable right now. Eg: if the first Required Signoff is added directly, and a Scheduled Change addition fails, we can't directly delete the Required Signoff. Ultimately, I think being able to batch multiple changes together in a single HTTP request (which would put them in a single Transaction) is the right fix.
- When new changes are Scheduled, the Required Signoffs they need don't show up until after a refresh (because they aren't returned by that endpoint). This is annoying in testing, because you often use a User who holds multiple Roles, which means you don't get the automatic signoff from #246. In production, Users will likely only hold one Role, and others will be loading the page from scratch before signing off anyways. I do think we should fix this, but I didn't want to block on it. If you think we should, we can do that.
- History is not implemented at all. This didn't seem like a blocker, and I didn't want to further delay this project. Obviously needs to be fixed.


mozbhearsum:
Whoops, fixing the test failures.

mozbhearsum:
I'm going to have to fix this up tomorrow. Turns out that node doesn't support some of the new language features I'm using. Everything works fine in browser, so I think it's to review, but the frontend tests and some parts of the build fail because of it.

mozbhearsum:
No major changes with the fixes - just removed modern JS features because of the problems described in https://bugzilla.mozilla.org/show_bug.cgi?id=1339525.

mozbhearsum:
@JohanLorenzo was asking about ways to just see the UI parts of this. Here's a compare link against the rs-apis branch: https://github.com/mozbhearsum/balrog/compare/rs-apis...mozbhearsum:rs-ui

ui/app/js/controllers/required_signoffs_delete.js
```diff
@@ -0,0 +1,156 @@
...
+        promises.push(deferred.promise);
+        var data = {"product": $scope.product, "role": role_name, "data_version": $scope.to_delete[role_name]["data_version"],
+                    "csrf_token": csrf_token};
+
+        if ($scope.mode === "channel") {
+          data["channel"] = $scope.channel;
+        }
+        // Not a Scheduled Change, we'll have to create one to delete it
+        if ($scope.to_delete[role_name]["sc"] === null) {
+          data["when"] = new Date().getTime() + 5000;
```
JohanLorenzo: ~~I'm not sure to understand why we need to add 5s in this case. What does this time compensate?~~

Edit: Found the answer there https://github.com/mozilla/balrog/pull/253/files#diff-e729d8dd6ca1733c654cfbef188401beR285

I agree with this comment https://github.com/mozilla/balrog/pull/253/files#diff-45c43c60c890f3114299e2a5c7cd5e5dR6, an object that represents the payload would help in having this logic at one single place.

ui/app/js/controllers/required_signoffs_modal_base_controller.js
```diff
@@ -0,0 +1,349 @@
...
+              pending = current_roles[i]["sc"] ? true : false;
+              role = current_roles[i];
+              break;
+            }
+          }
+        }
+        // Otherwise (it's in both), it's an update.
+        else {
+          action = "update";
+          // There must be a better way to do this, but I can't find it.
```
JohanLorenzo: I'm not sure to follow this double loop. Do we assign the update role here? What does `role` do? I don't see it used afterwards.

ui/app/js/controllers/required_signoffs_modal_base_controller.js
```diff
@@ -0,0 +1,349 @@
...
+                  "sc_data_version": 1,
+                  "signoffs": {},
+                  "change_type": data["change_type"],
+                };
+              }
+            }
+            deferred.resolve();
+          };
+        };
+        var errorCallback = function(data, deferred) {
```
JohanLorenzo: Could we extract this callback in an Angular service? That'd prevent duplication with https://github.com/mozilla/balrog/pull/253/files#diff-0b367c91d764d4985fd850d6be9e3bd3R81

ui/app/js/controllers/required_signoffs_modal_base_controller.js
```diff
@@ -0,0 +1,349 @@
...
+      service = PermissionsRequiredSignoffs;
+    }
+    else {
+      $scope.errors["exception"] = "Couldn't detect mode";
+      return;
+    }
+
+    $scope.saving = true;
+    CSRF.getToken()
+    .then(function(csrf_token) {
```
JohanLorenzo: I'm sorry, I don't manage to follow the context thoroughly of this request. I guess splitting it in many async function would help understanding the high-level steps.

ui/app/js/filters/toarray_filter.js
```diff
@@ -0,0 +1,19 @@
+angular.module('app').filter('toArray', function () {
```
JohanLorenzo: Nice!

ui/app/templates/required_signoff_modal.html
```diff
@@ -0,0 +1,80 @@
...
+<div class="modal-body">
+  <form role="form">
+    <div class="form-group" ng-class="{'has-error': errors.product}" ng-show="!editing">
+      <label for="id_product">Product</label>
+      <input tabindex="1" type="text" class="form-control" id="id_product" ng-model="product" autocomplete="off"
+             typeahead="product for product in products | filter:$viewValue | limitTo:16">
+      <p class="help-block" ng-show="errors.product">{{ errors.product.join(", ") }}</p>
+    </div>
+    <div class="tabbable tabs-below" ng-show="!editing">
+      <ul class="nav nav-pills" role="tablist">
```
mozbhearsum: Good idea, I'll see what I can do.

ui/app/templates/required_signoffs.html
```diff
@@ -0,0 +1,128 @@
...
+                    ng-show="role['sc']['signoffs'].hasOwnProperty(current_user)">
+              Revoke your Signoff
+            </button>
+          </div>
+        </h4>
+      </div>
+      <div class="panel-body" ng-show="role['sc'] !== null">
+        <h4>Before it can be enacted, this change must be signed off on by:</h4>
+        <ul>
+          <li ng-repeat="(rs_role, signoffs_required) in role['sc']['required_signoffs']">
```
mozbhearsum: Yeah. This is because the Scheduled Changes endpoints don't return the Signoffs required. I talked about this a bit in https://github.com/mozilla/balrog/pull/253#issue-207340965. It's annoying, but I don't think a major issue in production. I can block on it you think it's important, though.

ui/app/templates/required_signoffs.html
```diff
@@ -0,0 +1,128 @@
...
+            {{ signoffs_required }} members of {{ rs_role }}
+          </li>
+        </ul>
+        <h4 ng-show="role['sc'] !== null && ! isEmpty(role['sc']['signoffs'])">This change has already been signed off by:</h4>
```
mozbhearsum: I agree that it's a bit weird that you can signoff on things that don't require you. I'll see if I can improve this. At the very least, maybe we can stop doing automatic signoffs if the users group is not required.

ui/app/js/controllers/required_signoffs_controller.js
```diff
@@ -0,0 +1,371 @@
...
+  };
+
+  $scope.editRequiredSignoffs = function(mode, channel) {
+    channel = channel ? channel : "";
+    $modal.open({
+      templateUrl: "required_signoff_modal.html",
+      controller: "EditRequiredSignoffsCtrl",
+      backdrop: "static",
+      resolve: {
+        mode: function() {
```
mozbhearsum: Sorry, I'm not sure what you mean here! Can you expand?

ui/app/js/controllers/required_signoffs_controller.js
```diff
@@ -0,0 +1,371 @@
...
+          return $scope.required_signoffs;
+        },
+        current_user: function() {
+          return $scope.current_user;
+        },
+      }
+    });
+  };
+
+  $scope.deleteRequiredSignoffs = function(required_signoffs, mode, channel) {
```
mozbhearsum: Hm...I'll see if I can reproduce this. I'm pretty sure it was working for me...

ui/app/js/controllers/required_signoffs_controller.js
JohanLorenzo: I mean we can now use a lambda-like function in JS. They are called [arrow functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions). This will make the code just a little bit smaller:
```js
mode: () => mode,
channel: () => channel,
```

ui/app/js/controllers/required_signoffs_modal_base_controller.js
mozbhearsum: I'll give a try - I agree that it's not easy to read :(

ui/app/js/controllers/required_signoffs_modal_base_controller.js
mozbhearsum: "role" ends up getting used around https://github.com/mozilla/balrog/pull/253/files/5aa69637fd5becb2316f4cb86104fb25d4c9ae2f#diff-e729d8dd6ca1733c654cfbef188401beR264 when we start making http requests. Depending on what we're doing, we may need to pull signoffs_required or scheduled change information from it. Maybe it would be clearer to assign those to separate variables...

ui/app/js/controllers/required_signoffs_modal_base_controller.js
mozbhearsum: I'm not sure. Most of what this does is manipulate things from the controller ($scope variables, and deferred resolutions), so it may be weird to do it elsewhere...

ui/app/js/controllers/required_signoffs_controller.js
mozbhearsum: I can't reproduce this as you describe it, but I managed to get this output under different circumstances. Specifically, if I had multiple changes scheduled, and only the one with the highest sc_id was signed off, I would get this output for an _earlier_ one, and the Agent would bail and not process the later on. Would you mind double checking to see if you can produce this bug with *only* one change being scheduled?

#252 fixes the problem I described, which I'll merge in before I update this PR.

ui/app/templates/permissions_modal.html
```diff
@@ -58,7 +58,7 @@ <h3 class="panel-title">
         <form role="form" ng-submit="grantRole()">
```
mozbhearsum: https://bugzilla.mozilla.org/show_bug.cgi?id=1342231 for the real error, https://bugzilla.mozilla.org/show_bug.cgi?id=1342234 for the fact that the error wasn't shown.

ui/app/templates/permissions_modal.html
```diff
@@ -58,7 +58,7 @@ <h3 class="panel-title">
...
           <div class="form-group" ng-class="{'has-error': errors.role}">
```
mozbhearsum: https://bugzilla.mozilla.org/show_bug.cgi?id=1342241 for this

ui/app/templates/permissions_modal.html
```diff
@@ -58,7 +58,7 @@ <h3 class="panel-title">
...
             <label for="id_role">Role</label>
```
mozbhearsum: Is this covered by https://bugzilla.mozilla.org/show_bug.cgi?id=739241 ?

ui/app/templates/permissions_modal.html
```diff
@@ -58,7 +58,7 @@ <h3 class="panel-title">
...
-            <input type="text" class="form-control" id="id_role" ng-model="role.role"
```
mozbhearsum: I think https://bugzilla.mozilla.org/show_bug.cgi?id=1341622 covers part of what you're talking about here.

I do agree that it's confusing to have the top level nav be called "Permissions", but the content talk about Users. Perhaps we should change the nav button, now that this section covers more than just Permissions?

ui/app/templates/permissions_modal.html
```diff
@@ -58,7 +58,7 @@ <h3 class="panel-title">
...
+            <input type="text" class="form-control" id="id_role" ng-model="role.role" autocomplete="off"
             typeahead="role for role in roles_list | filter:$viewValue | limitTo:16">
             <p class="help-block" ng-show="errors.role">{{ errors.role.join(', ') }}</p>
           </div>
```
mozbhearsum: I think the reason that the Save button is further up is because we have save/delete buttons for existing permissions in the bottom half of the modal. Maybe we should change the close button to red to make it more obvious? To be honest though, I think this UI could do with a bit of an overhaul. There's a lot of small issues with that might be fixed by rethinking it.

ui/app/templates/required_signoff_modal.html
mozbhearsum: I think I'm improved this slightly. I couldn't find something that looked awesome, but having some separator is better than nothing.

ui/app/templates/release_scheduled_change_modal.html
```diff
@@ -49,7 +49,7 @@ <h4 ng-show= "sc.change_type !== 'delete'">Release Details</h4>
     </div>
```
mozbhearsum: Fixed.

ui/app/templates/required_signoffs.html
mozbhearsum: I removed the automatic signoff unless it is useful.

ui/app/js/controllers/required_signoffs_controller.js
```diff
@@ -0,0 +1,371 @@
...
+        },
+        current_user: function() {
+          return $scope.current_user;
+        },
+      }
+    });
+  };
+
+  $scope.editRequiredSignoffs = function(mode, channel) {
+    channel = channel ? channel : "";
```
mozbhearsum: Fixed.

ui/app/js/controllers/required_signoffs_controller.js
mozbhearsum: Let me know if you can still reproduce this.

ui/app/js/controllers/required_signoffs_modal_base_controller.js
mozbhearsum: Thanks for all the help refactoring this. I _think_ it's a bit better now, but I'm curious what you think.

ui/app/js/controllers/required_signoffs_modal_base_controller.js
mozbhearsum: I think this is much better in the refactored version - what do you think?

ui/app/js/controllers/required_signoffs_modal_base_controller.js
mozbhearsum: This is better in the refactored version, too! We only end up running this code once, after all the promises have been resolved or failed.

mozbhearsum:
Test failures are due to more modern JS features ("=>" in this case). I'll sort them out.

mozbhearsum:
I think I've fixed all of the blocking issues at this point. Please let me know if I missed something, or there's something else.

ui/app/js/controllers/required_signoffs_modal_base_controller.js
```diff
@@ -68,218 +73,128 @@ function($scope, $modalInstance, $q, CSRF, ProductRequiredSignoffs, PermissionsR
...
+                    "signoffs_required": change["role"]["signoffs_required"],
                     "data_version": response["new_data_version"],
                     "sc": null,
                 };
               }
-              deferreds[role_name].resolve();
+              resolve();
             })
-            .error(errorCallback(data, deferreds[role_name]));
+            .error(function(response) { reject(response); });
```
mozbhearsum: Good idea, let's do that.

PR #255, opened 2017-02-15
Bug 1333876 - scheduled change history sometimes fails to load because of IndexError

NinadBhat:
While enacting scheduled change table was getting updated but not conditions table hence the number of rows in history table of both were different which led to the IndexError.
https://bugzilla.mozilla.org/show_bug.cgi?id=1333876

### DIFF
--- a/auslib/db.py
+++ b/auslib/db.py
@@ -1207,7 +1207,7 @@ class ScheduledChangeTable(AUSTable):
         # when the transaction is rolled back.
         # We explicitly avoid using ScheduledChangeTable's update() method here
         # because we don't want to trigger its validation of conditions. Doing so
-        # would raise any exception for any timestamp based changes,  because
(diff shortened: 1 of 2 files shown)


auslib/db.py
```diff
@@ -1207,8 +1207,13 @@ def enactChange(self, sc_id, enacted_by, transaction=None):
         # when the transaction is rolled back.
         # We explicitly avoid using ScheduledChangeTable's update() method here
         # because we don't want to trigger its validation of conditions. Doing so
-        # would raise any exception for any timestamp based changes, because
-        # they are already in the past when we're ready to enact them.
+        # would raise any exception for any timestamp based changes, be
+        # Updating in conditions table also so that heistory view can work
```
mozbhearsum: What happened with this comment? Line 1210 has an abruptly ended sentence now. Please fix the typo, too ("heistory" vs "history").

PR #256, opened 2017-02-20
Bug 1329525: Ungrouping of emails

jaehoonhwang:
Hello, I was trying references and in-reply-to, but none of them seems to have worked.

I couldn't get the `References` to even show in gmail and `Reply-To` or `In-Reply-To` didn't have any effect on conversation view.

I believe that gmail only group emails that have same subject line, so I created random string generator and just attach to it at the end of the subject header.

I was thinking that if we need to change a subject line, wouldn't it be better coming from one of the database?
So, for example, I would use `sc_id` for scheduled changes and such or something that these db changes share. (Primary keys perhaps?)

Anyway, I this commit works, but I don't know how I could do testing on it.
Any feedback would be appreciated!

This is by no means finished merging.

- [x] Create a test for the bug
- [x] Find better way to edit subject line.


mozbhearsum:
> Hello, I was trying references and in-reply-to, but none of them seems to have worked.
> 
> I couldn't get the References to even show in gmail and Reply-To or In-Reply-To didn't have any effect on conversation view.
>
> I believe that gmail only group emails that have same subject line, so I created random string generator and just attach to it at the end of the subject header.

I don't think "References" shows up in the UI even if it does have an effect how it is displayed. Reading https://developers.google.com/gmail/api/guides/threads makes me wonder if References and/or In-Reply-To get ignored if they are considered "invalid" (eg: maybe if they reference a message id that doesn't exist in the users inbox, they are ignored completely).

According to the link above, the subject *header* matching is definitely required to thread, but it is not the only requirement.

> I was thinking that if we need to change a subject line, wouldn't it be better coming from one of the database?
> So, for example, I would use sc_id for scheduled changes and such or something that these db changes share. (Primary keys perhaps?)

I think using identifying information is a pretty good idea. For things like Rules or Scheduled Changes, whose primary key is just a unique int, we may want to do something a bit better. Eg: "sc_id 20" doesn't really mean anything to people. Maybe we'd want to include alias (if it exists) for Rules, or base table primary key for Scheduled Changes.

Do you have any opinion on this, @nthomas-mozilla?

> Anyway, I this commit works, but I don't know how I could do testing on it.

You *should* be able to test it locally by setting these variables: https://github.com/mozilla/balrog/blob/master/docker-compose.yml#L23. Once set, try modifying a Rule - you should get e-mail about it. Depending on your SMTP server, you may or may not need to set SMTP_TLS=1.

mozbhearsum:
Oh, and there *are* tests for this! They don't actually send e-mail, but they do check to make sure the change notifiers are called, and that *they* send e-mail with the expected subject, body, etc: https://github.com/mozilla/balrog/blob/fd03557db8d4be1ed1c5986114fc3734da0a7da8/auslib/test/test_db.py#L3875

galbie:
Okay thank you I'll get on it.

nthomas-mozilla:
I tried setting headers to confirm we can't do that. I had to create a google group (in the email list style) because sending direct from Balrog to my email didn't thread two UPDATE emails. They were threaded when sent via the test list. Unfortunately adding headers like this:
    References: <123060692@bar.boz.net>
    In-Reply-To: <480530009@bar.boz.net>
didn't separate them 😢. Pretty sure those are valid according to RFC 2822, and sampling other email I've received.

So we're left with munging the subject. I'd like to avoid grouping changes when we don't have a sc_id to make the subject string unique, and only have the primary key for the rule or release. Do we have the change id available at email time ? That should be unique I think.

mozbhearsum:
> So we're left with munging the subject. I'd like to avoid grouping changes when we don't have a sc_id to make the subject string unique, and only have the primary key for the rule or release. Do we have the change id available at email time ? That should be unique I think.

At the moment, the change notifiers run prior to the history rows being inserted (eg: https://github.com/mozilla/balrog/blob/86c4113891a69bc8b4939c227b94d2d55222dde8/auslib/db.py#L380), but that could be changed. We need to make sure to handle tables that don't have history too, though.

And to make spell things out fully, it sounds like you're saying that:
- For Scheduled Changes, we should put sc_id + primary key in the subject
- For other things, we should put primary key + change_id in the subject

...which would mean that updates to the same scheduled change end up grouped (because neither sc_id nor primary pk will change), but any changes to other objects will always be ungrouped (because change_id + pk will always be unique). Do I have that right?

If you want to ungroup scheduled change updates, we'll need to add change_id there, too.

nthomas-mozilla:
On reflection, this is all getting very complicated to provide information which may not be of any value to a human, who are unlikely to recognize primary keys. Perhaps we should just append some whitespace padding, and then a random string to the subject.

jaehoonhwang:
Is there a specific length of random string you guys wanted?
I'm working with string length of 6 as this moment.

mozbhearsum:
6 characters should be fine, yeah.

jaehoonhwang:
I have created unique string generator based on length `n`.
Now I am trying to create a test cases for these unique subject, but only test I can do is with `PartialString()` which only checks for partial input arguments.
That's problem for creating my test cases, because I won't know what type of string the function would generate and knowing it would be defeat the purpose of random string generator.

mozbhearsum:
Testing with PartialString is probably OK, but mocking would be better. We use it quite a bit for our tests, so I'm surprised it's not working for you. Have you tried something like how we mock "time" in https://github.com/mozilla/balrog/blob/7eb6163efd2bb67b5ff1e8970bdecdcfec67d30c/auslib/test/test_db.py#L462 ? Eg:
```
@mock.patch("auslib.db.generate_random_string", mock.MagicMock(return_value="ABCDEF"))
def testUniqueSubject(self):
  ...
```

jaehoonhwang:
Sorry, I am fairly new with `mock` and your suggestion worked.
Thank you for the help.

nthomas-mozilla:
Awesome! 🍾

jaehoonhwang:
Thank you for the help!

rail:
Woot! No more grouping! Thanks @jaehoonhwang!

PR #257, opened 2017-02-21
Bug-1334133. Possible fix for correct validation of vendors and addons in the GMP and SystemAddons blobs. r=bhearsum

tyagi-iiitv:


tyagi-iiitv:
I don't know why my commit is showing changes as (+45,-42). I'm using Pycharm and I can see the correct **diff** locally, but when I push to git, the **diff** shows all the lines as newly added and all the previous lines as deleted.

aksareen:
Can you add test cases also to support and verify this change.

mozbhearsum:
PyCharm is probably ignoring whitespace changes in your local diff - I wouldn't worry about it.

+1 to adding a test or two. Having that will make sure we don't regress this in the future. You'll probably want to add tests to https://github.com/mozilla/balrog/blob/master/auslib/test/blobs/test_systemaddons.py and https://github.com/mozilla/balrog/blob/master/auslib/test/blobs/test_gmp.py that create blobs would be invalid under this schema change, and make sure that self.validate() returns False for them.

auslib/test/blobs/test_systemaddons.py
```diff
@@ -222,3 +222,26 @@ def testDoesNotContainForbiddenDomain(self):
...
+                    "filesize": 2,
+                    "hashValue": "3",
+                    "fileUrl": "http://a.com/blah"
+                }
+            }
+        }
+    }
+}
+""")
+        blob.validate('gg', self.whitelistedDomains)
```
aksareen: @tyagi-iiitv  . you're validaiting an invalid blob. So when you're invoking blob.validate method it is throwing BlobValidationError as expected. Simply place that part in a try catch and assert that the caught exception and message are as expected.

tyagi-iiitv: Don't go by the function_name, the blob inside is actually valid. I wrote this test_name to check for invalid blobs only but temporarily, it contains a valid blob for verification. (Sorry for this)

mozbhearsum: It looks like the blob is invalid because "uninstall" is not provided at the top level, which is specified as required at https://github.com/tyagi-iiitv/balrog/blob/1936206e36d970889871581141cf4f13e965bdd1/auslib/blobs/schemas/systemaddons.yml#L10.

tyagi-iiitv:
Okay, so here is a new problem. I am trying this blob to validate **without any change in the code of systemaddons.yml**. (Not even my changes in the previous commit)

```
blob = SystemAddonsBlob()
        blob.loadJSON("""
{
    "name": "fake",
    "schema_version": 5000,
    "hashFunction": "SHA512",
    "addons": {
        "c": {
            "platforms": {
                "p": {
                    "filesize": 2,
                    "hashValue": "3"
                },
                "q": {
                    "filesize": 4,
                    "hashValue": "5",
                    "fileUrl": "http://boring.com/blah"
                },
                "q2": {
                    "alias": "q"
                }
            }
        }
    }
}
""")

```

It shouldn't be validated because platform **p** hasn't got a **fileUrl**. Also, the addon **c** hasn't got a **version**. But the result for `blob.validate('gg', self.whitelistedDomains)` is true i.e. the blob is correct. 

I've tried including 
```
patterProperties:
    ^.+$:
        type: object

```

for each of **platforms, platformName and addonName** but it doesn't seem to work for me. 
Please suggest where I'm going wrong?

mozbhearsum:
The reason the validate is failing without any changes to the schema is because of the problem described in the bug - the deep down parts of the blob are never checked because of an error in the schema (in the case of SystemAddons, I believe it's because "addonName" is used at https://github.com/mozilla/balrog/blob/18d881ae973d366b990c56fbecd000892556b901/auslib/blobs/schemas/systemaddons.yml#L63 - there's no "addonName" property in real blobs, so nothing underneath it is checked).

It's hard to say what your attempted fix isn't working without seeing the code. Can you push a version of it where you've tried to replace addonName with patternProperties?

tyagi-iiitv:
Confirmed (test_systemaddons.py):

-  Blobs without any addons are validated correctly i.e. with addons: {} or with **uninstall** instead of **addons**. So, the problem is with the code of **addons** and beneath it.

tyagi-iiitv:
This PR is ready for review.

tyagi-iiitv:
Finally, tests passed. Ready for review now !!

auslib/test/blobs/test_systemaddons.py
```diff
@@ -222,3 +222,165 @@ def testDoesNotContainForbiddenDomain(self):
...
+                    "q": {
+                        "hashValue": "3",
+                        "fileUrl": "http://boring.com/blah"
+                    }
+                }
+            }
+        }
+    }
+    """)
+        self.assertRaises(Exception, blob.validate, 'gg', self.whitelistedDomains)
```
mozbhearsum: You've done a great job writing all of these tests, but because they are all testing different cases, they should each be put into their own test. The way things are right now, if the first test fails the rest will not run, so it's more difficult to get an overview of failures. Putting them into their own methods will ensure that each runs no matter what happens with the others.

tyagi-iiitv: Thanks for the review. It's always great to learn.

PR #259, opened 2017-02-23
Bug 1304555:Remove Product/Channel filter from Rules history pages

tpingili:


mozbhearsum:
Originally reviewed in #249 by @NinadBhat and I.

nthomas-mozilla:
Unfortunately this has broken the filter box on the rules page.

Steps to reproduce:
* update to bcd57207a76c7f3f6da04205675d716102096a4d
* docker-compose up
* load http://localhost:8080/rules
* try changing the filter, but it has no affect

NinadBhat:
It seems that reason for this is using ng-if [1]. Changing it to ng-show solves the problem. @tpingili will you be able to look at this anytime soon?
For more details one can read
http://stackoverflow.com/questions/21869283/when-to-favor-ng-if-vs-ng-show-ng-hide

[1] http://stackoverflow.com/questions/30787147/angularjs-ng-if-and-scopes

mozbhearsum:
I'm not sure how responsive or available @tpingili, and this isn't a regression that we could live with in production. I'm fixing it in #269.

PR #260, opened 2017-02-23
Bug 1301051:number fields sometimes steal scroll events, causing unwa…

tpingili:
…nted changes to their values


mozbhearsum:
Originally reviewed by @NinadBhat and I in #249.

PR #261, opened 2017-02-23
Bug 1310217 - teach the Balrog Agent how to look for and enact Required Signoffs

nurav:


agent/balrogagent/cmd.py
```diff
@@ -10,7 +10,9 @@
 
 SCHEDULED_CHANGE_ENDPOINTS = ['rules',
                               'releases',
-                              'permissions']
+                              'permissions',
+                              'required_signoffs/products',
```
mozbhearsum: This should be "product", not "products", per https://github.com/mozilla/balrog/blob/dc79a6b06ae1f38fd2d3fb20d8df20e5a7481d35/auslib/admin/base.py#L137.

PR #263, opened 2017-02-24
Bug 1332000 - test that downgrades actually work

aksareen:


auslib/test/test_db.py
```diff
@@ -4138,3 +4138,87 @@ def testColumnAttributesAreSameAsDb(self):
...
+        Tests that downgrades and upgrades work as expected. Since the DB will never
+        be rolled back beyond version 21 we treat it as the base version for all future versions from now.
+        Note: These tests run and verify migrations on a sqllite DB
+        whereas the actual migration happens on a mySQL DB.
+        """
+        # https://github.com/buildbot/buildbot/blob/87108ec4088dc7fd5394ac3c1d0bd3b465300d92/master/buildbot/db/model.py#L455
+        # http://code.google.com/p/sqlalchemy-migrate/issues/detail?id=100
+        # means  we cannot use the migrate.versioning.api module.  So the workaround for that hard-code
+        # the latest version.
+        latest_version = 23
```
mozbhearsum: Are you sure this is necessary? I had a look https://github.com/buildbot/buildbot/blob/87108ec4088dc7fd5394ac3c1d0bd3b465300d92/master/buildbot/db/model.py#L455, and it seems to say that sqlalchemy-migrate started including a version as of 0.7. We use 0.7.2, so it sounds like we should be able to discover it?

auslib/test/test_db.py
```diff
@@ -4138,3 +4138,87 @@ def testColumnAttributesAreSameAsDb(self):
...
+        db = self._get_migrated_db()
+        versions_migrate_tests_dict = {22: self._rules_version_length_migration_test,
+                                       21: self._fix_column_attributes_migration_test}
```
mozbhearsum: Is there no way to test this without implementing methods to deal with column attributes for each schema version? If so, it makes me wonder if we should just ignore column attributes for the moment - until we're on an sqlalchemy version that reflects them better.

aksareen: Not that I could think of at that time when we discussed this on IRC.  I didn't understood this line :  " **until we're on an sqlalchemy version that reflects them better**."

auslib/test/test_db.py
aksareen: I'll have to take a look at it. BTW There's a use case in my mind. If in future, say we bump up to version 26 and that's now the base version (for all upcoming versions thereon), and didn't have any checks for any previous versions , so the tests will start failing for versions 24-26, if someone forgot to add them to this. Now either approach can work here : Either we can fetch the latest version and have tests fail for missing version tests (IMHO the right way) , or let the test declare a constant version until someone adds a new test case and manually configures the latest_version. What do you think ?

mozbhearsum: I agree with you - it's better to automatically grab the latest version, and have the tests fail. This will force anybody who's making a migration to get the tests working again before we merge it, so it *should* be impossible to get into a state where we can't downgrade.

auslib/test/test_db.py
mozbhearsum: My understanding is that the version of SQLAlchemy that we're on doesn't reflect column attributes from sqlite -> memory properly, which is why we have some special handling for them in https://github.com/mozilla/balrog/blob/1bc8439bc7d4cf996738a069987f4a39d29f6edf/auslib/test/test_db.py#L4109.

Basically, I'm saying that we may be able to get rid of these per-version methods whne we get to a newer sqlalchemy version.

aksareen: Ah I see. So it depends on how soon are we updating our sqlalchemy version. I can add a small comment to delete these tests when we've updated it to a certain version i.e. we can let these tests run  only for the current sqlalchemy version. If we do this, we should open a new future bug before closing this.

mozbhearsum: That sounds good.

PR #266, opened 2017-02-27
Fix required_signoffs returned in API

mozbhearsum:
While testing out https://bugzilla.mozilla.org/show_bug.cgi?id=1310227 I discovered that the Scheduled Changes API only looks at the scheduled changes table when looking for required signoffs. This means that for some updates (ones where the current version of the row requires signoff, but the scheduled version doesn't), and all deletes returned the wrong required_signoffs.

The database layer still enforces things correctly, but this caused problems for the UI. Eg: the Signoff button not showing up or showing up when not needed.

I added new tests where it made sense - both the update and delete case for Rules and Permissions. Releases are unaffected here, because changing a release doesn't impact the signoffs required (those are determined by looking at the rules table). Required Signoffs are also unaffected, because the things that influence the signoffs required are part of the key - so they're the same on the current and scheduled version of a Required Signoff.

### DIFF
--- a/auslib/admin/views/scheduled_changes.py
+++ b/auslib/admin/views/scheduled_changes.py
@@ -29,7 +29,7 @@ class ScheduledChangesView(AdminView):
         for row in rows:
             scheduled_change = {"signoffs": {}, "required_signoffs": {}}
             base_row = {}
-            pk = {}
+            base_pk = {}
 
             for k, v in row.iteritems():
                 if k == "data_version":
@@ -39,7 +39,7 @@ class ScheduledChangesView(AdminView):
                         k = k.replace("base_", "")
                         base_row[k] = v
                         if getattr(self.table, k).primary_key:
-                            pk[k] = v
+                            base_pk[k] = v
                     scheduled_change[k] = v
 
             for signoff in self.sc_table.signoffs.select({"sc_id": row["sc_id"]}):
@@ -51,7 +51,7 @@ class ScheduledChangesView(AdminView):
                 # We don't need to consider the existing version of a row for
                 # inserts, because it doesn't exist yet!
                 if row["change_type"] != "insert":
-                    affected_rows.append(self.table.select(where=pk)[0])
+                    affected_rows.append(self.table.select(where=base_pk)[0])
                 # We don't need to consider the future version of the row when
                 # looking for required signoffs, because it won't exist when
                 # enacted.
--- a/auslib/admin/views/scheduled_changes.py
+++ b/auslib/admin/views/scheduled_changes.py
@@ -29,6 +29,7 @@ class ScheduledChangesView(AdminView):
         for row in rows:
             scheduled_change = {"signoffs": {}, "required_signoffs": {}}
             base_row = {}
+            pk = {}
 
             for k, v in row.iteritems():
                 if k == "data_version":
@@ -37,6 +38,8 @@ class ScheduledChangesView(AdminView):
                     if k.startswith("base_"):
                         k = k.replace("base_", "")
                         base_row[k] = v
+                        if getattr(self.table, k).primary_key:
+                            pk[k] = v
                     scheduled_change[k] = v
 
             for signoff in self.sc_table.signoffs.select({"sc_id": row["sc_id"]}):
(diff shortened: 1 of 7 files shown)


auslib/test/admin/views/test_rules.py
```diff
@@ -778,67 +789,104 @@ def setUp(self):
...
             change_type="update")
-        dbo.rules.scheduled_changes.conditions.history.t.insert().execute(change_id=8, changed_by="bill", timestamp=7, sc_id=4, when=500000, data_version=2)
+        dbo.rules.scheduled_changes.conditions.t.insert().execute(sc_id=4, when=500000, data_version=2)
+        dbo.rules.scheduled_changes.conditions.history.t.insert().execute(change_id=6, changed_by="bill", timestamp=5, sc_id=4)
+        dbo.rules.scheduled_changes.conditions.history.t.insert().execute(change_id=7, changed_by="bill", timestamp=5, sc_id=4)
+        dbo.rules.scheduled_changes.conditions.history.t.insert().execute(change_id=8, changed_by="bill", timestamp=6, sc_id=4, when=500000, data_version=1)
+        dbo.rules.scheduled_changes.conditions.history.t.insert().execute(change_id=9, changed_by="bill", timestamp=7, sc_id=4, when=500000, data_version=2)
+
+        dbo.rules.scheduled_changes.t.insert().execute(
+            sc_id=5, scheduled_by="bill", data_version=1, complete=False, change_type="delete", base_rule_id=4, base_data_version=1
```
nthomas-mozilla: I see scheduled changes 5, 6, 7 being added but not being used anywhere (excluding the trivial get case). Am I missing something or where they intended for new enactChanges tests ?

mozbhearsum: This bug revolves around the GET for /scheduled_changes/X, so I didn't see any reason to test enacting them. It's an easy addition if you think there's value, though.

auslib/test/admin/views/test_rules.py
```diff
@@ -864,14 +912,38 @@ def testGetScheduledChanges(self):
...
+                    "data_version": None, "systemCapabilities": None, "telemetry_product": None, "telemetry_channel": None, "telemetry_uptake": None,
+                    "change_type": "insert", "signoffs": {}, "required_signoffs": {"relman": 1},
+                },
+                {
+                    "sc_id": 7, "when": 7500000, "scheduled_by": "bill", "complete": False, "sc_data_version": 1, "rule_id": 6, "priority": 40,
+                    "backgroundRate": 50, "product": "fake", "mapping": "a", "update_type": "minor", "version": None,
+                    "buildTarget": None, "alias": None, "channel": "g", "buildID": None, "locale": None, "osVersion": None,
+                    "distribution": None, "fallbackMapping": None, "distVersion": None, "headerArchitecture": None, "comment": None, "whitelist": None,
+                    "data_version": None, "systemCapabilities": None, "telemetry_product": None, "telemetry_channel": None, "telemetry_uptake": None,
+                    "change_type": "update", "signoffs": {}, "required_signoffs": {"releng": 1},
```
nthomas-mozilla: Neither 'update' has different signoff requirements for existing and new state (although you might argue this is covered in the permissions case). If this changed the channel to "k" then a relman signoff would be needed, and the test would cover slightly more ground.

auslib/test/admin/views/test_rules.py
nthomas-mozilla: I think you're right now that I've read more carefully. The separate implementations in web and db layers are worth keeping a close eye on though.

auslib/test/admin/views/test_rules.py
mozbhearsum: That's a good idea, let's do it!

auslib/admin/views/scheduled_changes.py
```diff
@@ -37,14 +38,26 @@ def get(self):
...
+                affected_rows = []
+                # We don't need to consider the existing version of a row for
+                # inserts, because it doesn't exist yet!
+                if row["change_type"] != "insert":
+                    affected_rows.append(self.table.select(where=base_pk)[0])
+                # We don't need to consider the future version of the row when
+                # looking for required signoffs, because it won't exist when
+                # enacted.
+                if row["change_type"] != "delete":
+                    affected_rows.append(base_row)
```
JohanLorenzo: I'm not sure to understand: can a row be neither an `insert` nor a `delete`? If so, should we insert it twice? If not, would it be worth raising an exception?

mozbhearsum: Yeah, change_type can be "insert", "update", or "delete". affected_rows here don't end up being inserted into the database or anything - they're just used for finding the required signoffs for this scheduled change. Does that make sense?

JohanLorenzo: Yes, it does! :D

PR #267, opened 2017-02-27
Fixes Bug #1334188

vedantrathore:
@bhearsum I am still working on this. I added a service to get the username from the api endpoint and passed to the controller but still it is showing as {} on the UI. We can verify this because it's showing on the console. I think It is because the call is asynchronous and the code which shows the name on the ui is executed before the call completes. any suggestions on how to improve this?

### DIFF
--- a/ui/app/js/controllers/rule_scheduled_change_new_controller.js
+++ b/ui/app/js/controllers/rule_scheduled_change_new_controller.js
@@ -66,6 +66,12 @@ function($scope, $http, $modalInstance, CSRF, Releases, Rules, scheduled_changes
         $scope.sc.sc_data_version = 1;
         $scope.sc.sc_id = response.sc_id;
         $scope.scheduled_changes.push($scope.sc);
+        Rules.getScheduledByUser()
+          .then(function successCallback(response) {
+            $scope.sc.scheduled_by = response.data.username;
+          },function errorCallback(response) {
+            console.log(response.data);
+          });
         $modalInstance.close();
       })
       .error(function(response, status) {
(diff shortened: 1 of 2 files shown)


ui/app/js/controllers/rule_scheduled_change_new_controller.js
```diff
@@ -66,6 +66,12 @@ function($scope, $http, $modalInstance, CSRF, Releases, Rules, scheduled_changes
         $scope.sc.sc_data_version = 1;
         $scope.sc.sc_id = response.sc_id;
         $scope.scheduled_changes.push($scope.sc);
+        $scope.sc.scheduled_by = Rules.getScheduledByUser()
+          .then(function successCallback(response) {
+            return response.data.username;
+          },function errorCallback(response) {
+            console.log(response.data);
+          });
```
mozbhearsum: I think the problem here is that you've made $scope.sc.scheduled_by a reference to the Rules.getScheduledByUser() function. The return from the successCallback doesn't get assigned to $scope.sc.scheduled_by - it goes nowhere. Try assigning to $scope.sc.scheduled_by within the successCallback instead.

vedantrathore:
@mozbhearsum this works fine now.

ui/app/js/services/rules_service.js
```diff
@@ -53,6 +53,9 @@ angular.module("app").factory('Rules', function($http) {
       data.csrf_token = csrf_token;
       return $http.post("/api/scheduled_changes/rules", data);
     },
+    getScheduledByUser: function () {
+      return $http.get("/api/users/current");
```
mozbhearsum: You should just use Permissions.getCurrentUser (https://github.com/mozilla/balrog/blob/3c6711fa6a7e72e641742ee610213a39533002d8/ui/app/js/services/permissions_service.js#L6) instead - there's no need to duplicate this here.

mozbhearsum:
@vedantrathore - I'm closing this due to inactivity. Feel free to re-open if you want to continue working on it.

PR #268, opened 2017-02-28
UI for signing off on Scheduled Changes for Rules, Releases, and Permissions

mozbhearsum:
This is mostly just glue code that uses the Signoff and Revocation dialogs that were created in #253. I also merged in #266, because it's hard to test certain cases without it. Because of that, this diff is massive. You can have a look at https://github.com/mozbhearsum/balrog/compare/rs-ui-and-bad-rs...mozbhearsum:signoff-ui?expand=1 to see _just_ the changes from this.


mozbhearsum:
This is up to date now, should be much easier to review!

ui/spec/controllers/rule_scheduled_changes_controller_spec.js
```diff
@@ -20,6 +20,17 @@ describe("controller: RuleScheduledChangesController", function() {
...
 
+  var current_user = {
+    "permissions": {
+      "admin": {
+        "data_version": 1,
+        "options": null,
+      },
+    },
+    "roles": {},
+    "username": "superduperadmin",
```
JohanLorenzo: /me now understands the meaning :)

ui/app/js/services/releases_service.js
```diff
@@ -114,6 +114,16 @@ angular.module("app").factory('Releases', function($http, $q) {
       url += '&csrf_token=' + encodeURIComponent(csrf_token);
       return $http.delete(url);
      },
+
+    signoffOnScheduledChange: function(sc_id, data) {
+      var url = "/api/scheduled_changes/releases/" + sc_id + "/signoffs";
```
JohanLorenzo: Proposal for improvement: I'm not sure. Maybe we can extract the url generation into its own service. For instance:
```js
 angular.module("app").factory('ScheduledChanges', function() {
   var service = {
      signoffUrl: function(signoff_type, sc_id) {
           return "/api/scheduled_changes/" + signoff_type +"/"  + sc_id + "/signoffs"
      }

      revokeSignoffUrl: () => {}; // TODO implement this one
   };
   return service;
});
```

What do you think?

mozbhearsum: I'll give it a shot and see how it looks.

ui/app/js/controllers/permission_scheduled_changes_controller.js
```diff
@@ -112,6 +134,72 @@ function($scope, $routeParams, $location, $timeout, Permissions, Rules, Search,
     });
   };
 
+  $scope.signoff = function(sc) {
+    var modalInstance = $modal.open({
+      templateUrl: "signoff_modal.html",
+      controller: "SignoffCtrl",
+      backdrop: "static",
+      resolve: {
```
mozbhearsum: I'll give this a try, too. It might be a bit weird since they take different args, though.

mozbhearsum: It looks like this isn't really practical, since these (particurarly in the required signoffs controller) depend on args being passed in from the templates. It's possible to factor some of this out, but the added indirection makes it harder to read IMO.

JohanLorenzo: Alright!

ui/app/js/services/releases_service.js
mozbhearsum: Done!

PR #269, opened 2017-03-03
Fix Rules drop down regression from #259

mozbhearsum:


mozbhearsum:
@NinadBhat spelled out the problem and fix for this in https://github.com/mozilla/balrog/pull/259#issuecomment-283281966.

PR #274, opened 2017-03-08
Bug: 1287536. Returning the exceptions from __heartbeat with 502 and a message. r=bhearsum.

tyagi-iiitv:


tyagi-iiitv:
I'm not able to write a test for raising a 502 exception. Please help me with that. Also, is this code okay? or should I return the error message in the response as well?

auslib/dockerflow.py
```diff
@@ -29,6 +29,10 @@ def heartbeat():
         database_entry_value = heartbeat_database_fn(dbo)
         return Response(str(database_entry_value), headers={"Cache-Control": "no-cache"})
 
+    @app.errorhandler(502)
```
mozbhearsum: Because we have control over the function that's returning the error, I don't think using app.errorhandler is the right thing to do. Instead, you can put a try/catch inside of the heartbeat function above, and return a 502 from in the exception case.

This wouldn't work anyways, because the current state of things means 500s are raised for errors, not 502s.

Once you do that, you should notice that the "testHeartbeatWithException" tests from https://github.com/mozilla/balrog/blob/master/auslib/test/admin/views/test_dockerflow.py and https://github.com/mozilla/balrog/blob/master/auslib/test/web/test_dockerflow.py#L25 start failing, which you'll need to update.

As for what to return in the body, the ideal thing is to return the exception name and message, but we need to be certain that there's no possibility of leaking SQLAlchemy credentials from doing so. Can you try to find out if SQLAlchemy ever shows database credentials in Exceptions before we go that route?

mozbhearsum:
> I'm not able to understand how the cr.side_effect = Exception() thing is causing the exception to occur in test_dockerflow.py? Please help me with this as I'm not able to write a test for catching the 500 error.

Remember: your goal is to generate a *502* error for exceptions, not a 500.

In that test, "cr" is a mocked version of the function that the Dockerflow endpoints end up calling. When you give a mocked object a "side_effect" that is an Exception, it causes that Exception to be raised when that function is called. For example:
```
>>> import mock
>>> import time
>>> print time.time()
1489070275.78
>>> with mock.patch("time.time") as mocked_time:
...   mocked_time.side_effect = Exception("Boom!")
...   print time.time()
... 
Traceback (most recent call last):
  File "<stdin>", line 3, in <module>
  File "/home/bhearsum/.virtualenvs/balrog/local/lib/python2.7/site-packages/mock/mock.py", line 1062, in __call__
    return _mock_self._mock_call(*args, **kwargs)
  File "/home/bhearsum/.virtualenvs/balrog/local/lib/python2.7/site-packages/mock/mock.py", line 1118, in _mock_call
    raise effect
Exception: Boom!
```

So, that means that when the test makes a request to /__heartbeat__, an Exception will get raised. With the current code, this will go uncaught in the Dockerflow endpoint, and end converting to a 500 automatically by Flask. When you start catching the Exception in the Dockerflow endpoint, you'll be able to returning a 502. In the unit test, you can verify this by checking the value of "ret.status_code", like we do testHeartbeat.

Does that help?

tyagi-iiitv:
For the test that I've put in test_dockerflow.py, I'm getting **ret.status_code = 200** while I've explicitly mentioned the status code for exceptions to be returned as 500. I think I'm wrong in the way I'm initializing the status code. Please suggest?

auslib/dockerflow.py
```diff
@@ -26,8 +26,12 @@ def heartbeat():
...
         Respond to /__heartbeat__ with a HTTP 200 or 5xx on error. This should
         depend on services like the database to also ensure they are healthy."""
-        database_entry_value = heartbeat_database_fn(dbo)
-        return Response(str(database_entry_value), headers={"Cache-Control": "no-cache"})
+        try:
+            database_entry_value = heartbeat_database_fn(dbo)
+            return Response(str(database_entry_value), headers={"Cache-Control": "no-cache"})
+        except Exception as e:
+            e.status_code = 500
+            return Response(e)
```
mozbhearsum: First of all, please remember that you're supposed to be returning a 502, _not_ a 500.

The reason this isn't working is beacuse you're not constructing a response correctly. Have a look at http://flask.pocoo.org/docs/0.10/api/#flask.Response to see the API of that class. You can also see some examples of setting a custom status code and body with Response if you look around the Balrog code base.

auslib/test/admin/views/test_dockerflow.py
```diff
@@ -31,7 +31,9 @@ def testHeartbeatWithException(self):
             cr.side_effect = Exception("kabom!")
             # Because there's no web server between us and the endpoint, we receive
             # the Exception directly instead of a 500 error
-            self.assertRaises(Exception, self.client.get, "/__heartbeat__")
+            # self.assertRaises(Exception, self.client.get, "/__heartbeat__")
+            ret = self.client.get("/__heartbeat__")
+            self.assertEqual(ret.status_code, 500)
             self.assertEqual(cr.call_count, 1)
```
mozbhearsum: Please ensure you test the response body, too, because the bug asks us to ensure that 502s return useful information.

tyagi-iiitv:
okay, so I'm done with generating the 502 status code and logging the traceback of the exception. For the data that exception generates, see the logs at [http://pastebin.com/9s1WQ64h](http://pastebin.com/9s1WQ64h). Now for the sqlalchemy part, I didn't see any credentials listed in the logs. But I'm not sure that I'm doing it the right way.. Am I doing it right?

auslib/dockerflow.py
```diff
@@ -26,8 +28,12 @@ def heartbeat():
...
         Respond to /__heartbeat__ with a HTTP 200 or 5xx on error. This should
         depend on services like the database to also ensure they are healthy."""
-        database_entry_value = heartbeat_database_fn(dbo)
-        return Response(str(database_entry_value), headers={"Cache-Control": "no-cache"})
+        try:
+            database_entry_value = heartbeat_database_fn(dbo)
+            return Response(str(database_entry_value), headers={"Cache-Control": "no-cache"})
+        except Exception as e:
+            logging.exception(e)
+            return Response(status=502)
```
mozbhearsum: You're on the right track here, but remember the bug asks for "useful information", so you'll want to provide the "response" arg to Response as well. Perhaps something similar to https://github.com/mozilla/balrog/blob/8f8057168b0ab13177468cac93b85af4cfcb2472/auslib/admin/views/permissions.py#L89 ?

auslib/test/admin/views/test_dockerflow.py
```diff
@@ -31,7 +31,8 @@ def testHeartbeatWithException(self):
             cr.side_effect = Exception("kabom!")
             # Because there's no web server between us and the endpoint, we receive
             # the Exception directly instead of a 500 error
-            self.assertRaises(Exception, self.client.get, "/__heartbeat__")
+            ret = self.client.get("/__heartbeat__")
+            self.assertEqual(ret.status_code, 502)
             self.assertEqual(cr.call_count, 1)
```
mozbhearsum: It's also a good idea to check "ret.data" to make sure it contains the expected information in the body. You can see an example of this in https://github.com/mozilla/balrog/blob/8f8057168b0ab13177468cac93b85af4cfcb2472/auslib/test/admin/views/test_permissions.py#L255.

tyagi-iiitv:
Ready for review !!

auslib/dockerflow.py
```diff
@@ -26,8 +28,12 @@ def heartbeat():
...
         Respond to /__heartbeat__ with a HTTP 200 or 5xx on error. This should
         depend on services like the database to also ensure they are healthy."""
-        database_entry_value = heartbeat_database_fn(dbo)
-        return Response(str(database_entry_value), headers={"Cache-Control": "no-cache"})
+        try:
+            database_entry_value = heartbeat_database_fn(dbo)
+            return Response(str(database_entry_value), headers={"Cache-Control": "no-cache"})
+        except Exception as e:
+            logging.exception("Error occured: %s", e.message)
+            return Response(status=502, response=e.message)
```
mozbhearsum: @mostlygeek - do you have any strong feelings about including the original Exception message? In local testing, this generates errors like this if I use the wrong db credentials:
> (OperationalError) (1045, "Access denied for user 'balrogadmin'@'172.17.0.4' (using password: YES)")

If this were to happen with the public app in production, the database server IP would be publicly available. I'm not sure how else to provide useful error messages for exceptions, though.

mostlygeek: The database is inaccessible from the public internet so no big deal. I'm OK with it. 
@jvehent any comments?

jvehent: If I understand correctly, you're concerned about leaking internal IPs in public error message. Is that correct?
If so, this isn't a concern. We've long accepted that internal IPs and hostnames are not privileged information. Just make sure not to return any configuration secrets in the error ;)

mozbhearsum: Yup, I was specifically asking about IPs here.

It's hard to guarantee that we're not leaking configuration secrets when blindly returning an Exception message, but SQLAlchemy doesn't print passwords AFAICT, so I think that's very unlikely. I don't think we have any secrets on the public app other than the database password anyways.

jvehent: Why are you returning exceptions to clients? Wouldn't it be better to ship those into sentry and return a generic "an error occured" to the public?

mozbhearsum: In https://bugzilla.mozilla.org/show_bug.cgi?id=1287536, @mostlygeek requested that we return a 502 with a useful error message. I'm good with either, though.

jvehent: I think a generic, midly informative, message with an ID you could map to a stacktrace in sentry would alleviate the risk of leaking confidential data to the public.

mozbhearsum: Would this still satisfy your needs, @mostlygeek?

mostlygeek: yup. the error message is really to help ops figure out where to start looking for errors. In balrog's case it's most likely be the db not working.

mozbhearsum:
@tyagi-iiitv - sorry for the false start here, but this will need to be worked slightly. Instead of sending part of the Exception in the response, we want to send a generic error. However, we must ensure that Exceptions from __heartbeat__ are propagated to Sentry. You'll need a separate class to do that part, because the error handlers must be enabled. https://github.com/mozilla/balrog/blob/master/auslib/test/web/test_client.py#L1025 has an example of this (note the difference between that class' constructor, and TestDockerflowEndpoints/ClientTestBase.

tyagi-iiitv:
I don't know how to test the sentry code that I've included in dockerflow.py. Also, I'm not sure whether the changes I've made are correct or not. Please verify and suggest the changes.

auslib/dockerflow.py
```diff
@@ -1,9 +1,15 @@
...
+
 from os import path
 
 from flask import Response, jsonify
 
 from auslib.global_state import dbo
 
+from raven.contrib.flask import Sentry
+
+sentry = Sentry()
```
mozbhearsum: You shouldn't need to do any of the Sentry set-up yourself - it is already set-up by other parts of Balrog for the entire application.

auslib/dockerflow.py
```diff
@@ -26,8 +32,14 @@ def heartbeat():
...
-        database_entry_value = heartbeat_database_fn(dbo)
-        return Response(str(database_entry_value), headers={"Cache-Control": "no-cache"})
+        try:
+            database_entry_value = heartbeat_database_fn(dbo)
+            return Response(str(database_entry_value), headers={"Cache-Control": "no-cache"})
+        except Exception as e:
+            logging.exception("Error Occured %s", e.message)
+            return Response(status=502, response="Error occured")
+            if sentry.client:
+                sentry.captureMessage(e.message)
```
mozbhearsum: Same here - Sentry will automatically capture exceptions if a 500 is returned. This is after the return statement anyways, so it will never run :)

auslib/test/web/test_dockerflow.py
```diff
@@ -27,10 +27,25 @@ def testHeartbeatWithException(self):
...
+class TestHeartbeatExceptionSentry(ClientTestWithErrorHandlers):
+
+    def testHeartbeatWithException(self):
+        with mock.patch("auslib.global_state.dbo.rules.countRules") as sentry:
+            sentry.side_effect = Exception("kabom!")
+            # Because there's no web server between us and the endpoint, we receive
+            # the Exception directly instead of a 500 error
+            ret = self.client.get("/__heartbeat__")
+            self.assertEqual(ret.status_code, 502)
+            self.assertEqual(ret.data, "kabom!")
```
mozbhearsum: I think you need to update the "kabom" here :).

auslib/test/web/test_dockerflow.py
```diff
@@ -27,10 +27,25 @@ def testHeartbeatWithException(self):
...
+            self.assertEqual(ret.data, "Error occured")
             self.assertEqual(cr.call_count, 1)
 
     def testLbHeartbeat(self):
         ret = self.client.get("/__lbheartbeat__")
         self.assertEqual(ret.status_code, 200)
         self.assertEqual(ret.headers["Cache-Control"], "no-cache")
+
+
+class TestHeartbeatExceptionSentry(ClientTestWithErrorHandlers):
```
mozbhearsum: Good job adding this test, but be carefula bout inheriting from ClientTestWithErrorHandlers. Becasue that class has tests of its own, they end up getting re-run as part of this. To avoid this, you need to do one of two things:
- Move ClientTestWithErrorHandlers to a shared base class
- Don't inherit from ClientTestWithErrorHandlers, and copy its setUp to this class.

tyagi-iiitv:
Thanks for the feedback but I need to clarify a few things here:

- How can I send the exception message from dockerflow.py to sentry? If it's automatically sent to sentry with the 502 exception, then I can send the error message in the error response. If it's not automatically sent to sentry, then I think there would be two seperate responses, one for user with a generic "Error occured" message and one for sentry with the error details. How can I do that?

- I can't understand what error handlers do and why exactly I've to create a new class in TestDockerflow.py which works with error handlers activated to test Sentry. In all, I'm not able to understand how Sentry works  :)

mostlygeek:
Not sure if I'm too late to the game with some guidance here. 

For balrog, I think it's enough for __heartbeat__ to just check the database when it's called. If the app can't talk to the database, 502 with an error message, "can't connect to db" (or something). A full error/stack trace into sentry would be an extra bonus but it's not necessary.

If we add more dependent services in the future just add another check to __heartbeat__. 

Also, I recommending adding a Cache-Control header of 30 seconds to 1 minute to prevent an easy DDOS against the endpoint. We have nginx in front of balrog with a microcache.

mozbhearsum:
> For balrog, I think it's enough for __heartbeat__ to just check the database when it's called. If the app can't talk to the database, 502 with an error message, "can't connect to db" (or something). A full error/stack trace into sentry would be an extra bonus but it's not necessary.

That should be easy to do here - just need to change the fixed error message. Can you do that @tyagi-iiitv?

> Also, I recommending adding a Cache-Control header of 30 seconds to 1 minute to prevent an easy DDOS against the endpoint. We have nginx in front of balrog with a microcache.

This as well. The line at https://github.com/mozilla/balrog/pull/274/files#diff-fd21452374dd087f63213eb2eb819727R37 just needs to be adjusted. Have a look at https://github.com/mozilla/balrog/blob/master/auslib/web/views/client.py#L18 for an example of how to set CACHE_CONTROL correctly.

tyagi-iiitv:
yes, I'll work on it. So now I don't we don't require to send the error message to Sentry, right?

tyagi-iiitv:
Ready for review!!

auslib/dockerflow.py
```diff
@@ -26,8 +26,11 @@ def heartbeat():
         """Per the Dockerflow spec:
         Respond to /__heartbeat__ with a HTTP 200 or 5xx on error. This should
         depend on services like the database to also ensure they are healthy."""
-        database_entry_value = heartbeat_database_fn(dbo)
-        return Response(str(database_entry_value), headers={"Cache-Control": "no-cache"})
+        try:
+            database_entry_value = heartbeat_database_fn(dbo)
+            return Response(str(database_entry_value), headers={"Cache-Control": "no-cache"})
```
mozbhearsum: This Cache-Control should be set to the same value as below. Using no-cache here means we're still vulnerable to DoS when the database is up. Other than that, this looks good to me!

tyagi-iiitv:
Done !!

mozbhearsum:
Thanks for your persistence here!

PR #276, opened 2017-03-09
B1325377 - Public API to Balrog's database

ex-Dev:
Hey Ben, 
Now we can talk about this bug.

Curretly the following changes was made:
-Publi api resides in auslib.web.public and ClientRequestView is now the auslib.web.public.api module (functions are module based);
-The swagger specification describes all the public api endpoint (auslib.web.public.api.yml);
-The mostly tests were fixed;
-Connexion is configured in auslib.web.public.base line 13;
- I don't have touched in admin api.
-New shared endpoints (getters) is the next step to develop. I intent to put they in auslib.web at module level on __init__.

I wait for your feedback.
Thanks!


auslib/web/public/base.py
```diff
@@ -1,19 +1,22 @@
...
 
 from raven.contrib.flask import Sentry
 
 from auslib.AUS import AUS
 from auslib.dockerflow import create_dockerflow_endpoints
 
-app = Flask(__name__)
+connexion_app = connexion.App(__name__)
+connexion_app.add_api('api.yml')
+app = connexion_app.app
```
mozbhearsum: I'm a little surprised that Connexion forces you to use its App instead of providing a middleware, but this is OK.

auslib/web/public/base.py
```diff
@@ -78,6 +81,15 @@ def contributejson():
...
+    # By default, we want a cache that can be shared across requests from
+    # different users ("public").
+    # and a maximum age of 90 seconds, to keep our TTL low.
+    # We bumped this from 60s -> 90s in November, 2016.
+    setattr(app, 'cacheControl', app.config.get("CACHE_CONTROL", "public, max-age=90"))
+
+
 @app.route('/update/1/%PRODUCT%/%VERSION%/%BUILD_ID%/%BUILD_TARGET%/%LOCALE%/%CHANNEL%/update.xml')
 @app.route('/update/2/%PRODUCT%/%VERSION%/%BUILD_ID%/%BUILD_TARGET%/%LOCALE%/%CHANNEL%/%OS_VERSION%/update.xml')
 @app.route('/update/3/%PRODUCT%/%VERSION%/%BUILD_ID%/%BUILD_TARGET%/%LOCALE%/%CHANNEL%/%OS_VERSION%/%DISTRIBUTION%/%DISTRIBUTION_VERSION%/update.xml')
```
mozbhearsum: Can these routes by handled by the Swagger spec, too?

auslib/web/public/api.yml
```diff
@@ -0,0 +1,474 @@
+swagger: '2.0'
```
mozbhearsum: I think you'll need to add this file to MANIFEST.in to make sure it ends up in built packages.

auslib/web/public/api.py
```diff
@@ -0,0 +1,157 @@
+import urllib
```
mozbhearsum: Naming quibble: I don't think "api.py" is the right name for this - we should save that for when there's a real public API (eg: one with endpoints like /releases and /rules). I think we should stick with "client.py" for this, because they are endpoints that exist to serve the update client in Firefox.

auslib/web/public/api.py
```diff
@@ -0,0 +1,157 @@
...
+def get_update_blob_4(**url):
+    return get_update_blob(**url)
+
+
+def get_update_blob_5(**url):
+    return get_update_blob(**url)
+
+
+def get_update_blob_6(**url):
+    return get_update_blob(**url)
```
ex-Dev: I tried it, but connexion requires different function names.

auslib/web/public/base.py
ex-Dev: So far, it is just to maintain everything working

auslib/web/public/api.yml
```diff
@@ -0,0 +1,474 @@
...
+        - name: channel
+          in: path
+          description: The update channel of the application request an update.
+          type: string
+        # Underlying code depends on osVersion being set. Since this route only
+        # exists to support ancient queries, and all newer versions have osVersion
+        # in them it's easier to set this here than make the all of the underlying
+        # code support queries without it.
+        - name: queryVersion
+          in: query
```
ex-Dev: Yeah, I could not get the same  flask "defaults" behavior, yet.

auslib/web/public/api.py
mozbhearsum: Ah, that's too bad.

ex-Dev: I was reading the connexion code today and saw that the feature to reuse the operation ids already exists, but the connexion team does not expose it yet.
https://github.com/zalando/connexion/pull/275
https://github.com/zalando/connexion/pull/278

I extended the App and Api class to use this feature, but for some reason the mock tests stopped to track the calls (but the call occurs).
For the following test, the Dalek exception does not thrown:
https://github.com/ex-Dev/balrog/blob/B1325377/auslib/test/web/test_client.py#L1027

What you think about extend connexion to handle this?

ex-Dev:
I should had test the bellow url before than made my changes, but I forgot:

http://127.0.0.1:9090/update/3/Firefox/33.0/20141202185629/Darwin_x86_64-gcc3-u-i386-x86_64/en-US/release/default/default/default/update.xml?force=1

Currently I get:

`<?xml version="1.0"?>
<updates>
    <update type="minor" displayVersion="51.0" appVersion="51.0" platformVersion="51.0" buildID="20170118123726" detailsURL="https://www.mozilla.org/en-US/firefox/51.0/releasenotes/">
        <patch type="complete" URL="http://download.mozilla.org/?product=firefox-51.0-complete&amp;os=osx&amp;lang=en-US&amp;force=1" hashFunction="sha512" hashValue="824bdf897b9672d2995dd814a6c4306fa3445da7bfbc4cb52d1dfeb38fd41e1a6ca2d31a24047d240fd2a1c880c4b956e5bdf2d757cfc68a300fe7472a55d30c" size="83382822"/>
    </update>
</updates>`

auslib/web/public/api.py
mozbhearsum: Nice find there. Were these tests passing before you started re-using operation ids?

This is very strange, it looks like you're mocking correctly, yet the function still runs. It makes me wonder if connexion is creating a copy of the function under the hood, or something. If so, this might be very difficult to deal with :(.

mozbhearsum:
The response in your previous comment looks right to me. Once you figure out the issue with the get_update_blob tests, I'll give this an in depth review. I was also thinking it would be a good idea to split this bug into two PRs:
1) Convert existing endpoints to Swagger (what you're already doing in this one)
2) Add the new endpoints to directly query Rules/Releases (which you haven't started on yet)

ex-Dev:
Ok!
I'll investigate more the mock case.
But so far, I know that if I change from "get_update_blob" to "getQueryFromURL" (first call of get_update_blob), the mock works.

mozbhearsum:
It definitely sounds like connexion is mangling the operation in some way, in that case. If mocking getQueryFromURL works, that's a fine workaround - all we're trying to do is ensure the endpoint raises an exception in some way, we don't really care how.

auslib/web/balrog_api.py
```diff
@@ -0,0 +1,76 @@
+import connexion
+from connexion.operation import Operation
+from connexion.resolver import Resolver
+
+
+RANDOM_DIGITS = 6
+
+
+class BalrogApi(connexion.Api):
```
mozbhearsum: What's the purpose of this class?

ex-Dev: The feature to reuse function name (OperationId) is disabled by default. So I had to inherit from base connexion classes.

mozbhearsum: Ah, I see! Is there a way to do it without replicating all this code from Connexion? Unfortunately, this code is a maintenance nightmare, it will require updates whenever Connexion internals change :(. It's actually better to use separate methods for endpoint than do this.

ex-Dev: OK, I'll made changes this weekend.
I sent an e-mail to connexion devs asking if they plan expose this feature. Waiting answers haha.

mozbhearsum: Sounds good, thanks for working through this!

ex-Dev:
I made some changes:

-Added a custom validator to handle flask defaults{}, from the path an not from the query:
http://connexion.readthedocs.io/en/latest/request.html#custom-validators
https://github.com/mozilla/balrog/blob/master/auslib/web/base.py#L101

-Don't reusing OperationId;

-Currently I have a problem with Avast fixes. For some reason when using connexion "request.args" does not changes but the "request.environ["QUERY_STRING"]" does:
https://github.com/mozilla/balrog/blob/master/auslib/web/views/client.py#L39
https://github.com/mozilla/balrog/blob/master/auslib/web/views/client.py#L51

Makes sense, because request.args is a cached_property, but why works with simple flask routes and does not work with connexion? Any suggestion?
https://github.com/pallets/werkzeug/blob/master/werkzeug/wrappers.py#L439

mozbhearsum:
I haven't looked into it deeply, but your theory that request.args isn't changing because it is a cached property seems plausible. You may also want to try using `connexion.request.args` instead of Flask's version, per https://github.com/zalando/connexion/blob/master/docs/request.rst#request-handling.

If the avast stuff needs to be reworked to make this work, that's totally fine - I'm not tied to the current implementation. Maybe there's a more elegant way to do it.

auslib/web/api_validator.py
```diff
@@ -0,0 +1,26 @@
...
+
+CONNEXION_IGNORE_VALIDATION_AVAST = ["avast", "force"]
+
+
+class BalrogParameterValidator(ParameterValidator):
+    def __init__(self, *args, **kwargs):
+        super(BalrogParameterValidator, self).__init__(*args, **kwargs)
+
+    def validate_path_parameter(self, args, param):
+        val = args.get(param['name'].replace('-', '_'))
```
mozbhearsum: What's the purpose of this?

ex-Dev: If a path parameter is not supplied, is required and have a default value, the default value is used.
Like here: https://github.com/mozilla/balrog/blob/master/auslib/web/base.py#L101

ex-Dev:
I tried the `connexion.request.args`, but does not work too. I'll work on avast validation.
In addition, do you have other considerations or changes in mind?

auslib/web/api_validator.py
mozbhearsum: I think the part that's confusing me is the dash to underscore replacement.

auslib/web/public/base.py
```diff
@@ -0,0 +1,97 @@
...
+from auslib.web.api_validator import BalrogParameterValidator
+
+log = logging.getLogger(__name__)
+
+validator_map = {
+    'parameter': BalrogParameterValidator
+}
+
+connexion_app = connexion.App(
+    __name__, specification_dir='.', validator_map=validator_map)
```
mozbhearsum: Can we turn on response validation here?

auslib/web/public/client.py
```diff
@@ -0,0 +1,159 @@
...
+AUS = AUS()
+LOG = logging.getLogger(__name__)
+
+update_blob_functions = ["get_update_blob_1",
+                         "get_update_blob_2",
+                         "get_update_blob_3",
+                         "get_update_blob_3_esrpre",
+                         "get_update_blob_4",
+                         "get_update_blob_5",
+                         "get_update_blob_6"]
```
mozbhearsum: This should probably move to the bottom, near where they get used.

mozbhearsum:
Overall, I think things are looking pretty good. I really like the new layout for the web apps (having both underneath "web") in particular. I think my only real concern is around api_validator.py. The code seems pretty hacky...but I'm also not sure what else to suggest. If there's any alternatives you can think of to that, I'd love to hear them.

auslib/web/api_validator.py
ex-Dev: Ah, It is only to maintain the base class behavior.

auslib/web/public/base.py
ex-Dev: This is the Swagger spec validation level.

ex-Dev:
Hi Ben, about the custom validator, this is a supported feature by connexion (http://connexion.readthedocs.io/en/latest/request.html#custom-validators).

The last updates was:
-Avast fixes rewritted;
-Connexion version updated to 1.1.9 (https://github.com/zalando/connexion/releases/tag/1.1.9).

mozbhearsum:
@ex-Dev - it looks like this is more or less ready as far as replacing the existing endpoints with swagger ones goes. Do you think you could update it against master so I can give it a final review? The new endpoints for the public API can be tackled in a follow-up, to reduce risk.

ex-Dev:
Branch updated against balrog/master, waiting for review, suggestions and request for changes.
Thanks!

mozbhearsum:
Don't worry about squashing - I'll squash merge to master when this is done. I probably won't be able to give this a good luck for a week or so, I got caught up in a high priority project.

auslib/web/api_validator.py
```diff
@@ -0,0 +1,24 @@
...
+        super(BalrogParameterValidator, self).__init__(*args, **kwargs)
+
+    def validate_path_parameter(self, param, request):
+        # Keep the base class behavior.
+        val = request.path_params.get(param['name'].replace('-', '_'))
+
+        # Supporting the flask add_url_rule "defaults" parameter to the swagger path parameter.
+        if val is None and 'default' in param:
+            val = param['default']
+            request.path_params[param['name']] = val
```
mozbhearsum: I'm still a little confused about why this is necessary after a reread. According to the Swagger spec, defaults are supported...and Connexion also seems to support default parameters. I see that we get an error like this if we remove your validator:
```
<Response streamed [400 BAD REQUEST]>
{
  "detail": "Missing path parameter 'queryVersion'", 
  "status": 400, 
  "title": "Bad Request", 
  "type": "about:blank"
}

```

But I don't quite understanding why - can you explain that a bit?

auslib/web/public/api.yml
```diff
@@ -0,0 +1,539 @@
...
+        # Underlying code depends on osVersion being set. Since this route only
+        # exists to support ancient queries, and all newer versions have osVersion
+        # in them it's easier to set this here than make the all of the underlying
+        # code support queries without it.
+        - name: queryVersion
+          in: path
+          description: queryVersion
+          required: true
+          type: integer
+          default: 2
```
mozbhearsum: Looks like this is set incorrectly - should be 1.

auslib/web/public/api.yml
```diff
@@ -0,0 +1,539 @@
...
+          type: string
+        - name: channel
+          in: path
+          description: The update channel of the application request an update.
+          required: true
+          type: string
+        # Underlying code depends on osVersion being set. Since this route only
+        # exists to support ancient queries, and all newer versions have osVersion
+        # in them it's easier to set this here than make the all of the underlying
+        # code support queries without it.
```
mozbhearsum: This comment needs to move closer to osVersion.

auslib/web/public/api.yml
```diff
@@ -0,0 +1,539 @@
...
+          description: queryVersion
+          required: true
+          type: integer
+          default: 2
+        - name: osVersion
+          in: path
+          description: The OS Version of the application requesting an update. This field is primarily used to point desupported operating systems to their last supported build.
+          required: true
+          type: string
+          default: ""
```
mozbhearsum: Looks like you're missing "force" on this and the /update/2 endpoint.

auslib/web/public/api.yml
```diff
@@ -0,0 +1,539 @@
+swagger: '2.0'
+info:
```
mozbhearsum: Is there a reason you didn't add the dockerflow, contribute.json, and robots.txt endpoints to the spec?

auslib/web/public/base.py
mozbhearsum: Are you sure about this? I think you need to set at least validate_responses=True, and maybe strict_validation=True as well, per http://connexion.readthedocs.io/en/latest/response.html#response-validation.

auslib/web/api_validator.py
ex-Dev: Yeah, defaults are supported, but for query parameters, not for path parameters.
See http://swagger.io/specification/#parameter-object-44

required: "...If the parameter is in "path", this property is required and its value MUST be true...."

default: "...Note: "default" has no meaning for required parameters..."

If you remove the validator, the defaults parameters are not included in the path_parameters dict (see line 22).

auslib/web/public/base.py
ex-Dev: Ah! I saw that you had asked about it above, but I had understood another thing.

auslib/web/public/api.yml
ex-Dev: My reference was this:
https://github.com/mozilla/balrog/blob/master/auslib/web/base.py#L101

Need to change to 1?

ex-Dev:
New commit added!

auslib/web/public/api.yml
ex-Dev: Ah, I forgot those endpoints..

auslib/web/public/api.yml
mozbhearsum: Yeah, it looks like the original implementation was buggy! I don't think it ends up getting used, but it's probably best to get it right.

mozbhearsum:
Unfortunately I've had to back this out in #313 due to some issues with update URLs that contain unicode. I don't think it's anything major, but it may have been causing problems for some System Addon and GMP updates. I left some more details in https://bugzilla.mozilla.org/show_bug.cgi?id=1325377#c8

PR #277, opened 2017-03-10
Bug 1331992: Update the requirement files with working hash values and add dependencies.r=bhearsum

tyagi-iiitv:


tyagi-iiitv:
Updates on hash checking:

- When running **python setup.py install**, there is **NO** hash checking.
- If running **pip install .** inside the balrog directory, then also it eventually runs **python setup.py install** on all the dependencies and hence, **NO** hash checking is done.

So, the correct way to check for the right packages using hash checking and also installing balrog is by running **pip install -r requirements.txt** which supports hash checking and then run **pip install --no-deps .** to run setup.py without installing dependencies.

requirements-test.txt
```diff
@@ -1,24 +1,33 @@
 certifi==2016.8.31 \
-#    --hash=sha256:f7708a42d86f29ccc7c8c4ff9d34a8d854d8d78eb2973d1f28406bb43d6b8919
+    --hash=sha256:172844a3e2896c924e6aa2657b2e56586a724f27edd0e9592be9353f3a3f6c73
```
mozbhearsum: I'm surprised to see so many of these hashes changing - where are you getting them from?

requirements.txt
```diff
@@ -1,48 +1,48 @@
...
 flask==0.10.1 \
-#    --hash=sha256:4c83829ff83d408b5e1d4995472265411d2c414112298f2eb4b359d9e4563373
+    --hash=sha256:4c83829ff83d408b5e1d4995472265411d2c414112298f2eb4b359d9e4563373
 flask-compress==1.3.0 \
-#    --hash=sha256:e6c52f1e56b59e8702aed6eb73c6fb0bffe942e5ca188f10e54a33ec11bc5ed4
+    --hash=sha256:e6c52f1e56b59e8702aed6eb73c6fb0bffe942e5ca188f10e54a33ec11bc5ed4
 flask-wtf==0.12 \
-#    --hash=sha256:bd99316c97ed1d1cb90b8f0c242c86420a891a6a2058f20717e424bf5b0bb80e --hash=sha256:99c5d75663d6bfcbf4dd0d71ffcb7313f64ae90e6f89e09ad99080107033e8b6
+    --hash=sha256:bd99316c97ed1d1cb90b8f0c242c86420a891a6a2058f20717e424bf5b0bb80e --hash=sha256:99c5d75663d6bfcbf4dd0d71ffcb7313f64ae90e6f89e09ad99080107033e8b6
 # Uncomment me when Python 2.6 is desupported.
```
mozbhearsum: Since you're (correctly) uncommenting functools, can you kill this comment too?

requirements-test.txt
tyagi-iiitv: Yes, Even I was surprised, I hope I'm doing it the right way. So this is what I did:

- Uncomment the hashes in requirements.txt file
- pip uninstall -r requirements.txt
- pip install -r requirements.txt
    - I get errors in this step where the hash isn't matching, and the expected hash is displayed by pip. So, I replace the given hash with the suggested hash. See the logs http://pastebin.com/Pcac94Ff

Also, If I'm right in detecting the right hash values, then we should think about implementing hash checking in balrog in development phase because with hash values changing, this can result in frequent failures while deploying.

requirements.txt
tyagi-iiitv: Done !!

tyagi-iiitv:
> If there's a way to also ensure we always verify hashes in local development, that would be good too. But I think it's more important that "python setup.py install" continues to function in a non-surprising way for developers. Do you think there's a way to do that, and verify hashes?

To let setup.py work as usual, I've made changes so that it ignore the comment and hash-values line. See below:
```
with open(path.join(here, "requirements.txt")) as f:
    requirements = []
    for line in f:
        # Skip comment lines and hash values
        if (not line.strip().startswith("#") and not line.strip().startswith("--")):
            requirements.append(line.split()[0])
```
This will not change the functionality of setup.py. However, to ensure that setup.py works along with hash checking, we've to tweak with the Balrog code to run `pip install -r requirements.txt` before running `setup.py install`.

requirements-test.txt
mozbhearsum: I agree that it's good to verify hashes in local development if we can - and I think we are for anyone using Docker, because it uses pip to install dependencies, and doesn't every install the Balrog package.

mozbhearsum:
re: tweaking Balrog code to run pip install before setup.py - which code are you talking about?

requirements-test.txt
mozbhearsum: It looks like the hash differences are because some packages have multiple formats. Eg, certifi is available as a .whl and a .tar.gz on https://pypi.python.org/pypi?:action=show_md5&digest=0931a7d2576083826efa419d695117a3:
```
➜  sha256sum *
172844a3e2896c924e6aa2657b2e56586a724f27edd0e9592be9353f3a3f6c73  certifi-2016.8.31-py2.py3-none-any.whl
f7708a42d86f29ccc7c8c4ff9d34a8d854d8d78eb2973d1f28406bb43d6b8919  certifi-2016.8.31.tar.gz
```

I know it's going to be a hassle, but can you make sure all hashes from each package are specified?

tyagi-iiitv:
Ready for review !!

setup.py
```diff
@@ -9,8 +9,8 @@
 with open(path.join(here, "requirements.txt")) as f:
     requirements = []
     for line in f:
-        # Skip comment lines
-        if not line.startswith("#"):
+        # Skip comment lines and hash values
+        if (not line.strip().startswith("#") and not line.strip().startswith("--")):
```
mozbhearsum: Now that we're not commenting out hashes, you can get rid of the first part of this if statement. I think you'll need to add the new part to agent/setup.py, too.

tyagi-iiitv:
Done adding changes .. Ready for review.

PR #280, opened 2017-03-17
Bug 1312499 - figure out how to match multiple substrings in one rule field

sirMackk:
This PR is a work in progress! I'll rebase and edit this PR message before it's ready for review and fix any tests - I just wanted to get a bit of feedback before continuing the work.

I went with rail's suggestion to the keep the comma-separated format and introduce "&&" and "!" to the syntax - which makes the whole thing a lot simpler and naturally imposes some constraints so that rules won't get out of hand. That said, I wanted to clarify two things:

1. Looking at the [bugzilla entry](https://bugzilla.mozilla.org/show_bug.cgi?id=1312499), it looks to me that we only want to introduce this simplified boolean matching to the osVersion and systemCapabilities fields - is this correct?
2. Has this functionality been implemented in the clients yet? Like, is Balrog getting any GET requests that feature extended systemCapabilities information? I'm mainly trying to pinpoint how the URL will look then - I'm currently assuming that it'll be something like `.../GenuineIntel,family6,model61,SSE3/...`. 

I'll be squaring away the tests and code in what I have so far and once I figure out the two things above, I'll add them in and make this a real PR.

Many thanks!

### DIFF
--- a/auslib/db.py
+++ b/auslib/db.py
@@ -1539,36 +1539,38 @@ class Rules(AUSTable):
             return True
         return string_compare(queryBuildID, ruleBuildID)
 
-    def _simpleBooleanMatchesSubRule(self, subRuleString, preparedQuery):
+    def _simpleBooleanMatchesSubRule(self, subRuleString, queryString, substring):
+        """Performs the actual logical 'AND' operation on a rule as well as partial/full string matching
+           for each section of a rule.
+           If all parts of the subRuleString match the queryString, then we have successfully resolved the
+           logical 'AND' operation and return True.
+           Partial matching makes use of Python's "<substring> in <string>" functionality, giving us the ability
+           for an incoming rule to match only a substring of a rule.
+           Full matching makes use of Python's "<string> in <list>" functionality, giving us the ability for
+           an incoming rule to exactly match the whole rule. Currently, incoming rules are comma-separated strings."""
         for rule in subRuleString:
-            if rule not in preparedQuery:
+            if substring and rule not in queryString:
+                return False
+            elif not substring and rule not in queryString.split(','):
                 return False
         return True
 
     def _simpleBooleanMatchesRule(self, ruleString, queryString, substring=True):
         """Decides whether a column from a rule matches an incoming one using simplified boolean logic.
-        Only two operators are supported: '&&' (and), ',' (or). A rule like 'AMD,SSE' will match incoming
-        rules that contain either 'AMD' or 'SSE'. A rule like 'AMD&&SSE' will only match incoming rules
-        that contain both 'AMD' and 'SSE'.
-        This function can do substring matching or full string matching. When doing substring matching, a rule
-        specifying 'AMD,Windows 10' WILL match an incoming rule such as 'Windows 10.1.2'. When doing full string
-        matching, a rule specifying 'AMD,SSE' will NOT match an incoming rule that contains 'SSE3', but WILL match
-        an incoming rule that contains either 'AMD' or 'SSE3'.
-        """
+           Only two operators are supported: '&&' (and), ',' (or). A rule like 'AMD,SSE' will match incoming
+           rules that contain either 'AMD' or 'SSE'. A rule like 'AMD&&SSE' will only match incoming rules
+           that contain both 'AMD' and 'SSE'.
+           This function can do substring matching or full string matching. When doing substring matching, a rule
+           specifying 'AMD,Windows 10' WILL match an incoming rule such as 'Windows 10.1.2'. When doing full string
+           matching, a rule specifying 'AMD,SSE' will NOT match an incoming rule that contains 'SSE3', but WILL match
+           an incoming rule that contains either 'AMD' or 'SSE3'."""
         if ruleString is None:
             return True
 
         decomposedRules = [[rule.strip() for rule in subRule.split('&&')] for subRule in ruleString.split(',')]
-        # If we want to do substring matching (ie. inexact matching), we preserve the queryString as a string,
-        # but if we want to do exact matching, we turn the queryString into a list. This works by using Python's
-        # "in" operator that does inexact matching for strings, but exact matching for lists.
-        if substring:
-            preparedQuery = queryString
-        else:
-            preparedQuery = queryString.split(',')
 
         for subRule in decomposedRules:
-            if self._simpleBooleanMatchesSubRule(subRule, preparedQuery):
+            if self._simpleBooleanMatchesSubRule(subRule, queryString, substring):
                 # We can immediately return True on the first match because this loop is iterating over an OR expression
                 # so we need just one match to pass.
                 return True
--- a/auslib/db.py
+++ b/auslib/db.py
@@ -1539,18 +1539,28 @@ class Rules(AUSTable):
             return True
         return string_compare(queryBuildID, ruleBuildID)
 
-    def _csvMatchesRule(self, ruleString, queryString, substring=True):
-        """Decides whether a column from a rule matches an incoming one.
-           Some columns in a rule may specify multiple values delimited by a
-           comma. Once split we do a full or substring match against the query
-           string. Because we support substring matches, there's no need
-           to support globbing as well."""
+    def _simpleBooleanMatchesSubRule(self, subRuleString, queryStringList):
+        for rule in subRuleString:
+            if rule not in queryStringList:
+                return False
+        return True
+
+    def _simpleBooleanMatchesRule(self, ruleString, queryString):
+        """Decides whether a column from a rule matches an incoming one using simplified boolean logic.
+        Only two operators are supported: '&&' (and), ',' (or). A rule like 'AMD,SSE' will match incoming
+        rules that contain either 'AMD' or 'SSE'. A rule like 'AMD&&SSE' will only match incoming rules
+        that contain both 'AMD' and 'SSE'.
+        Rules are matched to full strings, so if we have a rule that specifies 'AMD,SSE', it will NOT be
+        matched by an incoming rule that contains 'SSE3'.
+        """
         if ruleString is None:
             return True
-        for part in ruleString.split(','):
-            if substring and part in queryString:
-                return True
-            elif part == queryString:
+
+        decomposedRules = [[rule.strip() for rule in subRule.split('&&')] for subRule in ruleString.split(',')]
+        decomposedQueryString = queryString.split(',')
+
+        for subRule in decomposedRules:
+            if self._simpleBooleanMatchesSubRule(subRule, decomposedQueryString):
                 return True
         return False
 
@@ -1645,11 +1655,11 @@ class Rules(AUSTable):
             # To help keep the rules table compact, multiple OS versions may be
             # specified in a single rule. They are comma delimited, so we need to
             # break them out and create clauses for each one.
-            if not self._csvMatchesRule(rule['osVersion'], updateQuery['osVersion']):
+            if not self._simpleBooleanMatchesRule(rule['osVersion'], updateQuery['osVersion']):
                 self.log.debug("%s doesn't match %s", rule['osVersion'], updateQuery['osVersion'])
                 continue
             # Same deal for system capabilities
-            if not self._csvMatchesRule(rule['systemCapabilities'], updateQuery.get('systemCapabilities', ""), substring=False):
+            if not self._simpleBooleanMatchesRule(rule['systemCapabilities'], updateQuery.get('systemCapabilities', "")):
                 self.log.debug("%s doesn't match %s", rule['systemCapabilities'], updateQuery.get('systemCapabilities'))
                 continue
             # Locales may be a comma delimited rule too, exact matches only
--- a/auslib/db.py
+++ b/auslib/db.py
@@ -1539,28 +1539,38 @@ class Rules(AUSTable):
             return True
         return string_compare(queryBuildID, ruleBuildID)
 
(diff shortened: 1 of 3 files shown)


auslib/test/test_db.py
```diff
@@ -1892,6 +1892,16 @@ def setUp(self):
...
         self.paths.t.insert().execute(rule_id=10, priority=100, buildTarget="g", mapping="g", fallbackMapping='fallback', backgroundRate=100,
                                       update_type="z", product="foo", channel="foo", data_version=1)
+
+        self.paths.t.insert().execute(rule_id=11, priority=100, buildTarget='h', mapping='h', systemCapabilities='GenuineIntel,SSE2', update_type='z',
+                                      product='a', channel='a', data_version=1)
+        self.paths.t.insert().execute(rule_id=12, priority=100, buildTarget='i', mapping='i', systemCapabilities='GenuineIntel&&SSE2', update_type='z',
+                                      product='a', channel='a', data_version=1, version='5.0')
+        self.paths.t.insert().execute(rule_id=13, priority=100, buildTarget='j', mapping='j', systemCapabilities='!SSE3', update_type='z', product='a',
+                                      channel='a', data_version=1)
+        self.paths.t.insert().execute(rule_id=14, priority=100, buildTarget='k', mapping='k', systemCapabilities='!GenuineIntel&&SSE3,SSE2', update_type='z',
```
mozbhearsum: Could we support whitespace (even if it's just a single space) around the operators? It would greatly improve readability IMO.

mozbhearsum:
> I went with rail's suggestion to the keep the comma-separated format and introduce "&&" and "!" to the syntax - which makes the whole thing a lot simpler and naturally imposes some constraints so that rules won't get out of hand. That said, I wanted to clarify two things:

Hmm, I don't remember talking about "!", can your refresh my memory? It looks like you're adding support for negative matches too, which I think is out of scope for this bug.

> Looking at the bugzilla entry, it looks to me that we only want to introduce this simplified boolean matching to the osVersion and systemCapabilities fields - is this correct?

Yeah, we'll only apply this to these two fields, at least for now. None of the other fields contain more than one piece of information at the same time, so it's not useful for them.

> Has this functionality been implemented in the clients yet? Like, is Balrog getting any GET requests that feature extended systemCapabilities information? I'm mainly trying to pinpoint how the URL will look then - I'm currently assuming that it'll be something like .../GenuineIntel,family6,model61,SSE3/....

That level of CPU information hasn't been implemented in the update client yet, but I suspect it will look something like what you've got above. It's possible it might be more like: "(GenuineIntel)(family6)(model61)(SSE3)" - we've been using paranthesis lately to make partial string matching less dangerous.

auslib/db.py
```diff
@@ -1649,7 +1670,7 @@ def getRawMatches():
                 self.log.debug("%s doesn't match %s", rule['osVersion'], updateQuery['osVersion'])
                 continue
             # Same deal for system capabilities
-            if not self._csvMatchesRule(rule['systemCapabilities'], updateQuery.get('systemCapabilities', ""), substring=False):
```
mozbhearsum: If this ends up making the substring mode of _csvMatchesRule unused, please remove support for it. Dead code--!

auslib/db.py
```diff
@@ -1539,6 +1539,27 @@ def _buildIDMatchesRule(self, ruleBuildID, queryBuildID):
...
+    def _simpleBooleanMatchesSubRule(self, subRuleString, queryString):
+        for rule in subRuleString:
+            if rule.startswith('!'):
+                if rule[1:] in queryString:
+                    return False
+            elif rule not in queryString:
+                return False
+        return True
+
+    def _simpleBooleanMatchesRule(self, ruleString, queryString):
```
mozbhearsum: This function could use a docstring.

auslib/db.py
```diff
@@ -1539,6 +1539,27 @@ def _buildIDMatchesRule(self, ruleBuildID, queryBuildID):
             return True
         return string_compare(queryBuildID, ruleBuildID)
 
+    def _processBooleanRuleString(self, rule_str):
+        return [rule.split('&&') for rule in rule_str.split(',')]
```
mozbhearsum: I think this whole section would read better if you put this in _simpleBooleanMatchesRule instead of its own function. If it was multiple lines of code it might be worth doing, but currently this just seems to add indirection with no benefit.

auslib/db.py
```diff
@@ -1649,7 +1670,7 @@ def getRawMatches():
                 self.log.debug("%s doesn't match %s", rule['osVersion'], updateQuery['osVersion'])
```
mozbhearsum: Don't forget to apply this to OS_VERSION too.

auslib/test/test_db.py
```diff
@@ -2103,6 +2118,128 @@ def testGetRulesMatchingQuerySystemCapabilitiesNoSubstringMatch(self):
...
+            dict(product='a', channel='a', buildTarget='h', buildID='', locale='',
+                 osVersion='', distribution='', distVersion='', headerArchitecture='',
+                 force=False, queryVersion=6, systemCapabilities='SSE2', version='5.0'),
+            fallbackChannel=''
+        )
+        no_rules = self.paths.getRulesMatchingQuery(
+            dict(product='a', channel='a', buildTarget='h', buildID='', locale='',
+                 osVersion='', distribution='', distVersion='', headerArchitecture='',
+                 force=False, queryVersion=6, systemCapabilities='SSE3', version='5.0'),
+            fallbackChannel=''
```
mozbhearsum: Can you test the "SSE" case here too? We need to make sure we can contain to match "SSE" without matching "SSE2" or "SSE3".

auslib/db.py
sirMackk: Done, removed!

auslib/test/test_db.py
sirMackk: Awesome catch - I added these cases and changed the code to do exact matches instead of substring matches so that a rule containing "SSE" won't match an incoming rule like "SSE3".

auslib/db.py
sirMackk: Thanks - I did that and updated 3 of the osVersion tests. I was a bit on the fence about this, since this behavior is already tested using systemCapabilities, _but_ I thought I'd it to specifically test the behavior of osVersion matching.

sirMackk:
Thanks for the awesome feedback! I added the changes that you requested: I got rid of the NOT operator, switched osVersion matching from csv to boolean, and made it work with the client update strings you pointed out. Getting it to work with the updated "parenthesized" update string will be easy once that comes around. Last but not least, I updated the docs and docstrings (super big thanks for pointing this out, it's easy to omit this, but this is really important).

auslib/db.py
```diff
@@ -1539,18 +1539,28 @@ def _buildIDMatchesRule(self, ruleBuildID, queryBuildID):
...
+            if rule not in queryStringList:
+                return False
+        return True
+
+    def _simpleBooleanMatchesRule(self, ruleString, queryString):
+        """Decides whether a column from a rule matches an incoming one using simplified boolean logic.
+        Only two operators are supported: '&&' (and), ',' (or). A rule like 'AMD,SSE' will match incoming
+        rules that contain either 'AMD' or 'SSE'. A rule like 'AMD&&SSE' will only match incoming rules
+        that contain both 'AMD' and 'SSE'.
+        Rules are matched to full strings, so if we have a rule that specifies 'AMD,SSE', it will NOT be
```
sirMackk: Ah, thanks for clarifying that! I thought we were moving to exact string matches across the board, but I was wrong. I added back the ability to do substring matches - OS_VERSION does this, whereas SYSTEM_REQUIREMENTS does full string matching.

auslib/db.py
```diff
@@ -1539,18 +1539,28 @@ def _buildIDMatchesRule(self, ruleBuildID, queryBuildID):
...
+        """
         if ruleString is None:
             return True
-        for part in ruleString.split(','):
-            if substring and part in queryString:
-                return True
-            elif part == queryString:
+
+        decomposedRules = [[rule.strip() for rule in subRule.split('&&')] for subRule in ruleString.split(',')]
+        decomposedQueryString = queryString.split(',')
```
sirMackk: Currently, splitting the queryString allows us to toggle between substring and full string matching using the 'in' operator. I changed this part to accommodate both types of matching. When we want to do substring matching, we leave the queryString as a string so that the 'in' operator matches any substring in the queryString. But if we want exact matching, we split the string into a list, so that the 'in' operator only returns True if the exact string appears in the resulting list. 

You're right about this being future-proof if we decide to match against multiple parts of SYSTEM_REQUIREMENTS, even if we switch to using parentheses, but let me know if this is premature and I'll figure out something simpler :).

auslib/test/test_db.py
```diff
@@ -2024,7 +2041,7 @@ def testGetRulesMatchingQueryReturnBackgroundThrottled(self):
     def testGetRulesMatchingQueryOsVersionSubstring(self):
         rules = self.paths.getRulesMatchingQuery(
             dict(product="a", version='5.0', channel="a", buildTarget='d',
-                 buildID='', locale='', osVersion='foo 1.2.3', distribution='',
+                 buildID='', locale='', osVersion='foo 1', distribution='',
```
sirMackk: I reverted this and two other tests that made use of substring matching, saved me a lot of time in testing whether substring matching works correctly.

sirMackk:
Thanks for the comments - I made the requested changes and added one question (the one related to decomposing the queryString). Sorry for the long radio silence, but I was moving and now I'm permanently back on the east coast.

auslib/db.py
mozbhearsum: Ah, I see. I think it's actually best to avoid doing this for two reasons:
1) There's no reason to support lists of values in the query yet (and maintaining code that serves no purpose is wasteful)
2) It's a bit "magical" to have _simpleBooleanMatchesSubRule operate different depending on the type of the "preparedQuery". It would be better to use an explicit flag for this, when we get there.

sirMackk:
Yeah, you're right - Ill make the changes and make sure they're ready for
the Friday train :)

On Apr 3, 2017 12:34 PM, "Ben Hearsum" <notifications@github.com> wrote:

*@mozbhearsum* commented on this pull request.
------------------------------

In auslib/db.py
<https://github.com/mozilla/balrog/pull/280#discussion_r109463914>:

>          if ruleString is None:
             return True
-        for part in ruleString.split(','):
-            if substring and part in queryString:
-                return True
-            elif part == queryString:
+
+        decomposedRules = [[rule.strip() for rule in
subRule.split('&&')] for subRule in ruleString.split(',')]
+        decomposedQueryString = queryString.split(',')

Ah, I see. I think it's actually best to avoid doing this for two reasons:

   1. There's no reason to support lists of values in the query yet (and
   maintaining code that serves no purpose is wasteful)
   2. It's a bit "magical" to have _simpleBooleanMatchesSubRule operate
   different depending on the type of the "preparedQuery". It would be better
   to use an explicit flag for this, when we get there.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/mozilla/balrog/pull/280#discussion_r109463914>, or mute
the thread

mozbhearsum:
Sounds good. If you happen to know it, could you also post the commit hash of the previous commit? Follow-up reviews are much easier when I can do an interdiff.

sirMackk:
Sure thing, the prev commit (from 03.17) is `c27a21efed554eaa14006cfd9ad949f91a273ae2`. I'll hold off from rebasing.

mozbhearsum:
We definitely need multiple things in the query to do AND'ing, but I don't think it's a safe assumption that they will be comma separated. For example, here is an OS_VERSION that I pulled from the logs: "Windows_NT%2010.0.0.0%20(x64)(noBug1296630v1)(nowebsense)". We may end up doing something like (SSE)(AMD) for SYSTEM_REQUIREMENTS instead of comma separating.

I think I see what you're getting at now though -- for full string matches (like we do in SYSTEM_REQUIREMENTS) we must have some sort of separator to split the string apart, and then do full string matches for each section -- do I have that right?

sirMackk:
>  for full string matches (like we do in SYSTEM_REQUIREMENTS) we must have some sort of separator to split the string apart, and then do full string matches for each section

Yes, that's exactly it! Sorry that I didn't put it more clearly earlier on.

docs/source/database.rst
```diff
@@ -88,17 +88,17 @@ Following tables show columns according to different Categories:
...
-  |                        |                    | update. This field is primarily used to point    | comma separated list of         | "Darwin 6,Darwin 7         |
-  |                        |                    | desupported operating systems to their last      | partial strings to match on     | ,Darwin 8"                 |
-  |                        |                    | supported build.                                 |                                 |                            |
+  |                        | osVersion          | The OS Version of the application requesting an  | Simplified boolean string       | "Windows_NT 5.0" or        |
+  |                        |                    | update. This field is primarily used to point    | match. '&&' ANDs terms while    | "Darwin 6, Darwin 7        |
+  |                        |                    | desupported operating systems to their last      | ',' ORs them. Terms are matched | , Darwin 8"                |
+  |                        |                    | supported build.                                 | using exact strings.            |                            |
```
mozbhearsum: Looks like this needs an update, since we're doing substring matching here.

mozbhearsum:
Thank you so much for doing this @sirMackk! We've been eager to have this feature for quite awhile.

sirMackk:
The pleasure is all mine, @mozbhearsum ! Please let me know if anything related to this needs any tweaking or ironing out :).

PR #282, opened 2017-03-22
create docker images with version tags

relud:
this will create docker images like mozilla/balrog:2.24 which jenkins can identify and automatically deploy to stage


relud:
my logic in this was flawed, closing in favor of a different solution

PR #285, opened 2017-03-24
Remove support for revision in SystemAddon superblobs.

mozbhearsum:
We're pushing forward on using multifile updates for System Addons in Balrog. While setting up some sample rules for this, I realized that the SuperBlob requires "revision" to be set - because we originally designed this feature when that was going to be a thing. I'm pretty sure we've decided not to use revision numbers at this point, so this should be safe to kill. @rhelmer - is that right?

### DIFF
--- a/auslib/blobs/schemas/superblob.yml
+++ b/auslib/blobs/schemas/superblob.yml
@@ -57,9 +57,6 @@
           description: Blob identifier number
           enum:
             - 4000
-        revision:
-          type: number
-          description: Revision number of adds collection
         blobs:
           type: array
           items:
--- a/auslib/blobs/schemas/superblob.yml
+++ b/auslib/blobs/schemas/superblob.yml
@@ -39,13 +39,17 @@
       required:
         - name
         - schema_version
-        - revision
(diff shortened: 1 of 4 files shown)


rhelmer:
Revision numbers would still be useful to do things like throttled roll-out, but I think we've come up with better ways to do this sort of thing (eg push the add-on to everybody and have them locally a/b test, use Shield Studies/Telemetry Experiments first, etc) so nbd to me if we decide not to do it.

auslib/blobs/schemas/superblob.yml
```diff
@@ -48,6 +47,11 @@
...
           minLength: 1
           maxLength: 100
+        product:
+          type: string
+          description: The product that the blob is associated to
+          minLength: 1
+          maxLength: 100
         schema_version:
           type: number
           description: Blob identifier number
```
JohanLorenzo: Should we keep the revision defined  3 lines below?

mozbhearsum: Whoops, of course not!

PR #287, opened 2017-03-26
Bug 1342234: Fix errors that were not shown while adding permissions and roles

nitinprakash96:


aksareen:
Hi @nitinprakash96 , looks like your changes are causing the front-end tests to fail. Here's the log message : not ok 50 PhantomJS 2.1 - controller: UserPermissionsCtrl opening the edit user modal should should all defaults.
    ---
        message: >
            Expected { permissions : {  }, roles : {  } } to equal { permissions : {  } }.
        Log: |
    ...
Nit: Always run the following command in cli to check all tests are passing before posting a PR : $ ./run-tests.sh [frontend/backend]

nitinprakash96:
fixed.!

ui/app/js/controllers/user_edit_controller.js
```diff
@@ -100,11 +100,14 @@ function ($scope, $modalInstance, CSRF, Permissions, user, users) {
...
         };
+        $scope.errors = {
+          roles: {}
+        };
         sweetAlert("Saved", "Role granted.", "success");
       })
       .error(function(response) {
         if (typeof response === 'object') {
-          $scope.errors = response;
+          $scope.errors.role = response;
```
mozbhearsum: I see you've added a "roles" object to $scope.errors, above, but you're reference "role" here, that probably isn't helping.

ui/app/js/controllers/user_edit_controller.js
```diff
@@ -100,11 +100,14 @@ function ($scope, $modalInstance, CSRF, Permissions, user, users) {
           role: '',
           data_version: ''
         };
+        $scope.errors = {
+          roles: {}
```
mozbhearsum: I think this needs to be added on line 15, too.

nitinprakash96: It does. But I don't understand why it causes the frontend tests to fail.

ui/app/js/controllers/user_edit_controller.js
nitinprakash96: Oh. This might've been a typo. I'll add 'role' to $scope.errors.

ui/app/js/controllers/user_edit_controller.js
mozbhearsum: Have you had a look at the failing tests? Sometimes they need updates when you change related code: https://github.com/mozilla/balrog/blob/master/ui/spec/controllers/user_edit_controller_spec.js

nitinprakash96: Updating the spec file did the job. Thanks for suggestion @mozbhearsum

nitinprakash96:
I don't understand. It totally works fine on my system.
ui

mozbhearsum:
Ah, I see the confusion. The problem is that that dialog says "see fields highlighted in red", but there are none, so the user doesn't really have any indication of what they did wrong. So, either we need the error to show up in the Roles dialog somewhere, or the error needs to be shown in the "Form submission error" pop-up.

nitinprakash96:
sorry for the delay @mozbhearsum . I've added error messages in the roles dialog box and in permission.

ui/app/js/controllers/user_edit_controller.js
```diff
@@ -148,7 +152,7 @@ function ($scope, $modalInstance, CSRF, Permissions, user, users) {
       })
       .error(function(response) {
         if (typeof response === 'object') {
-          $scope.errors = response;
+          $scope.errors.permission = response;
```
mozbhearsum: After rereading some other code, I'm not sure this is the right thing to do. It ends up with keys like: errors.permission.permission and errors.permission.options, but the templates end up looking for things like errors.options.

nitinprakash96: There can be cases I suppose. If we enter an incorrect "permission" and "options" then the error message displayed will be "No JSON object decoded".
But if we re-enter the same permission for a given username it will only be displaying the sweetalert probably rather than highlighting the field.

mozbhearsum: It sounds like there might be another bug if there's a case where "No JSON object decoded" is shown...

nitinprakash96: I think that the error shown is reasonable as the "options" field will be incorrect. And to be able to fix that we can consider errors.permission in my opinion. Since this way we can take care of both incorrect "options" field and an unknown permission. It will also fix the case where a user re-enters the same permission for a given username.

mozbhearsum:
Closing due to inactivity. Feel free to re-open if you'd like to finish this up.

PR #288, opened 2017-03-28
Add support for creating rule scheduled changes

mozbhearsum:
Pretty simple, mostly a copy of the existing Rule class. I also discovered that the current backend method doesn't support non-JSON arguments when peeking to find change_type, so I fixed that as well.

@srfraser - are you comfortable reviewing this?


mozbhearsum:
@rail - can you stamp this too? It looks Simon can't do the official approval for some reason.

PR #290, opened 2017-03-29
[ Bug #1350970] Disable non-error output from docker containers in local development

andela-aolasode:
## What does this PR do?

Change and add log level for docker containers

1. Add log level for container balrog ui and balrog db
2. Change log level for container balrog admin, balrog pub, balrog agent from DEBUG to WARNING

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -71,7 +71,6 @@ balrogui:
     - .cache:/cache
   environment:
     - WEB_PORT=8080
-    - LOG_LEVEL=WARNING
   entrypoint:
     - /bin/bash
     - /app/docker-entrypoint.sh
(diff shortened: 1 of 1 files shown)


njirap:
Nitpick: To me your branch name feels unnecessarily verbose.. I believe the bug number would be sufficient instead of an entire sentence. E.g something like `bg_1350970` in this case.

docker-compose.yml
```diff
@@ -71,6 +71,7 @@ balrogui:
     - .cache:/cache
   environment:
     - WEB_PORT=8080
+    - LOG_LEVEL=WARNING
```
mozbhearsum: I don't think this will do anything for the ui and db containers. Remove this and then it's good to go!

andela-aolasode: I have implemented the changes. The log level for ui and db containers have been removed.

mozbhearsum:
Looks good, thank you very much!

PR #291, opened 2017-03-29
[Bug #1304026] Make the order of enacting changes predictable

andela-akoya:
## What does this PR do?
Make the order in which scheduled changes are enacted predictable.

1.  Add function to sort the scheduled changes based on priority and time.

### DIFF
--- a/agent/balrogagent/test/test_cmd.py
+++ b/agent/balrogagent/test/test_cmd.py
@@ -220,3 +220,77 @@ class TestRunAgent(asynctest.TestCase):
         self.assertEquals(telemetry_is_ready.call_count, 0)
         self.assertEquals(time_is_ready.call_count, 1)
         self.assertEquals(request.call_count, 5)
+
+    @asynctest.patch("time.time")
+    async def testRightEnactOrderForMultipleChangesOneEndpoint(self, time, time_is_ready, telemetry_is_ready, request):
+        time.return_value = 999999999
+        time_is_ready.return_value = True
+        sc = {'releases': [{"priority": 1, "sc_id": 4, "when": 234, "telemetry_uptake": None,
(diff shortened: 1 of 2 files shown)


njirap:
Nitpick: To me your branch name feels unnecessarily verbose.. I believe the bug number would be sufficient instead of an entire sentence. E.g something like `bg_1304026` in this case.

agent/balrogagent/test/test_cmd.py
```diff
@@ -171,11 +171,11 @@ def side_effect(balrog_api_root, endpoint, auth, loop, method='GET'):
...
         time_is_ready.return_value = True
-        sc = {'releases': [{"sc_id": 4, "when": 234, "telemetry_uptake": None,
+        sc = {'releases': [{"priority": None, "sc_id": 4, "when": 234, "telemetry_uptake": None,
                             "telemetry_product": None, "telemetry_channel": None},
-                           {"sc_id": 5, "when": 234, "telemetry_uptake": None,
+                           {"priority": None, "sc_id": 5, "when": 234, "telemetry_uptake": None,
                             "telemetry_product": None, "telemetry_channel": None},
-                           {"sc_id": 6, "when": 234, "telemetry_uptake": None,
+                           {"priority": None, "sc_id": 6, "when": 234, "telemetry_uptake": None,
                             "telemetry_product": None, "telemetry_channel": None}]}
```
mozbhearsum: We need at least one test that has multiple different "priority" and "when" set, and we need to make sure that the "enact" endpoint is called in the right order. This will be similar to this test, but the order of the entries of "called_endpoints" will be important.

andela-akoya: The test has been implemented

PR #292, opened 2017-03-29
[Bug #1325442] Remove update/delete button on completed scheduled changes for the UI

andela-odurodola:
## What does this PR do?
It removes the update and delete buttons on completed scheduled changes (for Rules, Permissions and Releases).

Set the condition for the `ng-show` attribute of the update/delete button to be False on completed scheduled changes.

## Screenshot

# Before the update/delete button was removed
screen shot 2017-03-29 at 3 32 35 pm

# After the update/delete button was removed
screen shot 2017-03-29 at 3 31 40 pm

### DIFF
--- a/ui/app/templates/permission_scheduled_changes.html
+++ b/ui/app/templates/permission_scheduled_changes.html
@@ -50,8 +50,8 @@
                 ng-show="! isEmpty(sc['required_signoffs']) && sc['signoffs'].hasOwnProperty(current_user)">
             Revoke your Signoff
         </button>
-        <button class="btn btn-default btn-xs" ng-click="openScheduledUpdateModal(sc)">Update</button>
-        <button class="btn btn-default btn-xs" ng-click="openDeleteModal(sc)">Delete</button>
+        <button ng-show="!sc.complete" class="btn btn-default btn-xs" ng-click="openScheduledUpdateModal(sc)">Update</button>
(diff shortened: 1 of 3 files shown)


njirap:
LGTM.

Nitpick: To me your branch name feels unnecessarily verbose.. I believe the bug number would be sufficient instead of an entire sentence. E.g something like `bg_1325442` in this case.

cc @andela-iikikin

ui/app/templates/permission_scheduled_changes.html
```diff
@@ -50,8 +50,8 @@ <h3 class="panel-title">
                 ng-show="! isEmpty(sc['required_signoffs']) && sc['signoffs'].hasOwnProperty(current_user)">
             Revoke your Signoff
         </button>
-        <button class="btn btn-default btn-xs" ng-click="openScheduledUpdateModal(sc)">Update</button>
-        <button class="btn btn-default btn-xs" ng-click="openDeleteModal(sc)">Delete</button>
+        <button ng-show="!scheduled_changes_rules_count && !sc.complete" class="btn btn-default btn-xs" ng-click="openScheduledUpdateModal(sc)">Update</button>
+        <button ng-show="!scheduled_changes_rules_count && !sc.complete" class="btn btn-default btn-xs" ng-click="openDeleteModal(sc)">Delete</button>
```
njirap: why are we checking for the `scheduled_changes_rules_count` in the permissions?

andela-odurodola: It was on oversight. It will be removed

PR #293, opened 2017-04-02
Bug 1342191

tyagi-iiitv:
balrog agent fails to process subsequent changes if an exception is hit for an earlier one. r=bhearsum

### DIFF
--- a/agent/balrogagent/cmd.py
+++ b/agent/balrogagent/cmd.py
@@ -62,11 +62,12 @@ async def run_agent(loop, balrog_api_root, balrog_username, balrog_password, tel
                         resp.close()
                     else:
                         logging.debug("Change %s is not ready", change["sc_id"])
+                        raise
 
         except:
             logging.error("Encountered exception:", exc_info=True)
-            if raise_exceptions:
-                raise
+            pass
+
         finally:
             if not once:
                 await asyncio.sleep(sleeptime)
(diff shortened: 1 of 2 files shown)


agent/balrogagent/cmd.py
```diff
@@ -62,11 +62,12 @@
                         resp.close()
                     else:
                         logging.debug("Change %s is not ready", change["sc_id"])
+                        raise
```
tyagi-iiitv: I tried to raise a virtual exception for the test case here

mozbhearsum:
It looks like you're still having a bit of trouble getting the tests set-up properly - is there anything I can do to help?

tyagi-iiitv:
Ohh not really .. I am on a vacation and will be back next week. Is this urgent ?

mozbhearsum:
Not at all, take your time!

tyagi-iiitv:
Hi Ben, I actually tried testing this one, and based on the test, I've a few suggestions which I listed in the review comments. Please have a look at them and see if replacing **raise** with **pass** will work in this case. Because if we use **raise**, we can't continue with the loop execution after an exception is raised.

mozbhearsum:
I had a reread of everything here and I'm not quite sure where you're getting stuck. Can you expand?

A couple of overall comments though:
- "raise_exceptions" should probably go away - raising them is incompatible with fixing this bug
- I think you're right that "pass" needs to be used in the "except" block, but make sure you still log the exceptions beforehand.
- The tests are going to the tricky part - you need a test that makes two requests, the first of which needs to cause an exception to be raised, the second of which should not.

mozbhearsum:
Closing due to inactivity. Feel free to re-open if you'd like to finish this up.

PR #295, opened 2017-04-05
Bug 1261061: Stop using the "dbo" wrapper in db.py

andela-oadeniran:
## What does this PR DO?
 It eliminates the use of dbo wrapper in the db.py file.
## Description of tasks to be Completed
Use db passed to each class, instead of dbo which is an instance of the DBWrapper.

### DIFF
--- a/auslib/db.py
+++ b/auslib/db.py
@@ -2573,7 +2573,7 @@ class AUSDatabase(object):
 
     def upgrade(self, version=None):
         # This method was taken from Buildbot:
-        # https://github.com/buildbot/buildbot/blob/87108ec4088dc7fd5394ac3c1d0bd3b465300d92/master/builselft/db/model.py#L455
+        # https://github.com/buildbot/buildbot/blob/87108ec4088dc7fd5394ac3c1d0bd3b465300d92/master/buildbot/db/model.py#L455
         # http://code.google.com/p/sqlalchemy-migrate/issues/detail?id=100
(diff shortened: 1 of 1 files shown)


auslib/db.py
```diff
@@ -2572,7 +2573,7 @@ def create(self, version=None):
 
     def upgrade(self, version=None):
         # This method was taken from Buildbot:
-        # https://github.com/buildbot/buildbot/blob/87108ec4088dc7fd5394ac3c1d0bd3b465300d92/master/buildbot/db/model.py#L455
+        # https://github.com/buildbot/buildbot/blob/87108ec4088dc7fd5394ac3c1d0bd3b465300d92/master/builselft/db/model.py#L455
```
NinadBhat: Any reason you changed this url?

andela-oadeniran: It was a typo I would correct that immediately

auslib/db.py
```diff
@@ -20,7 +20,7 @@
 import dictdiffer
 import dictdiffer.merge
 
-from auslib.global_state import cache, dbo
```
njirap: did you guys delete `dbo` from `auslib.global_state` or is it being used elsewhere apart from here?

andela-oadeniran: We thought of that but it is used in other files.

mozbhearsum: Yeah, we can't get rid of it entirely - it's required by the web layer. Good thought though!

PR #296, opened 2017-04-05
[bg_1342231] Disable role granting for users without permissions

andela-bolajide:
## What does this PR do?

It disables role granting for users without permissions. Users without permission shouldn't be able to have roles.

First, initialized a boolean variable `showRoles` to be used as a condition to display the Roles tab in the permissions modal. After this, we then used angular directives (ng-show and ng-hide) to display the roles tab based on the boolean we set in the controller.

#### Before disabling role granting for users without permissions
When a user is added or when updating a user with no permissions, we can see that the roles tab is active.
old

#### After disabling role granting for users without permissions
update1

update2

### DIFF
--- a/ui/app/js/controllers/user_edit_controller.js
+++ b/ui/app/js/controllers/user_edit_controller.js
@@ -8,7 +8,6 @@ function ($scope, $modalInstance, CSRF, Permissions, user, users) {
   $scope.is_edit = true;
   $scope.original_user = user;
   $scope.user = angular.copy(user);
-  $scope.showRole = true;
   $scope.permission = {
     permission: '',
     options_as_json: ''
@@ -146,7 +145,6 @@ function ($scope, $modalInstance, CSRF, Permissions, user, users) {
           permissions: {}
         };
         sweetAlert("Saved", "Permission added.", "success");
-        $scope.showRole = !$scope.showRole;
       })
       .error(function(response) {
         if (typeof response === 'object') {
--- a/ui/app/js/controllers/user_edit_controller.js
+++ b/ui/app/js/controllers/user_edit_controller.js
@@ -8,6 +8,7 @@ function ($scope, $modalInstance, CSRF, Permissions, user, users) {
   $scope.is_edit = true;
   $scope.original_user = user;
   $scope.user = angular.copy(user);
+  $scope.showRole = true;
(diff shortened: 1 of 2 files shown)


ui/app/templates/permissions_modal.html
```diff
@@ -10,7 +10,7 @@ <h3 class="modal-title">{{ user.username }}</h3>
           <a href="#" ng-click="currentItemTab = 1;">Permissions</a>
         </li>
         <li ng-class="{active: currentItemTab == 2}">
-          <a href="#" ng-click="currentItemTab = 2;">Roles</a>
+          <a href="#" ng-click="currentItemTab = 2;" ng-hide="showRole" ng-show='user.permissions.length > 0'>Roles</a>
```
njirap: how about something like `ng-show='user.permissions.length > 0' && !showRole`

njirap: Or `ng-hide='!user.permissions.length > 0 && showRole'`

njirap: Maybe the entire `li` item should be conditioned and not the inner link.. i.e put these conditions at Line 12 above..

ui/app/js/controllers/user_edit_controller.js
```diff
@@ -145,6 +146,7 @@ function ($scope, $modalInstance, CSRF, Permissions, user, users) {
           permissions: {}
         };
         sweetAlert("Saved", "Permission added.", "success");
+        $scope.showRole = !$scope.showRole;
```
njirap: What is happening here? What I see is that when a user gets the first permission then `showRole`becomes `false`, when they get the second then `showRole` becomes `true` etc so `showRole` keeps toggling as permission is added..

njirap:
What do you think about adding code when a permission is being revoked for `showRole` to be `false` when the user no longer has any more permissions?

andela-bolajide:
When a user no longer has any permission, it is automatically deleted. Moreover with the new update we have, there wouldn't be a need for the variable, showRole anymore.
We just only show the tab when the current user has atleast 1 permission.

ui/app/templates/permissions_modal.html
andela-bolajide: Done.

PR #297, opened 2017-04-05
Bug_1344716: Add health checks to Docker-compose file

andela-odurodola:
## What does this PR do?
It uses Docker "healthcheck" feature  to delay containers startup until the healthcheck command is successful.

- Upgrade docker-compose.yml from version 1 to version 2.1

- Add a Dockerfile to build mysql in order to install netcat.

- Write an healthcheck to determine the health status of balrogdb and balrogadmin by scanning their ports.

- Add the "depends_on" feature to implement containers startup's dependency.


Dockerfile.datab
```diff
@@ -0,0 +1,7 @@
+FROM mysql:5.7
```
mozbhearsum: Tiny nit: "Dockerfile.database" or "Dockerfile.db" would be a better name for this file.

docker-compose.yml
```diff
@@ -1,88 +1,120 @@
...
+      - SMTP_USERNAME
+      - SMTP_PASSWORD
+      - SMTP_TLS
+      - NOTIFY_TO_ADDR
+      - NOTIFY_FROM_ADDR
+    healthcheck:
+        test: nc -z -v balrogadmin 7070
+        interval: 5s
+        timeout: 30s
+        retries: 3
```
mozbhearsum: This seems to fail sometimes, at least on my machine (and I suspect it would fail more often on slower machines). It's probably best to use a high number of retries (at least 10). Same thing goes for the db healthcheck.

PR #304, opened 2017-04-19
Bug-1355474. Add a "view data" button for releases scheduled changes

dilekuzulmez:


mozbhearsum:
Let me try to clarify the update and delete cases a bit. The overarching goal here is to show useful information about what will happen when a scheduled change is enacted.

Releases can be very, very big (100kb even), which means that when we talk and think about mkaing changes to them, it's easiest to do so by comparing the old and new version. If I'm a user looking at a scheduled update to a Release, the most important thing I want to know is "what does this update change"? A diff is the most efficient way to read this. To get this information you will need to change the backend to add support for diffing a current release vs. a scheduled change, and then add support to the frontend for retrieving it. We do something similar for generating diffs of Releases History. It might be best to have this done in a separate pull request...it looks like more work than I thought it would be.

When looking at scheduled deletions of a Release, it doesn't really help to show the new version of the "data", beacuse it will always be empty. The most important thing I want to know in that scenario is "what is being deleted"? Showing the current version of the data answers this question. The backend can already tell you this by calling Releases.getRelease() with the name, so this should be easy.

PR #305, opened 2017-04-20
[Bug_1281520] Rip Whitelist Support

andela-bolajide:
## What Does this PR do?
Whitelisting was added earlier in balrog to support a very specific B2G use case and since there is no use for it anymore, removal of the whitelist is essential so as to avoid redundancy.

- Remove the "whitelist" column in the rules, rules_history, rules_scheduled_changes, and rules_scheduled_changes_history tables.

- Create migration script to remove whitelist columns from live database.

- Remove references to whitelisting in the ui, tests and other classes.


auslib/test/test_db.py
```diff
@@ -2592,8 +2592,10 @@ def setUp(self):
                                       data_version=1)
         self.rules.t.insert().execute(rule_id=2, product="b", channel="h", mapping="c", fallbackMapping="h", backgroundRate=100, priority=100,
                                       update_type="minor", data_version=1)
-        self.rules.t.insert().execute(rule_id=3, product="b", channel="h", mapping="c", whitelist="h", backgroundRate=100, priority=100,
-                                      update_type="minor", data_version=1)
+        # because i've removed the whitelist support, rule_id 1 and 3 are now the same thing so i commented it out.
+
+        # self.rules.t.insert().execute(rule_id=3, product="b", channel="h", mapping="c", backgroundRate=100, priority=100,
+        #                               update_type="minor", data_version=1)
```
mozbhearsum: Please just rid of this now-unnecessary rule instead of commenting it out.

andela-bolajide: I've done that ben.

PR #310, opened 2017-05-03
Bg1276289 reenable coveralls

andela-bolajide:
## What Does this PR do?

Currently, coverage results are not sent to [coveralls](https://coveralls.io). This PR re-enables it by writing a script to manually send coverage data after successful test completion.

First, coveralls is to be added to the requirements.txt so as to install it when building the docker image for testing, then the repo token is gotten via the curl command from a secret url and saved to .coveralls.yml so coveralls can get it. After all this the `coveralls` command is then run.

## Bug Reference
https://bugzilla.mozilla.org/show_bug.cgi?id=1276289

### DIFF
--- a/scripts/run.sh
+++ b/scripts/run.sh
@@ -73,6 +73,13 @@ elif [ $1 == "test" ]; then
 
         if [[ $backend_rc == 0 && $frontend_rc == 0 ]]; then
             echo "All tests pass!!!"
+            if [[ $GITHUB_HEAD_REPO_URL == "https://github.com/mozilla/balrog.git" ]];
+            then
+              password_url="taskcluster/secrets/v1/secret/repo:github.com/mozilla/balrog:coveralls"
+              repo_token=$(curl ${password_url} | python -c 'import json, sys; a = json.load(sys.stdin); print a["secret"]["repo_token"]')
+              echo 'repo_token:' $repo_token >> .coveralls.yml
+              coveralls
+            fi
             exit 0
         else
             echo "FAIL FAIL FAIL FAIL FAIL FAIL FAIL FAIL. Some tests failed, see above for details."
--- a/scripts/run.sh
+++ b/scripts/run.sh
@@ -79,6 +79,7 @@ elif [ $1 == "test" ]; then
               repo_token=$(curl ${password_url} | python -c 'import json, sys; a = json.load(sys.stdin); print a["secret"]["repo_token"]')
               echo 'repo_token:' $repo_token >> .coveralls.yml
               coveralls
+              echo "Coverage successfully sent to coveralls.io"
             fi
             exit 0
         else


andela-bolajide:
Hey @bhearsum, i'm having problems with where to access the `repo_token`, i followed the process used here: https://github.com/mozilla/balrog/blob/master/scripts/push-dockerimage.sh#L10 but i'm not sure if the repo_token will be stored in taskcluster secret.

scripts/run.sh
```diff
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+password_url="taskcluster/secrets/v1/secret/repo:github.com/mozilla/balrog:dockerhub"
```
mozbhearsum: This will need to be :coveralls instead of :dockerhub. I'm working on getting the secret added still, but it will be there soon!

scripts/run.sh
```diff
@@ -10,6 +13,9 @@ build_front_end() {
 run_back_end_tests() {
   cd /app
   tox $@
+  echo 'repo_token:' $repo_token >> .coveralls.yml
```
mozbhearsum: Is anything else needed in this file?

scripts/run.sh
```diff
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+password_url="taskcluster/secrets/v1/secret/repo:github.com/mozilla/balrog:dockerhub"
+repo_token=$(curl ${password_url} | python -c 'import json, sys; a = json.load(sys.stdin); print a["secret"]["repo_token"]')
```
mozbhearsum: I think you're going to need to do something to make sure this (and the actual "coveralls" command) only gets run for Taskcluster. You probably should put it inside an "if" block in run_backend_tests, and make it conditional on something you set in https://github.com/mozilla/balrog/blob/master/.taskcluster.yml#L9.

scripts/run.sh
andela-bolajide: No, just the repo_token is all that is needed in the `.coveralls.yml` file. When the `coveralls` command is run, it looks for the file and gets the token.

scripts/run.sh
andela-bolajide: I'll correct this.

andela-bolajide: if coveralls is run on pull request, yes it will override the coverage for master. I think you can override that from coveralls.io , by going to the repo settings on coveralls and disabling Pull Request Comment.

screen shot 2017-05-03 at 6 38 12 pm

mozbhearsum:
OK, I added the repo token to the secrets service. Once you update the script it should be able to submit.

scripts/run.sh
andela-bolajide: Oh! I get. So if the test isn't run on taskcluster, it won't send coverage at all.

andela-bolajide: I'll do that soon.

mozbhearsum: Yeah, it's up to you to do this though. You can probably check for $GITHUB_HEAD_REPO_URL and only run coveralls if that is set.

PR #312, opened 2017-05-03
Bug 1336452 : Migrating /rules , /users and /csrf_token to connexion app

aksareen:
Relevant code for initial review and also to serve as background for this bug : https://pastebin.mozilla.org/9020651


auslib/web/admin/swagger/api.yaml
```diff
@@ -0,0 +1,441 @@
...
+            allOf:
+              - $ref: '#/definitions/CSRFModel'
+              - $ref: '#/definitions/RulesBase'
+
+            # will fail testMissingField test in test_rules.py if test is not modified.
+            # Because only outputs first missing field : backgroundRate
+            # ignores further check i.e. check for priority and returns error.
+            required:
+              - backgroundRate
+              - priority
```
mozbhearsum: Is there a reason these are here and not in RulesBase? I'm surprised that they are treated differently than update_type.

auslib/web/admin/swagger/api.yaml
```diff
@@ -0,0 +1,441 @@
...
+      - locale
+      - mapping
+      - osVersion
+      - priority
+      - product
+      - rule_id
+      - systemCapabilities
+      - update_type
+      - version
+      #- whitelist
```
mozbhearsum: Whitelist is dead, this can go :)

auslib/web/admin/swagger/api.yaml
```diff
@@ -0,0 +1,441 @@
...
+    description: Balrog's Rule Attributes
+    type: object
+
+    required:
+      - update_type # Required during both POSTing and GET(fetch) Operations
+
+    #I was initially hopeful that x-nullable would serve the purpose at class level to define
+    # fields that can expect null values, but my research so
+    # far seems to show that currently x-nullable works at the parameter level but not
+    # within an object/model. Instead declaring 'type' as array like [string, null] or [integer, null]
```
mozbhearsum: After reading this comment I'm a bit surprised to see null type + x-nullable in each parameter. Is there a reason for that?

auslib/web/admin/swagger/api.yaml
```diff
@@ -0,0 +1,441 @@
...
+        maxLength: 75
+        example: beta
+
+      buildTarget:
+        description: "The 'build target' of the application requesting an update. This is usually related to the target platform the app was built for."
+        type: ["string", "null"]
+        x-nullable: true
+        minLength: 0
+        maxLength: 75
+        example: Darwin_x86_64-gcc3- u-i386-x86_64
```
mozbhearsum: Nit: unwanted space here.

auslib/web/admin/swagger/api.yaml
```diff
@@ -0,0 +1,441 @@
...
+    # fields that can expect null values, but my research so
+    # far seems to show that currently x-nullable works at the parameter level but not
+    # within an object/model. Instead declaring 'type' as array like [string, null] or [integer, null]
+
+    # Cannot have additionalProperties as false for strict validation ,else it disallows csrf_token or any other model's properties
+    #additionalProperties: false
+
+    properties:
+      rule_id:
+        description: The id of the rule. This id is necessary to make changes to the rule through the REST API.
```
mozbhearsum: Does it make sense to link to docs here, or do we only need to do that in the actual endpoints?

auslib/web/admin/swagger/api.yaml
```diff
@@ -0,0 +1,441 @@
...
+        description: "The update_type to use in the XML response. It's very rare for a rule to use anything other than 'minor' these days."
+        type: string
+        x-nullable: false
+        minLength: 0
+        maxLength: 15
+        enum: [ "major", "minor"]
+        example: minor
+
+      product:
+        description: The name of the application requesting an update.
```
mozbhearsum: I think these, and the other strings you pulled from https://github.com/mozilla/balrog/blob/master/auslib/web/public/api.yml need to be reworded slightly. There's no "application requesting an update" in the context of the admin API. It may be worth mentioning some of the special matching in fields like version, locale, etc. too.

auslib/web/admin/swagger/api.yaml
```diff
@@ -0,0 +1,441 @@
...
+
+      comment:
+        description: A string describing the purpose of the rule. Not always necessary for obvious rules.
+        type: ["string", "null"]
+        x-nullable: true
+        minLength: 0
+        maxLength: 500
+        example: lorem ipso facto
+
+      whitelist:
```
mozbhearsum: This field is dead, this can go!

auslib/test/admin/views/test_rules.py
```diff
@@ -141,9 +141,22 @@ def testNewRuleWithoutPermission(self):
...
         # But we still need to pass product, because permission checking
         # is done before what we're testing
-        ret = self._post('/rules', data=dict({'product': 'a'}))
+        post_data = {'product': 'a'}
+        ret = self._post('/rules', data=post_data)
+        self.assertEquals(ret.status_code, 400, "Status Code: %d, Data: %s" % (ret.status_code, ret.data))
+        # Connexion and manual request.json validation methods will preempt and throw 400 error as
+        # soon as the first field invalidates. Only one field may be present in the response.data depending
+        # upon the order of fields sorted in swagger yaml file and in views.
+        self.assertTrue('title' in ret.data, msg=ret.data)
```
aksareen: Forgot to remove this line. 
The 400 response is json which is structured like one below : 
{
         "detail": "'update_type' is a required property",
         "status": 400,
         "title": "Bad Request",
         "type": "about:blank"
       }

auslib/web/admin/swagger/api.yaml
```diff
@@ -0,0 +1,441 @@
+swagger: '2.0'
+info:
+  version: '2.27'
+  title: "Balrog's Admin API"
+  termsOfService: "https://wiki.mozilla.org/Balrog"
+  license:
+    name: "Mozilla Public License 2.0"
+    url: "https://www.mozilla.org/en-US/MPL/2.0/"
+  contact:
+    name: "Ben Hearsum"
```
aksareen: modifying

auslib/web/admin/swagger/api.yaml
```diff
@@ -0,0 +1,441 @@
...
+    url: "https://github.com/mozilla/balrog"
+    email: "ben@hearsum.ca"
```
aksareen: modifying

auslib/web/admin/swagger/api.yaml
```diff
@@ -0,0 +1,441 @@
...
+  description: "Mozilla's Update Server. APIs for Balrog" #TODO write in detail using Github Flavored Markdown
+
+schemes:
+  - http
```
aksareen: _Definition from swagger docs_ : **schemes** : 
> The transfer protocol of the API. Values MUST be from the list: "http", "https", "ws", "wss". If the schemes is not included, the default scheme to be used is the one used to access the Swagger definition itself.

Will add https and test it out

auslib/web/admin/swagger/api.yaml
```diff
@@ -0,0 +1,441 @@
...
+
+tags:
+  - name: "Rules"
+    description: "Balrog Rules Description."
+    externalDocs:
+      description: "Find out more about Rules"
+      url: "http://mozilla-balrog.readthedocs.io/en/latest/database.html#rules"
+
+  - name: "Users"
+    description: "Data about admin Users"
```
aksareen: deleting admin keyword

auslib/web/admin/swagger/api.yaml
```diff
@@ -0,0 +1,441 @@
...
+      tags:
+        - "CSRF Token"
+      operationId: auslib.web.admin.views.mapper.csrf_get
+      consumes: []
+      produces:
+        - text/html
+      deprecated: false
+      responses:
+        '200':
+          description: csrf token generate
```
aksareen: modifying

auslib/web/admin/swagger/api.yaml
```diff
@@ -0,0 +1,441 @@
...
+        - Rules
+      operationId: auslib.web.admin.views.mapper.rules_get
+      parameters:
+        - name: product
+          in: query
+          description: Name of the product to filter the rules against.
+          type: string
+          x-nullable: true
+          required: false
+        - name: format
```
aksareen: this parameter is only passed when API is invoked from the test cases.

auslib/web/admin/swagger/api.yaml
```diff
@@ -0,0 +1,441 @@
...
+                items:
+                  type: string
+                  minItems: 0
+                example: ["balrogadmin"]
+          #examples:
+          #  application/json:
+          #    users: ["balrogadmin"]
+# NOTE: 'examples' object in Response objects is no longer supported.
+# However it hasn't been documented properly in swagger docs
+# To see example values, set the 'example' value in the Model
```
aksareen: Nope,  Deleting it now.

auslib/web/admin/swagger/api.yaml
aksareen: I was still trying to still experiment with x-nullable at object/model level. Doesn't work for current version of connexion (1.1.9) , so will remove that from all the fields. Also  declaring 'type' as array like [string, null] is in compliance with jsonschema. [Jsonschema : type](https://spacetelescope.github.io/understanding-json-schema/reference/type.html)

auslib/web/admin/swagger/api.yaml
aksareen: I didn't get it ?

auslib/web/admin/swagger/api.yaml
aksareen: Deleting it

auslib/web/admin/views/mapper.py
```diff
@@ -0,0 +1,76 @@
...
+
+def rules_post():
+    return RulesAPIView().post()
+
+
+def users_get():
+    return UsersView().get()
+
+
+@draft4_format_checker.checks(format="buildID", raises=ValidationError)
```
aksareen: Moving them out into a new validators file

auslib/web/admin/views/mapper.py
```diff
@@ -0,0 +1,76 @@
...
+def operator_validator(field_value):
+    log.debug('starting in operator_validator: buildID is %s' % field_value)
+    if not isinstance(field_value, str_types):
+        return True
+    # empty input is fine
+    if field_value is None or field_value == '':
+        return True
```
aksareen: Actually we're overriding that validation with our own custom validation check, hence it needs to check whether the field value conforms to string type or not. Took reference from these [examples](https://github.com/Julian/jsonschema/blob/ee2de6d0e0b6fed6f580cf2d744fb2790fe06a54/jsonschema/_format.py)

auslib/web/admin/views/rules.py
```diff
@@ -46,32 +47,37 @@ def get(self, **kwargs):
...
-                    systemCapabilities=form.systemCapabilities.data,
-                    distVersion=form.distVersion.data,
-                    comment=form.comment.data,
-                    update_type=form.update_type.data,
-                    headerArchitecture=form.headerArchitecture.data)
+        # Replaces wtfForms validations
+        nullable_field_dict = dict()
+        for key in request.json:
+            if isinstance(request.json[key], str_types):
+                nullable_field_dict[key] = None if request.json[key].strip() == '' else request.json[key].strip()
```
aksareen: let me check whether wtfForms stripped white spaces or not. AFAIK empty string was passed as None in python or NULL in DB.

auslib/test/admin/views/test_rules.py
mozbhearsum: Ah, I see. So this is the format of responses if Connexion hits a validation error. We'll need to make sure the UI handles this, too.

auslib/web/admin/swagger/api.yaml
mozbhearsum: Which test cases? I don't see any using it in this pull request.

auslib/web/admin/swagger/api.yaml
mozbhearsum: The [string, null] part makes a lot of sense, I was just confused about why there was both. Carry on!

auslib/web/admin/swagger/api.yaml
mozbhearsum: I was just wondering if there should be links to documentation in this section, like you have in the endpoints.

auslib/web/admin/views/mapper.py
mozbhearsum: Does this mean that the type, x-nullable, and min/max length defined in the yml file don't do anything?

mozbhearsum:
Also, apologies for the bitrot after #276 got backed out. I'm hoping that will re-land in the next week or so, so you may want to defer updating this branch from master.

auslib/test/admin/views/test_rules.py
aksareen: That is a bit more trickier to manage. This response is only seen in test cases or when we attempt to invoke the API apart from the UI page , eg. a curl request.
When I try to submit a new rule from UI page I get [this](https://pastebin.mozilla.org/9020893) in the logs. 

Two observations from it, UI forms send a blank string value i.e. '' instead of null or none to the server via the wtfForms. Connexion acting in middle-ware for our flask app  intercepts the request automatically, validates the values based on the swagger spec file and immediately throws error. However, it fails to specify the field name which was invalid and neither does it highlight the invalid fields on UI page.

auslib/web/admin/swagger/api.yaml
aksareen: [Here](https://github.com/mozilla/balrog/blob/master/auslib/test/admin/views/base.py#L148)

auslib/web/admin/swagger/api.yaml
aksareen: We had discussed during the proposal submission phase that we require hierarchical models because , example : Rules model, priority and backgroundRate are both required when creating a new rule, but not required when updating a rule. 
This schema inherits RuleBase and then modifies it to mandate priority and backgroundRate both as required during posting a new rule along with update_type

auslib/web/admin/swagger/api.yaml
```diff
@@ -0,0 +1,441 @@
...
+definitions:
+  CSRFModel:
+    title: CSRF Token
+    description: CSRF Token in POST
+    type: object
+
+    #Existing Test cases fail when we don't provide this value. Hence keep it out of required section.
+    #Make it optional.
+    #required:
+    #- csrf_token
```
aksareen: I couldn't find csrf_token being explicitly passed in the test client's post request. [Link](https://github.com/mozilla/balrog/blob/master/auslib/test/admin/views/base.py#L155) . We can try and modify tests to pass a random csrf_token and make it required argument even during executing test cases.

auslib/web/admin/swagger/api.yaml
aksareen: yeah I can add an externalDocs object here.

auslib/web/admin/swagger/api.yaml
aksareen: That's the definition of product in the documentation : 
>  Product 	The name of the application requesting an update. 	Exact string match only “Firefox” or “B2G”
How should I describe it ?

auslib/web/admin/views/mapper.py
aksareen: All of them are used in validation too. x-nullable is used [here](https://github.com/zalando/connexion/blob/53908d30e566ccb81d73d3e27b461f407a8a8d31/connexion/utils.py#L101) . min/max length constraints are checked if the type is string. Format is used for special purpose or custom validation as per our use-case. However, the jsonschema designers made sure that if someone incorrectly specifies format for an invalid data type (eg. int instead of string), then perhaps it shouldn't throw a validation error. [examples](https://github.com/Julian/jsonschema/blob/ee2de6d0e0b6fed6f580cf2d744fb2790fe06a54/jsonschema/_format.py#L151)

As for the checking if value if None or an empty string, this is done because UI sends a blank string value, and since buildID is nullable field, hence allowing blank values to pass through the validation checks and then when processing them [here](https://github.com/aksareen/balrog/blob/first_set_of_apis_migrate/auslib/web/admin/views/rules.py#L56) , we can simply pass them into DB as NULL.

auslib/web/admin/views/rules.py
```diff
@@ -46,32 +47,37 @@ def get(self, **kwargs):
...
-                    buildTarget=form.buildTarget.data,
-                    osVersion=form.osVersion.data,
-                    systemCapabilities=form.systemCapabilities.data,
-                    distVersion=form.distVersion.data,
-                    comment=form.comment.data,
-                    update_type=form.update_type.data,
-                    headerArchitecture=form.headerArchitecture.data)
+        # Replaces wtfForms validations
+        nullable_field_dict = dict()
+        for key in request.json:
```
aksareen: nope, forgot about it. connexion.requests is just a proxy to flask.request object

auslib/test/admin/views/test_rules.py
aksareen: One approach can be to use custom validator for requests to customize the responses generated when error occurs. Inheriting [this](https://github.com/zalando/connexion/blob/master/connexion/decorators/validation.py#L156) validator method and re-writing it like ex-Dev did for re-writing and one of his parameter validators according to his use-case. 

Also, reason why connexion throws exception as soon as it encounters error in validation, in contrast to wtfForms validation which runs the validation on all fields and returns a dictionary of field key names and their exceptions is because when it invokes jsonschema's validate() method (which is currently based on json schema Draft4Validator) , it in turns throws exception as soon as it encounters. [Link](https://github.com/Julian/jsonschema/blob/master/jsonschema/validators.py#L128).  This implies that we can only highlight or throw one field error at a time from now in our  UI forms .

mozbhearsum: Only throwing one error at a time isn't the end of the world (though it's annoying), but it's important to get the field name to show a useful error message at least. Custom validators seem like a reasonable approach to do this.

auslib/web/admin/swagger/api.yaml
mozbhearsum: Ah, I'm pretty sure this is dead. I don't see anything in the tests overriding "format", nor any handling of it in the admin app. Best to remove it instead.

auslib/web/admin/swagger/api.yaml
mozbhearsum: Ah, I see. I think the fact that "update_type" is required in GETs was confusing me. That's totally correct, I just never thought of it that way!

auslib/web/admin/swagger/api.yaml
mozbhearsum: It's important that csrf_token is required in production, so we'll need some sort of workaround for this.

auslib/web/admin/swagger/api.yaml
mozbhearsum: Hm, if that's what the docs have this is probably fine for now.

auslib/web/admin/views/mapper.py
mozbhearsum: It's unfortunate that the UI sends an empty string (it should really send null for empty fields) - we should fix that at some future point. I'm still surprised that you need to handle None here though. Does the validator even get called when the value is None/null? Typically validators aren't called when the data isn't present, but maybe swagger/connexion work differently.

auslib/test/admin/views/test_rules.py
aksareen: The request object passed [here](https://github.com/zalando/connexion/blob/master/connexion/decorators/validation.py#L105) is an instance of [ConnexionRequest](https://github.com/zalando/connexion/blob/master/connexion/lifecycle.py#L2). Here's the [link](https://pastebin.mozilla.org/9021108) debug log when I post a new rule.

When validate method is invoked [here](https://github.com/zalando/connexion/blob/master/connexion/decorators/validation.py#L133), then the entire json data is validated in one go based on schema. Hence , jsonschema validation throws the first error it encounters and the message cannot be customized unless we edit jsonschema code. Also the ConnexionRequest class objects have no instance variable or field which can identify the invalid field's key or parameter name. Ex: 

> {
  "detail": "u'mindosr' is not one of ['major', 'minor']",
  "status": 400,
  "title": "Bad Request",
  "type": "about:blank"
}

The above response doesn't contain the key name that update_type field value is invalid. However if it is a 'required' property and is missing then the detail key says "'update_type' field is missing".

auslib/web/admin/views/mapper.py
aksareen: Let me check if None is passed at all or not

auslib/web/admin/swagger/api.yaml
aksareen: We'll need to modify the test client's post method in test cases to pass a random csrf_token each time.

auslib/web/admin/swagger/api.yaml
aksareen: Alright. I'll have to modify the test cases first.

auslib/test/admin/views/test_rules.py
mozbhearsum: @tarekziade - I wonder if you have any idea what to do here. The problem we're running into is that the errors returned by connexion when validation fails do not contain the name of the field, which means we can't place the errors into the UI appropriately. Do you know if there's a way to match up errors with fields, or some other recommended way of coping with this?

auslib/web/admin/swagger/api.yaml
mozbhearsum: Ah, I see. I'd forgotten that we were able to easily disable it for testing purposes with WTForms.

auslib/test/admin/views/test_rules.py
tarekziade: @mozbhearsum reading at the connexion code, looks like you need a custom RequestBodyValidator to set up the problem response correctly.  I am not sure what it entails, I'd ask the connexion folks :)

aksareen: @tarekziade , I've already tried that. The issue is that using a custom RequestBodyValidator requires me to inherit [this](https://github.com/zalando/connexion/blob/master/connexion/decorators/validation.py#L142) class and modify the validate_schema method. The issue here is that the form data is passed as a json object [here](https://github.com/zalando/connexion/blob/master/connexion/decorators/validation.py#L107) (keys are the form field names and values are the field's values). The json is passed to the jsonschema's schema validator method when validate method ,[here](https://github.com/zalando/connexion/blob/master/connexion/decorators/validation.py#L133) , is invoked. Now this inturn invokes jsonschema's [validate method](https://github.com/Julian/jsonschema/blob/master/jsonschema/validators.py#L128) which validates the input json based on swagger schema and raisesan error as soon as it encounters one. 
Now the tricky part : the error response from jsonschema does not specify the field name in the error.message .  This is an issue for us because as mentioned in example above if i pass invalid value to ,say update_type field (which can have either 'major' or 'minor' field value) then the error message just says something like this : 

> "u'mindosr' is not one of ['major', 'minor']",


Now, this means that the field name isn't specified above.  So, as far as I understood, I will have to modify the jsonschema code rather than using a custom RequestBodyValidator (though it can be used later for some custom validations besides this issue) which itself is a bit arduous and a maintenance nightmare for us. What do you think ? Is there a way to circumvent this issue ?

mozbhearsum: Thanks for the insight @tarekziade.

@aksareen - I wonder if it's worth asking the jsonschema folks if there's a way to get more specific information about errors? Or maybe looking through their issue tracker? https://github.com/Julian/jsonschema/issues/119 talks about a "schema_path" and "path" in the Exception object, maybe those might help?

auslib/web/admin/swagger/api.yaml
aksareen: Turns out the externalDocs object can be added, but according to swaggerhub it is unexpected (invalid too) and won't be used in rendering the UI when passed in the model. Most likely due to fact that if model inherits multiple models then multiple externalDocs will conflict and one single doc cannot be decided upon.

aksareen:
@mozbhearsum . So far we've been discussing how to handle the response format that is generated by connexion (which is a json object with error message in 'detail' key). And yesterday we planned to modify the UI so that it can be customized to accommodate connexion's error response via help of a label at the bottom of the form. However, this is the first validation done at connexion (middle) before the request is passed to the flask's actual endpoint-method. Now suppose in /rules POST method, we have a field of form.mapping which is a select field. 

Hence the choices i.e. release names  must be loaded from the release table and then validated if they were entered correctly or not. In case of an error, the API sends a 400 error response with error message that **can be set**  like this 

> return Response(status=200, response=str("Invalid mapping value. No release name exists")). 

I see 2 possible approaches :
(1)  The UI must be modified to check if the response is a json or string in case of error. If json then output response.detail in the error label, else set the error label to response in case of string.
(2) Use and code connexion's [problem](https://github.com/zalando/connexion/blob/master/connexion/problem.py#L33) exception response format for HTTP 400 error responses at flask app's API endpoint methods too. This means that every 400 error response should have same format as defined in problem i.e. json object having 4 keys : detail , title, and status . whether it is returned from connexion or the admin app.

What do you think ?

aksareen:
There's another issue with UI forms. Here's the log containing the request json : 

> balrogadmin_1  | 2017-05-12 07:03:26,234 - DEBUG - PID: 1095 - Request: 140170821867536 - connexion.apis.flask_api.get_request#256: Getting data and status code
> balrogadmin_1  | 2017-05-12 07:03:26,235 - DEBUG - PID: 1095 - Request: 140170821867536 - auslib.web.admin.views.validators.wrapper#36: http://balrogadmin:7070/rules validating schema...
> balrogadmin_1  | {u'_duplicate': False, u'product': u'', u'backgroundRate': u'89', u'csrf_token': u'1494575791##256c0e3a157037c23e86cbe4b0c08a01e31c0f66', u'priority': u'989', u'update_type': u'minor'}
> balrogadmin_1  | 2017-05-12 07:03:26,235 - ERROR - PID: 1095 - Request: 140170821867536 - auslib.web.admin.views.validators.validate_schema#59: http://balrogadmin:7070/rules validation error: backgroundRate: u'89' is not of type 'integer', 'null'
> balrogadmin_1  | 2017-05-12 07:03:26,235 - DEBUG - PID: 1095 - Request: 140170821867536 - connexion.apis.flask_api.get_response#143: Getting data and status code
> balrogadmin_1  | 2017-05-12 07:03:26,236 - DEBUG - PID: 1095 - Request: 140170821867536 - connexion.apis.flask_api.get_response#156: Got data and status code (400)
> balrogadmin_1  | [pid: 1095|app: 0|req: 39/39] 172.18.0.6 () {56 vars in 1014 bytes} [Fri May 12 07:03:26 2017] POST /api/rules => generated 144 bytes in 5 msecs (HTTP/1.1 400) 7 headers in 442 bytes (1 switches on core 0)


The integer fields i.e. priority and backgroundRate should be passed as integer from UI forms are getting passed as Unicode values. This causes connexion checks to fail the integer data type check. 
I'm starting to think connexion was not designed to work with any kind of UI or forms at all.

So, the only possible way to circumvent this is to declare all fields as string in swagger spec file. Create custom integer format validators for specific fields (which try to cast string into integer to verify the values). But this means in API endpoint methods in flask app, we have to add explicit checks and conversions if these specific fields are present. That's less generic than I was expecting as we're again handling checks for many fields in the code explicitly.

aksareen:
There's another issue with UI forms. Here's the log containing the request json : 

> balrogadmin_1  | 2017-05-12 07:03:26,234 - DEBUG - PID: 1095 - Request: 140170821867536 - connexion.apis.flask_api.get_request#256: Getting data and status code
> balrogadmin_1  | 2017-05-12 07:03:26,235 - DEBUG - PID: 1095 - Request: 140170821867536 - auslib.web.admin.views.validators.wrapper#36: http://balrogadmin:7070/rules validating schema...
> balrogadmin_1  | {u'_duplicate': False, u'product': u'', u'backgroundRate': u'89', u'csrf_token': u'1494575791##256c0e3a157037c23e86cbe4b0c08a01e31c0f66', u'priority': u'989', u'update_type': u'minor'}
> balrogadmin_1  | 2017-05-12 07:03:26,235 - ERROR - PID: 1095 - Request: 140170821867536 - auslib.web.admin.views.validators.validate_schema#59: http://balrogadmin:7070/rules validation error: backgroundRate: u'89' is not of type 'integer', 'null'
> balrogadmin_1  | 2017-05-12 07:03:26,235 - DEBUG - PID: 1095 - Request: 140170821867536 - connexion.apis.flask_api.get_response#143: Getting data and status code
> balrogadmin_1  | 2017-05-12 07:03:26,236 - DEBUG - PID: 1095 - Request: 140170821867536 - connexion.apis.flask_api.get_response#156: Got data and status code (400)
> balrogadmin_1  | [pid: 1095|app: 0|req: 39/39] 172.18.0.6 () {56 vars in 1014 bytes} [Fri May 12 07:03:26 2017] POST /api/rules => generated 144 bytes in 5 msecs (HTTP/1.1 400) 7 headers in 442 bytes (1 switches on core 0)


The integer fields i.e. priority and backgroundRate should be passed as integer from UI forms are getting passed as Unicode values. This causes connexion checks to fail the integer data type check. 
I'm starting to think connexion was not designed to work with any kind of UI or forms at all.

So, the only possible way to circumvent this is to declare all fields as string in swagger spec file as HTML forms do not define "int" type explicitly. Create custom integer format validators for specific fields (which try to cast string into integer to verify the values). But this means in API endpoint methods in flask app, we have to add explicit checks and conversions if these specific fields are present. That's less generic than I was expecting as we're again handling checks for many fields in the code explicitly.

But then the swagger spec file will show that these fields are string type even though they are present as integers in DB.

aksareen:
With respect to [last comment](https://github.com/mozilla/balrog/pull/312#issuecomment-301002486) , even if we change the data type in swagger spec from string to integer for requests from UI, then we cannot use the same base model for validating the responses as well, because the response from DB will have priority and backgroundRate as integer. So , in that scenario two different models will have to be considered, one for validating Requests and other one for validating Responses.  It will be something 

like 

`RequestModel(BaseModel) : 
    priority: type : string 
    backgroundRate : type string
`
`ResponseModel(BaseModel) : 
    priority: type : integer 
    backgroundRate : type integer
`

This will certainly complicate things a lot.

aksareen:
I think a I found a better and simpler way.  Simply, changing data_type of priority and backgroundRate in swagger spec from ['int','null'] to ['int','string','null'] . This way we will be able to get both string and int formats and convert string into int in requests and continue to serve int in response. Though the maximum and minimum constraints will no longer be applicable, however, we can write our own custom format validator for these fields like we already do for operator and buildID. :D

auslib/web/admin/swagger/api.yaml
mozbhearsum: Ah, OK. No big deal, it was just a random thought.

auslib/test/admin/views/test_rules.py
```diff
@@ -141,16 +141,29 @@ def testNewRuleWithoutPermission(self):
...
+        # upon the order of fields sorted in swagger yaml file and in views.
+        post_data['update_type'] = 'minor'
+        ret = self._post('/rules', data=post_data)
+
         self.assertEquals(ret.status_code, 400, "Status Code: %d, Data: %s" % (ret.status_code, ret.data))
         self.assertTrue('backgroundRate' in ret.data, msg=ret.data)
+        post_data['backgroundRate'] = 10
+        ret = self._post('/rules', data=post_data)
+
+        self.assertEquals(ret.status_code, 400, "Status Code: %d, Data: %s" % (ret.status_code, ret.data))
```
mozbhearsum: Good job fixing this up. This test is much less brittle now!

auslib/web/admin/swagger/api.yaml
```diff
@@ -0,0 +1,423 @@
...
+    name: "Mozilla Public License 2.0"
+    url: "https://www.mozilla.org/en-US/MPL/2.0/"
+  contact:
+    name: "Mozilla Release Engineering"
+    url: "https://github.com/mozilla/balrog"
+    email: "release-engineering@lists.mozilla.org"
+  description: "Mozilla's Update Server. APIs for Balrog" #TODO write in detail using Github Flavored Markdown
+
+schemes:
+  - http
```
mozbhearsum: Does removing http break anything (eg: local development)? If it doesn't, please remove it.

auslib/web/admin/swagger/api.yaml
mozbhearsum: This can be addressed in a follow-up. Maybe we can file a new bug and make it a good first bug.

auslib/web/admin/swagger/api.yaml
```diff
@@ -0,0 +1,423 @@
...
+        - name: rule
+          in: body
+          description: New Rule object data
+          required: true
+          schema:
+            allOf:
+              - $ref: '#/definitions/CSRFModel'
+              - $ref: '#/definitions/RulesBase'
+
+            #If you change the order, make sure to modify test cases too.
```
mozbhearsum: Good idea to add a reminder about this.

auslib/web/admin/swagger/api.yaml
```diff
@@ -0,0 +1,423 @@
...
+      rule_id:
+        description: The id of the rule. This id is necessary to make changes to the rule through the REST API.
+        type: integer
+        format: int32
+        readOnly: true
+        example: 10
+
+      priority:
+        description: The priority of the rule, relative to other rules. If multiple rules match an incoming request based on the Matchable columns, the rule with the highest priority is chosen.
+        type: ["integer", "string", "null"]
```
mozbhearsum: Please make sure you file a follow-up bug for this. I'm okay with allowing strings for now, but it's definitely something I'd like to have fixed later (not necessarily by you).

auslib/web/admin/views/validators.py
```diff
@@ -0,0 +1,147 @@
...
+        x = int(field_value)
+    except:
+        raise JsonschemaValidationError('Invalid input %s for priority.Not an integer.' % field_value)
+    if x < 0:
+        raise JsonschemaValidationError("Priority field value %s shouldn't be less than 0" % field_value)
+    return True
+
+
+@draft4_format_checker.checks(format="backgroundRate", raises=JsonschemaValidationError)
+def background_rate_validator(field_value):
```
mozbhearsum: Can you refactor this to minimize duplication between priority_validator and background_rate_validator? It seems like you could support both of them by writing a function that allows a minimum and maximum value to be specified.

auslib/web/admin/swagger/api.yaml
aksareen: yeah it's needed in local development and testing.

auslib/web/admin/swagger/api.yaml
aksareen: about the "format" paramter above, I forgot to delete it from swagger spec. Already modified the test cases to no longer use it anymore.

auslib/web/admin/views/validators.py
aksareen: The format name is same as field name to improve readability. However, the format checker method is independent from the field name and type. So here only the field's data value is passed to the method. There's no way of verifying the field name or its original type. That's why different format checker methods are needed for these fields.

auslib/web/admin/swagger/api.yaml
aksareen: Alright will file the bug in Balrog:FrontEnd. However what should be the title of the bug ?

mozbhearsum: Something like "only allow integers for priority and backgroundRate".

aksareen: Alright once this is merged , then I should file a bug so that I can write the pointers to the assignee where to make code changes both in front and back end.

auslib/web/admin/views/validators.py
aksareen: As discussed moving the checker into a helper method

auslib/web/admin/views/validators.py
```diff
@@ -124,24 +93,23 @@ def priority_validator(field_value):
     try:
         x = int(field_value)
     except:
-        raise JsonschemaValidationError('Invalid input %s for priority.Not an integer.' % field_value)
-    if x < 0:
-        raise JsonschemaValidationError("Priority field value %s shouldn't be less than 0" % field_value)
+        raise JsonschemaValidationError(message=("Invalid input for %s. Not an integer." % field_name))
+    if min_val is not None and x < min_val:
+        raise JsonschemaValidationError(
+            message=("%s field value should be an integer >= %s" % field_name, min_val))
```
mozbhearsum: I think you need () around the formatting parameters here - this is currently causing 500s like:
balrogadmin_1  |   File "./auslib/web/admin/views/validators.py", line 99, in integer_and_range_validator
balrogadmin_1  |     message=("%s field value should be an integer >= %s" % field_name, min_val))
balrogadmin_1  | TypeError: not enough arguments for format string


Looks fine other than that.

PR #314, opened 2017-05-06
B1325377 - Public API to Balrog's database

ex-Dev:
PR reopened to fix unicode error in route.
-Unicode strings was encoded to utf-8 on validation level.
-New unit test to try url with unicode values.


auslib/web/api_validator.py
```diff
@@ -0,0 +1,30 @@
...
+        val = request.path_params.get(param['name'].replace('-', '_'))
+
+        # Supporting the flask add_url_rule "defaults" parameter to the swagger path parameter.
+        if val is None and 'default' in param:
+            val = param['default']
+            request.path_params[param['name']] = val
+
+        # Encode strings in utf-8.
+        if param['type'] == 'string':
+            val = codecs.encode(val, 'utf-8')
```
mozbhearsum: After speaking with @catlee , I don't think assuming utf-8 is a great thing to do here. We may get all sorts of different encodings, and we have no idea what will be correct. Maybe you can look into how Werkzeug or Flask does this now, and replicate that behaviour?

ex-Dev: I had considered these two resoures before using utf-8:

-https://docs.python.org/2/howto/unicode.html#encodings
"UTF-8 has several convenient properties:
1) It can handle any Unicode code point."

and

-http://flask.pocoo.org/docs/0.12/unicode/

Today I took a look in werkzeug docs and found this:
https://github.com/pallets/werkzeug/blob/0bc61df6e1ae9f2ffdf5d066aa3cd9d5ebcb307d/docs/unicode.rst#unicode-in-http

Seems that there is a open PR to handle strings as UTF-8 too in connexion:
zalando/connexion#349

mozbhearsum: I'm not exactly sure why those Werkzeug docs talk about assuming UTF-8, but I don't think it's what we want here. As far as we're concerned, non-ASCII data in the URL is garbage, and we should do our best to ignore it. How about using 'val.encode("ascii", "ignore")', which will throwaway any unparseable characters? This will get the field into the most usable state.

ex-Dev: Ah, if non-ASCII chars are garbage, it's OK, I'll make more tests tonight, probably 'ignore' as parameter will lose these chars.
Maybe Werkzeug assumes UTF-8 because data does not be lost.

ex-Dev:
Now, only ascii char are valid for url.

auslib/test/web/test_unicode.py
```diff
@@ -0,0 +1,21 @@
...
+        dbo.create()
+        self.client = app.test_client()
+
+    def testUnicodeInRoute(self):
+        url = "/update/3/GMP/53.0/20170421105455/Linux_x86_64-gcc3/null/{}/Linux%204.4.0-53-generic%20(GTK%203.18.9,libpulse%208.0.0)/mint/1.0/update.xml"
+        # Test an arbitrary number of unicode chars
+        cdec = step = 42
+        while cdec < 4200:
+            channel = ''.join(unichr(c) for c in range(cdec, cdec + step) if unichr(c).isalpha())
+            cdec += step
```
mozbhearsum: Nice idea for testing this - I like it!

PR #316, opened 2017-05-10
Preemptively bump to next release version.

mozbhearsum:


mozbhearsum:
@rail - can you stamp this when you have time? I'm doing them right after production pushes these days to get them out of the critical path, so no rush.

PR #317, opened 2017-05-18
Bug 1336452 : migrate all remaining Rules and Users APIs

aksareen:
PR to migrate the remaining /rules/* and /users/* APIs


aksareen:
Some test cases are failing. Will reslove it ASAP

auslib/test/admin/views/test_rules.py
```diff
@@ -316,13 +340,14 @@ def testPostRuleOutdatedData(self):
...
-        ret2 = self._put('/rules/1', data=dict(backgroundRate=71, mapping='d', priority=73, data_version=1, product='Firefox', channel='nightly'))
+        ret2 = self._put('/rules/1', data=dict(backgroundRate=71, update_type='minor', mapping='d', priority=73,
+                                               data_version=1, product='Firefox', channel='nightly'))
         self.assertEquals(ret2.status_code, 400, "Status Code: %d, Data: %s" % (ret2.status_code, ret2.data))
 
     def testPostByAlias(self):
         # Make some changes to a rule
         ret = self._post('/rules/frodo', data=dict(backgroundRate=71, mapping='d', priority=73, data_version=1,
-                                                   product='Firefox', channel='nightly'))
+                                                   product='Firefox', channel='nightly', update_type="minor"))
```
mozbhearsum: Making update_type a required argument for rules that already exist is going to be problematic. We've got tools and scripts that make changes to existing Rules that will break because of this. Is it possible to make update_type required for new rules, but optional for existing ones? Maybe by allowing type to be string or null like you've done or other fields?

auslib/test/admin/views/test_rules.py
```diff
@@ -307,7 +326,12 @@ def testPutRuleOutdatedData(self):
 
     def testPostRuleOutdatedData(self):
         # Make changes to a rule
-        ret = self._post('/rules/1', data=dict(backgroundRate=71, mapping='d', priority=73, data_version=1, product='Firefox', channel='nightly'))
+        ret = self._post('/rules/1', data=dict(backgroundRate=71,
+                                               mapping='d', priority=73,
+                                               data_version=1, product='Firefox', channel='nightly'))
+        self.assertEquals(ret.status_code, 400, "Status Code: %d, Data: %s" % (ret.status_code, ret.data))
+        ret = self._post('/rules/1', data=dict(backgroundRate=71, mapping='d', priority=73, data_version=1,
+                                               product='Firefox', channel='nightly', update_type='major'))
```
mozbhearsum: No need to add an extra post() here - just add update_type to the existing one. Same goes for the other tests where you're doing this. Unless the test is explicitly testing the update_type case, there's no reason to check or verify it.

auslib/test/admin/views/test_rules.py
```diff
@@ -505,11 +536,14 @@ def testDeleteRuleOutdatedData(self):
...
 
     def testDeleteRuleByAlias(self):
-        ret = self._delete('/rules/frodo', qs=dict(data_version=1))
+        ret = self._delete('/rules/frodo', qs={"data_version": 1})
         self.assertEquals(ret.status_code, 200, msg=ret.data)
 
     def testDeleteRule404(self):
         ret = self._delete("/rules/112")
+        # Missing data_version field
+        self.assertEquals(ret.status_code, 400)
```
mozbhearsum: Same thing goes here -- this test's purpose isn't to verify the data_version handling, it's to verify 404 handling, so get rid of the extra delete() and pass data_version in the existing one.

auslib/web/admin/swagger/api.yaml
```diff
@@ -91,21 +108,25 @@ paths:
                 example: ["balrogadmin"]
   /rules:
     get:
-      summary: Returns filtered list of Rules in Balrog's DB
-      description: "Returns filtered list Rules in Balrog’s database
-      inside of a JSON Object. Returns all rules if 'product' query parameter isn't passed"
+      summary: Returns filtered list of Rules from Balrog's DB
+      description: >
+        Returns filtered list Rules in Balrog’s database inside of a JSON Object. Returns all rules if 'product' query parameter isn't passed. [Docs](http://mozilla-balrog.readthedocs.io/en/latest/admin_api.html#id1)
```
mozbhearsum: Nice improvements to the docs here.

auslib/web/admin/swagger/api.yaml
```diff
@@ -168,30 +192,229 @@ paths:
...
+      produces:
+        - application/json
+        - text/html
+      externalDocs:
+        url: "http://mozilla-balrog.readthedocs.io/en/latest/admin_api.html#id2"
+        description: "Returns the entire rule identified by the id or alias given in JSON format."
+      parameters:
+        - name: id_or_alias
+          in: path
+          description: rule_id or alias field value to filter the rule against
```
mozbhearsum: I don't think it makes sense to use the word "filter" here. How about "rule_id or alias to update"?

auslib/web/admin/views/base.py
```diff
@@ -25,34 +23,39 @@ def decorated(*args, **kwargs):
...
-                msg = "Couldn't perform the request %s. Outdated Data Version. old_data_version doesn't match current data_version" % messages
+                msg = "Couldn't perform the request %s. Outdated Data Version. " \
+                      "old_data_version doesn't match current data_version" % messages
                 logging.warning("Bad input: %s", msg)
                 logging.warning(e)
-                return Response(status=400, response=json.dumps({"exception": msg}), mimetype="application/json")
+                # using connexion.problem results in TypeError: 'ConnexionResponse' object is not callable
+                # hence using flask.Response but modifying response's json data into connexion.problem format
+                # for validation purpose
+                return problem(400, "Bad Request", "OutDatedError", ext={"exception": msg})
```
mozbhearsum: This should probably be "OutdatedDataError".

auslib/web/admin/views/base.py
```diff
@@ -25,34 +23,39 @@ def decorated(*args, **kwargs):
...
             except OutdatedDataError as e:
-                msg = "Couldn't perform the request %s. Outdated Data Version. old_data_version doesn't match current data_version" % messages
+                msg = "Couldn't perform the request %s. Outdated Data Version. " \
+                      "old_data_version doesn't match current data_version" % messages
                 logging.warning("Bad input: %s", msg)
                 logging.warning(e)
-                return Response(status=400, response=json.dumps({"exception": msg}), mimetype="application/json")
+                # using connexion.problem results in TypeError: 'ConnexionResponse' object is not callable
+                # hence using flask.Response but modifying response's json data into connexion.problem format
+                # for validation purpose
```
mozbhearsum: This comment could be improved. It talks about "using flask.Response", but there's no references to it here. Maybe this should be documented in problem.py instead?

auslib/web/admin/views/problem.py
```diff
@@ -0,0 +1,44 @@
+import json
+from flask import Response
+
+
+def problem(status, title, detail, type=None, instance=None, headers=None, ext=None):
```
mozbhearsum: It's unfortunate that you need to implement this yourself. What exactly went wrong with using Connexion's problem?

auslib/web/admin/views/rules.py
```diff
@@ -46,25 +74,11 @@ def get(self, **kwargs):
...
-        mapping_values = [y for x, y in mapping_choices if x == what.get("mapping")]
+        what, mapping_values = process_rule_form(connexion.request.json)
         if what.get('mapping', None) is not None and len(mapping_values) != 1:
             return problem(400, 'Bad Request', 'Invalid mapping value. No release name found in DB')
-
-        for i in ["priority", "backgroundRate"]:
-            if what.get(i, None):
-                what[i] = int(what[i])
+        # Solves Bug https://bugzilla.mozilla.org/show_bug.cgi?id=1361158
+        what.pop("csrf_token", None)
```
mozbhearsum: Thanks for fixing this!

auslib/web/admin/views/rules.py
```diff
@@ -89,40 +103,37 @@ def _post(self, id_or_alias, transaction, changed_by):
...
-            return Response(status=400, response=json.dumps(form.errors))
+        # Solves https://bugzilla.mozilla.org/show_bug.cgi?id=1361158
+        edit_rule_dict.pop("csrf_token", None)
 
         what = dict()
         # We need to be able to support changing AND removing parts of a rule,
         # and because of how Flask's request object and WTForm's defaults work
         # this gets a little hary.
-        for k, v in form.data.iteritems():
+        for k, v in edit_rule_dict.iteritems():
```
mozbhearsum: I think you can get rid of this entire block if you pop data_version like you're doing above for csrf_token. Now that the request.json/request.form handling is gone, the only thing this block does is filter out data_version.

auslib/test/admin/views/test_rules.py
aksareen: yeah it can easily done.

auslib/web/admin/views/problem.py
aksareen: The response error is like this 

> using connexion.problem results in TypeError: 'ConnexionResponse' object is not callable .

I couldn't debug it since the error was sporadic. It worked for some APIs but surprisingly not for all everytime. I tried to debug it but couldn't in the end. So that's why I decided to re-write the same method. This way it solves my issue for now and in future if we want to re-use connexion.problem all we have to do is simple one line change in the import statement from 'auslib.admin.views.problem" to "connexion.problem" .

auslib/web/admin/views/rules.py
aksareen: I think it does the opposite. Except for data_version, it filters out the rest of the keys into a dictionary (**what) .

auslib/web/admin/views/base.py
aksareen: alright. moving this comment to problem.py too

auslib/web/admin/views/base.py
aksareen: changing it

auslib/web/admin/views/problem.py
mozbhearsum: Very strange. Maybe something to follow-up on if there ends up being extra time in the end. This is certainly fine for now though.

auslib/web/admin/views/rules.py
aksareen: oh yeah. didn't see the optimization

auslib/web/admin/views/problem.py
aksareen: Alright will try to follow up on it in week 13 or during the follow-up weeks

aksareen:
It's weird that Taskcluster says the backend tests are failing,but I cannot access the logs.  However, when I run the tests locally, they always pass successfully. here is the [log](https://pastebin.mozilla.org/9022267)

auslib/web/admin/swagger/api.yaml
```diff
@@ -168,49 +193,499 @@ paths:
...
+            application/json:
+              title: Bad Request
+              detail: Error when trying to Revert to Rule revisions
+              status: 400
+
+  /rules/columns/{column}:
+    get:
+      summary: Returns previous versions of the rule.
+      description: >
+        Returns previous versions of the rule identified by the id or alias given in a JSON Object. [Docs](http://mozilla-balrog.readthedocs.io/en/latest/admin_api.html#id4)
```
mozbhearsum: I think you forget to update these descriptions :)

auslib/web/admin/swagger/api.yaml
```diff
@@ -168,49 +193,499 @@ paths:
...
+          minimum: 1
+          x-nullable: true
+          required: false
+
+      responses:
+        '200':
+          description: Returns JSON object of list of rules with their count
+          schema:
+            type: object
+            allOf:
```
mozbhearsum: Just making sure I understand this correctly: RulesListGET and HistoryModel combine to form the full schema?

auslib/web/admin/swagger/api.yaml
```diff
@@ -259,21 +757,87 @@ definitions:
...
+            whitelist: null
+
+  UserRoles:
+    title: User Roles object definition
+    description: Users may hold any number of Roles. Roles are used when signing off on Scheduled Changes.
+    type: object
+    properties:
+      data_version:
+        description: the version number of user role
+        type: ["integer", "string"]
```
mozbhearsum: Is there a reason string is allowed here, and for rule_id further down?

aksareen: same issue that occurred with all other integer fields such as priority and backgroundRate. The rule_id is passed into request body when duplicate existing rule form submits data.

auslib/web/admin/swagger/api.yaml
aksareen: oh yeah. Updating it.

auslib/web/admin/swagger/api.yaml
aksareen: yes

auslib/web/admin/swagger/api.yaml
mozbhearsum: Ah, OK. Can you make sure there's a bug on file for this?

aksareen: https://bugzilla.mozilla.org/show_bug.cgi?id=1365658

auslib/test/admin/views/test_rules.py
```diff
@@ -209,6 +209,19 @@ def testValidationEmptyInput(self):
...
+    def testInvalidMapping(self):
+        data_dict = dict(backgroundRate=31, mapping='random', priority=33, product='a', update_type='minor')
+        ret = self._post('/rules', data=data_dict)
+        self.assertEquals(ret.status_code, 400, "Status Code: %d, Data: %s" % (ret.status_code, ret.data))
+        self.assertIn("mapping", ret.data, "Status Code: %d, Data: %s" % (ret.status_code, ret.data))
+
+    def testPutDataVersionLessThanOne(self):
+        # Throw 400 error when data_version is less than 1.
+        ret = self._put('/rules/1', data=dict(backgroundRate=71, mapping='d',
+                                              priority=73, data_version=1,
```
mozbhearsum: Shouldn't data version be zero here? Or is this test misnamed?

auslib/web/admin/base.py
```diff
@@ -101,17 +101,9 @@ def add_security_headers(response):
...
-app.add_url_rule("/users/<username>/permissions", view_func=PermissionsView.as_view("user_permissions"))
 app.add_url_rule("/users/<username>/permissions/<permission>", view_func=SpecificPermissionView.as_view("specific_permission"))
-app.add_url_rule("/users/<username>/roles", view_func=UserRolesView.as_view("user_roles"))
-app.add_url_rule("/users/<username>/roles/<role>", view_func=UserRoleView.as_view("user_role"))
 # Normal operations (get/update/delete) on rules can be done by id or alias...
-app.add_url_rule("/rules/<id_or_alias>", view_func=SingleRuleView.as_view("rule"))
-app.add_url_rule("/rules/columns/<column>", view_func=SingleRuleColumnView.as_view("rule_columns"))
-# ...but anything to do with history must be done by id, beacuse alias may change over time
-app.add_url_rule("/rules/<int:rule_id>/revisions", view_func=RuleHistoryAPIView.as_view("rules_revisions"))
+# ...but anything to do with history must be done by id, because alias may change over time
```
mozbhearsum: This comment no longer applies here, please remove it.

auslib/web/admin/swagger/api.yaml
```diff
@@ -36,129 +50,418 @@ tags:
...
+
+  roleParam:
+    name: role
+    in: path
+    description: user role to grant access to
+    type: string
+    maxLength: 50
+    required: true
+
+  rule_id_Or_alias_param:
```
mozbhearsum: Nit: "Or" should probably be all lower case.

auslib/web/admin/swagger/api.yaml
```diff
@@ -36,129 +50,418 @@ tags:
...
-# basePath: "/api"
-# host: "localhost:8080"
-# Swagger-UI bug : fix TODO week 13
+  - name: "Releases"
+    description: "Balrog Releases Description"
+    externalDocs:
+      description: "Find out more about Releases"
+      url: "http://mozilla-balrog.readthedocs.io/en/latest/database.html#releases"
+
+parameters:
```
mozbhearsum: What's the difference between this and the "definitions" section? I'm also a bit confused as to why only a few things are in here. Eg: why aren't product and channel defined here as well?

auslib/web/admin/swagger/api.yaml
```diff
@@ -36,129 +50,418 @@ tags:
...
+    name: data_version
+    in: query
+    description: data version of object.
+    type: integer
+    minimum: 1
+    required: true
+
+  roleParam:
+    name: role
+    in: path
```
mozbhearsum: It seems a little bit strange to have "in" and "required" defined outside of a specific endpoint/operation. For example, role is in the path when granting/revoking a role, but in the body when doing signoffs. It it possible to define these parts in the endpoint, and the other parts here?

auslib/web/admin/swagger/api.yaml
```diff
@@ -36,129 +50,418 @@ tags:
...
+              username: balrogadmin
+        '404':
+          description: No permission found for given username
+          schema:
+            $ref: "#/definitions/4xxClientErrorResponseModel"
+          examples:
+            application/json:
+              detail: "No permission found for username"
+              status: 404
+        '403':
```
mozbhearsum: Are you sure this is a valid response? As far as I know we have no restrictions around who can view permissions.

auslib/web/admin/swagger/api.yaml
```diff
@@ -168,49 +471,471 @@ paths:
...
+        '404':
+          description: Requested rule did not exist
+          schema:
+            $ref: '#/definitions/4xxClientErrorResponseModel'
+          examples:
+            application/json:
+              detail: Requested rule does not exist
+              status: 404
+        '400':
+          description: "Bad Request's Response"
```
mozbhearsum: How about something like "Error in form data" for the 400s?

auslib/web/admin/swagger/api.yaml
```diff
@@ -168,49 +471,471 @@ paths:
...
+                format: int32
+                minimum: 0
+                example: 1
+            # limits the extent of validation of column_key array
+            # oneOf isn't supported in swagger 2.0 but will be in 3.0
+            # https://github.com/OAI/OpenAPI-Specification/issues/333
+            additionalProperties:
+              type: array
+            maxProperties: 2
+            minProperties: 2
```
mozbhearsum: Clever solution here!

auslib/web/admin/swagger/api.yaml
```diff
@@ -232,9 +957,182 @@ definitions:
...
+                type: ['object', 'null']
+                description: Options available for the given permission
+                properties:
+                  actions:
+                    type: array
+                    uniqueItems: true
+                    items:
+                      type: string
+                      enum: ["enact"]
+
```
mozbhearsum: It must've been a pain to fully implement all of these permissions schemas, very good work though!

auslib/test/admin/views/test_rules.py
aksareen: input value data_version is incorrect. Changing it to 0

auslib/web/admin/swagger/api.yaml
aksareen: This is a parameters definition object defined by swagger. [Link](http://swagger.io/specification/#parametersDefinitionsObject). Any parameter that can be reused across multiple APIs should be defined here. However, if a query parameter like 'product' that is only used in GET /rules need not be defined globally

auslib/web/admin/swagger/api.yaml
aksareen: body parameters are defined in the definitions section as data models. Path and query parameters are defined in the parameters section. I created roleParam specifically to prevent defining it redundantly multiple times in the multiple user api path parameter. For signoffs, I'll probably define a new model or extend existing one when I start working on that.

auslib/web/admin/swagger/api.yaml
aksareen: It's getting raised on this [line](https://github.com/mozilla/balrog/blob/master/auslib/web/admin/views/permissions.py#L45)

aksareen:
@mozbhearsum . Addressed the above comments in [this commit](https://github.com/mozilla/balrog/pull/317/commits/01fd2e8a79f34010cfd025ccf0d385b113fe9808) . Commits on top of it will now cover the final users/ api i.e. specific permissions view

auslib/web/admin/swagger/api.yaml
mozbhearsum: So if "product" ends up getting used across multiple endpoints it should be moved to this section?

auslib/web/admin/swagger/api.yaml
mozbhearsum: What happens if we have a parameter that is in the path for one endpoint, and in the query for another? And are you sure that body parameters aren't supported here? http://swagger.io/specification/#parameterObject seems to say that "body" is allowed as a value for "in".

auslib/web/admin/swagger/api.yaml
mozbhearsum: Sorry, mea culpa!

auslib/web/admin/swagger/api.yaml
aksareen: if we have a parameter that is in the path for one endpoint, and in the query for another, then in that particular case separate definitions are needed where in first 

> in: path

is defined and second 

> in: query

will be required so that swagger can differentiate between those two parameters having only the same names. The request json is supported and validated via the 'body' parameter only. Inside the body parameter we define a schema (either as standalone or a collection or hierarchy of models). This schema validates the parameters only in the request.json

auslib/web/admin/swagger/api.yaml
aksareen: If it does and we can use the same "description" so as not to confuse the clients then yes.

auslib/web/admin/views/rules.py
```diff
@@ -88,41 +103,20 @@ def _post(self, id_or_alias, transaction, changed_by):
...
-            # same as a modification here.
-            if (request.json and k in request.json) or k in request.form:
-                what[k] = v
+            return problem(status=404, title="Not Found", detail="Requested rule wasn't found",
+                           ext={"exception": "Requested rule does not exist"})
+
+        what, mapping_values = process_rule_form(connexion.request.json)
+
+        if what.get('mapping', None) is not None and len(mapping_values) != 1:
+            return problem(400, 'Bad Request', 'Invalid mapping value. No release name found in DB')
```
mozbhearsum: Can you check fallbackMapping here, too?

PR #319, opened 2017-05-22
Devedition changes

nthomas-mozilla:
I'm not sure if we'll end up using Balrog staging given the VPN situation but the first commit here adds in the staging equivalent of archive.mozilla.org when STAGING is set in the environment (CloudOps would have set that). Also tweaks the SYSTEM_ACCOUNTS to use that.

The second sets up Devedition as a separate product, as an extension of the approach with bouncer, and what Rail has so far on jamun (ie it's submitting with `Data sent: {'product': 'Devedition', ...` in the updates job).

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -34,6 +34,7 @@ services:
       - SMTP_TLS
       - NOTIFY_TO_ADDR
       - NOTIFY_FROM_ADDR
+      - STAGING
     healthcheck:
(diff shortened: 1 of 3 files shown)


nthomas-mozilla:
https://bugzilla.mozilla.org/show_bug.cgi?id=1367304 to set STAGING=1 on our staging instances.

PR #320, opened 2017-05-22
B1325377 - Public API - Part 2

ex-Dev:
PR opened, but is not finalized yet, just to be more easy the review and discuss if the approach to access the admin stuff is the expected.

Endpoints planned to implement:
/rules
/rules/{id_or_alias}
/rules/{id_or_alias}/revisions
/releases
/releases/{release}
/releases/{release}/revisions

There are other one that I forgot?

TODO: 
-Create the public endpoints to releases and unittests;
-Improve the spec (yml)


auslib/web/public/api.yml
```diff
@@ -587,3 +587,210 @@ paths:
...
+      parameters:
+        - name: release
+          in: path
+          required: true
+          type: string
+      responses:
+        200:
+          description: "Release History"
+
+definitions:
```
mozbhearsum: @aksareen has been creating some definitions for Rules as part of #312 and #317. Is there some way we can share definitions between the apps? If not, I have a few comments about these.

auslib/web/public/releases.py
```diff
@@ -0,0 +1,17 @@
+from auslib.web.admin.views.releases import (
+    ReleasesAPIView, SingleReleaseView, ReleaseHistoryView)
```
mozbhearsum: Unfortunately, this isn't a great idea. The admin and public app are siblings, so they should not be importing each others code. I think most of the endpoints are small enough that copying/reimplementing would be OK, but you're also welcome to create a base or helpers methods that implement each one, and use that for both implementations.

auslib/web/public/api.yml
aksareen: AFAIK the model definitions cannot be shared between files even within the same app in connexion. So if we have two yaml spec files within the app, they are independent of each other. Although jsonschema has defined external references to models to be declared in other files according to jsonschema-draft4 but those references need to be parsed/build explicitly by the program before loading up the swagger.json . Connexion does not do any of that right now.

auslib/web/public/api.yml
```diff
@@ -587,3 +587,210 @@ paths:
...
+          type: string
+        - name: names_only
+          in: query
+          required: false
+          type: string
+      responses:
+        200:
+          description: ""
+          schema:
+            $ref: "#/definitions/RulesWrapper"
```
srfraser: Should this be ReleasesWrapper?

auslib/web/public/api.yml
mozbhearsum: Hmm, it looks like you can pass a dict as the "specification" (https://github.com/zalando/connexion/blob/93c06711ed02b453121818f65b25ac3c863aefdd/connexion/apps/abstract.py#L97). I wonder if we can put the definitions in a common file, and manually combine that + the endpoints instead of passing a filename?

auslib/web/public/api.yml
ex-Dev: You right, I opened this PR mainly for validade the access to admin api, from public api. There are a lot of details to be implemented yet.

auslib/web/public/api.yml
ex-Dev: I saw the @aksareen, actually I'll use his yml as a guide by linking docs, provide examples etc. 

Indeed the connexion accepts a dict as parameter, I didn't relize it. We can merge the dicts...I'll try it.

ex-Dev:
@mozbhearsum, @aksareen I'll go back to this bug tomorrow, so yesterday I push a commit with a simple test to try using connexion with a dict, seems there is no problem to do it.

So I created a "spec builder" to allow us reusing yamls, you can see a use example in tests.

What do you think?

aksareen:
Quite impressive since we can have distribution of model across multiple files. Two questions : does pyyaml already have​ a spec builder that we can use instead of writing one if our own ? 

After building all the distributed yanl files can we somehow locally save the final file for debugging purposes every time it is built  ?

ex-Dev:
Dunno if pyyaml does this job, I need research it. My intention with builder was allow composing only to enpoints and definitions.

Yeah, we can save the output spec, by performing yaml.dump().

mozbhearsum:
pyyaml won't have a spec builder - yaml is just a format (like json), it's not anything specific to swagger. I agree that it would be useful to save the final spec somewhere. In fact, I wonder if we should generate it when we start the app or perhaps build the docker images, and use the generated one instead of the one we construct in memory? I agree that having the full version will make debugging easier.

auslib/util/swagger.py
```diff
@@ -0,0 +1,50 @@
...
+        self._main_spec = yaml.load(stream)
+        return self
+
+    def add_spec_part(self, file_name):
+        stream = file(file_name, 'r')
+        self._spec_parts.append(yaml.load(stream))
+        return self
+
+    def build(self):
+        main_endpoints = self._get_dict(self._main_spec, self.endpoints_key)
```
mozbhearsum: I think it might be good to add some safeguards against overwriting something that may exist in more than one spec. It could lead to confusing behaviour if we allow that.

auslib/test/common/test_spec_builder.py
```diff
@@ -0,0 +1,31 @@
...
+
+    def testBuildSpec(self):
+        main_spec = path.join(self.resources_dir, "api.yml")
+        part_1 = path.join(self.resources_dir, "part_1.yml")
+        part_2 = path.join(self.resources_dir, "part_2.yml")
+
+        spec = SpecBuilder().add_main_spec(main_spec)\
+                            .add_spec_part(part_1)\
+                            .add_spec_part(part_2)\
+                            .build()
```
mozbhearsum: I quite like the interface for this, nice work.

auslib/util/swagger.py
```diff
@@ -0,0 +1,50 @@
...
+            content = f.read()
+        return content
+
+    def add_main_spec(self, file_name):
+        stream = file(file_name, 'r')
+        self._main_spec = yaml.load(stream)
+        return self
+
+    def add_spec_part(self, file_name):
+        stream = file(file_name, 'r')
```
mozbhearsum: I'm curious why you decided to separate "main" and "parts" rather than just treat everything as parts. I'm not sure I see the benefit.

auslib/web/public/base.py
```diff
@@ -20,7 +21,12 @@
 }
 
 connexion_app = connexion.App(__name__, specification_dir='.', validator_map=validator_map)
-connexion_app.add_api('api.yml', validate_responses=True, strict_validation=True)
+
+import os
+f = os.path.join(os.path.dirname(__file__), 'api.yml')
+stream = file(f, 'r')
+spec = yaml.load(stream)
+connexion_app.add_api(spec, validate_responses=True, strict_validation=True)
```
mozbhearsum: I assume this is just to verify that a dict works for now, and this will change to SpecBuilder later?

ex-Dev: Yes, this was just to validate before create the SpecBuilder.

auslib/util/swagger.py
ex-Dev: I considered that parts must have only endpoints (/paths) and definitions for simplicity. 
The "Main spec" has others  supported swagger elements and can have endpoints and definitions too.

mozbhearsum: Ah, OK. It sounds like this is a logical separation, but not enforced in any way. Eg: if I happened to put something that wasn't an endpoint or definition in a "part", it would still work I think? Just trying to clarify here.

ex-Dev: Yes, the add_spec_part function would still work, but the build function does not will merges anything to the "paths" and "definitions" dict.
I can put a validation in the add_spec_part function and raise a exception when "paths" nor "definitions" exists.

mozbhearsum: That sounds like a great idea, let's do it!

ex-Dev:
@mozbhearsum 
Reading the Balrog's documentation, I realized some endpoints that maybe makes sense be part of public api too:

https://github.com/mozilla/balrog/blob/master/auslib/web/admin/views/releases.py#L184
https://github.com/mozilla/balrog/blob/master/auslib/web/admin/views/releases.py#L321
https://github.com/mozilla/balrog/blob/master/auslib/web/admin/views/releases.py#L493

What you think?

mozbhearsum:
I don't think exposing read_only or /releases/columns makes sense. The former is only helpful to know if you have write access and the latter is just a helper we use when building the UI. The single locale endpoint is useful though, good idea!

auslib/web/common/rules_spec.yml
```diff
@@ -0,0 +1,124 @@
+paths:
+  /rules:
```
mozbhearsum: I got to thinking about this API a bit more and I think we should make it a bit more defined, and versioned. How about putting all of the new endpoints behind: /api/v1?

ex-Dev: Looks good, I'll working more on balrog next saturday(3).
About the new endpoints implementations living in auslib.web.common, it's ok?

https://github.com/ex-Dev/balrog/blob/38c7cb84489664d5df379c26a46fc8e186a34255/auslib/web/common/rules.py#L13

https://github.com/ex-Dev/balrog/blob/38c7cb84489664d5df379c26a46fc8e186a34255/auslib/web/common/releases.py#L14

mozbhearsum: That seems like a good place to me.

auslib/util/swagger.py
ex-Dev: Only the  main_spec will be updated and returned. For while I'll raise an exception when an item with same key is found, but maybe we can want that an item can be overwritten, in this case, we can warn in log.

ex-Dev:
I pushed some commits on the last weekend, fell free to review what is done. About release blobs, I'm not sure if we must create a response model for all release blob variants, what you think?

We can't set the application basePath to '/v1'. All update endpoints would be under the '/v1' too if we do this. So I'll put statically '/v1/releases'...

auslib/util/swagger.py
mozbhearsum: That sounds fine.

mozbhearsum:
I don't think it's worthwhile doing specs for all the different blob types, unless you can somehow use the ones in https://github.com/mozilla/balrog/tree/master/auslib/blobs/schemas. It's way too much to reimplement otherwise.

ex-Dev:
I set the base path of public api to /v1, all endpoint under the public api will have /v1 in your paths.
The clients that request updates do not able to use /v1/update... right? If it right, I thought to rewrite the url in wsgi file, something like this:

[uwsgi]
http = :$(PORT)
wsgi-file = /app/uwsgi/public.wsgi
route-uri = ^/update/(.*) rewrite:/v1/$1

Currently I don't know how uwsgi routes works, for me this means:
if path starts with /update..., rewrite to /v1/path

But it does not work on my local tests.

mozbhearsum:
I don't think we should be doing any special magic in the routing for /v1. It's important that clients include it in their requests so that when we change the API (ie: create /v2), clients that are explicitly using /v1 do not break.

ex-Dev:
I'm curious, currently the clients have a way to know that we change, for instance:
/update/1/{product}/{version}/{buildID}/{buildTarget}/{locale}/{channel}/update.xml
to
**/v1/**update/1/{product}/{version}/{buildID}/{buildTarget}/{locale}/{channel}/update.xml
?

I can use basePath /v1 for all endpoints with no fear? haha

mozbhearsum:
Oh, we shouldn't do anything with the /update endpoints. Those are fine as is (and are already versioned in the second section of the URL). I'm only talking about putting the *new* /rules and /releases endpoints behind a /v1/ version.

ex-Dev:
Yeah, that's the problem. Basepath is for all app endpoints. Due to this I thought to rewrite all urls that start with /update to /v1/update internally.
The other option is write /v1 for all non "update" paths, but does not seems a good idea.

mozbhearsum:
What's wrong with putting /v1 in all of the non-update endpoints? It seems like an OK solution to me.

ex-Dev:
Indeed there is no problem , just when swagger "basePath" is set to /v1 (for instance), we versioned on api level. If /v1 was set as /v1/rules, /v1/releases, we versioned on endpoint level. But ok I'll put /v1 for the non-update endpoints.

ex-Dev:
@mozbhearsum, I think that the final review can start.
As the last weeks was busy, I code a little through the days, so maybe I forgot something :)

auslib/test/common/test_releases.py
```diff
@@ -0,0 +1,61 @@
+import json
+from auslib.test.common.base import CommonTestBase
+
+
+class TestPublicReleasesAPI(CommonTestBase):
+    def test_get_releases(self):
+        ret = self.public_client.get("/api/v1/releases")
+        got = json.loads(ret.data)
+        self.assertEquals(len(got["releases"]), 3)
+        self.assertIsInstance(got["releases"][0], dict)
```
mozbhearsum: You don't need to fully verify the blobs, but verifying a few things on it instead of just the length and type would be good. In particular, the name, schema_version, and somthing in the platform and/or locale section. This will make the tests stronger, and guard against false positives if some of the underlying test data changes. Please do this for all tests that contain blobs.

auslib/test/common/test_rules.py
```diff
@@ -0,0 +1,69 @@
+import json
+from auslib.test.common.base import CommonTestBase
+
+
+class TestPublicRulesAPI(CommonTestBase):
+    def test_get_rules(self):
+        ret = self.public_client.get("/api/v1/rules")
+        got = json.loads(ret.data)
+        self.assertEquals(got["count"], 3)
```
mozbhearsum: Similar thing here - please verify at least some parts of each rule returned.

auslib/test/common/base.py
```diff
@@ -0,0 +1,123 @@
...
+
+        admin_app.config["WTF_CSRF_ENABLED"] = False
+        self.admin_client = admin_app.test_client()
+
+        dbo.setDb('sqlite:///:memory:')
+        dbo.create()
+        dbo.setDomainWhitelist({'a.com': ('q')})
+        dbo.permissions.t.insert().execute(permission='admin', username='bill', data_version=1)
+        dbo.rules.t.insert().execute(priority=90, backgroundRate=100, mapping='b', update_type='minor', product='Fennec',
+                                     data_version=1, alias="moz-releng")
```
mozbhearsum: When setting up test data like this, it would be best to set rule_id yourself instead of relying on autoincrement.

auslib/test/common/test_rules.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def test_get_rules_by_product(self):
+        product = "Fennec"
+        ret = self.public_client.get("/api/v1/rules?product={}".format(product))
+        self.assertEqual(ret.status_code, 200, ret.data)
+        got = json.loads(ret.data)
+        self.assertTrue(got, "Empty dict")
```
mozbhearsum: Can you use a more verbose error message here? Something that includes both the inputs and outputs. Eg: "No rules returned when querying for product X".

auslib/test/common/test_rules.py
```diff
@@ -0,0 +1,69 @@
...
+                        update_type="minor", product="Fennec", data_version=1,
+                        alias="moz-releng", buildID=None, buildTarget=None,
+                        channel=None, comment=None, distVersion=None,
+                        distribution=None, fallbackMapping=None,
+                        headerArchitecture=None, locale=None, version=None,
+                        systemCapabilities=None, osVersion=None)
+        ret = self.public_client.get("/api/v1/rules/moz-releng")
+        self.assertEqual(ret.status_code, 200, ret.data)
+        got = json.loads(ret.data)
+        del got["rule_id"]
```
mozbhearsum: This is not the right way to do this :). You should include rule_id in "expected" instead.

auslib/test/common/base.py
mozbhearsum: When setting up test data like this, it would be best to set rule_id yourself instead of relying on autoincrement.

auslib/util/swagger.py
```diff
@@ -0,0 +1,119 @@
+import yaml
+import auslib
+
+from os import path
+
+AUS_DIR = path.dirname(auslib.__file__)
+
+
+class SpecBuilder():
```
mozbhearsum: This class is really great overall, but I think it would be good if you kept all Balrog-specific knowledge out of it. Concretely, this means that consumers should be passing full filepaths (or already read files) to it, and the version should be passed explicitly instead of reference to a version file.

auslib/util/swagger.py
```diff
@@ -0,0 +1,119 @@
...
+            elif isinstance(v, dict):
+                self._get_dict(v, key)
+        return {}
+
+    def _load_spec(self, file_name):
+        with file(file_name, 'r') as spec_file:
+            return yaml.load(spec_file)
+
+    def _try_change_version(self, info_dict):
+        version_key = 'version'
```
mozbhearsum: What is the purpose of this method?

auslib/util/swagger.py
```diff
@@ -0,0 +1,119 @@
...
+        if info_dict and version_key in info_dict:
+            version = self._get_version()
+            if version:
+                info_dict[version_key] = version
+
+    def add_main_spec(self, file_name):
+        """
+        Adds the current application specification file.
+
```
mozbhearsum: If I remember correctly, the "main" spec is required to have a bunch of top level keys that won't be present in the spec "parts". If that's the case, please enhance this docstring to reflect that.

auslib/util/swagger.py
```diff
@@ -0,0 +1,119 @@
...
+AUS_DIR = path.dirname(auslib.__file__)
+
+
+class SpecBuilder():
+    def __init__(self,
+                 app,
+                 endpoints_key="paths",
+                 definitions_key="definitions",
+                 parameters_key="parameters",
```
mozbhearsum: Are these really something you want consumers to be able to override? If I set the endpoints_key to "my_endpoints", would the spec still work? If they truly should be overridable, they belong as parameters to add_spec_part. If not, they should be class attributes.

auslib/util/swagger.py
```diff
@@ -0,0 +1,119 @@
...
+                self.endpoints_key not in spec_part and
+                self.parameters_key not in spec_part):
+            raise ValueError("Mergeable swagger keys was not found in: {}".format(file_name))
+
+        self._spec_parts.append(spec_part)
+        return self
+
+    def _test_merge(self, main_dict, part_dict):
+        for k, v in part_dict.items():
+            assert k not in main_dict, "Key {} already exists in spec file.".format(k)
```
mozbhearsum: Good job checking for this. You should probably raise a ValueError instead, though.

auslib/util/swagger.py
```diff
@@ -0,0 +1,119 @@
...
+            self._test_merge(main_definitions, part_definitions)
+            main_definitions.update(part_definitions)
+
+    def _update_parameters_definitions(self, main_parameters, spec_part):
+        if self.parameters_key in spec_part:
+            if not main_parameters:
+                self._main_spec[self.parameters_key] = main_parameters
+            part_parameters = spec_part[self.parameters_key]
+            self._test_merge(main_parameters, part_parameters)
+            main_parameters.update(part_parameters)
```
mozbhearsum: Have you considered writing a helper to factor out the common logic of these three methods? Not a blocker or requirement by any means, just a thought.

auslib/util/swagger.py
```diff
@@ -0,0 +1,119 @@
...
+    def build(self):
+        self._try_change_version(self._get_dict(self._main_spec, 'info'))
+        main_endpoints = self._get_dict(self._main_spec, self.endpoints_key)
+        main_definitions = self._get_dict(self._main_spec, self.definitions_key)
+        main_parameters = self._get_dict(self._main_spec, self.parameters_key)
+        for spec_part in self._spec_parts:
+            self._update_endpoints(main_endpoints, spec_part)
+            self._update_definitions(main_definitions, spec_part)
+            self._update_parameters_definitions(main_parameters, spec_part)
+        return self._main_spec
```
mozbhearsum: It's a bit strange to be returning this when so many other methods return "self". I wonder if it would be better to make this class a subclass of dict so you could return self here as well? It seems like a Spec is really just a dict with some extra helpers to make construction easier.

auslib/web/common/history.py
```diff
@@ -0,0 +1,107 @@
...
+            self.obj_not_found_msg = obj_not_found_msg
+        return self
+
+    def with_history_filters_callback(self, history_filters_callback):
+        self.fn_history_filters = history_filters_callback
+        return self
+
+    def with_process_revisions_callback(self, process_revisions_callback):
+        self.fn_process_revisions = process_revisions_callback
+        return self
```
mozbhearsum: I'm not sure I see the benefit to these methods. Is there a reason you don't just pass the callbacks directly to get_history instead?

auslib/web/common/releases_spec.yml
```diff
@@ -0,0 +1,129 @@
...
+        - name: page
+          in: query
+          required: false
+          type: integer
+          format: int32
+        - name: limit
+          in: query
+          required: false
+          type: integer
+          format: int32
```
mozbhearsum: Is there a reason page and limit are defined here instead of in parameters?

auslib/web/common/releases_spec.yml
```diff
@@ -0,0 +1,129 @@
...
+      produces:
+        - application/json
+      parameters:
+        - $ref: '#/parameters/releaseParam'
+        - name: platform
+          description: Platform
+          in: path
+          required: true
+          type: string
+        - $ref: '#/parameters/localeParam'
```
mozbhearsum: I don't think localeParam is right here. This endpoint returns a JSON data structure with details about a locale's update, not a simple locale string.

auslib/web/common/releases_spec.yml
```diff
@@ -0,0 +1,129 @@
...
+      parameters:
+        - $ref: '#/parameters/releaseParam'
+        - name: pretty
+          description: If "pretty" is present in the query string and set to true, it will be pretty formatted.
+          in: query
+          required: false
+          type: integer
+      responses:
+        200:
+          description: Release Blob
```
mozbhearsum: X-Data-Version should be in the headers here, too.

auslib/web/common/releases_spec.yml
```diff
@@ -0,0 +1,129 @@
...
+        - application/json
+      parameters:
+        - $ref: '#/parameters/releaseParam'
+        - name: pretty
+          description: If "pretty" is present in the query string and set to true, it will be pretty formatted.
+          in: query
+          required: false
+          type: integer
+      responses:
+        200:
```
mozbhearsum: 404 is also a valid response.

auslib/web/common/releases_spec.yml
```diff
@@ -0,0 +1,129 @@
...
+    type: object
+    properties:
+      names:
+        type: array
+        items:
+          type: string
+      releases:
+        type: array
+        items:
+          $ref: '#/definitions/Release'
```
mozbhearsum: Is there a reason that the "Release" definition exists when this is the only reference to it?

auslib/web/common/rules_spec.yml
```diff
@@ -0,0 +1,196 @@
...
+            $ref: '#/definitions/RuleHistoryWrapper'
+        400:
+          description: Bad input.
+
+definitions:
+  Rule:
+    type: object
+    required:
+      - rule_id
+      - update_type
```
mozbhearsum: priroity and backgroundRate are also required.

auslib/web/public/api.yml
```diff
@@ -1,45 +1,35 @@
 swagger: '2.0'
 info:
+  version: '2.30'
   title: Balrog Public API
-  description: Balrog Application Update Service
-  version: "1.0.0"
+  termsOfService: https://wiki.mozilla.org/Balrog
```
mozbhearsum: No need to include this, there are no terms of service.

auslib/web/public/api.yml
```diff
@@ -582,8 +320,94 @@ paths:
...
     get:
-      operationId: "auslib.web.public.dockerflow.version"
+      operationId: 'auslib.web.public.dockerflow.version'
       description: Respond to /__version__ with the contents of /app/version.json.
       responses:
         200:
           description: version.json content.
+
+parameters:
+  productParam:
```
mozbhearsum: Minor nit: There's no need to include "Param" at the end of each of these. Anytime they are referenced, they are proceeded by "paramaters" so it ends up being redundant.

auslib/util/swagger.py
ex-Dev: This method try get the Balrog version from version file, if he is present.

auslib/web/common/history.py
```diff
@@ -0,0 +1,107 @@
+import json
```
ex-Dev: Yes, this was my understand, that admin will reuse some endpoints. Is not right?

auslib/test/common/base.py
```diff
@@ -0,0 +1,123 @@
...
+    @classmethod
+    def tearDownClass(cls):
+        app.error_handler_spec = cls.error_spec
+
+    def setUp(self):
+        app.config['DEBUG'] = True
+        self.public_client = app.test_client()
+
+        admin_app.config["WTF_CSRF_ENABLED"] = False
+        self.admin_client = admin_app.test_client()
```
ex-Dev: Indeed, my intention was test exactly like admin, to ensure that all is ok.
I'll refactor all test cases.

auslib/web/common/history.py
ex-Dev: Nothing special, normally I create helpers and builders fluent. I'll move it.

auslib/web/common/releases_spec.yml
ex-Dev: Maybe I did not understand what you mean, but Yes, this endpoint returns exactly this. 
As in: https://github.com/mozilla/balrog/blob/master/auslib/web/admin/views/releases.py#L191

auslib/web/common/releases_spec.yml
ex-Dev: Because the other endpoints not use this parameter. But have no restrictions to put in parameters.

auslib/web/common/history.py
mozbhearsum: I think that makes a lot of sense, I was just making sure I understood :). It's probably worth adding a comment and/or filing a follow-up bug about converting it (better not to do it here - this PR is big enough already).

auslib/web/common/releases_spec.yml
mozbhearsum: Oh, of course. For some reason I had it in my head that we used these across all history endpoints...

auslib/web/common/releases_spec.yml
mozbhearsum: I think I got confused between the input parameters and the response. Carry on!

auslib/web/common/releases_spec.yml
ex-Dev: I checked it, you right, page and limit are in history of rules_spec and releases_spec. Maybe at the end of Public and Admin API connexion migration, we can have a better view that what is common through endpoints, and define a way to organize that to avoid lead us to a high granularity.

ex-Dev:
@mozbhearsum, 
Up to now, there is no conflict with master, let me know when releases and rules from admin api is ready to merge.

aksareen:
@ex-Dev : rules and release admin APIs have been merged with the master branch. Here's the merged PR [link](https://github.com/mozilla/balrog/pull/336)

ex-Dev:
Ah, nice! I did not follow the last changes. Thanks @aksareen!
So I wait for review.

ex-Dev:
Sure! Should I file a bug on bugzilla for this or just submit a patch here?

mozbhearsum:
Filing a new bug is probably the best thing to do.

ex-Dev:
Bug created https://bugzilla.mozilla.org/show_bug.cgi?id=1376304, but I can't assign it to me.

PR #321, opened 2017-05-24
Fix bug 1363058 - don't send empty strings for null values when a field's value is removed

harikishen:
I was considering something like this for https://bugzilla.mozilla.org/show_bug.cgi?id=1363058  @aksareen @mozbhearsum

### DIFF
--- a/ui/app/js/services/rules_service.js
+++ b/ui/app/js/services/rules_service.js
@@ -1,4 +1,4 @@
-angular.module("app").factory('Rules', function($http, ScheduledChanges) {
+angular.module("app").factory('Rules', function($http, ScheduledChanges, Helpers) {
   // these routes map to stubbed API endpoints in config/server.js
   var service = {
     getRules: function() {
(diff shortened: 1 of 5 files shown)


ui/app/js/services/rules_service.js
```diff
@@ -19,6 +19,13 @@ angular.module("app").factory('Rules', function($http, ScheduledChanges) {
     },
     updateRule: function(id, data, csrf_token) {
       data.csrf_token = csrf_token;
+      for (var i in data) {
+        if (data.hasOwnProperty(i)) {
+          if (data[i] === "") {
```
mozbhearsum: What will happen if data[i] is not a string? As far as I know we don't have any nullable fields that aren't strings, but it would be good to make sure this doesn't cause errors integers or objects. The approach looks fine otherwise though.

harikishen:
@mozbhearsum I've added the check for strings and updated all objects.

harikishen:
@mozbhearsum where do you usually place the helper functions ???

PR #323, opened 2017-05-29
Bug 1336452 : Specific users view + Release History Diffs & View API migration

aksareen:


auslib/web/admin/swagger/api.yaml
```diff
@@ -108,6 +131,110 @@ parameters:
...
+              example: 1
+            options:
+              type: ["string", "null"]
+              format: JSONStringField
+              description: JSON field specifying the actions that user can perform and the prdouct on which it has current permission to do so
+              example: "{'actions':['modify']}"
+          # accomodate 'options_as_json' and 'permission' which are duplicate values of 'options'
+          # and 'permission path parameter' respectively, getting passed in request body
+          # No need to validate the duplicate values again
+          additionalProperties: true
```
mozbhearsum: This sounds like another case of the UI passing things it shouldn't. The backend doesn't look for either of these things in the form (as far as I know), so we shouldn't allow them to be passed. It makes the interface confusing and validation less useful.

auslib/web/admin/swagger/api.yaml
```diff
@@ -108,6 +131,110 @@ parameters:
...
+      allOf:
+        - $ref: '#/definitions/CSRFModel'
+        - type: object
+          properties:
+            data_version:
+              type: integer
+              format: int32
+              description: data_version of permission
+              minimum: 1
+              example: 1
```
mozbhearsum: I think data_version and options both need to be set to required: true.

auslib/web/admin/views/permissions.py
```diff
@@ -66,67 +65,79 @@ def get(self, username, permission):
...
-                    self.log.warning("Bad input: %s", form.errors)
-                    return Response(status=400, response=json.dumps(form.errors))
-                dbo.permissions.update(where={"username": username, "permission": permission}, what={"options": form.options.data},
-                                       changed_by=changed_by, old_data_version=form.data_version.data, transaction=transaction)
-                new_data_version = dbo.permissions.getPermission(username=username, permission=permission, transaction=transaction)['data_version']
+                # Existing Permission
+                if not connexion.request.json.get("data_version"):
+                    return problem(400, "Bad Request", "'data_version' is missing from request body")
+
+                options_dict = None if not connexion.request.json.get("options") else json.loads(
```
mozbhearsum: Can you split this (and the subsequent similar code) into multiple lines? Something like:
```
options_dict = None
if connexion.request.json.get("options"):
    options_dict = json.loads(...)
```

...is much easier to read.

auslib/web/admin/swagger/api.yaml
```diff
@@ -108,6 +131,110 @@ parameters:
     minimum: 0
     required: true
 
+  change_id_param:
+    name: change_id
+    in: path
+    description: Unique id to identify the identify the change
+    type: string
```
aksareen: Nope. [Here](https://github.com/mozilla/balrog/blob/master/auslib/web/admin/base.py#L116) it is defined as string type path parameter. For setting it as integer type we need to add int:change_id. Example for sc_id, [link](https://github.com/mozilla/balrog/blob/master/auslib/web/admin/base.py#L119)

auslib/web/admin/swagger/api.yaml
aksareen: No , data_version is only needed when updating an existing permission. [Link](https://github.com/mozilla/balrog/blob/master/auslib/web/admin/views/permissions.py#L76) whereas when creating a new permission it isn't required. [Link](https://github.com/mozilla/balrog/blob/master/auslib/web/admin/views/permissions.py#L85)

auslib/web/admin/swagger/api.yaml
mozbhearsum: That seems like a bug. It's definitely an integer in the database (https://github.com/mozilla/balrog/blob/2a245b6e25ff38a4b896fb6965d66d884bb9d79a/auslib/db.py#L629), and it's returned by the /rules/:rule_id/revisions endpoint as an integer. Can you make this an integer, too?

aksareen: surely

auslib/web/admin/swagger/api.yaml
aksareen: I was unable to debug the issue on UI. AFAIK , the duplicate value is getting set [here](https://github.com/mozilla/balrog/blob/master/ui/app/js/controllers/user_edit_controller.js#L249) but in next line the API is invoked. Also here's the definition of permission i.e. data object. [link](https://github.com/mozilla/balrog/blob/master/ui/app/js/controllers/user_edit_controller.js#L139) , it defines options_as_json and permission so that they can be displayed on the UI after adding or updating the permissions. Example: [link](https://github.com/mozilla/balrog/blob/master/ui/app/templates/permissions_modal.html#L52) , [link2](https://github.com/mozilla/balrog/blob/master/ui/app/templates/permissions_modal.html#L56). Not sure how to fix this issue on the UI

auslib/web/admin/swagger/api.yaml
```diff
@@ -108,6 +131,110 @@ parameters:
...
+              description: JSON field specifying the actions that user can perform and the prdouct on which it has current permission to do so
+              example: "{'actions':['modify']}"
+          # accomodate 'options_as_json' and 'permission' which are duplicate values of 'options'
+          # and 'permission path parameter' respectively, getting passed in request body
+          # No need to validate the duplicate values again
+          additionalProperties: true
+
+
+responses:
+  updateExistingPermission:
```
aksareen: Yeah, the descriptions and minimum is unique for updating the existing permissions. Specifically the description part so as not to confuse the client when they read the docs.

auslib/web/admin/swagger/api.yaml
```diff
@@ -108,6 +131,110 @@ parameters:
...
+    examples:
+      application/json:
+        detail: "Requested resource was not found"
+        status: 404
+
+  defaultResponse:
+    description: Default case for sporadic exceptions. Example, No remote username and password in basic authentication for posting new data.
+    examples:
+      application/json:
+        Error: "REMOTE_USERNAME required"
```
aksareen: the idea was to only define those 400x responses which are explicitly thrown from within the API methods. [Example](https://github.com/mozilla/balrog/blob/master/auslib/web/admin/views/rules.py#L98) , The handle general exceptions and required login decorator apply for all put,post and delete operations. To make the specification flexible enough so that if any new decorator is added in future returning exception in any format, we don't have to edit every API specification I thought it would be best to have a default case rather than specify 401,403 , 400 and 404 as it could confuse clients. Also ChangeScheduledError, UpdateMergeError and OutdatedDataError all throw 400 error with different exception messages , hence having only one specific example could be a bit misleading in the swagger specification.

auslib/web/admin/views/permissions.py
aksareen: surely

auslib/web/admin/swagger/api.yaml
mozbhearsum: I think the key here is making sure that only the required parameters are passed to the Permissions service. Eg: remove options_as_json from "permission" before passing it along.

auslib/web/admin/swagger/api.yaml
mozbhearsum: I think the part that confuses me is that I can't think of any other errors that would be caught by this other than the username/password one. Do you have another example?

auslib/web/admin/swagger/api.yaml
aksareen: Two cases.
1. I added a delete permission.options_as_json statement on this [line](https://github.com/mozilla/balrog/blob/master/ui/app/js/controllers/user_edit_controller.js#L249) to prevent the key being sent in request. It did not work.

2. Also, when adding a new permission the same line has already been commented out [here](https://github.com/mozilla/balrog/blob/master/ui/app/js/controllers/user_edit_controller.js#L129). Still permission and options_as_json can be seen getting passed in request body.

auslib/web/admin/swagger/api.yaml
aksareen: In updating an existing rule, if the data_version is outdated [here](https://github.com/mozilla/balrog/blob/master/auslib/web/admin/views/rules.py#L128) , then OutDatedDataVersion Error of 400 will be automatically handled [here](https://github.com/mozilla/balrog/blob/master/auslib/web/admin/views/base.py#L25). Same case for PermissionDenied 403 error. In future if any new HTTP exception code is added/modified or deleted, then we don't have to edit the entire schema every time.

auslib/web/admin/swagger/api.yaml
mozbhearsum: I'm still not sure I understand why this needs to be specific to permissions. You're using generic terms for parameters like change_id, why not here? If you change the descirption to say something like "data version of the updated object", this can be re-used for rule, release, scheduled change, etc. updates.

auslib/web/admin/swagger/api.yaml
```diff
@@ -108,6 +131,111 @@ parameters:
...
+              example: "{'actions':['modify']}"
+          # accomodate 'options_as_json' and 'permission' which are duplicate values of 'options'
+          # and 'permission path parameter' respectively, getting passed in request body
+          # No need to validate the duplicate values again
+          additionalProperties: true
+
+
+responses:
+  updateExistingPermission:
+    description: "If the user already has the given permission, update its options and the data_version"
```
mozbhearsum: I don't think this is a fitting description for a response, this seems like it would be more appropriate for a request decription?

auslib/web/admin/swagger/api.yaml
mozbhearsum: This parameter gets used by /users/{username}/permissions/{permission}, which only handle updates. Does this mean that data_version is not going to be checked for that endpoint?

auslib/web/admin/swagger/api.yaml
mozbhearsum: Given that you're only replicating existing behaviour, I'm not going to block on it.

auslib/web/admin/swagger/api.yaml
mozbhearsum: Can you clarify this?

auslib/web/admin/swagger/api.yaml
mozbhearsum: I still don't think we're talking about the same thing here. I fully agree that this default handler makes sense, I just don't think the "example" that talks about username/password is relevant. How about just using "Unhandled error" as the description?

auslib/web/admin/swagger/api.yaml
mozbhearsum: Can you address this?

auslib/web/admin/swagger/api.yaml
aksareen: It is going to be checked explicitly in the method body.
**Reason**: considering the PUT operation's [method](https://github.com/aksareen/balrog/blob/user_then_releases/auslib/web/admin/views/permissions.py#L73),  it handles two cases: First whether the input permission [already exists](https://github.com/aksareen/balrog/blob/user_then_releases/auslib/web/admin/views/permissions.py#L77) (if yes then update it using data_version as required parameter). If it [does not](https://github.com/aksareen/balrog/blob/user_then_releases/auslib/web/admin/views/permissions.py#L92) then it create a new permission (data_version isn't required in the input).

auslib/web/admin/swagger/api.yaml
aksareen: Already addressed [here](https://github.com/aksareen/balrog/blob/releases/auslib/web/admin/swagger/api.yaml#L250) in release branch, forgot to update the comment

auslib/web/admin/swagger/api.yaml
aksareen: As per last week's discussion, there's no need to define the 'default' case at all. Removing all the default responses

auslib/web/admin/swagger/api.yaml
aksareen: already addressed [here](https://github.com/aksareen/balrog/blob/releases/auslib/web/admin/swagger/api.yaml#L251) in release branch

auslib/web/admin/swagger/api.yaml
mozbhearsum: OK, this is fine.

mozbhearsum:
All of the required changes here ended up in #336 by accident. That will end up merging at the same time, so I'm r+'ing this one.

PR #324, opened 2017-05-30
Fix bug #1346212 - Cannot revert a rule change: Cannot revert a rule change: https://aus4-admin.mozilla.org/api/rules/undefined/revisions

harikishen:
@mozbhearsum The issue was with the parameter being passed to [Rules.revertRule()](https://github.com/harikishen/balrog/blob/ticket_1346212/ui/app/js/services/rules_service.js#L36-L40). I replaced the parameter `$scope.rule.id` with `$scope.rule.rule_id` in [rule_revert_controller.js](https://github.com/harikishen/balrog/blob/ticket_1346212/ui/app/js/controllers/rule_revert_controller.js#L11).

### DIFF
--- a/ui/app/js/controllers/rule_revert_controller.js
+++ b/ui/app/js/controllers/rule_revert_controller.js
@@ -10,12 +10,13 @@ function ($scope, $modalInstance, CSRF, Rules, revision) {
     .then(function(csrf_token) {
       Rules.revertRule($scope.rule.rule_id, $scope.rule.change_id, csrf_token)
       .success(function(response) {
-        $scope.saving = false;
         $modalInstance.close();
       })
       .error(function() {
-        $scope.saving = false;
         console.error(arguments);
+      })
+      .finally(function() {
+        $scope.saving = false;
       });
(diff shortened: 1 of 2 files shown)


harikishen:
@mozbhearsum the tests seem to fail but the feature works. Can you help me out here ?
this is the error https://pastebin.mozilla.org/9023044

ui/app/js/controllers/rule_revert_controller.js
```diff
@@ -8,14 +8,15 @@ function ($scope, $modalInstance, CSRF, Rules, revision) {
...
       .success(function(response) {
-        $scope.saving = false;
         $modalInstance.close();
       })
       .error(function() {
-        $scope.saving = false;
         console.error(arguments);
+      })
+      .finally(function() {
+        $scope.saving = false;
```
mozbhearsum: Good idea moving that here.

harikishen: @mozbhearsum thanks

PR #327, opened 2017-05-31
Bg1276289 reenable coveralls

andela-bolajide:
## What Does this PR do?

Currently, coverage results are not sent to coveralls. This PR re-enables it by writing a script to manually send coverage data after successful test completion.

First, coveralls is to be added to the requirements.txt so as to install it when building the docker image for testing, then the repo token is gotten via the curl command from a secret url and saved to .coveralls.yml so coveralls can get it. After all this the coveralls command is then run.

## Bug Reference

https://bugzilla.mozilla.org/show_bug.cgi?id=1276289


scripts/run.sh
```diff
@@ -72,15 +72,15 @@ elif [ $1 == "test" ]; then
         echo
 
         if [[ $backend_rc == 0 && $frontend_rc == 0 ]]; then
-            echo "All tests pass!!!"
-            if [[ $GITHUB_HEAD_REPO_URL == "https://github.com/mozilla/balrog.git" ]];
+            if [ '$GITHUB_HEAD_REPO_URL' == "https://github.com/mozilla/balrog.git" ];
```
mozbhearsum: I don't think this is going to work...single quotes prevent variable expansion.

andela-bolajide: I'll work something out soon.
I'm surprised as the previous way worked on my PC locally. I'm trying to reproduce it again using some conditions.

mozbhearsum: Huh, I'd be curious to see this. I can't make it work:
```
bhearsum@dib:~/repos/balrog$ export GITHUB_HEAD_REPO_URL='abc'
bhearsum@dib:~/repos/balrog$ if [ "$GITHUB_HEAD_REPO_URL" == "abc" ]; then echo "yes"; fi
yes
bhearsum@dib:~/repos/balrog$ if [ '$GITHUB_HEAD_REPO_URL' == "abc" ]; then echo "yes"; fi
bhearsum@dib:~/repos/balrog$ 
```

andela-bolajide: Can we have a pairing session during the week?

mozbhearsum: Sure, could you e-mail me with your availability?

andela-bolajide: Hey, do you think we can use another variable as a condition? Probably we define a variable in the taskcluster environment and check for that to know whether to send coverage or not.

andela-bolajide: If we set a new variable on taskcluster that the condition can be based on, i think that should work. What do you think @mozbhearsum ?

mozbhearsum: I don't think that's going to make a difference, but we can try.

PR #331, opened 2017-06-01
Backout patch that broke /rules/:rule API.

mozbhearsum:


ex-Dev:
@mozbhearsum, Is this directly related with new connexion impl?

mozbhearsum:
Kindof. The latest patch to the admin swagger api broken one of our submission scripts (because it doesn't json-encode the data). Nothing you need to worry about, I think.

ex-Dev:
Ah, OK, just to keep me informed for future issues.

PR #336, opened 2017-06-12
Bug 1336452 - Release APIs Migration

aksareen:
Temp PR to get the first review of changes made during migration of /release APIs. 

**NOTE:** Check [this commit](https://github.com/mozilla/balrog/commit/ac66c753202a69607b211cd7574012e043227e6b) to view the subset of changes for Release APIs only.

### DIFF
--- a/auslib/test/admin/views/test_releases.py
+++ b/auslib/test/admin/views/test_releases.py
@@ -414,7 +414,7 @@ class TestReleasesAPI_JSON(ViewTest):
         self.assertStatusCode(ret, 400)
 
     def testDeleteNonExistentRelease(self):
-        ret = self._delete("/releases/ueo")
+        ret = self._delete("/releases/ueo", qs=dict(data_version=1))
         self.assertStatusCode(ret, 404)
 
     def testDeleteWithoutPermission(self):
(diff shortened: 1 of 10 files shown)


auslib/web/admin/swagger/api.yaml
```diff
@@ -108,6 +187,173 @@ parameters:
...
+          format: int32
+          minimum: 0
+          example: 1
+      # limits the extent of validation of column_key array
+      # oneOf isn't supported in swagger 2.0 but will be in 3.0
+      # https://github.com/OAI/OpenAPI-Specification/issues/333
+      additionalProperties:
+        type: array
+      maxProperties: 2
+      minProperties: 2
```
aksareen: surely

auslib/web/admin/swagger/api.yaml
```diff
@@ -133,6 +379,492 @@ paths:
...
+            allOf:
+              - $ref: '#/definitions/CSRFModel'
+              - $ref: '#/definitions/DataVersionModel'
+              - type: object
+                description: read_only boolean flag
+                required:
+                  - read_only
+                properties:
+                  read_only:
+                    type: ["boolean", "string"]
```
aksareen: the input request json always contains the read_only key when submitted from the UI due to this [statement](https://github.com/aksareen/balrog/blob/f6e7c578b348c5a0d01f8b9f561603a8b3e8f9c1/ui/app/js/controllers/release_read_only_controller.js#L10), hence the blank string '' is treated as type: string instead of type: null.

auslib/web/admin/swagger/api.yaml
```diff
@@ -969,6 +1606,148 @@ definitions:
         minimum: 1
         example: 3
 
+  ReleaseBase:
```
aksareen: thanks :)

auslib/test/admin/views/test_releases.py
```diff
@@ -766,7 +766,15 @@ def testLocaleGet(self):
...
-        ret = self.client.put('/releases/d/builds/p/d', data=dict(product='a'))
+        data = json.dumps({
+            "complete": {
+                "filesize": 435,
+                "from": "*",
+                "hashValue": "abc",
+            }
+        })
+        inp_data = dict(csrf_token="lorem", data=data, product='d', data_version=1, schema_version=1)
+        ret = self.client.put('/releases/d/builds/p/d', data=json.dumps(inp_data), content_type="application/json")
```
aksareen: thanks :)

auslib/web/admin/swagger/api.yaml
mozbhearsum: Ah, OK. Another thing we should fix sometime later...can you add a TODO about this?

PR #342, opened 2017-06-22
Bug 1336452: Required Signoffs

aksareen:


auslib/web/admin/swagger/api.yaml
```diff
@@ -1597,74 +1963,79 @@ definitions:
...
+        maxLength: 50
+        example: releng
 
+  ChannelModel:
+    title: Channel
+    description: "model to define 'channel' field in request body."
+    type: object
+    properties:
+      channel:
+        description: The update channel of the application request an update.
```
mozbhearsum: Same thing here - there's no "application requesting an update" in this context.

auslib/web/admin/swagger/api.yaml
```diff
@@ -126,6 +132,33 @@ parameters:
...
+    description: roles are used when signing off on Scheduled Changes
+    type: string
+    minLength: 1
+    maxLength: 50
+    required: true
+
+  channelRequiredParam:
+    name: channel
+    in: query
+    description: The update channel of the application request an update.
```
aksareen: I borrowed the definition from the [database model in docs](http://mozilla-balrog.readthedocs.io/en/latest/database.html). What should I change it to ?

auslib/web/admin/swagger/api.yaml
```diff
@@ -841,7 +874,36 @@ paths:
                 type: array
                 minItems: 0
                 items:
-                  $ref: '#/definitions/ReleaseRevision'
+                  type: object
```
aksareen: Previously I had defined 'ReleaseRevision' model to define the object as element for every item in revisions array above. However, I realized that the ReleaseRevision model won't re-used at all in any other API, hence I moved it out from definitions.

auslib/web/admin/swagger/api.yaml
```diff
@@ -1597,74 +1963,79 @@ definitions:
 
   DataVersionModel:
     title: Data Version
-    description: version number of object
+    description: "model to re-use 'data_version' field."
```
aksareen: How about concat of both "model to re-use 'data_version' field. Data_version is the version  number of object" ?

auslib/web/admin/swagger/api.yaml
```diff
@@ -1597,74 +1963,79 @@ definitions:
...
-  ReleaseBase:
-    title: Release Base
-    description: "To Balrog, a 'release' is data about a related set of builds. This does _not_ match up with the concept of a 'release' being on the 'beta', 'release' or 'esr' channel elsewhere. In Balrog, each set of nightlies on any branch is considered a release"
+  SignoffsRequiredModel:
+    title: Signoffs Required
+    description: "model to re-use 'signoffs_required' field."
+    type: object
+    properties:
+      signoffs_required:
+        description: minimum number of signoffs required to execute the change
```
aksareen: this model only defines the signoffs_required integer field defined in the [forms](https://github.com/mozilla/balrog/blob/eb35f9938d327719edc2fb330edc9476378caf41/auslib/web/admin/views/forms.py#L381) . Am not entirely getting your point above.

auslib/web/admin/swagger/api.yaml
aksareen: that's the same definition of channel in the docs as above. What should I change this to ?

auslib/web/admin/swagger/api.yaml
mozbhearsum: OK. In the future, this sort of thing would be better as its own PR, as the scope here was supposed to be just required signoffs.

auslib/web/admin/swagger/api.yaml
mozbhearsum: Good enough for now.

aksareen:
@mozbhearsum : The [UI issue](https://screenshots.firefoxusercontent.com/images/c10bfdd4-19e9-4299-ac20-7f93ba42eb4b.png) seems to be similar to one we encountered during Rules API migration. it seems that the input json of signoffs is passed as `{"signoffs_required": u'1'}` i.e. a string. Seems a custom validator needs to be added to verify the value as done previously for Rule model's fields

auslib/web/admin/swagger/api.yaml
mozbhearsum: What I'm trying to say is that the signoffs_required is linked with a role, and there may be multiple combinations of role+signoffs_required *before* a change can be executed, so having "to execute the change" here (as part of a single role) is a bit confusing. Perhaps something like "The number of signoffs required for a given role" would be better?

auslib/web/admin/swagger/api.yaml
mozbhearsum: Alright, I guess that's fine for now.

auslib/web/admin/swagger/api.yaml
aksareen: will do

PR #345, opened 2017-06-25
Bug 1336452: Abstracting 'count' field to a separate model

aksareen:


aksareen:
It's a bit tricky combining 'additionalProperties' and 'allOf' properties in the same schema. Since allOf [won't behave as inheritance](https://stackoverflow.com/questions/22689900/json-schema-allof-with-additionalproperties) in OOP , hence maxProperties and minProperties cannot be applied as easily in case of 'allOf' based schema as it can be in case of basic schema with additionalProperties

aksareen:
Update: I tried using allOf with column response. Won't work. Here's the [logs](https://pastebin.mozilla.org/9025834)

PR #346, opened 2017-06-30
Add support for "memory" to the update ping

mozbhearsum:
Three major parts to this:

1. Rename systemCapabilities -> instructionSet
2. Add "memory to the Rules table
3. Add some code that knows how to split %SYSTEM_CAPABILITIES% into parts.

As part of the instructionSet rename I also converted its matching logic back to a simple csv match, as there's no reason to support && matching anymore (new parts of systemCapabilities will go in new columns). This essentially reverts the non-osVersion parts of #280.

The memory support part should be pretty straightforward - it works very much like the buildID field, except we're using the new int_compare function instead of string_compare. (I was surprised to see the buildID used string_compare, and that it started failing when I switched it to int_compare...)

The splitting logic is a little complicated because it needs to support the instruction-set-only case, the plain csv case, and the explicit prefixing being added in https://bugzilla.mozilla.org/show_bug.cgi?id=1373367. The nice thing about the prefixing is that the splitting code will just ignore newly added parts until we add support for them.

I've done a fair bit of testing of this code, including:
* Comparing a ton of update requests from production vs. this code (with the latest db dump)
* Verifying that we can actually set-up Rules to migrate people from 32 -> 64-bit (https://screenshots.firefoxusercontent.com/images/d59ee584-cd42-4903-b4df-eca32e2cac0e.png)
* Explicitly checking that SSE deprecation continues to work.


auslib/web/public/client.py
```diff
@@ -29,6 +29,33 @@ def getHeaderArchitecture(buildTarget, ua):
         return 'Intel'
 
 
+def getSystemCapabilities(systemCapabilities):
```
ex-Dev: Can the name of this function be "getInstructionSet", since it extract the instruction set from systemCapabilities query string?

mozbhearsum: I don't think that's a great name, because it does more than that (it also returns memory, and probably other values in the future). I'm open to other suggestions though - now that I re-read this, a function called "getSystemCapabilities" that takes an argument of systemCapabilities is very weird...

ex-Dev: Ah, OK, I thought that memory seemed a special case of instruction set, something like "an hardware with more than 8GB supports 64bits version".

auslib/test/web/test_client.py
```diff
@@ -261,6 +296,37 @@ def setUp(self):
 }
 """))
 
+        dbo.rules.t.insert().execute(priority=90, backgroundRate=100, mapping="f", update_type="minor", product="f",
+                                     channel="a", memory="<=8000", data_version=1)
+        dbo.rules.t.insert().execute(priority=90, backgroundRate=100, mapping="f", update_type="minor", product="f",
+                                     channel="b", memory="9000", data_version=1)
+        dbo.rules.t.insert().execute(priority=90, backgroundRate=100, mapping="f", update_type="minor", product="f",
+                                     channel="c", memory=">10000", data_version=1)
```
nthomas-mozilla: You are very sneaky, sir ! I was going to complain that all the responses are the same so how do you know if the memory condition works, but then saw the channel differs and you'd get an empty response otherwise. 👏

nthomas-mozilla:
re the screenshot for the test rules, I meant to ask if you were planning any automation to create Firefox-<version>-build2-32-to-64 releases as new <version>'s come along, or will this be a watershed ?

auslib/db.py
```diff
@@ -1663,15 +1688,17 @@ def getRawMatches():
...
                 self.log.debug("%s doesn't match %s", rule['osVersion'], updateQuery['osVersion'])
                 continue
-            # Same deal for system capabilities
-            if not self._simpleBooleanMatchesRule(rule['systemCapabilities'], updateQuery.get('systemCapabilities', ""), substring=False):
-                self.log.debug("%s doesn't match %s", rule['systemCapabilities'], updateQuery.get('systemCapabilities'))
+            if not self._csvMatchesRule(rule['instructionSet'], updateQuery.get('instructionSet', ""), substring=False):
+                self.log.debug("%s doesn't match %s", rule['instructionSet'], updateQuery.get('instructionSet'))
                 continue
             # Locales may be a comma delimited rule too, exact matches only
             if not self._localeMatchesRule(rule['locale'], updateQuery['locale']):
```
mozbhearsum: That's a good point -- I suspect we evolved a couple of similar ways to do the same thing over time, and haven't refactored yet.

auslib/web/admin/swagger/api.yaml
```diff
@@ -1832,11 +1832,12 @@ paths:
                   headerArchitecture: null
```
mozbhearsum: Yeah, I think that's the best way to do it. We should probably change the spec/API versions to "1.0" for now, instead of inheriting the Balrog app version.

auslib/test/web/test_client.py
mozbhearsum: 😏😏😏

mozbhearsum:
I'm not planning to do any automation at this time. As far as I know, the current plan is to do a one-off watershed at 56 to migrate everyone, and that's it. We decided in Hawai'i (IIRC) that we wouldn't want to continually migrate people yet to because we want to allow anyone who decides they don't want 64-bit to pave-over with a 32-bit build and stay on that. I imagine we'll do additional migrations later, but there's nothing in the cards yet. (Tangentially related, Robert Strong was talking yesterday about adding a "has been migrated" indication to the update ping so we can avoid migrating people multiple times.)

If we ever get to a point where we always want to migrate some 32-bit users to 64-bit we'll definitely need automation. The way things stand right now it would require multiple Releases per Firefox version :(.

auslib/db.py
mozbhearsum: Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1379221 for this.

docs/source/database.rst
```diff
@@ -84,35 +84,32 @@ Following tables show columns according to different Categories:
...
   |                        +--------------------+--------------------------------------------------+---------------------------------+----------------------------+
-  |                        | Version            | The version of the application requesting an     | Exact string match or exact     | "36.0" or "36.0,36.1,36.2" |
+  |                        | memory             | The amount of RAM, in megabytes, that the client | Exact match or operator plus    | 8096, <8096, >=8096        |
+  |                        |                    | requesting the update has                        | memory to compare the incoming  |                            |
+  |                        |                    |                                                  | one against                     |                            |
+  |                        +--------------------+--------------------------------------------------+---------------------------------+----------------------------+
+  |                        | version            | The version of the application requesting an     | Exact string match or exact     | "36.0" or "36.0,36.1,36.2" |
```
mozbhearsum: Strange, it looks like this cell ends up getting rendered as <dt> and <dd>, which has some special styling. I'll dig into it.

auslib/web/admin/swagger/api.yaml
```diff
@@ -2478,13 +2481,20 @@ definitions:
...
-      systemCapabilities:
+      instructionSet:
         description: The supported hardware features of the application requesting an update. This field is primarily used to point desupported users based on their hardware. Example - users who do not support SSE2
         type: ["string", "null"]
         minLength: 0
         maxLength: 1000
         example: 'SSE'
 
+      memory:
+        description: The amount of RAM, in megabytes, that the client requesting the update has
```
mozbhearsum: That's a good point, I'll improve these.

mozbhearsum:
I believe I've addressed everything you've pointed out. If I missed anything (or you have comments on any of the new stuff) feel free to let me know and I'll fix them in a follow-up.

auslib/db.py
```diff
@@ -1663,15 +1688,17 @@ def getRawMatches():
...
             # To help keep the rules table compact, multiple OS versions may be
             # specified in a single rule. They are comma delimited, so we need to
             # break them out and create clauses for each one.
             if not self._simpleBooleanMatchesRule(rule['osVersion'], updateQuery['osVersion']):
                 self.log.debug("%s doesn't match %s", rule['osVersion'], updateQuery['osVersion'])
                 continue
-            # Same deal for system capabilities
-            if not self._simpleBooleanMatchesRule(rule['systemCapabilities'], updateQuery.get('systemCapabilities', ""), substring=False):
-                self.log.debug("%s doesn't match %s", rule['systemCapabilities'], updateQuery.get('systemCapabilities'))
+            if not self._csvMatchesRule(rule['instructionSet'], updateQuery.get('instructionSet', ""), substring=False):
```
nthomas-mozilla: I had one more thought. During deployment, what happens when the db is migrated and a canary public node is up, but the rest of the public nodes on the old code ? Perhaps the latter will hit a lot of errors when rule['systemCapabilities'] doesn't exist.

mozbhearsum: Ack, that's an important point. Even if we used .get() here we'd still end up not properly evaluating rules that use systemCapabilities/instructionSet. I'll need to do a bit of reworking to make this a two-part landing.

PR #350, opened 2017-07-10
Remove systemCapabilities column

mozbhearsum:
This is part 3 of 3 to give us a smooth update path for #346. It is built on top of #348 and #349 and simply removes the last remnants of the systemCapabilities column.


mozbhearsum:
That taskcluster failure is fallout from the work I was doing on https://bugzilla.mozilla.org/show_bug.cgi?id=1276289 - now fixed.

PR #351, opened 2017-07-11
Make items-per-site filters on rules page

ROGAL12321:
My proposition of feature in bugzilla n.o. :

https://bugzilla.mozilla.org/show_bug.cgi?id=1300171

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/balrog/351)

### DIFF
--- a/ui/app/css/rules.less
+++ b/ui/app/css/rules.less
@@ -1,0 +1,8 @@
+.rules-filters__label {
+  display: block;
+}
+
+.rules-filers__wrapper {
+  padding-bottom: 20px;
+  font-size: 18px;
+}
--- a/ui/app/js/controllers/releases_controller.js
+++ b/ui/app/js/controllers/releases_controller.js
@@ -9,7 +9,8 @@ function($scope, $routeParams, $location, $timeout, Releases, Search, $modal, Pa
   $scope.release_name = $routeParams.name;
 
   $scope.currentPage = 1;
-  $scope.pageSize = 10;
+  $scope.pagesize_options = [10, 25, 50, 100];
+  $scope.pageSize = $scope.pagesize_options[0];
   $scope.maxSize = 10;
 
   function loadPage(newPage) {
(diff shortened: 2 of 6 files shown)


ui/app/js/controllers/rules_controller.js
```diff
@@ -10,7 +10,8 @@ function($scope, $routeParams, $location, $timeout, Rules, Search, $modal, $rout
   $scope.pr_ch_options = [];
 
   $scope.currentPage = 1;
-  $scope.pageSize = 20;
+  $scope.pagesize_options = [10, 25, 50, 100]
```
mozbhearsum: I'm surprised it doesn't cause a test failure (or a failure in actual usage), but this is throwing an error:
```
balrogui_1     |    app/js/controllers/rules_controller.js
balrogui_1     |      13 |  $scope.pagesize_options = [10, 25, 50, 100]
balrogui_1     |                                                       ^ Missing semicolon.
balrogui_1     | 

```

ui/app/css/rules.less
```diff
@@ -0,0 +1,8 @@
+.rules-filters__label {
```
mozbhearsum: I appreciate what you're trying to do with these css changes, but I think they actually take away from the page a bit. Users are primarily focused on the rules themselves -- adding all the extra whitespace/size to the controls takes away from that. These controls are sometimes sometimes used right after the page loads, but most of the time spent on the page is down in the rules themselves -- so keeping the focus on them is preferable.

ui/app/js/controllers/releases_controller.js
```diff
@@ -75,6 +77,11 @@ function($scope, $routeParams, $location, $timeout, Releases, Search, $modal, Pa
   }
   $scope.ordering_str = $scope.ordering_options[0];
 
+
+  $scope.$watch('pageSize', function(value) {
+    $scope.pageSize = value;
```
mozbhearsum: Can you store and retrieve this in localStorage like we do for the product/channel drop down (https://github.com/mozilla/balrog/blob/d4b030dcf1c8fc097aaee6ffd7ac6366d70fcbf8/ui/app/js/controllers/rules_controller.js#L82). I suspect that different users may have different preferences for page size, so it would be nice to save them.

mozbhearsum:
Are you planning to finish this up at some point @ROGAL12321?

mozbhearsum:
Closing due to inactivity. Feel free to continue this work at a later time if you're interested.

PR #352, opened 2017-07-13
Initial Update

mozbhearsum:


mozbhearsum:
So this looks like it's ready to go now. I had to make a few code and test changes to code with some of the new package versions, but AFAICT everything works fine. I did a bunch of testing through the UI, as well as running some of the submission scripts locally - all worked fine.

There's a few new warnings we're seeing now, but nothing that's going to break anything. We get this from the balrog agent, presumably because we don't manage our own session:
> balrogagent_1  | 2017-07-13 19:43:50,300 - ERROR - asyncio.default_exception_handler#1258: Unclosed client session

And there's some deprecation warnings like:

> balrogadmin_1  | /usr/local/lib/python2.7/site-packages/werkzeug/local.py:347: DeprecationWarning: json is deprecated.  Use get_json() instead.
> balrogadmin_1  |   return getattr(self._get_current_object(), name)
> 

I think it's OK to ignore them for now.

auslib/test/admin/views/test_csrf.py
```diff
@@ -9,19 +9,11 @@ class TestCSRFEndpoint(ViewTest):
     def setUp(self):
         ViewTest.setUp(self)
         app.config['WTF_CSRF_ENABLED'] = True
-        # Normally we'd just use mock.patch to do this, but it's not working
```
JohanLorenzo: Yay clean up!

PR #355, opened 2017-07-20
Bug 1336452: Scheduled Changes First group of APIs

aksareen:


auslib/web/admin/swagger/api.yaml
```diff
@@ -254,6 +276,19 @@ parameters:
...
+    in: body
+    description: request body for posting a signoff to a scheduled change
+    required: true
+    schema:
+      type: object
+      allOf:
+        - $ref: '#/definitions/CSRFModel'
+        - $ref: '#/definitions/RoleModel'
+      required:
+        - role
```
aksareen: it is already a required field specified in the CSRFModel. here's the [link](https://github.com/aksareen/balrog/blob/ddee69f55db14de816c84d3615426661c3dce846/auslib/web/admin/swagger/api.yaml#L2794) . Role is explicitly required in some APIs and not in others, hence specified on this [line](https://github.com/aksareen/balrog/blob/ddee69f55db14de816c84d3615426661c3dce846/auslib/web/admin/swagger/api.yaml#L290)

auslib/web/admin/swagger/api.yaml
```diff
@@ -1893,90 +1906,970 @@ paths:
...
+      consumes:
+        - application/json
+      produces:
+        - text/html
+        - application/json
+      externalDocs:
+        url: "http://mozilla-balrog.readthedocs.io/en/latest/admin_api.html#scheduled-changes-object"
+        description: "create scheduled change for rule."
+      parameters:
+        - name: change_type
```
aksareen: because this one is the query (optional) parameter passed for the /sc/rules post operation and already handled [here](https://github.com/aksareen/balrog/blob/ddee69f55db14de816c84d3615426661c3dce846/auslib/web/admin/views/rules.py#L296).

auslib/web/admin/swagger/api.yaml
```diff
@@ -1893,90 +1906,970 @@ paths:
...
+          in: query
```
aksareen: same reason as above.

auslib/web/admin/swagger/api.yaml
```diff
@@ -1893,90 +1906,970 @@ paths:
...
+          type: string
+          enum:
+            - insert
+            - delete
+            - update
+          minLength: 1
+          maxLength: 20
+          x-nullable: true
+          required: false
+        - name: sc_rule_body
```
aksareen: this is a query parameter which may not be required if the change_type is passed in the request.json body

auslib/web/admin/swagger/api.yaml
```diff
@@ -1893,90 +1906,970 @@ paths:
...
+                        - $ref: '#/definitions/SCIdModel'
+                        - $ref: '#/definitions/SCDataVersion'
+                        - $ref: '#/definitions/SCGetModel'
+                        - $ref: '#/definitions/DataVersionNullable'
+                        - $ref: '#/definitions/OptionsModel'
+                        - $ref: '#/definitions/PermissionsModel'
+                      required:
+                        - change_type
+                        - data_version
+                        - when
```
aksareen: some models like [CSRFModel](https://github.com/aksareen/balrog/blob/ddee69f55db14de816c84d3615426661c3dce846/auslib/web/admin/swagger/api.yaml#L2794) already have required fields defined in their model definitions. So using the model via allOf or any other inheritance method automatically implies that the required fields conditions from that model are automatically inherited too.  Hence we need not explicitly define another list of required fields (which is also redundant) in the API request json body's 'required' field list again.

auslib/web/admin/swagger/api.yaml
```diff
@@ -1893,90 +1906,970 @@ paths:
...
+      consumes:
+        - application/json
+      produces:
+        - text/html
+        - application/json
+      externalDocs:
+        url: "http://mozilla-balrog.readthedocs.io/en/latest/admin_api.html#scheduled-changes-object"
+        description: "create scheduled change for releases."
+      parameters:
+        - name: sc_permission_body
```
aksareen: what should I change it into ? BTW in case of body parameters , 'name' is entirely useless and is only needed for uniquely identifying parameter name by swagger but never used in handling the request.json in python code.

auslib/web/admin/swagger/api.yaml
```diff
@@ -1893,90 +1906,970 @@ paths:
...
+      count:
+        description: count of total values fetched from DB
+        type: integer
+        minimum: 0
+        format: int32
+        example: 1
+
+  DataVersionModel:
+    title: Data Version
+    description: "model to re-use 'data_version' field."
```
aksareen: This model is used multiple API operations. Those API request/response json bodies that have 'required' set to true explicitly define in their own 'required' list in API's definition sections, ([example](https://github.com/aksareen/balrog/blob/ddee69f55db14de816c84d3615426661c3dce846/auslib/web/admin/swagger/api.yaml#L759), while some APIs have it as optional , hence not always required. [Example](https://github.com/aksareen/balrog/blob/ddee69f55db14de816c84d3615426661c3dce846/auslib/web/admin/swagger/api.yaml#L300), [example](https://github.com/aksareen/balrog/blob/ddee69f55db14de816c84d3615426661c3dce846/auslib/web/admin/swagger/api.yaml#L3132)

auslib/web/admin/swagger/api.yaml
```diff
@@ -1893,90 +1906,970 @@ paths:
...
+    description: "model to re-use 'signoffs_required' integer field."
+    type: object
+    properties:
+      signoffs_required:
         description: The number of signoffs required for a given role
         type: ["integer", "string"]
         format: signoffs_required
         example: 2
 
+  SignoffsRequiredNullable:
```
aksareen: I did see instances of this field getting set as null when operating GET on /scheduled_changes/required_signoffs/permissions or /scheduled_changes/required_signoffs/product. Also , one of the test cases failed when field was set to non-nullable. here's the [test case](https://github.com/aksareen/balrog/blob/ddee69f55db14de816c84d3615426661c3dce846/auslib/test/admin/views/test_required_signoffs.py#L651)

auslib/web/admin/swagger/api.yaml
```diff
@@ -2345,22 +3372,75 @@ definitions:
...
+        description: "The update_type to use in the XML response. 'minor' is almost always the correct value these days."
+        type: string
+        minLength: 0
+        maxLength: 15
+        enum: [ "major", "minor"]
+        example: minor
 
-    # Cannot have additionalProperties as false for strict validation ,else it disallows csrf_token or any other model's properties to be added
-    #additionalProperties: false
+  RulesNonNullable:
```
aksareen: what should I re-name it into ?

auslib/web/admin/views/scheduled_changes.py
```diff
@@ -123,7 +121,7 @@ def _delete(self, sc_id, transaction, changed_by):
         if not sc:
             return Response(status=404, response="Scheduled change does not exist")
 
-        form = DbEditableForm(request.args)
+        form = DbEditableForm(connexion.request.args)
```
aksareen: Removing flask.request caused the sc/:namespace/<int:sc_id> API operations to break since request package was no longer defined. This form is being used in the  ScheduledChangeView group of APIs. They haven't been migrated in this commit.

auslib/web/admin/views/permissions.py
```diff
@@ -144,23 +143,36 @@ def _delete(self, username, permission, changed_by, transaction):
...
 class PermissionScheduledChangesView(ScheduledChangesView):
+    """/scheduled_changes/permissions"""
+
     def __init__(self):
         super(PermissionScheduledChangesView, self).__init__("permissions", dbo.permissions)
 
     @requirelogin
     def _post(self, transaction, changed_by):
-        change_type = connexion.request.json.get("change_type")
+        what = connexion.request.json
```
aksareen: in that case this will cause us to maintain two list of fields for each "change_type". First, the swagger list of allowed fields is the exhaustive/super-set and each change_type will need to have explicit list of fields allowed, rest will still have to be filtered out. 

Copying extra fields will imply that 3 separate list of allowed field values for each change_type defined in python code. It will be same as maintaining a Form like before.

auslib/web/admin/views/scheduled_changes.py
```diff
@@ -64,17 +65,14 @@ def get(self):
...
+        what.pop("csrf_token", None)
 
-        # Forms can normally be accessed as a dict through form.data,
-        # but because some of the Forms we end up using have a Field
-        # called "data", this gets overridden, so we need to construct
-        # a dict ourselves.
-        columns = {k: v.data for k, v in form._fields.iteritems()}
-        sc_id = self.sc_table.insert(changed_by, transaction, **columns)
+        # Validate 'when' is not in past
+        if what.get("when", None) and int(what.get("when")) < getMillisecondTimestamp():
```
aksareen: sc_when_validator if declared on a field will automatically validate the request and response bodies both. In case of response bodies , it will always return an error when attempting to fetch a S.C.(which was enacted) and its 'when' value (stored in DB) was in past.

aksareen:
@bhearsum  : regarding the comment above

> In addition to the comments inline below, I found an issue when testing through the UI: When I tried to schedule a delete for a rule I got: "comment isn't required when scheduling a 'delete' change". The error didn't show up in the UI, only in the error console.

Looks like the UI is broken and is sending extra fields in the delete request. Hence 'comment' which is not needed in scheduling a delete SC is also passed in request.json body. Looks like a quick fix to this problem is to pop out the non-required fields when "change_type" == "delete"

auslib/web/admin/swagger/api.yaml
mozbhearsum: Ah, I didn't realize that the "required" sections stacked like that. Sweet!

auslib/web/admin/swagger/api.yaml
mozbhearsum: Hm, I don't think we actually pass change_type in the query -- we put it in the body. The code you linked to was added to support this client class: https://github.com/mozilla/balrog/blob/master/client/balrogclient/api.py#L242.

auslib/web/admin/swagger/api.yaml
mozbhearsum: Hm, I don't think we actually pass change_type in the query -- we put it in the body. The code you linked to was added to support this client class: https://github.com/mozilla/balrog/blob/master/client/balrogclient/api.py#L242.

auslib/web/admin/swagger/api.yaml
mozbhearsum: Sorry - I think I added this comment to the wrong line. It was meant for change_type.

auslib/web/admin/swagger/api.yaml
mozbhearsum: OK, this inconsistency is a bit confusing, but it's not a blocker. I think in the long term it might be a good idea to keep "required" out of the models, and make it explicit in every endpoint (even if that means repeating it in places).

auslib/web/admin/swagger/api.yaml
mozbhearsum: Yeah, I realize it's just a label, but it can lead to a bit of confusion (I think it's shown in the swagger UI, too?). In any case, this is the releases API, so sc_release_body is probably more appropriate.

auslib/web/admin/swagger/api.yaml
mozbhearsum: OK

auslib/web/admin/swagger/api.yaml
mozbhearsum: Ah, I see. I'd forgotten about the DELETE case for the required signoffs table. Carry on!

auslib/web/admin/swagger/api.yaml
mozbhearsum: Maybe RulesAlwaysRequired, if that's actually the case?

auslib/web/admin/views/permissions.py
mozbhearsum: You can do this by inverting the logic a bit, and copying any field that's _not_ dependent on change_type first, and then copying the change_type specific ones later.

auslib/web/admin/views/scheduled_changes.py
mozbhearsum: Got it.

auslib/web/admin/views/scheduled_changes.py
mozbhearsum: Ah, so this will be removed when you finish converting this endpoint? If so, sounds good for now.

auslib/web/admin/swagger/api.yaml
aksareen: that's why we have to define both types of parameters (json body and query) in this case else requests will start failing for the client class. Also the query parameter has 'required' set as False too , hence it is optional.

auslib/web/admin/swagger/api.yaml
aksareen: Not always. 'change_type' can be fetched from both query and body on this [line](https://github.com/aksareen/balrog/blob/ddee69f55db14de816c84d3615426661c3dce846/auslib/web/admin/views/rules.py#L296) . Therefore, cannot have change_type in 'required' list of fields since it will break client requests. 

Instead this condition is handled explicitly in the web layer code [here](https://github.com/aksareen/balrog/blob/ddee69f55db14de816c84d3615426661c3dce846/auslib/web/admin/views/rules.py#L330)

auslib/web/admin/swagger/api.yaml
aksareen: oh my bad, forgot to change label name for release API

auslib/web/admin/swagger/api.yaml
aksareen: Okay, that will make things a bit redundant but it looks the best time to do this after fully migrating all the APIs first.

auslib/web/admin/swagger/api.yaml
aksareen: No that's not correct. Some fields aren't always required. Seems like a misnomer. Also priority and backgroundRate are actually nullable in many cases when handling request/response in /rules APIs. 

Any name that can automatically signify that model's fields can be of any type but 'null' will do here

auslib/web/admin/views/permissions.py
aksareen: copying specific fields based on change types is same as using explicit wtfform for each change_type. (which filters out specific fields from request.json and copies into form object). 

Since wtfforms already had inheritance, they were easily reusing 'when' and 'telemtry_%s' forms from forms.py for every S.C. operation.

**Pros**: Simple and easier to maintain in some cases where no. of fields are low like for sc/permissions and sc/releases.

Also, another issue is that additionalProperties has to be always true(in order to use allOf swagger feature to re-use fields via models) i.e. any no. of extra parameters can be passed into the requests (UI passes extra json body arguments in case of DELETE operations) therefore, filtering out only required fields (maintained in explicit list for each of 3 change_type) for given change_type value seems better option rather than throw error for extra-field-present-in-json-body.

**Cons**: Harder to maintain when dozens of fields need to be maintained for each change_type. Each change_type needs an explicit list. Ex: /sc/rules. Easier to simply copy entire request.json and filter out some specific fields then. 

Also, in case extra parameters get passed into the 'what' dictionary won't matter, since in the DB layer only the fields that match name of DB columns will get updated, rest will be simply ignored. Just need to make sure that some specific fields values are changed or filtered out first in web layer (around 3-5 fields) and simply pass the rest of the request.json dictionary . Easier to maintain than having explicit lists for each change_type

auslib/web/admin/swagger/api.yaml
mozbhearsum: Alright

mozbhearsum:
To follow-up on @aksareen's previous comment - I put together a proposal for an alternative way to do filtering (https://github.com/mozbhearsum/balrog/commit/0a0b7c2042b07fb5a7241f082c30f5b654e67dd1) and I think we're on the same page there now.

auslib/web/admin/views/permissions.py
```diff
@@ -144,23 +143,36 @@ def _delete(self, username, permission, changed_by, transaction):
...
+            what["options"] = json.loads(what.get("options"))
+
+        if change_type in ["update", "delete"]:
+            if not what.get("data_version", None):
+                return problem(400, "Bad Request", "Missing field", ext={"exception": "data_version is missing"})
+            else:
+                what["data_version"] = int(what["data_version"])
+
+        elif change_type != "insert":
+            return problem(400, "Bad Request", "Invalid or missing change_type")
```
mozbhearsum: This error is better raised in scheduled_changes.py rather than repeating it in each object type.

auslib/web/admin/views/permissions.py
```diff
@@ -144,23 +143,36 @@ def _delete(self, username, permission, changed_by, transaction):
...
-            form = ScheduledChangeExistingPermissionForm()
-        elif change_type == "insert":
-            form = ScheduledChangeNewPermissionForm()
-        elif change_type == "delete":
-            form = ScheduledChangeDeletePermissionForm()
-        else:
-            return Response(status=400, response="Invalid or missing change_type")
+        what = {}
+        for field in connexion.request.json:
+            if change_type == "delete" and field == "options" or change_type == "insert" and field == "data_version":
```
mozbhearsum: If I remember correctly, we decided to ignore options in this case because the UI currently passes it? If so, please add a comment as a reminder.

mozbhearsum: I'm surprised to see data_version being ignored here - is there a case where that is actually passed when change_type is insert?

auslib/web/admin/views/permissions.py
```diff
@@ -144,23 +143,36 @@ def _delete(self, username, permission, changed_by, transaction):
...
 class PermissionScheduledChangesView(ScheduledChangesView):
+    """/scheduled_changes/permissions"""
+
     def __init__(self):
         super(PermissionScheduledChangesView, self).__init__("permissions", dbo.permissions)
 
     @requirelogin
     def _post(self, transaction, changed_by):
         change_type = connexion.request.json.get("change_type")
+        connexion.request.json.pop("csrf_token", None)
```
mozbhearsum: Don't pop this here, ignore it when you iterate over it. Removing items from dicts like this can make debugging a lot more difficult. For example: if I'm looking at a client request and see csrf_token in it, but then don't see it if I print connexion.request.json in this function, it's very confusing.

auslib/web/admin/views/rules.py
```diff
@@ -285,45 +283,60 @@ def get(self, column):
...
+        # TODO: change the below operation from filter/pop to throw Error when extra fields are passed.
         elif change_type == "delete":
-            form = ScheduledChangeDeleteRuleForm()
-
-        else:
-            return Response(status=400, response="Invalid or missing change_type")
+            for key in list(what):
+                if key not in ["telemetry_product", "telemetry_channel", "telemetry_uptake", "when", "rule_id",
+                               "data_version", "change_type"]:
+                    what.pop(key, None)
```
mozbhearsum: As I said before, *please* do this before copying the values into "what", not by removing them after. This should be happening in the loop at line 300.

auslib/web/admin/views/scheduled_changes.py
mozbhearsum: Can you confirm that you'll be removing the forms later?

auslib/web/admin/views/permissions.py
aksareen: I thought since I'm ignoring csrf_token everytime when constricting what dict, rather simply delete it. Will do the suggested change.

auslib/web/admin/views/permissions.py
aksareen: not via the UI forms AFAIK. However, any client can still pass it accidentally and the newly created rule can end up with a non-single (not 1) value. Hence the check

auslib/web/admin/views/permissions.py
aksareen: Makes sense. Also I think we can safely remove the else check entirely since both change_type field either present in [query](https://github.com/aksareen/balrog/blob/6d5dc12245ab906bc47de15d4fa8994862f7e892/auslib/web/admin/swagger/api.yaml#L2009) or in [body](https://github.com/aksareen/balrog/blob/6d5dc12245ab906bc47de15d4fa8994862f7e892/auslib/web/admin/swagger/api.yaml#L3043).

auslib/web/admin/views/rules.py
aksareen: Oh mea culpa. Rather than popping it out, simply ignoring the rest of fields while constructing 'what 'if change_type == 'delete'

auslib/web/admin/views/scheduled_changes.py
aksareen: this particular form is being used for "/scheduled_changes/:namespace/:sc_id" endpoints. These endpoints will be migrated in the upcoming PR. Then once all the form-usage instances have been purged from the code, we can safely delete the models from forms.py and by the end of all API migrations, the entire forms.py file

auslib/web/admin/views/permissions.py
aksareen: will do.

auslib/web/admin/views/permissions.py
aksareen: Just tested out the via proof of concept. It fails for /sc/rules APIs since it checks first for change_type in body then automatically uses the value passed in query args without validating whether the value was passed or is it using the default 'None'  . Hence an explicit check is needed in the Parent Class.

auslib/web/admin/views/permissions.py
mozbhearsum: We don't have any non-UI clients for these endpoints with the exception of a script that creates a scheduled Rule change with change_type="update". I'd rather not encourage more bad client behaviour -- any new clients that are passing bad data should fix themselves.

So, please remove any precautionary filtering like this, except for the Rule case noted above, and for cases where we know the UI is already doing bad things.

aksareen: removing these checks

aksareen: UI is  passing 'options' as null for change_type == delete. I must have made some err while testing the UI initially. Seems only /sc/rules API is sending extra fields for delete change_type. Removing this check too.

auslib/web/admin/swagger/api.yaml
```diff
@@ -1893,82 +1916,956 @@ paths:
...
+          in: body
+          description: request body for sc required signoffs of product
+          required: true
+          schema:
+            type: object
+            allOf:
+              - $ref: '#/definitions/CSRFModel'
+              - $ref: '#/definitions/SCChangeType'
+              - $ref: '#/definitions/SCTime'
+              - $ref: '#/definitions/DataVersionModel'
```
mozbhearsum: I think using DataVersionModel here is causing some problems with the UI. When I tried to add a new role requirement for an existing signoff I got this: https://screenshots.firefox.com/XPk0lWEWCYG1HH15/localhost

The requests we make for changes like that contain the following POST data:

> {"product":"Firefox","role":"relman","csrf_token":"xxxxxxx","data_version":null,"channel":"beta","when":1501599718293,"change_type":"insert","signoffs_required":"1"}

Note that data_version is null because adding a new required signoff ends up a new row in the database - not modifying an existing one.

aksareen: Modifying it to Nullable-Data-Version-Model

aksareen: I just realized that data_version as Null isn't supposed to be passed at all since for change_type == 'insert' code[here](https://github.com/mozilla/balrog/blob/bf996837d2f19e88b9f9135a20b7e0563e11d559/auslib/web/admin/views/required_signoffs.py#L129) the wtfForm ScheduledChangeNewProductRequiredSignoffForm ([link](https://github.com/mozilla/balrog/blob/bf996837d2f19e88b9f9135a20b7e0563e11d559/auslib/web/admin/views/forms.py#L392) does not use data_version field at all. It looks like another UI (similar to rules) issue where extra arguments are passed as Null when they aren't supposed to be passed at all for a particular change_type. Looks like I need to add a TODO comment in swagger spec to fix UI later since data_version should not be passed as 'null' during a POST operation for 'update' or 'delete' change_types (since insert does not use data_version field at all).

PR #361, opened 2017-08-01
bug 1335861: add notifications for most tables

mozbhearsum:
Per https://bugzilla.mozilla.org/show_bug.cgi?id=1335861#c0.

### DIFF
--- a/auslib/db.py
+++ b/auslib/db.py
@@ -2537,23 +2537,12 @@ class AUSDatabase(object):
     engine = None
     migrate_repo = path.join(path.dirname(__file__), "migrate")
 
-    def __init__(self, dburi=None, mysql_traditional_mode=False, notify_tables=None):
+    def __init__(self, dburi=None, mysql_traditional_mode=False):
         """Create a new AUSDatabase. Before this object is useful, dburi must be
            set, either through the constructor or setDburi()"""
         if dburi:
             self.setDburi(dburi, mysql_traditional_mode)
         self.log = logging.getLogger(self.__class__.__name__)
-        if notify_tables:
-            self.notify_tables = notify_tables
-        else:
-            self.notify_tables = [
-                self.rules, self.rules.scheduled_changes, self.rules.scheduled_changes.signoffs,
-                self.permissions, self.permissions.user_roles, self.permissions.scheduled_changes, self.permissions.scheduled_changes.signoffs,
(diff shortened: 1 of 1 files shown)


auslib/db.py
```diff
@@ -2537,12 +2537,23 @@ class AUSDatabase(object):
...
 
-    def __init__(self, dburi=None, mysql_traditional_mode=False):
+    def __init__(self, dburi=None, mysql_traditional_mode=False, notify_tables=None):
         """Create a new AUSDatabase. Before this object is useful, dburi must be
            set, either through the constructor or setDburi()"""
         if dburi:
             self.setDburi(dburi, mysql_traditional_mode)
         self.log = logging.getLogger(self.__class__.__name__)
+        if notify_tables:
+            self.notify_tables = notify_tables
```
nthomas-mozilla: How would this work ? Seems a bit like you're passing references to tables/attributes that don't exist yet, but perhaps I've forgotten how we instantiate all this.

auslib/db.py
```diff
@@ -2537,12 +2537,23 @@ class AUSDatabase(object):
...
+        else:
+            self.notify_tables = [
+                self.rules, self.rules.scheduled_changes, self.rules.scheduled_changes.signoffs,
+                self.permissions, self.permissions.user_roles, self.permissions.scheduled_changes, self.permissions.scheduled_changes.signoffs,
+                self.permissionsRequiredSignoffs, self.productRequiredSignoffs.scheduled_changes, self.productRequiredSignoffs.scheduled_changes.signoffs,
```
nthomas-mozilla: I think you mean self.productRequiredSignoffs at the start of this line.

auslib/db.py
mozbhearsum: This works because in practice, AUSDatabase is always constructed with dburi being passed, via https://github.com/mozilla/balrog/blob/4d6f673a850cbd2924eb55145f047fadfceabeea/auslib/global_state.py#L18.

However, it's quite clear that if someone were to construct an AUSDatabase without dburi, this would fail. I didn't realize this when I was writing it - I'll fix it up.

auslib/db.py
mozbhearsum: Whoops, yup.

mozbhearsum:
How's this, @nthomas-mozilla?

nthomas-mozilla:
Ship it!

PR #364, opened 2017-08-03
Bug 1379285 handle requests that set bad values for "force" query parameter

hemant1612:


auslib/web/public/client.py
```diff
@@ -68,7 +68,10 @@ def getCleanQueryFromURL(url):
         if len(force_split) < 2:
             force_split = force_value.split('%3F', 1)
 
-        query['force'] = int(force_split[0])
+        if force_split[0]:
+            query['force'] = 1
+        else:
+            query['force'] = 0
```
mozbhearsum: I don't think this is the right place to do this, actually. getCleanQueryFromURL merely makes the query parseable by fixing some known bad strings that get sent. I think the right place for this fix is at https://github.com/mozilla/balrog/blob/4d6f673a850cbd2924eb55145f047fadfceabeea/auslib/web/public/client.py#L92.

You should add some test cases to https://github.com/mozilla/balrog/blob/master/auslib/test/web/test_client.py to be sure, though. Some values we've seen for force are:
- nothing
- '1" name="Firefox 54.0" isOSUpdate="false" isCompleteUpdate="false" installDate="1498012260998" displayVersion="54.0" channel="release" buildID="20170608105825" appVersion="54.0"'
- '1)'
- '"|sleep	7	#'

Having test cases for those should be enough to ensure this works.

auslib/web/public/client.py
```diff
@@ -68,7 +68,10 @@ def getCleanQueryFromURL(url):
         if len(force_split) < 2:
             force_split = force_value.split('%3F', 1)
 
-        query['force'] = int(force_split[0])
+        if force_split[0]:
```
mozbhearsum: In the bug, I suggest that we should only ever set force to True if it is explicitly "1" - not for any other value. I think this will set it to True whenever it's not an empty string.

mozbhearsum:
@hemant1612 - are you planning to finish this up at some point?

hemant1612:
Hi, really sorry for being late.Actually this was my first python bug and I have no experience with testing.
As far as I understood from above comments:

`query['force'] = (int(query.get('force', 0)) == 1) `
Here we are setting `query['force']` to `true` if we get 'force' in query as `'1'` or `1`.Right?
We want to set it `true` only when we have `force ` in query as `1` and not `'1'`.

so `query['force'] = (query.get('force', 0) == 1) ` should suffice.
Please correct me if I am wrong here,

allan-silva:
Take care when fall here:
https://github.com/mozilla/balrog/blob/master/auslib/web/public/client.py#L60
And here:
 https://github.com/mozilla/balrog/blob/master/auslib/web/public/client.py#L67
 
force will be 1 not '1'

auslib/web/public/client.py
```diff
@@ -89,7 +89,7 @@ def getQueryFromURL(url):
     query['osVersion'] = urllib.unquote(query['osVersion'])
     ua = request.headers.get('User-Agent')
     query['headerArchitecture'] = getHeaderArchitecture(query['buildTarget'], ua)
-    query['force'] = (int(query.get('force', 0)) == 1)
+    query['force'] = (query.get('force', 0) == "1")
```
allan-silva: @hemant1612, glasserc is going to finish a new great feature for this flag in #382, "force" behavior will change a little and override this code region.

mozbhearsum:
@hemant1612 - I'm sorry, but as @allan-silva said, this ended up being implemented in another pull request. I sincerely apologize for not being able to merge this after the time and effort you put in. If you'd like to look at something else, I'll make sure you're the only one looking at the next bug. If you'd rather not look at something else after this, I understand.

PR #367, opened 2017-08-07
improves query matching #1379244

bhavishyagopesh:
Presently `_matchesRegex` provides too much freedom this tries to limit that.

### DIFF
--- a/auslib/db.py
+++ b/auslib/db.py
@@ -1487,11 +1487,21 @@ class Rules(AUSTable):
     def _matchesRegex(self, foo, bar):
         # Expand wildcards and use ^/$ to make sure we don't succeed on partial
         # matches. Eg, 3.6* matches 3.6, 3.6.1, 3.6b3, etc.
-        test = foo.replace('.', '\.').replace('*', '.*')
-        test = '^%s$' % test
-        if re.match(test, bar):
+        # Channel length must be strictly greater than two
+        # And globbing is allowed at the end of channel-name only
+        if foo.endswith('*'):
+            if(len(foo) >= 3):
+                test = foo.replace('.', '\.').replace('*', '\*', foo.count('*') - 1)
+                test = '^{}.*$'.format(test[:-1])
+                if re.match(test, bar):
+                    return True
+                return False
+            else:
+                return False
+        elif (foo == bar):
             return True
-        return False
+        else:
+            return False
 
     def _channelMatchesRule(self, ruleChannel, queryChannel, fallbackChannel):
         """Decides whether a channel from the rules matches an incoming one.
--- a/auslib/test/test_db.py
+++ b/auslib/test/test_db.py
@@ -2422,32 +2422,6 @@ class TestRulesSpecial(unittest.TestCase, RulesTestMixin, MemoryDatabaseMixin):
         rules = self._stripNullColumns(rules)
         self.assertEquals(rules, expected)
 
-        # To ensure length of ruleChannel is >=3
-        expected = [dict(rule_id=2, priority=100, backgroundRate=100, channel='r*', update_type='z', data_version=1)]
-        rules = self.rules.getRulesMatchingQuery(
-            dict(name='', product='', version='3.0', channel='releasetest',
-                 buildTarget='', buildID='', locale='', osVersion='', distribution='',
-                 distVersion='', headerArchitecture='',
-                 force=False, queryVersion=3,
-                 ),
-            fallbackChannel='releasetest'
-        )
-        rules = self._stripNullColumns(rules)
-        self.assertNotEquals(rules, expected)
-
-        # To ensure globbing at end only
-        expected = [dict(rule_id=2, priority=100, backgroundRate=100, channel='r*test*', update_type='z', data_version=1)]
-        rules = self.rules.getRulesMatchingQuery(
-            dict(name='', product='', version='3.0', channel='releasetest-cck-blah',
-                 buildTarget='', buildID='', locale='', osVersion='', distribution='',
-                 distVersion='', headerArchitecture='',
-                 force=False, queryVersion=3,
-                 ),
-            fallbackChannel='releasetest'
-        )
-        rules = self._stripNullColumns(rules)
-        self.assertNotEquals(rules, expected)
-
     def testGetRulesMatchingBuildIDComparison(self):
         expected = [dict(rule_id=3, priority=100, backgroundRate=100, buildID='>=20010101222222', update_type='z', data_version=1)]
         rules = self.rules.getRulesMatchingQuery(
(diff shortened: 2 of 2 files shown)


auslib/db.py
```diff
@@ -1487,11 +1487,22 @@ def getPotentialRequiredSignoffs(self, affected_rows, transaction=None):
...
+            test = test + '.*'
+            test = '^%s$' % test
+            if re.match(test, bar):
+                return True
+            return False
+        else:
+            test = foo.replace('.', '\.').replace('*', '\*')
+            test = '^%s$' % test
+            if re.match(test, bar):
+                return True
```
mozbhearsum: I don't think there's any need to do regex matching in this else block. As far as I can tell, we only want to match exact strings, so you may as well just do "if test == bar" instead -- this is much faster than using a regex.

auslib/db.py
```diff
@@ -1487,11 +1487,22 @@ def getPotentialRequiredSignoffs(self, affected_rows, transaction=None):
     def _matchesRegex(self, foo, bar):
         # Expand wildcards and use ^/$ to make sure we don't succeed on partial
         # matches. Eg, 3.6* matches 3.6, 3.6.1, 3.6b3, etc.
-        test = foo.replace('.', '\.').replace('*', '.*')
-        test = '^%s$' % test
-        if re.match(test, bar):
-            return True
-        return False
+        if foo.endswith('*'):
+            if(re.match('^.{0,2}\*$', foo)):
```
mozbhearsum: It's a great idea to limit the globbing to strings of a certain length, but I would suggest a couple of changes:
* Use a simple len() comparison to determine if the string is long enough. Let's limit globbing to strings that have at least two other characters, so this can be "if len(foo) >= 3" (2 characters + a *, which we already know is the final character).
* If this check fails, we should go into else block instead of returning False.

auslib/db.py
```diff
@@ -1487,11 +1487,22 @@ def getPotentialRequiredSignoffs(self, affected_rows, transaction=None):
     def _matchesRegex(self, foo, bar):
         # Expand wildcards and use ^/$ to make sure we don't succeed on partial
         # matches. Eg, 3.6* matches 3.6, 3.6.1, 3.6b3, etc.
```
mozbhearsum: This comment could use an update to reflect the new logic.

auslib/db.py
```diff
@@ -1487,11 +1487,22 @@ def getPotentialRequiredSignoffs(self, affected_rows, transaction=None):
...
-        if re.match(test, bar):
-            return True
-        return False
+        if foo.endswith('*'):
+            if(re.match('^.{0,2}\*$', foo)):
+                return False
+            test = foo.replace('.', '\.').replace('*', '\*', foo.count('*')-1)
+            test = test[:-1]
+            test = test + '.*'
+            test = '^%s$' % test
```
mozbhearsum: This section might read a bit better with string construction, eg:
`test = '^{}.*$'.format(test[:-1])`

auslib/db.py
```diff
@@ -1487,11 +1487,22 @@ def getPotentialRequiredSignoffs(self, affected_rows, transaction=None):
...
         # matches. Eg, 3.6* matches 3.6, 3.6.1, 3.6b3, etc.
-        test = foo.replace('.', '\.').replace('*', '.*')
-        test = '^%s$' % test
-        if re.match(test, bar):
-            return True
-        return False
+        if foo.endswith('*'):
+            if(re.match('^.{0,2}\*$', foo)):
+                return False
+            test = foo.replace('.', '\.').replace('*', '\*', foo.count('*')-1)
```
mozbhearsum: Nice work escaping the non-trailing *!

bhavishyagopesh:
ping @mozbhearsum

mozbhearsum:
I'll have a look shortly. In the future, please avoid rebasing open pull requests though...it makes it harder to review, and it stops Github from sending e-mail about the changes.

auslib/test/test_db.py
```diff
@@ -2422,6 +2422,32 @@ def testGetRulesMatchingQueryChannelGlobbing(self):
         rules = self._stripNullColumns(rules)
         self.assertEquals(rules, expected)
 
+        # To ensure length of ruleChannel is >=3
+        expected = [dict(rule_id=2, priority=100, backgroundRate=100, channel='r*', update_type='z', data_version=1)]
```
mozbhearsum: Please put each of these cases into their own method. When you include them in existing method, it means they won't run at all if an earlier assert fails.

auslib/test/test_db.py
```diff
@@ -2422,6 +2422,32 @@ def testGetRulesMatchingQueryChannelGlobbing(self):
...
+        rules = self.rules.getRulesMatchingQuery(
+            dict(name='', product='', version='3.0', channel='releasetest',
+                 buildTarget='', buildID='', locale='', osVersion='', distribution='',
+                 distVersion='', headerArchitecture='',
+                 force=False, queryVersion=3,
+                 ),
+            fallbackChannel='releasetest'
+        )
+        rules = self._stripNullColumns(rules)
+        self.assertNotEquals(rules, expected)
```
mozbhearsum: I don't think you want to be using assertNotEquals. You should be using assertEquals, and ensuring that the data that the rules table is initialized with (https://github.com/mozilla/balrog/blob/a10b66786774262397a51e77931525ea2c0cbbdc/auslib/test/test_db.py#L2297) includes data that lets your tests pass. For example, you will probably need to add an additional rule with channel set to "r*", and then expect that that rule will be returned by getRulesMatchingQuery(). Please let me know if you have trouble with that - I know the tests can be intimidating at first.

auslib/test/test_db.py
```diff
@@ -2422,6 +2422,32 @@ def testGetRulesMatchingQueryChannelGlobbing(self):
...
+
+        # To ensure globbing at end only
+        expected = [dict(rule_id=2, priority=100, backgroundRate=100, channel='r*test*', update_type='z', data_version=1)]
```
mozbhearsum: This is a good test case. Like the other one, you'll want to add this rule as part of setUp(). However, I think you'll want to expect no rules to match when you pass "channel='releasetest-cck-blah'" to getRulesMatchingQuery (because "r*test*" isn't a match). Does that make sense?

auslib/db.py
```diff
@@ -1487,11 +1487,21 @@ def getPotentialRequiredSignoffs(self, affected_rows, transaction=None):
...
+                if re.match(test, bar):
+                    return True
+                return False
+            else:
+                return False
+        elif (foo == bar):
             return True
-        return False
+        else:
+            return False
```
mozbhearsum: This block is much nicer now - great work!

auslib/test/test_db.py
```diff
@@ -2698,6 +2710,33 @@ def testProceedIfNotReadOnly(self):
...
+                 distVersion='', headerArchitecture='',
+                 force=False, queryVersion=3,
+                 ),
+            fallbackChannel='releasetest'
+        )
+        rules = self._stripNullColumns(rules)
+        self.assertEquals(rules, expected)
+
+        # To ensure globbing at end only
+        expected = [dict(rule_id=6, priority=100, backgroundRate=100, channel='r*test*', update_type='z', data_version=1)]
```
mozbhearsum: As I mentioned earlier, please make sure each test case is in its own function - otherwise they may not always be run.

auslib/test/test_db.py
```diff
@@ -2698,6 +2710,33 @@ def testProceedIfNotReadOnly(self):
...
+        rules = self.rules.getRulesMatchingQuery(
+            dict(name='', product='', version='3.0', channel='r*test-cck-blah',
```
mozbhearsum: This is a good test for the positive case. Can you also add one for the negative case (where channel is set to something like 'raaaatest', which should have no rules matching)?

bhavishyagopesh:
@mozbhearsum I'm not sure if you want this is a separate commit, so (without rebasing or amend) what should be the workflow?

mozbhearsum:
Separate commits are ideal when continuing work on an existing PR. Rebasing or squashing merging (which is what I do for this repo) will happen once the PR is approved :).

PR #368, opened 2017-08-08
Bug 1336452: Scheduled changes Remaining APIs

aksareen:
Final PR to migrate all remaining admin APIs i.e. SC Edit and Revision APIs.

### DIFF
--- a/auslib/web/admin/swagger/api.yaml
+++ b/auslib/web/admin/swagger/api.yaml
@@ -2219,6 +2219,7 @@ paths:
               - change_type
               - permission
               - username
+              - when
       responses:
         '200':
           description: Returns sc_id of change
@@ -2460,6 +2461,7 @@ paths:
             required:
               - change_type
               - name
+              - when
       responses:
         '200':
           description: Returns sc_id of change
@@ -2710,6 +2712,7 @@ paths:
               - product
               - channel
               - role
+              - when
       responses:
         '200':
           description: Returns sc_id of change
@@ -2957,6 +2960,7 @@ paths:
               - change_type
               - product
               - role
+              - when
       responses:
         '200':
           description: Returns sc_id of change
--- a/auslib/web/admin/swagger/api.yaml
+++ b/auslib/web/admin/swagger/api.yaml
@@ -2034,6 +2034,72 @@ paths:
         '400':
           $ref: '#/responses/invalidFormData'
 
+  /scheduled_changes/rules/{sc_id}:
+    post:
+      summary: Modifies existing scheduled change for a Rule.
+      description: >
+        Modifies existing scheduled change for a rule. [Docs](http://mozilla-balrog.readthedocs.io/en/latest/admin_api.html#scheduled-changes-object-sc-id)
+      tags:
+        - Scheduled Changes
+      operationId: auslib.web.admin.views.mapper.scheduled_change_rules_post
+      consumes:
+        - application/json
+      produces:
+        - text/html
+        - application/json
+      externalDocs:
+        url: "http://mozilla-balrog.readthedocs.io/en/latest/admin_api.html#scheduled-changes-object-sc-id"
+        description: "Modifies existing scheduled change for a Rule."
+      parameters:
+        - $ref: '#/parameters/sc_id_param'
+        - name: sc_rule_body
+          in: body
+          description: request body for editing sc rule
+          required: true
+          schema:
+            type: object
+            allOf:
+              - $ref: '#/definitions/CSRFModel'
+              - $ref: '#/definitions/SCChangeType'
+              - $ref: '#/definitions/SCTime'
+              - $ref: '#/definitions/SCUptake'
+              - $ref: '#/definitions/SCDataVersion'
+              - $ref: '#/definitions/RulesNonNullable'
+              - $ref: '#/definitions/RulesBase'
+      responses:
+        '200':
+          $ref: '#/responses/updateExistingObject'
+        '400':
+          $ref: '#/responses/invalidFormData'
+        '404':
+          $ref: '#/responses/resourceNotFound'
+
+    delete:
+      summary: Deletes the given Scheduled Change for the Rule.
+      description: >
+        Deletes the given Scheduled Change for the named Rule object. 'data_version' must be provided. [Docs](http://mozilla-balrog.readthedocs.io/en/latest/admin_api.html#scheduled-changes-object-sc-id)
+      tags:
+        - Scheduled Changes
+      operationId: auslib.web.admin.views.mapper.scheduled_change_rules_delete
+      consumes: []
+      produces:
+        - application/json
+        - text/html
+      externalDocs:
+        url: "http://mozilla-balrog.readthedocs.io/en/latest/admin_api.html#scheduled-changes-object-sc-id"
+        description: Deletes the given Scheduled Change for the Rule.
+      parameters:
+        - $ref: '#/parameters/sc_id_param'
+        - $ref: '#/parameters/dataVersionParam'
+        - $ref: '#/parameters/csrfTokenParam'
+      responses:
+        '200':
+          description: Successfully deleted the given scheduled change.
+        '400':
+          $ref: '#/responses/invalidFormData'
+        '404':
+          $ref: '#/responses/resourceNotFound'
+
   /scheduled_changes/permissions:
     get:
       summary: Returns scheduled changes for permissions.
@@ -2080,6 +2146,8 @@ paths:
                         - change_type
                         - data_version
                         - when
+                        - permission
+                        - username
     post:
       summary: create scheduled change for permissions.
       description: >
@@ -2111,6 +2179,8 @@ paths:
               - $ref: '#/definitions/OptionsJSON'
             required:
               - change_type
+              - permission
+              - username
       responses:
         '200':
           description: Returns sc_id of change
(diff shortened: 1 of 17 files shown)


auslib/web/admin/swagger/api.yaml
```diff
@@ -2977,31 +3675,38 @@ definitions:
           minimum: 1
           example: 2
 
-  SCGetModel:
-    title: SC GET
-    description: "model for defining common fields in scheduled changes GET operations"
+  SCMiscModel:
```
mozbhearsum: Please find a better name for this model - "misc" is not descriptive.

auslib/web/admin/views/rules.py
```diff
@@ -320,42 +318,68 @@ def _post(self, transaction, changed_by):
...
-            form = EditScheduledChangeNewRuleForm()
+            return problem(404, "Not Found", "Unknown sc_id",
+                           ext={"exception": "No scheduled change for rule found for given sc_id"})
+
+        what = {}
+        for field in connexion.request.get_json():
+            # Unlike when scheduling a new change to an existing rule, rule_id is not
+            # required (or even allowed) when modifying a scheduled change for an
+            # existing rule. Allowing it to be modified would be confusing.
+            if (field in ["csrf_token", "rule_id"] or (change_type == "insert" and field == "data_version") or
```
mozbhearsum: As discussed over video, ignore this - I was getting confused between creating and editing scheduled changes (rule_id is only required for the former).

auslib/web/admin/views/rules.py
```diff
@@ -75,10 +73,13 @@ def _post(self, transaction, changed_by):
...
 
-        if what.get('mapping', None) is not None and len(mapping_values) != 1:
+        if not what.get('mapping', None):
+            return problem(400, 'Bad Request', 'mapping value cannot be set to null/empty')
+
+        elif what.get('mapping', None) and len(mapping_values) != 1:
             return problem(400, 'Bad Request', 'Invalid mapping value. No release name found in DB')
 
-        if what.get('fallbackMapping', None) is not None and len(fallback_mapping_values) != 1:
+        elif what.get('fallbackMapping') is not None and len(fallback_mapping_values) != 1:
```
aksareen: just to improve readability and couple these statements as a single check

auslib/web/admin/views/permissions.py
```diff
@@ -172,17 +170,46 @@ def _post(self, transaction, changed_by):
...
-            form = EditScheduledChangeNewPermissionForm()
+            return problem(404, "Not Found", "Unknown sc_id",
+                           ext={"exception": "No scheduled change for permission found for given sc_id"})
+
+        # TODO: UI passes too many extra non-required fields apart from 'change_type' in request body
+        # Only required fields must be passed to DB layer
+        what = {}
+        for field in connexion.request.get_json():
+
+            # When editing an existing Scheduled Change for a Permission only options may be
```
aksareen: changing it

auslib/web/admin/views/releases.py
```diff
@@ -548,24 +546,47 @@ def _post(self, transaction, changed_by):
...
-        if form.data.data:
-            form.data.data = createBlob(form.data.data)
-        return super(ReleaseScheduledChangeView, self)._post(sc_id, form, transaction, changed_by)
+            return problem(404, "Not Found", "Unknown sc_id",
+                           ext={"exception": "No scheduled change for release found for given sc_id"})
+
+        what = {}
+        for field in connexion.request.get_json():
+            # Only data may be changed when editing an existing Scheduled Change for
+            # a Release. Name cannot be changed because it is a PK field, and product
```
aksareen: changing it

auslib/web/admin/views/rules.py
```diff
@@ -117,10 +118,13 @@ def _post(self, id_or_alias, transaction, changed_by):
...
 
-        if what.get('mapping', None) is not None and len(mapping_values) != 1:
+        if 'mapping' in what and not what.get('mapping', None):
+            return problem(400, 'Bad Request', 'mapping value cannot be set to null/empty')
+
+        elif what.get('mapping', None) and len(mapping_values) != 1:
             return problem(400, 'Bad Request', 'Invalid mapping value. No release name found in DB')
 
-        if what.get('fallbackMapping', None) is not None and len(fallback_mapping_values) != 1:
+        elif what.get('fallbackMapping') is not None and len(fallback_mapping_values) != 1:
```
aksareen: reason being that empty string "" can be passed in the input and comparing truthfulness is to be a better option of handling both None and '' together. ( int(0) or bool(False) will never be passed for a string field 'mapping'). Hence this check : what.get('mapping') covers all.

When 

> 'mapping' in what and not what.get('mapping', None) 

condition is present , it implies that if mapping was present in request.json and if has a Null/Blank value then execute the following statement.

Whereas:
> not what.get('mapping', None) 
statement checks has 2 checks in 1:  (a) if mapping was passed in request.json, then check if it's value was blank/empty . (b) if mapping wasn't passed at all in request.json then perform  following statement . This condition is only used in "change_type" == "insert" cases.

fallbackMapping can be null/blank but mapping cannot be.

auslib/web/admin/views/rules.py
```diff
@@ -318,44 +322,84 @@ def _post(self, transaction, changed_by):
...
-            form.mapping.choices = [(item['name'], item['name']) for item in release_names]
-            form.mapping.choices.insert(0, ('', 'NULL'))
+        if change_type == "update" and not what.get("data_version", None):
+            return problem(400, "Bad Request", "Missing field", ext={"exception": "data_version is missing"})
 
-        return super(RuleScheduledChangeView, self)._post(sc_id, form, transaction, changed_by)
+        elif change_type == "insert":
+            # edit scheduled change for new rule
+            for field in ["update_type", "backgroundRate", "priority"]:
+                if field in what and not what.get(field):
```
aksareen: yup. editing this condition

auslib/web/admin/views/rules.py
mozbhearsum: I don't think you need to worry about checking empty strings initially. If those make it past the not null checks, mapping_values will be zero, and we will get an "Invalid mapping value" - won't we?

aksareen: No we won't due to this [line](https://github.com/mozilla/balrog/blob/master/auslib/web/admin/views/rules.py#L30) , I'll have to edit it to move it [here](https://github.com/mozilla/balrog/blob/master/auslib/web/admin/views/rules.py#L45) so that blank field values are invalid for mapping but not for fallbackMapping. Should I do it ?

aksareen: Update: This [line](https://github.com/mozilla/balrog/blob/master/auslib/web/admin/views/rules.py#L36) will convert all blank fields to None so, we need only check for null values for 'mapping' and 'fallbackMapping'. But we still need to check that 'mapping_values' didn't have [null] by moving this [line](https://github.com/mozilla/balrog/blob/master/auslib/web/admin/views/rules.py#L30) to [here](https://github.com/mozilla/balrog/blob/master/auslib/web/admin/views/rules.py#L45)

aksareen: Update 2: this [line](https://github.com/mozilla/balrog/blob/master/auslib/web/admin/views/rules.py#L30) is irrelevant because [here](https://github.com/mozilla/balrog/blob/master/auslib/web/admin/views/rules.py#L36) we replace all blank field values to None , hence while constructing the 'mapping_values' list on these lines [1](https://github.com/mozilla/balrog/blob/master/auslib/web/admin/views/rules.py#L44), [2](https://github.com/mozilla/balrog/blob/master/auslib/web/admin/views/rules.py#L45), x i.e. key will never match a blank value from rule_form_dict.get("mapping", None) since it has been set to None if it was blank initially and hence the list will be empty. So no issues. If mapping value was indeed provided and list was empty then 400 response will be sent.

auslib/web/admin/views/rules.py
```diff
@@ -117,10 +118,14 @@ def _post(self, id_or_alias, transaction, changed_by):
 
         what, mapping_values, fallback_mapping_values = process_rule_form(connexion.request.get_json())
 
-        if what.get('mapping', None) is not None and len(mapping_values) != 1:
+        # If 'mapping' key is present in request body but is either blank/null
+        if 'mapping' in what and not what.get('mapping', None):
```
mozbhearsum: I'm still finding these blocks really difficult to read. I think it would help if you expanded them out a bit, and used separate `if` blocks for mapping and fallbackMapping. Eg, something like:
```
if "mapping" in what:
  if what.get("mapping"):
    if len(mapping_values) != 1:
      ...
  else:
    ...

if what.get("fallbackMapping") is not None and len(fallback_mapping_values) != 1:
  ...
```

auslib/web/admin/views/rules.py
```diff
@@ -318,44 +323,89 @@ def _post(self, transaction, changed_by):
...
+                                                     "when scheduling insertion of a new rule" % field})
 
         if change_type in ["update", "insert"]:
             rule_dict, mapping_values, fallback_mapping_values = process_rule_form(what)
             what = rule_dict
-            if what.get('mapping') is not None and len(mapping_values) != 1:
+
+            # Either change_type == "insert" and 'mapping' key-value is blank/null/not-present-in-request-body
+            # OR 'mapping' key is present in request body but is either blank/null
+            if not what.get('mapping', None) and (change_type == "insert" or ('mapping' in what)):
```
mozbhearsum: I think this block can be expanded out a bit to be made more readable, too.

auslib/web/admin/views/rules.py
aksareen: refactoring it.

auslib/web/admin/views/rules.py
aksareen: expanding it

PR #370, opened 2017-08-10
Remove queryVersion handle from validator.

ex-Dev:
Remove flask "defaults" workaround from validator. Keeping balrog spec more closely to swagger v2.


auslib/test/web/test_client.py
```diff
@@ -907,6 +908,54 @@ def testNonSubstitutedUrlVariablesReturn404(self):
                 self.assertEqual(ret.status_code, 404)
                 self.assertFalse(mock_cr_view.called)
 
+    def testUpdateQueryVersion(self):
+        version_key = 'queryVersion'
+        url_params = {}
+        update_query_version('/update/1/a/b/c.xml', url_params)
```
mozbhearsum: This looks like an excellent use for hypothesis! Rather than writing all of these out by hand, you can apply a decorator like:
```
@given(integers(min_value=1, max_value=X))
def testUpdateQueryVersion(self, qv):
```

...which will set ```qv``` to numerous different values for you.

auslib/web/public/client.py
```diff
@@ -93,7 +93,31 @@ def getQueryFromURL(url):
     return query
 
 
+def extract_query_version(request_url):
+    version = 0
+    pattern = '^.*/update/([1-6])/.*\.xml$'
```
mozbhearsum: It's probably good to futureproof this could by accepting any integer in this field.

auslib/web/public/client.py
```diff
@@ -93,7 +93,31 @@ def getQueryFromURL(url):
...
+    match = re.match(pattern, request_url)
+    if match:
+        version = int(match.group(1))
+    return version
+
+
+def update_query_version(request_url, url_params):
```
mozbhearsum: Could you rework this to be a pure function (return the query version instead of updating an input parameter)? This makes the code more readable and testable.

auslib/test/web/test_client.py
```diff
@@ -910,6 +911,23 @@ def testNonSubstitutedUrlVariablesReturn404(self):
...
+        self.assertEqual(query_version, v1_qv)
+
+    @given(just("x1"), just("1x"))
+    def testUpdateInvalidQueryVersion(self, invalid_prefix, invalid_suffix):
+        query_version = extract_query_version('/update/x1/a/b/c/update.xml')
+        self.assertEqual(query_version, 0)
+        query_version = extract_query_version('/update/1x/a/b/c/update.xml')
+        self.assertEqual(query_version, 0)
+
+    @given(integers(min_value=1, max_value=100))
```
mozbhearsum: I _think_ you can combine this with the first test by using: `@example`, like https://hypothesis.readthedocs.io/en/latest/details.html#providing-explicit-examples. r=me with that changed.

PR #371, opened 2017-08-14
Fix Bug 1370220

vikasmahato:


auslib/web/admin/views/permissions.py
```diff
@@ -44,7 +44,8 @@ def get(self, username):
                                detail="You are not authorized to view permissions of other users.")
 
         permissions = dbo.permissions.getUserPermissions(username)
-        if not permissions:
+
+        if username != current_user and not permissions:
```
mozbhearsum: This fixes the bug, but it actually introduces a new one: if a User without permissions queries for their own permissions by name (eg: a GET to "/users/vikas"), a 200 is returned instead of a 404. We _only_ want a 200 when no permissions exist for the specific case of "/users/current". If a user is explicitly named, we should return a 404 if they have no permissions.

auslib/test/admin/views/test_permissions.py
```diff
@@ -57,6 +57,17 @@ def testGetCurrentUserWithoutRoles(self):
...
+        ret = self._get("/users/current", username="vikas")
+        self.assertEqual(ret.status_code, 200)
+        data = json.loads(ret.data)
+        expected = {
+            "username": "vikas",
+            "permissions": {},
+            "roles": {},
+        }
+        self.assertEqual(data, expected)
+
```
mozbhearsum: This new test is good, nice work!

auslib/web/admin/views/permissions.py
mozbhearsum: Those tests are both correct, but keep in mind that even if someone _else_ was querying "/users/vikas", it should still return a 404. Put another way: the only thing that factors into the decision is whether or not the username being queried is "current" or not.

Does that make sense?

vikasmahato:
@mozbhearsum Can youplease review my pull request? I have updated it.

nthomas-mozilla:
Ben will probably have a look next week.

mozbhearsum:
Looks good now - nice work!

vikasmahato:
Thanks

PR #372, opened 2017-08-16
bug 1386756: add support for mig64 query parameter

mozbhearsum:
This adds support for the mig64 query parameter as described in https://bugzilla.mozilla.org/show_bug.cgi?id=1386756#c0. Between that, the comments, and the doc updates I _think_ this should be reasonable self explanatory - please let me know if anything is unclear.

In addition to the unit tests I verified this by ensuring that I could create and update both Rules and Rule Scheduled Changes with all mig64 values, and that every possibility in the truth table was correct.

I'm not sure the UI is the best, but I couldn't come up with any better idea than a drop down. Because the parameter is so opaque I tried to add some helpful text to the UI.

### DIFF
--- a/auslib/db.py
+++ b/auslib/db.py
@@ -1466,8 +1466,10 @@ class Rules(AUSTable):
             self.table.append_column(Column("mig64", BIT))
         else:
             # It's not ideal to disable the constraint here, but if we leave it enabled
-            # the database cannot be downgraded (t
-            self.table.append_column(Column("mig64", Boolean(create_constraint=False)))
+            # the database cannot be downgraded (dropping the column will try to remove
+            # the column, but not the constraint). This is sqlite-only, so it only
+            # applies to unit tests, not deployed instances.
+            self.table.append_column(Column("mig64", Integer))
 
         AUSTable.__init__(self, db, dialect, scheduled_changes=True)
 
--- a/auslib/db.py
+++ b/auslib/db.py
@@ -1461,6 +1461,14 @@ class Rules(AUSTable):
                            Column('comment', String(500)),
                            )
 
+        if dialect == 'mysql':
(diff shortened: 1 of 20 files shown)


ui/app/templates/rule_scheduled_change_modal.html
```diff
@@ -205,10 +205,29 @@ <h4 ng-show= "sc.change_type !== 'delete'">Rule Details</h4>
         </div>
       </div>
       <div class="col-md-6" ng-show= "sc.change_type !== 'delete'">
+        <div class="form-group" ng-class="{'has-error': errors.mig64}">
```
nthomas-mozilla: BTW, I noticed that mig64 has different position in Edit Rule and Schedule Change dialogs, which turns out to be because Alias isn't in the latter. Is that deliberate ?

auslib/db.py
```diff
@@ -1703,6 +1731,26 @@ def getRawMatches():
             if not self._localeMatchesRule(rule['locale'], updateQuery['locale']):
                 self.log.debug("%s doesn't match %s", rule['locale'], updateQuery['locale'])
                 continue
+            # As with all other columns, if mig64 isn't present in the Rule, the Rule matches.
+            # Unlike other columns, mig64 is optional in the update query, so we must handle False,
+            # True, and None values. Note that None in the updateQuery is treated as "unknown",
```
mozbhearsum: Will do.

ui/app/templates/rule_scheduled_change_modal.html
mozbhearsum: Hmm, that's a good point. I'll move it ahead of Alias for consistency - it makes sense to keep all of the fields that affect rule table parsing grouped together.

PR #373, opened 2017-08-18
Return 404 for non-update endpoints

allan-silva:
Bugzilla: 1391294
Return 404 for non-update endpoints

### DIFF
--- a/auslib/test/web/test_client.py
+++ b/auslib/test/web/test_client.py
@@ -1054,6 +1054,8 @@ class ClientTestWithErrorHandlers(ClientTestCommon):
     @given(just('/whizzybang'))
     @example(path='/')
     @example(path='/api/v1')
+    @example(path='/api/v1/releases/')
+    @example(path='/api/v1/rules/')
     def test404ResponseForNonUpdateEndpoint(self, path):
         ret = self.client.get(path)
         self.assertEqual(ret.status_code, 404)
--- a/auslib/test/web/test_client.py
+++ b/auslib/test/web/test_client.py
@@ -1048,8 +1048,15 @@ class ClientTestWithErrorHandlers(ClientTestCommon):
             self.assertEqual(ret.headers.get("X-Content-Type-Options"), "nosniff")
 
     def testEmptySnippetOn404(self):
-        ret = self.client.get('/whizzybang')
+        ret = self.client.get('/update/whizzybang')
         self.assertUpdatesAreEmpty(ret)
+
+    @given(just('/whizzybang'))
+    @example(path='/')
+    @example(path='/api/v1')
+    def test404ResponseForNonUpdateEndpoint(self, path):
+        ret = self.client.get(path)
+        self.assertEqual(ret.status_code, 404)
 
     def testErrorMessageOn500(self):
         with mock.patch('auslib.web.public.client.getQueryFromURL') as m:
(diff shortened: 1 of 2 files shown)


auslib/test/web/test_client.py
```diff
@@ -1048,9 +1048,16 @@ def testXContentTypeOptionsIsSetFor500(self):
             self.assertEqual(ret.headers.get("X-Content-Type-Options"), "nosniff")
 
     def testEmptySnippetOn404(self):
-        ret = self.client.get('/whizzybang')
+        ret = self.client.get('/update/whizzybang')
         self.assertUpdatesAreEmpty(ret)
 
+    @given(just('/whizzybang'))
+    @example(path='/')
+    @example(path='/api/v1')
```
mozbhearsum: It would be good to include something under /api/v1/rules/ and /api/v1/releases/, too.

auslib/web/public/base.py
```diff
@@ -61,10 +61,12 @@ def apply_security_headers(response):
...
-    """We don't return 404s in AUS. Instead, we return empty XML files"""
-    response = make_response('<?xml version="1.0"?>\n<updates>\n</updates>')
-    response.mimetype = 'text/xml'
-    return response
+    if re.match("^/update", request.path):
+        """We don't return 404s in AUS. Instead, we return empty XML files"""
+        response = make_response('<?xml version="1.0"?>\n<updates>\n</updates>')
+        response.mimetype = 'text/xml'
+        return response
+    return Response(status=404, mimetype="text/plain")
```
mozbhearsum: I think it would be better to return the original error instead of making your own - you may be losing some information by doing this. You might need to raise it instead of returning, I don't recall exactly how this works...

auslib/web/public/base.py
```diff
@@ -61,10 +61,12 @@ def apply_security_headers(response):
 
 @app.errorhandler(404)
 def fourohfour(error):
-    """We don't return 404s in AUS. Instead, we return empty XML files"""
-    response = make_response('<?xml version="1.0"?>\n<updates>\n</updates>')
-    response.mimetype = 'text/xml'
-    return response
+    if re.match("^/update", request.path):
+        """We don't return 404s in AUS. Instead, we return empty XML files"""
```
mozbhearsum: This comment could use an update now that we *do* return some 404s.

auslib/web/public/base.py
ex-Dev: The error parameter is a NotFound exception. I'll test what happens when the exception is raised.

ex-Dev: The error handle expects an exception, see the flow:
https://github.com/pallets/flask/blob/c74f46979a8b8358437bd7f76e478d04248a9c72/flask/app.py#L1727
https://github.com/pallets/flask/blob/c74f46979a8b8358437bd7f76e478d04248a9c72/flask/app.py#L1606
https://github.com/pallets/flask/blob/c74f46979a8b8358437bd7f76e478d04248a9c72/flask/app.py#L1638
https://github.com/pallets/flask/blob/c74f46979a8b8358437bd7f76e478d04248a9c72/flask/app.py#L1561

mozbhearsum:
I think this looks fine now...I'm going to take another look with fresh eyes next week before merging.

PR #374, opened 2017-08-20
fix bug-1379277

bhavishyagopesh:
I'm not sure this has to be done in this way or not. Writing an error handler is considered better but I'm not sure...need instructions!


bhavishyagopesh:
I thought `testFieldViewBadValuesBadChangeId` does that kind of testing...How should the new test be different from `testFieldViewBadValuesBadChangeId`

mozbhearsum:
@bhavishyagopesh You're right - they're pretty much the same. The only difference is that "data" is actually a column that exists on the Releases table, whereas "whatever" doesn't, so your new test is exclusively verifying that bad change ids are handled correctly.

PR #377, opened 2017-08-22
Remove unsubstituted_url_variables handler

glasserc:
This was added by https://bugzilla.mozilla.org/show_bug.cgi?id=1282569
because we were apparently encountering 400s tens of thousands of
times per day. However, it isn't clear why. Removing the handler falls
back to the real one, which fails to find any information and
generates the exact same response. This adds a tiny amount of database
load but simplifies the code.

### DIFF
--- a/auslib/test/web/test_client.py
+++ b/auslib/test/web/test_client.py
@@ -893,24 +893,6 @@ class ClientTest(ClientTestBase):
         ret = self.client.get('/update/3/product_that_should_not_be_updated/1.0/1/p/l/a/a/a/a/update.xml')
         self.assertUpdatesAreEmpty(ret)
 
-    def testNonSubstitutedUrlVariablesReturn404(self):
-        request1 = '/update/1/%PRODUCT%/%VERSION%/%BUILD_ID%/%BUILD_TARGET%/%LOCALE%/%CHANNEL%/update.xml'
-        request2 = '/update/2/%PRODUCT%/%VERSION%/%BUILD_ID%/%BUILD_TARGET%/%LOCALE%/%CHANNEL%/%OS_VERSION%/update.xml'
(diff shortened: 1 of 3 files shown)


mozbhearsum:
For reasons I don't understand, this has caused the endpoints in question to generate 400s in stage, eg: https://aus4.stage.mozaws.net/update/2/%PRODUCT%/%VERSION%/%BUILD_ID%/%BUILD_TARGET%/%LOCALE%/%CHANNEL%/%OS_VERSION%/update.xml

Curiously, the same endpoints return the 200 empty xml when I test locally.

glasserc:
That's really strange. The fact that it's an nginx 400 rather than a Balrog 400 seems suspicious. Do we have access to logs or anything? Can we see what translated request is hitting the application and/or what the application responded with? I tried this locally too and got a 200 empty XML. Is it something with the prod database being different from my local database? I guess the public API translates BadDataError into a 400.

mozbhearsum:
I think it's unlikely that database state is a factor. The local dev environment imports prod dumps, so your state should be very similar to stage/prod. I'll try to dig up some logs.

mozbhearsum:
Looking at this again, I think my test request was invalid. If I urlencode the "%", I get the expected empty XML. In local dev, we only have uwsgi in the mix, which is probably less strict than nginx about this.

glasserc:
Does that mean we believe the 400 response is correct? I think either way, it's outside the scope of this handler, so removing it is still OK, right?

mozbhearsum:
I think that there's nothing we can do about the 400's when there's a non-escaped "%" in the URL -- nginx seems to balk at that. Definitely outside the scope of the handler though, as you said. Both the escaped and unescaped forms of the URLs have exactly the same behaviour before and after this PR.

PR #378, opened 2017-08-22
Get rid of spec-e2e

glasserc:
All of this e2e code was developed back in the dark ages of 2013 ([example](https://github.com/mozilla/balrog-ui/commit/eb033b8fd7beeff941302535a1a4ac9d12768354)) and appears to have become quite bit-rotten. Get rid of it because it's the only reason we seem to have a CoffeeScript dependency. Mozilla is, essentially, the JavaScript company, so let's drop CoffeeScript if it doesn't buy us anything. (We still indirectly depend on CoffeeScript through lineman, but one step at a time.)


mozbhearsum:
Thank you, this kind of clean is great!

PR #379, opened 2017-08-25
Fix silly typos affecting editing scheduled deletions

mozbhearsum:
Discovered this while doing some other work - it's fallout from #372.


mozbhearsum:
No worries, so did I! :)

PR #382, opened 2017-08-28
Force a fallback mapping

glasserc:
Currently it's possible to cause a request to definitely receive the up-to-date mapping using the `force=1` argument. However, it isn't possible to cause a request to definitely receive the fallback mapping. Add this by adding a `force=0` argument.

While we're here, refactor the AUS tests a bit.

This PR, being somewhat involved, may be easiest to read commit-by-commit.

Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1391668.

### DIFF
--- a/auslib/AUS.py
+++ b/auslib/AUS.py
@@ -35,10 +35,8 @@ def getFallbackChannel(channel):
 class AUSRandom:
     """Abstract getting a randint to make it easier to test the range of
     possible values"""
-
-    def __init__(self, min=0, max=99):
-        self.min = min
-        self.max = max
+    min = 0
+    max = 99
 
     def getInt(self):
         return randint(self.min, self.max)
--- a/auslib/AUS.py
+++ b/auslib/AUS.py
@@ -6,6 +6,17 @@ import logging
 from auslib.global_state import dbo
 
 
+class ForceResult(object):
+    """Enumerated "result" class that represents a non-random result chosen by a caller."""
+    def __init__(self, name):
+        self.name = name
+
+
+# Magic constants that callers can use to choose a specific "random" result.
+SUCCEED = ForceResult('succeed')
+FAIL = ForceResult('fail')
+
+
 def isSpecialURL(url, specialForceHosts):
     if not specialForceHosts:
         return False
@@ -76,9 +87,9 @@ class AUS:
         # serve every request an update
         # backgroundRate=100 means all requests are served
         # backgroundRate=25 means only one quarter of requests are served
-        if not updateQuery['force'] and rule['backgroundRate'] < 100:
+        if not updateQuery['force'] == SUCCEED and rule['backgroundRate'] < 100:
             self.log.debug("backgroundRate < 100, rolling the dice")
-            if self.rand() >= rule['backgroundRate']:
+            if updateQuery['force'] == FAIL or self.rand() >= rule['backgroundRate']:
                 fallbackReleaseName = rule['fallbackMapping']
                 if fallbackReleaseName:
                     release = dbo.releases.getReleases(name=fallbackReleaseName, limit=1)[0]
--- a/auslib/AUS.py
+++ b/auslib/AUS.py
@@ -44,9 +44,9 @@ class AUSRandom:
 
 class AUS:
 
-    def __init__(self):
+    def __init__(self, rand=None):
         self.specialForceHosts = None
-        self.rand = AUSRandom()
+        self.rand = rand or AUSRandom()
         self.log = logging.getLogger(self.__class__.__name__)
 
     def evaluateRules(self, updateQuery):
--- a/auslib/AUS.py
+++ b/auslib/AUS.py
@@ -8,13 +8,14 @@ from auslib.global_state import dbo
 
 class ForceResult(object):
     """Enumerated "result" class that represents a non-random result chosen by a caller."""
-    def __init__(self, name):
+    def __init__(self, name, query_value):
         self.name = name
+        self.query_value = query_value
 
 
 # Magic constants that callers can use to choose a specific "random" result.
-SUCCEED = ForceResult('succeed')
-FAIL = ForceResult('fail')
+SUCCEED = ForceResult('succeed', '1')
+FAIL = ForceResult('fail', '-1')
 
 
 def isSpecialURL(url, specialForceHosts):
--- a/auslib/AUS.py
+++ b/auslib/AUS.py
@@ -43,9 +43,6 @@ class AUSRandom:
     def getInt(self):
         return randint(self.min, self.max)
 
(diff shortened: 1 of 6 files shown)


auslib/web/public/client.py
```diff
@@ -85,7 +85,8 @@ def getQueryFromURL(url):
     query['osVersion'] = urllib.unquote(query['osVersion'])
     ua = request.headers.get('User-Agent')
     query['headerArchitecture'] = getHeaderArchitecture(query['buildTarget'], ua)
-    query['force'] = (int(query.get('force', 0)) == 1)
+    force = query.get('force')
+    query['force'] = {'1': SUCCEED, '0': FAIL}.get(force)
```
allan-silva: How there are two specific options to always get a fallback release or always get the up-to-date release, maybe the force values  can be different of zero (i.e 1 succed, 2 fail), and ignore all other "false" cases. Maybe change query string name to something more close to this context.
What you think?

glasserc: I don't have strong opinions on the API itself. However, I assumed we had to maintain backwards compatibility with Firefoxes that are already "in the wild", which will continue passing `force=1`. But my inclination is that "force" isn't the worst name, since we're forcing the random number to either be successful (not be throttled) or fail (be throttled). And of course we have to distinguish "force to succeed" and "force to fail" from "no forcing". It seems more natural to me to have 0 represent force to fail and no query argument, or an empty string, represent "no forcing", rather than 2 represent force to fail and 0 represent "no forcing".

Of course, internally we have to have three distinct values to represent these three requests, and for those I chose to define two arbitrary constants `SUCCESS` and `FAIL`, neither of which are falsy, and use `None` for "no forcing". I could have just as easily used `True` and `False`, but I think it's good to define a unique enumeration for a unique use like this one.

allan-silva: Oh!Yeah, makes all sense the querystring name not change, sorry!
I commented mainly by the "force=0", this seems say "do not force anything".

glasserc: I'd much prefer `force=succeed` and `force=fail`, but I'm not sure how to make that happen. I think `force=1` and `force=fail` would be a bit strange but I'm willing to go for it if you are. `force=1` and `force=-1` might be a compromise.

allan-silva: I like force=-1

auslib/AUS.py
```diff
@@ -35,23 +46,18 @@ def getFallbackChannel(channel):
 class AUSRandom:
     """Abstract getting a randint to make it easier to test the range of
```
mozbhearsum: The new version of this class is merely a tiny wrapper around randint() now - can we just call it directly, instead? I don't know that enforcing the min/max values really matters all that much since the caller is just a few lines down.

auslib/AUS.py
```diff
@@ -35,23 +46,18 @@ def getFallbackChannel(channel):
...
-        return range(self.min, self.max + 1)
-
 
 class AUS:
 
-    def __init__(self):
+    def __init__(self, rand=None):
         self.specialForceHosts = None
-        self.rand = AUSRandom()
+        self.rand = rand or AUSRandom().getInt
```
mozbhearsum: We tend to use mocking instead of dependency injection in Balrog - can you do that here?

auslib/test/test_AUS.py
```diff
@@ -3,56 +3,49 @@
...
+            name='fallback', product='c', data_version=1,
+            data=createBlob({"name": "fallback", "extv": "1.0", "schema_version": 1, "platforms": {"a": {"buildID": "1", "locales": {"a": {}}}}}))
+
+    def tearDown(self):
+        dbo.reset()
 
+    def random_aus_test(self, background_rate, force=None, fallback=False):
+        mapping = 'b'
+        with mock.patch('auslib.db.Rules.getRulesMatchingQuery') as m:
+            fallback = fallback and 'fallback'  # convert True to string
```
mozbhearsum: I'm not sure I understand this line. The string 'fallback' will always evaluate to True, so it seems like this is the same as bool(fallback)? If so, I don't think this will set the value correctly. The returned value from the database should be the name of the fallback release ("fallback" in these tests) if present, or None if there isn't a fallbackMapping specified.

I could be misunderstanding something here, so feel free to correct me.

mozbhearsum:
One more thing I just noticed: the pass-force-to-some-hosts isn't working right now. We end up passing "force=1" even if force=-1. Eg:
```
$ curl 'http://localhost:9090/update/3/Firefox/54.0/20141202185629/Darwin_x86_64-gcc3-u-i386-x86_64/en-US/release/default/default/default/update.xml?force=-1'
<?xml version="1.0"?>
<updates>
    <update type="minor" displayVersion="54.0.1" appVersion="54.0.1" platformVersion="54.0.1" buildID="20170628075643" detailsURL="https://www.mozilla.org/en-US/firefox/54.0.1/releasenotes/">
        <patch type="complete" URL="http://download.mozilla.org/?product=firefox-54.0.1-complete&amp;os=osx&amp;lang=en-US&amp;force=1" hashFunction="sha512" hashValue="d8290246666a10d66c50c4afece0c1b32f13c78b6f09f05c8d7ad87ea92e1351cac0ee2458bc364afed8ba11209eaaaf4e91993a0225a9d20fa311c755be479d" size="58080672"/>
    </update>
</updates>
```

This is the code that handles that: https://github.com/mozilla/balrog/blob/master/auslib/blobs/base.py#L123

auslib/AUS.py
glasserc: I thought it was nice to have a clear "statement" that the min/max values the random number generator should return are 0 and 99. But I've gotten rid of it.

auslib/AUS.py
glasserc: OK.

auslib/test/test_AUS.py
glasserc: Another way to write this line of code is

```python
if fallback:
    fallback = 'fallback'
```

In other words, the `and` operator takes the first falsy value, or the last value if all are truthy. Some examples:

```python
>>> 'abc' and 'def'
'def'
>>> "" and 'def'
""
>>> False and 'def'
False
>>> True and 'def'
'def'
```

Here, the point is to convert `True` (meaning "yes, turn on `fallback`, please") becomes the string `'fallback'`, since `True` is more convenient at the API layer but not a valid fallback release (since it has to be a name, i.e. string).

glasserc:
I didn't know about the special hosts thing. Thanks for pointing it out. I've addressed this too.

auslib/test/test_AUS.py
mozbhearsum: Huh, TIL. Thanks for the detailed explanation.

PR #384, opened 2017-09-04
Fix:#1379285 handle requests that set bad values for "force" query parameter

Rajesh-Veeranki:
Ref: https://bugzilla.mozilla.org/show_bug.cgi?id=1379285

### DIFF
--- a/auslib/test/web/test_client.py
+++ b/auslib/test/web/test_client.py
@@ -6,8 +6,8 @@ import unittest
 from xml.dom import minidom
 import json
 
-from hypothesis import example, given
-from hypothesis.strategies import integers, just
+from hypothesis import assume, example, given
+from hypothesis.strategies import integers, just, text
 
 import auslib.web.public.client as client_api
 from auslib.web.public.client import extract_query_version
@@ -804,6 +804,23 @@ class ClientTest(ClientTestBase):
     </update>
 </updates>
 """)
+
+    @given(text(max_size=128))
+    @example('')
+    @example('1" name="Firefox 54.0" isOSUpdate="false" installDate="1498012260998')
+    @example('1)')
+    @example('"|sleep 7 #')
+    def testForceParamWithBadInputs(self, x):
+        assume(x != '1')
+        force_output = """<?xml version="1.0"?>
+<updates>
+    <update type="minor" version="1.0" extensionVersion="1.0" buildID="2">
+        <patch type="complete" URL="http://a.com/z" hashFunction="sha512" hashValue="4" size="3"/>
+    </update>
+</updates>
+"""
+        ret = self.client.get("/update/4/b/1.0/1/p/l/a/a/a/a/1/update.xml?force=%s" % x)
+        self.assertUpdateEqual(ret, force_output)
 
     def testDeprecatedEsrVersionStyleGetsUpdates(self):
         ret = self.client.get('/update/3/b/1.0esrpre/1/p/l/a/a/a/a/update.xml')
--- a/auslib/test/web/test_client.py
+++ b/auslib/test/web/test_client.py
@@ -7,7 +7,7 @@ from xml.dom import minidom
 import json
 
 from hypothesis import assume, example, given
-from hypothesis.strategies import integers, just, text
+from hypothesis.strategies import characters, integers, just, text
 
 import auslib.web.public.client as client_api
 from auslib.web.public.client import extract_query_version
@@ -805,7 +805,7 @@ class ClientTest(ClientTestBase):
 </updates>
 """)
 
-    @given(text(max_size=128))
+    @given(text(alphabet=characters(blacklist_categories=('Cs', 'Po')), max_size=128))
     @example('')
     @example('1" name="Firefox 54.0" isOSUpdate="false" installDate="1498012260998')
     @example('1)')
--- a/requirements.txt
+++ b/requirements.txt
@@ -118,9 +118,9 @@ dictdiffer==0.6.1 \
 raven==6.1.0 \
     --hash=sha256:56dc9062dd42bca97350e5048ff417c914376366caa3b1b5f788b27ddc0a34b7 \
     --hash=sha256:02cabffb173b99d860a95d4908e8b1864aad1b8452146e13fd7e212aa576a884
-connexion==1.1.11 \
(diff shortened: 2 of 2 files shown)


allan-silva:
glasserc is going to finish a new feature for the "force" flag in #382 that will overrides the logic for it, however, you wrote good test cases, maintain your the tests would be very useful.

auslib/test/web/test_client.py
```diff
@@ -805,6 +805,24 @@ def testAvastURLsWithGoodQueryArgs(self):
...
+    </update>
+</updates>
+"""
+        inputs = [
+            '',
+            '1" name="Firefox 54.0" isOSUpdate="false" installDate="1498012260998',
+            '1)',
+            '"|sleep 7 #'
+        ]
+        for i in inputs:
```
allan-silva: You can use hypothesis package to help you on test cases like that.
Here an usage example:
https://github.com/mozilla/balrog/blob/master/auslib/test/web/test_client.py#L905

Take a look at Hypothesis docs too if you want:
https://hypothesis.readthedocs.io/en/latest/#

Rajesh-Veeranki: That looks interesting. Will use it and refactor the test cases.

Rajesh-Veeranki:
Thanks for the review. So should I remove my changes and just keep the tests?

allan-silva:
Yes, #382 was meged today.
You can update your branch and keep the test cases. The tests should pass if all is ok.

Rajesh-Veeranki:
I used `hypothesis.strategies.text` module for the test cases. It fails with an exception for the value `\x80`. Is this a valid value for the input? or should I limit test cases to only ASCII values? 

```
---------------------------------- Hypothesis ----------------------------------
Falsifying example: testForceParamWithBadInputs(self=<auslib.test.web.test_client.ClientTest testMethod=testForceParamWithBadInputs>, x=u'\x80')


value = '\x80', type = 'string'

    def make_type(value, type):
        type_func = TYPE_MAP[type]  # convert value to right type
>       return type_func(value)
E       UnicodeEncodeError: 'ascii' codec can't encode character u'\x80' in position 0: ordinal not in range(128)

.tox/py27/lib/python2.7/site-packages/connexion/decorators/parameter.py:51: UnicodeEncodeError
```

allan-silva:
Ah, Hypothesis uses unicode by default, but we don't have handle for unicode for query parameters (only for path parameters). 
If @mozbhearsum  have no objections, you can do something like this:

`@given(text(alphabet=characters(min_codepoint=0x0, max_codepoint=0x7F, blacklist_categories=('Cs', 'Po')), max_size=128))`

Basicly this uses only ascii chars and ignores caracters that breaks path.

allan-silva:
@Rajesh-Veeranki, your test found a problem that we had in the past for path parameters. Would be great handle unicode in query parameters in this bug if you want.

Path parametes are handled here: 
https://github.com/mozilla/balrog/blob/master/auslib/web/api_validator.py#L7
https://github.com/mozilla/balrog/blob/master/auslib/web/public/base.py#L26

Some connexion references:
http://connexion.readthedocs.io/en/latest/request.html#custom-validators
https://github.com/zalando/connexion/blob/master/connexion/decorators/validation.py#L165

Rajesh-Veeranki:
@allan-silva ,  Looks like `connexion` library fixed this issue in their new release `1.1.14` 
Related PR https://github.com/zalando/connexion/issues/485
Should we try updating the library and see if it fixes this issue?

allan-silva:
Yes, go ahead. We are in "1.1.11" yet.
After update, should not be necessary codepoint range anymore. Ensure that all test pass and this is OK for me.

allan-silva:
@Rajesh-Veeranki, you can try `./run-tests.sh` locally to test your changes

This code snipet removes the punctuation/other unicode category, may can help you:

`@given(text(alphabet=characters(blacklist_categories=('Cs', 'Po')), max_size=128))`

Rajesh-Veeranki:
I ran `./run-tests.sh auslib`. Earlier it seemed to have picked up `1.1.11` version. So I deleted `.tox` and re-ran. The `unicode` issue is solved but now we have a failing testcase with `&0` input

```
---------------------------------- Hypothesis ----------------------------------
Falsifying example: testForceParamWithBadInputs(self=<auslib.test.web.test_client.ClientTest testMethod=testForceParamWithBadInputs>, x=u'&0')
```

allan-silva:
Can you add the full output?

Tests fail even using `@given(text(alphabet=characters(blacklist_categories=('Cs', 'Po')), max_size=128))` ?

Rajesh-Veeranki:
It works on using `blacklist_categories=('Cs', 'Po')`
Modified my PR with these changes.

allan-silva:
@Rajesh-Veeranki, many thanks for the patch!
@mozbhearsum will give the final review and approvement.

auslib/test/web/test_client.py
```diff
@@ -805,6 +805,23 @@ def testAvastURLsWithGoodQueryArgs(self):
 </updates>
 """)
 
+    @given(text(alphabet=characters(blacklist_categories=('Cs', 'Po')), max_size=128))
+    @example('')
+    @example('1" name="Firefox 54.0" isOSUpdate="false" installDate="1498012260998')
+    @example('1)')
+    @example('"|sleep 7 #')
+    def testForceParamWithBadInputs(self, x):
+        assume(x != '1')
```
mozbhearsum: This is very neat - I didn't know about assume().

auslib/test/web/test_client.py
```diff
@@ -805,6 +805,23 @@ def testAvastURLsWithGoodQueryArgs(self):
 </updates>
 """)
 
+    @given(text(alphabet=characters(blacklist_categories=('Cs', 'Po')), max_size=128))
```
mozbhearsum: What's the reason to blacklist some punctuation categories but not others?

allan-silva: I suggested ignore the punctuation/other unicode category (Po), because it can insert caracters that connexion assumes as extra parameters that we not have in spec. Ignore "Cs" category is a default behavior of text() when alphabet is none.

mozbhearsum: Got it - thanks for the explanation.

Rajesh-Veeranki:
Thank you! :)

PR #385, opened 2017-09-04
bug 1353225: Remove strings on release revision pages.

Alweezy:
Currently, the release revision pages show a field called **Not mapped to** , which has no value. This PR removes the field.

### DIFF
--- a/auslib/test/admin/views/test_permissions.py
+++ b/auslib/test/admin/views/test_permissions.py
@@ -57,6 +57,21 @@ class TestCurrentUserAPI_JSON(ViewTest):
         }
         self.assertEqual(data, expected)
 
+    def testGetCurrentUserWithoutRolesWithoutPermissions(self):
+        ret = self._get("/users/current", username="vikas")
+        self.assertEqual(ret.status_code, 200)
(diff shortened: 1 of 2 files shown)


ui/app/templates/releases.html
```diff
@@ -96,9 +96,6 @@ <h5 title="Data version">
       </h5>
 
       <h5 title="Rules referencing it" ng-switch="release.rule_ids">
-        <b ng-show="!release.rule_ids.length">
-            Not mapped to.
-        </b>
```
mozbhearsum: Not quite right - this template is used by both the primary Releases page (http://localhost:8080/releases) and Releases History (eg: http://localhost:8080/releases/Firefox-55.0.3-build2). We want this string (or the one below) to continue to show for the normal Releases page, but _not_ for the history pages.

So, what you need to do is hide this entire h5 tag if the History page is being shown. We do something similar for some of the buttons on lines like https://github.com/Alweezy/balrog/blob/3b5462fd99d9688ecc60222e3412cfd161fc904c/ui/app/templates/releases.html#L79.

PR #387, opened 2017-09-05
Fix bug 1163200: Replace add a new user with add a new permission

andela-hngerebara:
Replace "add a new user" with "add a new permission"
Created pop a dialog that allows one to enter a username and permission

### DIFF
--- a/ui/app/js/controllers/permissions_controller.js
+++ b/ui/app/js/controllers/permissions_controller.js
@@ -5,7 +5,6 @@ function($scope, $routeParams, $location, $timeout, Permissions, Search, $modal,
 
   $scope.loading = true;
   $scope.failed = false;
-
   $scope.username = $routeParams.username;
   if ($scope.username) {
     // history of a specific rule
@@ -77,49 +76,49 @@ function($scope, $routeParams, $location, $timeout, Permissions, Search, $modal,
   /* End filtering */
 
   $scope.openUpdateModal = function(user) {
-
+    $scope.is_edit = true;
     var modalInstance = $modal.open({
       templateUrl: 'permissions_modal.html',
       controller: 'UserPermissionsCtrl',
       backdrop: 'static',
-      // size: size,  // can be lg or sm
+      size: 'md',  // can be lg or md, sm
       resolve: {
+        users: function () {
+          return $scope.users;
+        },
+        is_edit: function () {
+          return $scope.is_edit;
+        },
         user: function () {
           return user;
         },
-        users: function () {
-          return $scope.users;
-        }
       }
     });
-
-
   };
   /* End openUpdateModal */
 
   $scope.openNewModal = function() {
-
+    $scope.is_edit = false;
     var modalInstance = $modal.open({
-      templateUrl: 'user_modal.html',
-      controller: 'NewUserCtrl',
+      templateUrl: 'permissions_modal.html',
+      controller: 'UserPermissionsCtrl',
       backdrop: 'static',
-      // size: 'sm',
+      size: 'md',
       resolve: {
         users: function () {
           return $scope.users;
-        }
+        },
+        is_edit: function () {
+          return $scope.is_edit;
+        },
+        user: function () {
+          return $scope.user;
+        },
       }
     });
-
-    modalInstance.result.then(function (user) {
-      $scope.users.push(user);
-      $scope.openUpdateModal(user);
-    }, function () {
-      // modal closed
-    });
-
   };
-  /* End openNewModal */
+    /* End openNewModal */
+
 
 
   $scope.openNewScheduledPermissionChangeModal = function(user) {
--- a/ui/app/js/controllers/user_edit_controller.js
+++ b/ui/app/js/controllers/user_edit_controller.js
@@ -270,15 +270,15 @@ function ($scope, $modalInstance, CSRF, Permissions, users, is_edit, user) {
     $scope.usersaved = true;
     CSRF.getToken()
     .then(function(csrf_token) {
-      if (!$scope.user.username) {
-        $scope.errors.nullValue = 'The username field cannot be empty';
-        sweetAlert("failed", $scope.errors.nullValue, "error");
-        return false;
-      }
       permission.options = permission.options_as_json;
       Permissions.updatePermission($scope.user.username, permission, csrf_token)
       .success(function(response) {
         permission.data_version = response.new_data_version;
+        $scope.user.permissions.push($scope.permission);
+        $scope.permission = {
+          permission: '',
+          options_as_json: ''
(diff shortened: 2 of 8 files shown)


mozbhearsum:
I'm not able to review this yet because the "Add a new Permission" button throws an error:

> Error: [$injector:unpr] Unknown provider: userProvider <- user

ui/app/js/controllers/permissions_controller.js
```diff
@@ -76,50 +75,55 @@ function($scope, $routeParams, $location, $timeout, Permissions, Search, $modal,
...
-      controller: 'UserPermissionsCtrl',
-      backdrop: 'static',
-      // size: size,  // can be lg or sm
-      resolve: {
-        user: function () {
-          return user;
-        },
-        users: function () {
-          return $scope.users;
+    $scope.openUpdateModal = function(user) {
```
mozbhearsum: Can you please revert the whitespace changes here? They've caused the indentation to be different than the rest of the file.

ui/app/js/controllers/user_edit_controller.js
```diff
@@ -130,8 +139,12 @@ function ($scope, $modalInstance, CSRF, Permissions, user, users) {
     $scope.saving = true;
     CSRF.getToken()
     .then(function(csrf_token) {
+      if (!$scope.user.username) {
+        $scope.errors.nullValue = 'The username field cannot be empty';
+        sweetAlert("failed", $scope.errors.nullValue, "error");
+        return false;
+      }
```
mozbhearsum: This is a good check. Make sure you change $scope.saving back to false before returning, otherwise the "Save Changes" button will not be shown again.

ui/app/js/controllers/permissions_controller.js
```diff
@@ -76,50 +75,55 @@ function($scope, $routeParams, $location, $timeout, Permissions, Search, $modal,
...
-      $scope.openUpdateModal(user);
-    }, function () {
-      // modal closed
-    });
+      });
+      // $scope.users.push($scope.user);
+      // modalInstance.result.then(function (user) {
+      // }, function () {
+      //   // modal closed
+      // });
```
mozbhearsum: Please remove this code instead of commenting it out.

ui/app/js/controllers/user_edit_controller.js
```diff
@@ -281,10 +299,9 @@ function ($scope, $modalInstance, CSRF, Permissions, user, users) {
...
   $scope.cancel = function () {
-      if (!$scope.user.permissions.length) {
-        $scope.users.pop($scope.user.username);
-      }
-      $modalInstance.dismiss('cancel');
-
+    if (!$scope.is_edit){
+      $scope.users.push($scope.user);
+    }
+    $modalInstance.dismiss('cancel');
```
mozbhearsum: This causes a user with no permissions to be added to the Users list, which is a bit confusing. Did you intend this?

ui/app/js/controllers/permissions_controller.js
andela-hngerebara: Okay I will. Thanks for pointing it out

ui/app/js/controllers/user_edit_controller.js
andela-hngerebara: Okay

ui/app/js/controllers/user_edit_controller.js
andela-hngerebara: No.. I didn't notice the user was being added even without permission. Will fix it.

ui/app/js/controllers/permissions_controller.js
andela-hngerebara: Okay

ui/app/js/controllers/user_edit_controller.js
```diff
@@ -1,57 +1,66 @@
...
-  Permissions.getUserPermissions(user.username)
-  .then(function(permissions) {
-    _.forEach(permissions, function(p) {
-      if (p.options) {
-        p.options_as_json = JSON.stringify(p['options']);
-      }
+  if($scope.is_edit){
+    $scope.original_user = user;
+    $scope.user = angular.copy(user);
+    $scope.loading = false;
```
mozbhearsum: You should wait until after all of the network requests have finished to set this back to false (ie: in a .finally() block) - they are the most likely thing to cause loading to fail.

ui/app/js/controllers/user_edit_controller.js
```diff
@@ -1,57 +1,66 @@
...
+    .success(function(response) {
+      $scope.user.roles = response.roles;
+    })
+    .error(function(response) {
+      if (typeof response === 'object') {
+        $scope.errors = response;
+        sweetAlert(
+          "Form submission error",
+          "See fields highlighted in red.",
+          "error"
```
mozbhearsum: This code runs before the user has made any changes to the Form, so "Form submission error" isn't the right message to show. How about "Failed to load User Roles"?

ui/app/js/controllers/user_edit_controller.js
```diff
@@ -128,10 +142,18 @@ function ($scope, $modalInstance, CSRF, Permissions, user, users) {
   $scope.addPermission = function() {
     // $scope.permission.options = $scope.permission.options_as_json;
     $scope.saving = true;
+    $scope.usersaved = true;
```
mozbhearsum: Careful here - the user is not actually saved until the Permissions.addPermission request has successfully completed.

ui/app/templates/permissions_modal.html
```diff
@@ -15,10 +20,20 @@ <h3 class="modal-title">{{ user.username }}</h3>
...
+              <select name="singleSelect" class="form-control" id="id_permission" ng-model="permission.permission">
+                  <option value="">---Please select---</option>
+                  <option value="admin">Admin</option>
+                  <option value="rule">Rule</option>
+                  <option value="release">Release</option>
+                  <option value="release_read_only">Release Read Only</option>
+                  <option value="release_locale">Release Locale</option>
+                  <option value="required_signoff">Required Signoff</option>
+                  <option value="permission">Permission</option>
+                  <option value="scheduled_change">Scheduled Change</option>
```
mozbhearsum: Thank you for taking the initiative to add this - it's really convenient!

Hopaez:
@mozbhearsum  Thanks. I have made the changes.

ui/app/js/controllers/user_edit_controller.js
Hopaez: Okay.

andela-hngerebara:
ooh seen.. Fixed it. All good now

PR #389, opened 2017-09-05
Bug 1389340: When for scheduled changes should declare local timezone at start

tossj:
https://bugzilla.mozilla.org/show_bug.cgi?id=1389340


tossj:
@mozbhearsum Thanks for the feedback. I've changed my fix so only the places where time is input by the user displays 'When (local time)'.

PR #390, opened 2017-09-05
Bug 1355071 allow users to schedule changes to Rules and Releases without a timestamp

collin5:
- Adds a button on Rule Scheduled Change modal which sets timestamp to the near future from time of form submission.


collin5:
@mozbhearsum Oh sorry for the `Error: data.when.getTime is not a function` error, fixed it though. Also improved in the UI as you had suggested :) Thanks.

collin5:
@mozbhearsum  Implemented the requested changes, hope all is good now :). Thanks.

collin5:
@mozbhearsum  Thanks, done with the requested changes 👍

PR #391, opened 2017-09-06
Bug 1339893 show "All rules" if no value is selected on the product/channel filter

andela-ookwuolisa:
Adds "All rules" to the product/channel filter if there is no selected value.


andela-ookwuolisa:
Okay. Thanks for pointing that out. I'll fix it.

mozbhearsum:
This looks good now - I updated your branch to remove the unwanted .coverage files, though.

PR #392, opened 2017-09-06
bug 1380874: Remove signoff/revoke buttons from history.

Alweezy:
Currently, the history pages have the signoff/revoke signoff buttons. It is unnecessary to control permissions from history pages. This PR removes the buttons.

### DIFF
--- a/ui/app/templates/permission_scheduled_changes.html
+++ b/ui/app/templates/permission_scheduled_changes.html
@@ -41,12 +41,12 @@
 
     <h3 class="panel-title">
       <div style="float: right">
-        <i ng-show="!scheduled_changes_rules_count && !sc.complete" ng-if=" scheduled_changes_rules_count && $first && (currentPage == 1)">Current</i>
-        <button ng-show="!scheduled_changes_rules_count && !sc.complete" class="btn btn-xs btn-primary" ng-click="signoff(sc)"
+        <i ng-if=" scheduled_changes_rules_count && $first && (currentPage == 1)">Current</i>
+        <button ng-show="!scheduled_changes_rules_count" class="btn btn-xs btn-primary" ng-click="signoff(sc)"
                 ng-show="! isEmpty(sc['required_signoffs']) && ! sc['signoffs'].hasOwnProperty(current_user)">
             Signoff as...
         </button>
-        <button ng-show="!scheduled_changes_rules_count && !sc.complete" class="btn btn-xs btn-danger" ng-click="revokeSignoff(sc)"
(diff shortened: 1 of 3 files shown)


ui/app/templates/permission_scheduled_changes.html
```diff
@@ -41,12 +41,12 @@
 
     <h3 class="panel-title">
       <div style="float: right">
-        <i ng-if=" scheduled_changes_rules_count && $first && (currentPage == 1)">Current</i>
```
mozbhearsum: I don't think you want to change the ng-if on this one -- it already has the correct conditions for whether or not it should be displayed. Same thing on the other pages.

ui/app/templates/permission_scheduled_changes.html
```diff
@@ -41,12 +41,12 @@
...
-        <button class="btn btn-xs btn-primary" ng-click="signoff(sc)"
+        <i ng-show="!scheduled_changes_rules_count && !sc.complete" ng-if=" scheduled_changes_rules_count && $first && (currentPage == 1)">Current</i>
+        <button ng-show="!scheduled_changes_rules_count && !sc.complete" class="btn btn-xs btn-primary" ng-click="signoff(sc)"
```
mozbhearsum: These shouldn't be conditional on sc.complete, just checking scheduled_changes_rules_count is enough. That being set or not is a proxy for "are we on the history page or not", and we _never_ want the buttons shown on the history page.

ui/app/templates/permission_scheduled_changes.html
Alweezy: So the _Current_ should be showing?

Alweezy:
@mozbhearsum I have since made the changes you pointed out. Please make a review and comment on the same.

ui/app/templates/permission_scheduled_changes.html
```diff
@@ -42,11 +42,11 @@
     <h3 class="panel-title">
       <div style="float: right">
         <i ng-if=" scheduled_changes_rules_count && $first && (currentPage == 1)">Current</i>
-        <button class="btn btn-xs btn-primary" ng-click="signoff(sc)"
+        <button ng-show="!scheduled_changes_rules_count" class="btn btn-xs btn-primary" ng-click="signoff(sc)"
```
mozbhearsum: With this applied, the Signoff and Revoke buttons are _always_ shown on pages like http://localhost:8080/rules/scheduled_changes, probably because you've added a new ng-show that has overriden the existing one for all of these buttons. You need to integrate this condition with the ones from the existing "ng-show" attributes, such as the one on line 46.

Alweezy:
@mozbhearsum, I have effected the changes you pointed out, please check it out.

PR #393, opened 2017-09-07
bug 1311765: Add rule_id to INSERT to rules email

njirap:


auslib/db.py
```diff
@@ -403,14 +403,12 @@ def _prepareInsert(self, trans, changed_by, **columns):
...
-
-        if self.onInsert:
-            self.onInsert(self, "INSERT", changed_by, query, trans)
-
         ret = trans.execute(query)
         if self.history:
             for q in self.history.forInsert(ret.inserted_primary_key, data, changed_by):
                 trans.execute(q)
+        if self.onInsert:
+            self.onInsert(self, "INSERT", changed_by, query, trans, *ret.inserted_primary_key)
```
allan-silva: How primary keys can be composable, may be a good idea pass the entire list as parameter and use it on the target callback, instead of dereference it.

auslib/db.py
```diff
@@ -2529,6 +2527,7 @@ def bleet(table, type_, changed_by, query, transaction):
                 body.append(UTF8PrettyPrinter().pformat(row))
         elif type_ == "INSERT":
             body.append("Row to be inserted:")
+            query.parameters['rule_id'] = inserted_item_id
```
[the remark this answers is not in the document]
mozbhearsum: Yeah, this is a bit problematic - we can't reference table-specific columns here (otherwise we end up putting things like "rule_id" in e-mails about tables that don't have one. The primary key column names either need to be passed into make_change_notifier or whenever onInsert is called. I think former would be a bit cleaner.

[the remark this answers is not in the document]
njirap: Noted..working on this

ui/npm-debug.log
```diff
@@ -0,0 +1,38 @@
+0 info it worked if it ends with ok
```
allan-silva: Probably this file should not be here.

auslib/db.py
```diff
@@ -2504,7 +2503,7 @@ def send_email(relayhost, port, username, password, to_addr, from_addr, table, s
 
 
 def make_change_notifier(relayhost, port, username, password, to_addr, from_addr, use_tls):
-    def bleet(table, type_, changed_by, query, transaction):
+    def bleet(table, type_, changed_by, query, transaction, *table_args):
```
allan-silva: Maybe you can use kwargs, will not necessary guessing  the parameters indexes, just get by key if is in dict.

auslib/db.py
```diff
@@ -403,14 +403,13 @@ def _prepareInsert(self, trans, changed_by, **columns):
...
-
-        if self.onInsert:
-            self.onInsert(self, "INSERT", changed_by, query, trans)
-
         ret = trans.execute(query)
         if self.history:
             for q in self.history.forInsert(ret.inserted_primary_key, data, changed_by):
                 trans.execute(q)
+        if self.onInsert:
+            pk_column = self.t.primary_key.columns.keys()[0]
```
allan-silva: Right now, I have no access to sqlalchemy doc, but doing `self.t.primary_key.columns.keys()[0]`  will get part of a possible composable key or a composable key as whole?

njirap:
hmmm interesting..

ui/npm-debug.log
njirap: you are right sorry for this!

auslib/db.py
njirap: I believe `.keys()` returns a list of primary keys.. So if the table has only 1 pk then the `[0]` indexing gets it.

auslib/db.py
njirap: ok lemmie see

auslib/db.py
allan-silva: Yep, some tables can has more than one primary key column, so getting zero index will get only one of them.

njirap: Do we want all primary keys fields in the email?

allan-silva: Makes sense for me, once the function is generic. What you think?

@mozbhearsum any thought about that?

auslib/db.py
njirap: this has been implemented.. the args are now kwargs

njirap:
@mozbhearsum @allan-silva  in what way can I replicate the failure in the unit tests coz it appears checking that the "rule_id: " is there is not catching the failure. Thanks!

allan-silva:
@njirap, I tested by configuring SMTP server as my gmail account (you will need enable unsecure app in gmail, if you try it) and by changing a rule via UI.

auslib/db.py
mozbhearsum: Yeah, I think that makes sense.

mozbhearsum:
This looks like it's working for rule_id now, I managed to get an e-mail like:
```
Changed by: balrogadmin
Row to be inserted:
{'data_version': 1,
 'options': None,
 'permission': ['scheduled_change', 'ueoueouoe'],
 'username': 'ueoueouoe'}
```

However, it appears to be breaking the e-mails for tables that don't have autoincrement PKs. For example, here's an e-mail I got when adding a new permission:
```
Changed by: balrogadmin
Row to be inserted:
{'data_version': 1,
 'options': None,
 'permission': ['scheduled_change', 'ueoueouoe'],
 'username': 'ueoueouoe'}
```

As you can see, both the "permission" column and "username" column end up in the "permission" value.

Good progress though!

auslib/db.py
mozbhearsum: I think this is probably why the permissions e-mail broke -- you need to account for the fact that `ret.inserted_primary_key` may contain multiple columns.

njirap:
OK I'm working on the breakage!

njirap:
The multiple primary key table breakage has been fixed @mozbhearsum @allan-silva

njirap:
thanks @allan-silva I have rebased

njirap:
@allan-silva @mozbhearsum the latest 2 commits on master fail Taskcluster. I believe this here is the same failure inherited.

auslib/db.py
```diff
@@ -397,14 +397,15 @@ def _prepareInsert(self, trans, changed_by, **columns):
...
         query = self._insertStatement(**data)
-
-        if self.onInsert:
-            self.onInsert(self, "INSERT", changed_by, query, trans)
-
         ret = trans.execute(query)
         if self.history:
             for q in self.history.forInsert(ret.inserted_primary_key, data, changed_by):
                 trans.execute(q)
+        if self.onInsert:
```
mozbhearsum: Moving this code means that we're not sending e-mail until after the row has been modified. This is a really nice side effect of this patch. If you have time, do you think you can follow-up and do the same for update and delete?

PR #394, opened 2017-09-07
bug 1363452: view for showing scheduled changes in context of the rest of the rules

mozbhearsum:
This is a partial overhaul of the Rules interface that shows Scheduled Changes in-context with the Rules. I took the opportunity to do a few other small things as well. The highlights are:
- Show Scheduled Changes inline with their associated Rule. Unchanged fields have no highlighting, changed fields show a sort of "diff" with red/green.
- Add required signoff/completed signoffs, Signoff/Revoke buttons for signoffs for Rules with Scheduled Changes.
-- This won't work without a reload if you schedule a new change. This should be fixed when https://bugzilla.mozilla.org/show_bug.cgi?id=1355076 is merged.
- Add a button to allow scheduling the creation of new Rules on the Rules page; prepopulate with the selected product/channel if applicable.
- Make it possible to edit Scheduled Changes by clicking the "Pending..." buttons.
- Make it possible to delete Scheduled Changes within the Edit dialog.
- Remove the Order By filter, and always filter on priority (as far as I know, we haven't used the other filters since the product/channel filter came along).
- Factor out the display of Rule fields into the "rulefield" directive. This has the side effect of changing the styling of the fields on the right hand side of the cards (which was different than the left hand side until now).

This isn't a full on replacement for the Scheduled Changes UI, as there's no history integrated, but it should replace it for day to day usage. In order to make this easier to review, I've stood up an instance at http://54.90.211.22:8080/rules with some changes to Firefox release rules pending.

@JohanLorenzo, I think you're probably the best person to do the actual review of this, but I know there's a lot of people that may have thoughts here. @nthomas-mozilla, @kmoir, @MihaiTabara, @srfraser, @escapewindow - feel free to have a look and drop any questions or comments!


mozbhearsum:
Apologies for the test failures - I've fixed them up now.

nthomas-mozilla:
I really really like this change, it's a massive step forward. 

Here's some things I noticed in running this locally
- Alias is missing from dialogs to create and update rules, I think we talked about this on another recent PR so it might just be a merge
- Update/Delete/Duplicate/History buttons are missing unless there is a scheduled change against a rule. This is weird given it doesn't look like that on your instance; I'd rebuilt docker images and node_modules from scratch, and tried to remove other ignored files and .cache but possibly screwed it up. Should have a history button after a change is applied, even the prod dump is pretty bare of history before hand.
- Trying to change every single value in one go: change all values
  1. The product gets a nice diff, but channel shows the current value in black too
  2. 'Pending Update in [time]' isn't obviously a button to my eye. Maybe separate text for 'Pending Update in [time]' and a 'Modify change' button
  3. No red part of diff on all these rule parameters

Also tried simulating the win64 migration changes where had rules on beta-cdntest and then changed them to beta*. The scheduled change showed up for a filter of 'Firefox, beta-cdntest' but not 'Firefox, beta'. In the reverse direction it showed up both ways.

Finally, in terms of workflow, I suspect the list of scheduled changes will be used a lot less. It would be great to bookmark/direct link to a filter of Firefox,beta and so on (aka [bug 1337929](https://bugzilla.mozilla.org/show_bug.cgi?id=1337929)).

mozbhearsum:
Thanks for you thoughtful feedback as always, @nthomas-mozilla.

Sounds like there's a couple of issues I didn't catch after merging in the latest master, and you've identified some other bugs as well. Obviously the red/strikethrough should be consistent throughout, and the History button should always be shown.

For the Update/Delete/Duplicate buttons, they should definitely be shown regardless of Scheduled Change status, but once https://bugzilla.mozilla.org/show_bug.cgi?id=1355076 is fixed, we should start hiding them for anything that requires signoff (since you'll always get an error when you try to save changes from them).

I'll see what I can do for the pending change notification/button - I can see how it doesn't look like a button currently.

nthomas-mozilla:
I should probably clarify the button comment - it's styled like a button, just doesn't have a verb-y label.

ui/app/templates/rule_field.html
```diff
@@ -0,0 +1,19 @@
...
+    <span ng-show="help && (rule[fieldname] !== null || (showsc && rule.scheduled_change[fieldname] !== null))"
+          class="glyphicon glyphicon-question-sign" title="{{ help }}"></span>
+    :
+<b>
+<!-- we have two display modes: with and without scheduled changes.
+     if scheduled changes are being shown, we shown any fields that are
+     changing as a "diff", where the current value is struck out, and
+     the scheduled change version is in green. if scheduled changes aren't
+     being shown, we just show the current value plainly -->
+<span ng-style="!rule.rule_id && (showsc && rule.scheduled_change !== null && rule[fieldname] !== rule.scheduled_change[fieldname]) ? {'text-decoration': 'line-through', 'color': 'red'} : {};">
```
JohanLorenzo: I'm surprised rulefields aren't struck out, like @nthomas-mozilla reported. Could it be because of `rule.rule_id` or showsc being truthy/falsey?

ui/app/templates/rules.html
```diff
@@ -37,57 +43,110 @@
...
+        </button>
+        <button ng-show="!rule_id && rule.scheduled_change === null" class="btn btn-danger btn-xs"
+                ng-click="openNewScheduledDeleteModal(rule)">
+            Schedule for Deletion
+        </button>
+        <button ng-show="!rule_id && rule.scheduled_change !== null && rule.scheduled_change.change_type !== 'insert'"
+                class="btn btn-default btn-xs" ng-click="openUpdateModal(rule)">
+            Update
+        </button>
+        <button ng-show="!rule_id && rule.scheduled_change !== null && rule.scheduled_change.change_type !== 'insert'"
```
JohanLorenzo: I don't think we should show the Update/Delete buttons, if a scheduled change is active. If you ask for a deletion while a pending change is active, you end up with this (expected) error:

> "Couldn't perform the request DELETE due a conflict with a scheduled change. Cannot delete rows that have changes scheduled."

Duplicate/History are good to keep, thought, IMO.

ui/app/templates/rules.html
```diff
@@ -37,57 +43,110 @@
...
+           class="btn btn-default btn-xs" ng-href="/rules/{{ rule.rule_id }}">
+            History
+        </a>
       </div>
-
-      <i title="Product" ng-show="!rule.product">no product</i>
-      <b title="Product" ng-show="rule.product" ng-bind-html="highlightSearch(rule.product, 'product')"></b>
+      <i title="Product" ng-show="!rule.product && !rule.scheduled_change.product">no product</i>
+      <b title="Product">
+        <span ng-show="rule_id || !show_sc || rule.scheduled_change === null || rule.product === rule.scheduled_change.product">{{ rule.product }}</span>
```
JohanLorenzo: I think we can factorize similar conditions thanks to `rulefield` 😃

ui/app/templates/rules.html
```diff
@@ -109,61 +184,44 @@ <h5 title="FallBack Mapping">
...
-      <dd ng-if="rule.mig64 === false">No</dd>
-      <dt ng-show="rule.distVersion">Dist Version:</dt>
-      <dd>{{ rule.distVersion }}</dd>
-      <dt ng-show="rule.update_type">Update type:</dt>
-      <dd>{{ rule.update_type }}</dd>
-      <dt ng-show="rule.headerArchitecture">Header Architecture:</dt>
-      <dd>{{ rule.headerArchitecture }}</dd>
-    </dl>
+    <div style="float: right">
+      <rulefield showsc="show_sc" rule="rule" fieldname="alias" fieldtitle="Alias" help="Can be used in place of rule_id in REST operations"></rulefield>
```
JohanLorenzo: Nice factorization 👍

ui/app/templates/rules.html
```diff
@@ -37,57 +43,110 @@
...
       </select>
     </div>
-    <div class="row">
-      <label>Order by:</label>
-      <select ng-model="ordering_str" name="ordering_str"
-        ng-options="o.text for o in ordering_options track by o.value">
-      </select>
+    <div class="row" ng-if="!rule_id">
+      <label for="include_sc">Include Scheduled Changes?</label>
+      <input type="checkbox" id="include_sc" ng-model="show_sc" checked>
```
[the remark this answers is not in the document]
mozbhearsum: Are you saying that you think it's not useful to be able to see _only_ the active Rules? It was working at one point, I guess I broken it before posting this =\.

ui/app/templates/rules.html
mozbhearsum: I think you're right that we should hide the Delete button, but you can still update when there's a change scheduled, so long as you don't try to modify any of the same rows the Change does.

Good point about Duplicate though, that should be shown regardless.

ui/app/templates/rules.html
```diff
@@ -37,57 +43,110 @@
...
+
     <div class="left" style="float: left">
 
-      <h5 title="Mapping">
-        Mapping: <span ng-show="rule.mapping">
-            <b ng-bind-html="highlightSearch(rule.mapping, 'mapping')"></b>
+      <h5 title="Mapping" ng-show="rule.mapping || (show_sc && rule.scheduled_change !== null && rule.scheduled_change.mapping)">
+        Mapping: 
+          <span ng-show="rule.mapping">
+            <b ng-style="!rule_id && show_sc && rule.scheduled_change !== null && rule.mapping !== rule.scheduled_change.mapping ? {'text-decoration': 'line-through', 'color': 'red'}: ''">
```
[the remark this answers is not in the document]
mozbhearsum: Hmm, that's an interesting point. Would you still use the strikethrough/red for the old mapping when a Scheduled Change is changing it?

ui/app/templates/rules.html
mozbhearsum: Sorry, can you expand on this?

ui/app/js/controllers/rule_scheduled_change_edit_controller.js
```diff
@@ -84,7 +85,31 @@ function ($scope, $modalInstance, CSRF, Rules, Releases, sc) {
...
+  $scope.delete = function() {
+    $scope.saving = true;
+    CSRF.getToken()
+    .then(function(csrf_token) {
+      Rules.deleteScheduledChange($scope.sc.sc_id, $scope.sc, csrf_token)
+      .success(function(response) {
+        $modalInstance.close("delete");
+      })
+      .error(function(response) {
+        if (typeof response === 'object') {
```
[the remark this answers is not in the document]
mozbhearsum: Probably...it shouldn't happen, but history has shown it probably will at some point :)

ui/app/js/controllers/rules_controller.js
```diff
@@ -132,22 +124,38 @@ function($scope, $routeParams, $location, $timeout, Rules, Search, $modal, $rout
     return !!(false || $scope.filters.search.length);
   };
 
-  function escapeRegExp(string){
-    return string.replace(/([.*+?^=!:${}()|\[\]\/\\])/g, "\\$1");
-  }
+  $scope.isEmpty = function(obj) {
```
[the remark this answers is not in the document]
mozbhearsum: Yeah, that's a good point. I think I'm learned enough about Angular to do that without too much trouble now...

ui/app/templates/rules.html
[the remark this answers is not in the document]
JohanLorenzo: Oh, I think it can be. I thought it was a boilerplate for future improvements. It the fix is simple to do, then let's keep it :)

ui/app/templates/rules.html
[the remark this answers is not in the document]
JohanLorenzo: That's a good question :)

Strikethrough looks less confusing if that value is changed, I think.

ui/app/templates/rules.html
JohanLorenzo: Yeah, sorry about this.

Basically, the condition at this line, is almost identical to the one in `rulefield`. Moreover, we want similar behaviors on strikethroughs, per Nick's screenshot. So my point was: because the logic is the same, we can use rulefield here. What do you think?

mozbhearsum: We might be able to move it into a helper function and pass show_sc and rule to it...I'll investigate that.

ui/app/templates/rules.html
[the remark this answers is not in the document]
mozbhearsum: I haven't been able to come up with anything better here - I think if we _don't_ do the strikethrough on mapping, it's confusing in another way (why is everything else struck out, but not mapping?). Perhaps we should avoid doing strikethroughs altogether for deletes, and rely solely on the titlebar indicators?

ui/app/templates/rule_field.html
mozbhearsum: It turns out that this was because the "rule_id" we should be making this decision on is the one that comes from the URL (that let's us know "are we on the history page or not") - not the one that's part of each individual rule object.

ui/app/templates/rules.html
mozbhearsum: It may also be worth noting that once https://bugzilla.mozilla.org/show_bug.cgi?id=1355076 is fixed, the Update/Delete/Duplicate buttons should be hidden for any channel that has required signoffs (since you must use Scheduled Changes). It's still good to make sure works properly for Nightly though.

mozbhearsum:
Alright, I think I've address all of the feedback from both of you. I updated http://54.90.211.22:8080/rules with the new code and added some initial Scheduled Changes to the Firefox release rules. @JohanLorenzo - I know you wanted to have a second pass at this anyways, so let me know what you think of the latest version.

Thanks for encouraging me to refactor, by the way, I ended up finding a bunch of duplicated code to get rid of!

ui/app/js/controllers/parent_controller.js
```diff
@@ -44,4 +44,8 @@ angular.module("app").config(['$httpProvider', function($httpProvider) {
 angular.module("app").controller('ParentController',
 function($scope, Page) {
   $scope.Page = Page;
+  $scope.isEmpty = isEmpty;
```
JohanLorenzo: Wow, I'm positively surprised we don't need to declare a dependency on `helpers.js`. Nice!

ui/app/templates/rules.html
```diff
@@ -37,57 +43,116 @@
...
+
     <div class="left" style="float: left">
 
-      <h5 title="Mapping">
-        Mapping: <span ng-show="rule.mapping">
-            <b ng-bind-html="highlightSearch(rule.mapping, 'mapping')"></b>
+      <h5 title="Mapping" ng-show="rule.mapping || (show_sc && rule.scheduled_change !== null && rule.scheduled_change.mapping)">
+        Mapping:
+          <span ng-show="rule.mapping">
+            <b ng-style="!rule_id && show_sc && rule.scheduled_change !== null && rule.mapping !== rule.scheduled_change.mapping ? {'text-decoration': 'line-through', 'color': 'red'}: ''">
```
JohanLorenzo: I tried to schedule a deletion, and I found the UI clearer, even though mapping is struck out. I guess that's because every field is too.

This looks great!

ui/app/templates/rules.html
```diff
@@ -96,11 +161,27 @@ <h5 title="Mapping">
                 <i class="glyphicon glyphicon-pencil"></i>
                 Edit
             </a>
-        </span>
+          </span>
+          <span ng-show="!rule_id && show_sc && rule.scheduled_change !== null && rule.mapping !== rule.scheduled_change.mapping && rule.scheduled_change.mapping">
+            <b style="color: green" ng-show="show_sc && rule.scheduled_change !== null && rule.mapping !== rule.scheduled_change.mapping">
+              {{ rule.scheduled_change.mapping }}
+            </b>
+            <a class="btn btn-default btn-xs" href="" ng-click="openReleaseDataModal(rule.scheduled_change.mapping)">
```
JohanLorenzo: I like the fact that both releases are viewable!

ui/app/templates/rules.html
```diff
@@ -109,66 +190,51 @@ <h5 title="FallBack Mapping">
...
-      <dd ng-if="rule.mig64 === false">No</dd>
-      <dt ng-show="rule.distVersion">Dist Version:</dt>
-      <dd>{{ rule.distVersion }}</dd>
-      <dt ng-show="rule.update_type">Update type:</dt>
-      <dd>{{ rule.update_type }}</dd>
-      <dt ng-show="rule.headerArchitecture">Header Architecture:</dt>
-      <dd>{{ rule.headerArchitecture }}</dd>
-    </dl>
+    <div style="float: right">
+      <rulefield showdiff="!rule_id" showsc="show_sc" rule="rule" fieldname="alias" fieldtitle="Alias"
```
JohanLorenzo: Nice sequence of rulefields!

PR #397, opened 2017-09-11
bug 1376331: improve database initialization for dev environments

mozbhearsum:
This changes the database export and import scripts to make the set-up process much less error prone. The new database dump format includes the schema of the entire database (not just select tables), and the same set of data as before(1). The new import process starts with an empty database, imports the dump, and then upgrades the database to the latest schema version. This means that database initialization should work even when master and production do not have the same schema version.

I tested this this locally by re-dumping the most recent production dump and initializing with that. I also tested by downgrading the most recent dump, dumping that, and initializing with it. Both scenarios worked without issue, and gave me a schema version of 29 with the expected data. You'll notice I also removed the checked in sample data -- it's so out of date now that I don't think it's useful, and trying to maintain is more trouble than it's worth. You need to download a bunch of Docker images already to be a useful Balrog developer, so I don't think it's unreasonable to ask that you download a database dump too.

This is probably easiest to review as one pull request, but I intend to split it up when landing -- the new dump code will go first, and once the published dump is using the new format, I can commit the new initialization script.

(1) Almost the same set of data as before, actually. I fixed a bug where any Release ever referenced by a Scheduled Change was included into the dump.


scripts/manage-db.py
```diff
@@ -123,64 +123,79 @@ def extract_active_data(trans, url, dump_location='dump.sql'):
...
+        batch_generator = chunk_list(list(release_names), 30)
+        for batched_release_list in batch_generator:
+            qry = ", ".join("'" + names + "'" for names in batched_release_list)
             popen(_strip_multiple_spaces('%s --skip-add-drop-table --no-create-info %s '
                                          'releases --where="releases.name IN (%s)" \
                    >> %s' % (mysql_default_command, database, qry, dump_location)))
 
+    popen(
+        _strip_multiple_spaces('%s --skip-add-drop-table --no-create-info %s releases_history --where="releases_history.name=\'Firefox-mozilla-central-nightly-latest\' \
+        limit 50" >> %s' % (mysql_default_command, database, dump_location))
```
nthomas-mozilla: I see you've only mobved this, but does it grab the first 50 history rows or last 50 ? aka might need an ORDER clause too. Same for the next query.

scripts/initdb_and_run.sh
```diff
@@ -4,17 +4,18 @@ export LOCAL_DUMP="/app/scripts/prod_db_dump.sql"
 
 if [ ! -e /app/.cache/mysql/db.done ]; then
     echo "Initializing DB..."
-    python scripts/manage-db.py -d mysql://balrogadmin:balrogadmin@balrogdb/balrog create
     python scripts/import-db.py
```
nthomas-mozilla: Nit, import-db.py seems to pull down the prod db dump if it's missing/stale, so the name is a bit misleading.

scripts/initdb_and_run.sh
```diff
@@ -4,17 +4,18 @@ export LOCAL_DUMP="/app/scripts/prod_db_dump.sql"
...
 
     if [ -e "$LOCAL_DUMP" ]; then
       db_source="cat $LOCAL_DUMP"
     else
-      db_source="bunzip2 -c /app/scripts/sample-data.sql.bz2"
+      echo "Couldn't download production database dump. You must be connected to the internet the first time you start Balrog..."
```
nthomas-mozilla: I was going to say that i doubted we'd ever show this message with `set -e` in use, as import-db.py will already have errored out, but it turns out it uses `exit()` rather than `exit(1)`. Supposing it did exit with non-zero status then this whole if/else block isn't needed any more, and we can do `cat $LOCAL_DUMP | mysql ....`

scripts/manage-db.py
```diff
@@ -123,64 +123,79 @@ def extract_active_data(trans, url, dump_location='dump.sql'):
...
-    )
-
-    popen(
-        _strip_multiple_spaces('%s --skip-add-drop-table --no-create-info %s releases_history --where "name = ( \
-            SELECT rules.mapping \
-            FROM rules \
-            WHERE rules.alias=\'firefox-release\' \
-        ) LIMIT 50" >> %s' % (mysql_default_command, database, dump_location))
+        _strip_multiple_spaces("""
+            %s --skip-add-drop-table --no-create-info %s
```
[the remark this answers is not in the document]
mozbhearsum: Good idea, will do.

scripts/manage-db.py
mozbhearsum: I'll double check this. I imagine we want the 50 most recent ones...

scripts/initdb_and_run.sh
mozbhearsum: Yeah...probably a good time to improve on that.

scripts/initdb_and_run.sh
mozbhearsum: Good idea - I'll fix import-db.py to exit better, and then we can simplify the shell code \o/

mozbhearsum:
I think I've addressed all of your points, @nthomas-mozilla. I'll hold off a day or two before landing the new dumping code.

nthomas-mozilla:
👍

mozbhearsum:
The manage-db.py changes have been made in https://github.com/mozilla/balrog/commit/6d22c6df2f74dc85a5e864a0f8237a2a93285692. I'll keep this open to merge the remainder after a new-format dump is available.

PR #399, opened 2017-09-13
bug 1398237: adjust release process for new stage set-up

mozbhearsum:
This adjusts the Taskcluster jobs + docker image scripts to work with the new release process we've come up with in https://bugzilla.mozilla.org/show_bug.cgi?id=1398236. Specifically, it starts creating tags on Dockerhub with the Release version in them. The new stage environment will automatically update itself to the new images when this happens.

To support this, I had to make a few other changes as well
- Limit CI tasks to pull requests + pushes to the master branch (see the comment in .taskcluster.yml for why).
- Add routes to the Release-only tasks, to make them easier to find.

I know this is going to be painful to review because you can't really run any of it yourself. I have shadow set-up that I used in testing this, feel free to have a look at it. I can give you access if you want to play around yourself, too: https://github.com/testbhearsum/balrog, https://hub.docker.com/r/bhearsumtesttest/

### DIFF
--- a/scripts/push-dockerimage.sh
+++ b/scripts/push-dockerimage.sh
@@ -2,6 +2,8 @@
 
 set -e
 
+extra_tag=$1
+
 password_url="taskcluster/secrets/v1/secret/repo:github.com/mozilla/balrog:dockerhub"
 artifact_url="taskcluster/queue/v1/task/${TASK_ID}/runs/${RUN_ID}/artifacts/public/docker-image-shasum256.txt"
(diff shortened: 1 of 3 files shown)


scripts/push-dockerimage.sh
```diff
@@ -26,31 +32,29 @@ echo "{
     \"build\": \"https://tools.taskcluster.net/task-inspector/#${TASK_ID}\"
 }" > version.json
 
-# Initialize and update the UI submodule
-git submodule init
-git submodule update
```
rail: yay!

PR #401, opened 2017-09-14
Remove whitelists

glasserc:
There are a couple of places where we appear to whitelist fields of records before we serialize them as part of a response. The original rationale for this seems to be because we wanted to rename fields for the frontend; see https://github.com/mozilla/balrog/commit/4338af9338012201e4fc355c778305bc09e905b7#diff-2ca168a4f3b72f6f00f46f8d3fd513a4R237. However, no renaming happens any more, and even if there were, we would probably use something more formal like a Marshmallow schema. That means this code is essentially a no-op so get rid of it.

I had originally intended this to be part of a different PR but I'm going to end up taking a different approach there so I'll just open this while I'm figuring out what my next step is.

### DIFF
--- a/auslib/db.py
+++ b/auslib/db.py
@@ -28,19 +28,14 @@ from auslib.util.timestamp import getMillisecondTimestamp
 import logging
 
 
-def rowsToDicts(fn):
-    """Decorator that converts the result of any function returning a dict-like
-       object to an actual dict. Eg, converts read-only row objects to writable
-       dicts."""
-    def convertRows(*args, **kwargs):
-        ret = []
-        for row in fn(*args, **kwargs):
-            d = {}
-            for key in row.keys():
-                d[key] = row[key]
-            ret.append(d)
-        return ret
-    return convertRows
+def rows_to_dicts(rows):
+    """Converts SQL Alchemy result rows to dicts.
+
+    You might want this if you want to mutate objects (SQLAlchemy rows
+    are immutable), or if you want to serialize them to JSON
+    (SQLAlchemy rows get confused if you try to serialize them).
+    """
+    return map(dict, rows)
 
 
 class AlreadySetupError(Exception):
@@ -354,7 +349,6 @@ class AUSTable(object):
                 query = query.where(cond)
         return query
 
-    @rowsToDicts
     def select(self, where=None, transaction=None, **kwargs):
         """Perform a SELECT statement on this table.
            See AUSTable._selectStatement for possible arguments.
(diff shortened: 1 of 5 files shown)


auslib/web/common/releases.py
```diff
@@ -107,7 +87,7 @@ def get_release_history(release):
                                    order_by=order_by,
                                    get_object_callback=lambda: _get_release(release),
                                    history_filters_callback=_get_filters,
-                                   process_revisions_callback=_process_revisions)
+                                   process_revisions_callback=process_release_revisions)
```
allan-silva: Maybe you can pass `annotateRevisionDifferences` directly here and delete 
 `process_release_revisions`

glasserc: Do you mean by stripping the `data` field first, and then calling `annotateRevisionDifferences`, using something like `lambda revisions: annotateRevisionDifferences(map(strip_data, revisions))`? I think that's sufficiently complicated that I'd prefer to keep it as a `def`.

allan-silva: Ah, no!
I mean `process_revisions_callback=annotateRevisionDifferences`

glasserc: Does this mean we no longer `strip_data`? Is that OK? Could we get rid of stripping data from other places?

allan-silva: Oh, my mistake, I do not realized that line 80 is part of `process_release_revisions`.

auslib/db.py
```diff
@@ -28,19 +28,14 @@
...
-        return ret
-    return convertRows
+def rows_to_dicts(rows):
+    """Converts SQL Alchemy result rows to dicts.
+
+    You might want this if you want to mutate objects (SQLAlchemy rows
+    are immutable), or if you want to serialize them to JSON
+    (SQLAlchemy rows get confused if you try to serialize them).
+    """
+    return map(dict, rows)
```
mozbhearsum: Thank you for reworking this. The decorator seemed like a good idea when it was first written, but it's grown very long in the tooth.

auslib/web/common/releases.py
```diff
@@ -12,6 +12,18 @@
 log = logging.getLogger(__name__)
 
 
+def strip_data(release):
+    """Return a release with all the fields except data.
+
+    This is used in multiple APIs to save bandwidth and present a
+    simplified view of the release by removing its largest field,
+    data, which is of no use except when serving clients.
+    """
```
mozbhearsum: And thank you for all the great comments/docstrings - it makes it much easier to review!

PR #403, opened 2017-09-15
add dropdown to permissions schduled change

andela-hngerebara:
**What does this PR do?**
Add dropdown to permissions scheduled change

**How should this be manually tested?**

Once balrog is running on http://127.0.0.1:8080, navigate to permissions, click on schedule changes then "add a new scheduled permission change". You can select the permission of your choice from the list of permissions.

### DIFF
--- a/ui/app/templates/permission_scheduled_change_update_modal.html
+++ b/ui/app/templates/permission_scheduled_change_update_modal.html
@@ -16,7 +16,17 @@
       <div class="col-md-6" >
         <div class="form-group" ng-class="{'has-error': errors.permission}">
         <label for="id_permission">permission</label>
-        <input type="text" class="form-control" id="id_permission" ng-model="sc.permission">
+        <select name="singleSelect" class="form-control" id="id_permission" ng-model="sc.permission">
+            <option value="">---Please select---</option>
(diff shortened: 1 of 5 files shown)


ui/app/templates/permission_scheduled_change_update_modal.html
```diff
@@ -16,7 +16,9 @@ <h3 class="modal-title">  Edit Scheduled Change for {{ sc.username }}</h3>
       <div class="col-md-6" >
         <div class="form-group" ng-class="{'has-error': errors.permission}">
         <label for="id_permission">permission</label>
-        <input type="text" class="form-control" id="id_permission" ng-model="sc.permission">
+          <select name="singleSelect" class="form-control" id="id_permission" ng-model="sc.permission">
```
mozbhearsum: This ng-model doesn't appear to be working. When I edit an existing scheduled permission change, the drop down always has "--- Please select ---" instead of the actually selecting the permission.

andela-hngerebara: @mozbhearsum Didn't notice it earlier. Fixed it now.

PR #404, opened 2017-09-15
Update to Node 4.x

glasserc:
Move off of a no-longer-supported version of NodeJS to a version that
is still at least barely supported.

Debian's jessie-backports repo has an updated version of NodeJS, but
it wasn't possible to use npm with it because of
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=861178, which
prevents the installation of nodejs-dev, which prevents the
installation of node-gyp, on which npm depends. As a result, we
introduce a Nodesource repository, which seems to work fine.

This means NodeJS is no longer the limiting factor for ES6
support. Now PhantomJS is.

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -8,14 +8,21 @@ FROM python:2.7-slim
 MAINTAINER bhearsum@mozilla.com
 
 # Some versions of the python:2.7 Docker image remove libpcre3, which uwsgi needs for routing support to be enabled.
-# Node and npm are to build the frontend. nodejs-legacy is needed by this version of npm.
 # mysql-client is needed to import sample data
 # libmysqlclient-dev is required to use SQLAlchemy with MySQL, which we do in production.
 # libfontconfig1 is required by phantomjs
 # git is needed to send coverage information
+# gcc is needed to build Python SQL module as well as uwsgi.
+# Use nodesource repo to get more modern nodejs. Node and npm are to build the frontend.
 RUN apt-get -q update \
-    && apt-get -q --yes install netcat libpcre3 libpcre3-dev nodejs nodejs-legacy npm libmysqlclient-dev mysql-client \
-                                libfontconfig1 netcat git curl \
+    && apt-get -q --yes install apt-transport-https curl \
+    && curl -sLf -o /dev/null 'https://deb.nodesource.com/node_4.x/dists/jessie/Release' \
+    && curl -s https://deb.nodesource.com/gpgkey/nodesource.gpg.key | apt-key add - \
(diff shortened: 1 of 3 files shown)


glasserc:
I tested the UI lightly -- trying to update a rule that has signoffs, and one that doesn't, and scheduling a change to a user's permissions, and revoking/resigningoff on that change. However, I didn't test every single operation that the UI offers.

glasserc:
I settled on Node 4.x because starting at about Node 6.x, lineman tasks start failing and I didn't want to spend time investigating why. Also, of the versions listed at https://hub.docker.com/r/library/node/tags/, 4.x is available, but 5.x isn't.

Dockerfile.dev
```diff
@@ -8,14 +8,21 @@ FROM python:2.7-slim
...
-                                libfontconfig1 netcat git curl \
+    && apt-get -q --yes install apt-transport-https curl \
+    && curl -sLf -o /dev/null 'https://deb.nodesource.com/node_4.x/dists/jessie/Release' \
+    && curl -s https://deb.nodesource.com/gpgkey/nodesource.gpg.key | apt-key add - \
+    && echo 'deb https://deb.nodesource.com/node_4.x jessie main' > /etc/apt/sources.list.d/nodesource.list \
+    && apt-get update \
+    && apt-get -q --yes install nodejs gcc \
+                       netcat libpcre3 libpcre3-dev libmysqlclient-dev mysql-client \
+                       libfontconfig1 netcat git \
     && apt-get clean
```
mozbhearsum: These changes will need to get carried over to "Dockerfile" as well (that's the one we use to build the production images). We should probably explicitly test that image as well, to make sure it can still build the app.

glasserc:
I can't promise that this covers that, or that it even works, but here are the changes as best as I could understand them.

glasserc:
Using https://gist.github.com/mozbhearsum/97dc6feec452ae44347b7e495b2f2ef5, I was able to get the containers to build. Actually fetching something from localhost:8080 doesn't work (I get empty responses with logs like `balrogadmin_1  | [pid: 16|app: -1|req: -1/30] 172.19.0.1 () {44 vars in 853 bytes} [Mon Sep 18 16:35:12 2017] GET / => generated 0 bytes in 0 msecs (HTTP/1.1 500) 0 headers in 0 bytes (0 switches on core 0)`). However, it seems like the CI is failing for reasons that aren't clear to me, so I'm going to push this off for now.

mozbhearsum:
I'm closing this because it's unlikely that we'll ever do it since we've been talking about moving the UI out of Balrog itself and into Delivery Dashboard and/or other places. We can always re-open if it becomes desirable again in the future.

PR #406, opened 2017-09-18
[bug]make it possible to link to the Rules page with a specific filter

njirap:


ui/app/js/router.js
```diff
@@ -74,6 +74,12 @@ angular.module("app").config(function($routeProvider, $locationProvider) {
     reloadOnSearch: false
   })
 
+  .when('/rules/:id/:pr_ch_filter', {
```
mozbhearsum: I don't think /rules/:id is what you want here - that's the history page, not the primary Rules page. Maybe we can use query args instead, to avoid the need for an additional route altogether, and get rid of the csv values? Eg: /rules?product=Firefox&channel=beta

Alweezy:
@mozbhearsum please check this out, I have made the changes you suggested.

mozbhearsum:
I haven't looked at this yet - please resolve the merge conflicts and fix the test failures.

Alweezy:
@mozbhearsum I'm done fixing the code, please check it out.

Alweezy:
@mozbhearsum I have solved the issues you pointed out, please review this.

Alweezy:
@bhearsum please check this out, I have resolved the merge conflicts.

Alweezy:
@bhearsum the Filter responds faster to URL changes now. Please check this out and comment on the same, thanks.

Alweezy:
Sorry, I have noticed some undesired output on this, let me fix it first.

Alweezy:
@mozbhearsum this is working fine now, please review.

ui/app/js/controllers/rules_controller.js
```diff
@@ -216,6 +236,37 @@ function($scope, $routeParams, $location, $timeout, Rules, Search, $modal, $rout
     return true;
   };
 
+  $scope.$on('$routeChangeSuccess', function () {
+    $timeout(function () {
+      populateRulesProductsChannels();
```
mozbhearsum: Does this actually do anything? populateRulesProductsChannels has no side effects, and you're not using the return value anywhere.

ui/app/js/controllers/rules_controller.js
```diff
@@ -216,6 +236,37 @@ function($scope, $routeParams, $location, $timeout, Rules, Search, $modal, $rout
...
+        filterString = 'All rules';
+      }
+      if ($routeParams.product && $routeParams.channel && $scope.pr_ch_options.includes($routeParams.product + ',' + $routeParams.channel)) {
+        filterString = $routeParams.product + ',' + $routeParams.channel;
+      } else if ($routeParams.product && $routeParams.channel && !$scope.pr_ch_options.includes($routeParams.product + ',' + $routeParams.channel)) {
+        filterString = 'All rules';
+      }
+      $scope.pr_ch_filter = filterString;
+      $rootScope.pr_ch_filter = $scope.pr_ch_filter;
+    }, 2000);
```
mozbhearsum: As I'm sure you know, this timeout of 2000 is why the location isn't updated immediately upon page load. Why do you need it, though? Why not just execute all of this code directly without the $timeout()? I tried that locally and everything appeared to work still.

ui/app/js/controllers/rules_controller.js
```diff
@@ -103,7 +122,8 @@ function($scope, $routeParams, $location, $timeout, Rules, Search, $modal, $rout
           $scope.pr_ch_options.sort().unshift("All rules");
           $scope.pr_ch_filter = "All rules";
           if ($scope.pr_ch_options.includes(localStorage.getItem('pr_ch_filter'))){
-            $scope.pr_ch_filter = localStorage.getItem('pr_ch_filter');
+            $scope.pr_ch_filter = $rootScope.pr_ch_filter || localStorage.getItem('pr_ch_filter') || "All rules";
+            $rootScope.pr_ch_filter = $scope.pr_ch_filter;
```
mozbhearsum: Can you explain the rationale behind using $rootScope? I don't understand why it's needed. There's no reason to store both the explicit filter from a URL _and_ what's in localStorage. The logic should be:
- If there's a valid filter in the URL use that, and update localStorage with it.
- If there's a filter with an error in the URL, use "All rules", and update localStorage with it.
- If there's no filter in the URL, use what's in localStorage
- If there's no filter in the URL or in localStorage, use "All rules", and update localStorage with it.

And of course, whenever the user changes the filter, update both the URL and localStorage with it.

Please let me know if I'm missing something here.

Alweezy:
@mozbhearsum I have implemented the changes you pointed out, please check this out.

PR #407, opened 2017-09-18
Make a command that resets the stage database

mozbhearsum:
This should provide the Docker command we need to do the stage resets as described in https://bugzilla.mozilla.org/show_bug.cgi?id=1398236. As you can see, I've added multiple levels of safeguards against accidentally prod resets ($ENV check, magic word, and sleep time). I'm still a bit paranoid about it, but I'm not sure what else we could look for.

@relud - I can't tag you for official review, but please let me know if this interface will work for you. The intent is that the you should be able to run with "reset-stage-db destroythedatabase" as a command. (It won't work until #397 is fully landed, though.)


relud:
running the docker container with a special command that resets the db with the prod dump sounds perfect

relud:
we have automated backups from up to 1 day ago (that _really_ needs to be changed to 14 days, thank you for bringing this up), and I have restored a backup from that before. We also have the public db dump, which this could theoretically restore to a fresh db if we lost everything else, and we have restored that in stage before, with some intervention, but that would involve some degree of data loss. @mozbhearsum would know better what kind of data loss that would entail.

mozbhearsum:
The public dumps contain enough information to get us up and running again in an emergency, but there's some stuff that we either purposely exclude or haven't added support for dumping yet that would make them less than ideal to run with for very long. It's possible to rebuild than information (it's mostly Release metadata), but it wouldn't be easy.

It sounds like our official backup situation is pretty solid though, as long as we're satisified with where they are stored?

nthomas-mozilla:
Our in-AWS solution sounds reasonable to me with the # of backups is increased. Pushing a full backup outside of AWS would be more like a disaster recovery plan for "AWS is dead for many days". Do we have that in a doc somewhere ?

mozbhearsum:
I'm not aware of a doc that has that information. @relud - are you?

relud:
I don't have any docs with this information yet. I could add it to [cloudops' balrog mana page](https://mana.mozilla.org/wiki/display/SVCOPS/Balrog+%28aus5%2C+aus4%2C+...%29+Firefox+Update+Service#Balrog(aus5,aus4,...)FirefoxUpdateService-RDSDatabase) or add a child page to that.

nthomas-mozilla:
Getting off track this particular issue but DR is something to consider I think. We did a RelEng DR plan in 2016 when updates were still served from SCL3/PHX1. It noted using a multi-region setup after the transition to CloudOps hosting.

relud:
Yeah, we are planning to add support for multi-region, and adding an rds replica was the first step for that. We are going to find a way to get replication working, and hopefully add multi-region before December.

PR #408, opened 2017-09-18
Add "is signoff required" to UI overall

glasserc:
This is a WIP attempt to try to solve https://bugzilla.mozilla.org/show_bug.cgi?id=1355076 and https://bugzilla.mozilla.org/show_bug.cgi?id=1355074.

This takes the approach of disabling buttons for immediate actions as well as adding notes to modals where it makes sense. I've never written anything in Angular before so I'm hoping to get some "30% review" feedback on the UI approach as well as the implementation approach.

- [x] Implement signoff requirements logic in the client
- [x] Fetch product signoff requirements and use them on
  - [x] Rules list page
  - [x] Edit rule modal
  - (n/a) Delete rule modal -- as far as I know, it's impossible to trigger the delete rule modal except through the (now hidden) button for an existing rule
  - [x] New rule modal (incl. Duplicate rule modal)
  - [x] Rule scheduled change/delete modal
  - [x] Releases list page
  - [x] Edit release modal
  - (n/a) New release modal
  - [x] Releases scheduled change/delete modal
- [x] Fetch permission signoff requirements and use them on
  - [x] Permissions list page
  - [x] Permissions update modal
  - [x] Permissions scheduled update modal

### DIFF
--- a/auslib/web/admin/views/base.py
+++ b/auslib/web/admin/views/base.py
@@ -83,3 +83,12 @@ class AdminView(MethodView):
         self.log.debug("processing DELETE request to %s" % request.path)
         with dbo.begin() as trans:
             return self._delete(*args, transaction=trans, **kwargs)
+
+
+def serialize_signoff_requirements(requirements):
+    dct = {}
+    for rs in requirements:
+        signoffs_required = max(dct.get(rs["role"], 0), rs["signoffs_required"])
+        dct[rs["role"]] = signoffs_required
+
+    return dct
--- a/auslib/web/admin/views/scheduled_changes.py
+++ b/auslib/web/admin/views/scheduled_changes.py
@@ -50,7 +50,9 @@ class ScheduledChangesView(AdminView):
                 # We don't need to consider the existing version of a row for
                 # inserts, because it doesn't exist yet!
                 if row["change_type"] != "insert":
-                    affected_rows.append(self.table.select(where=base_pk)[0])
+                    base_row = self.table.select(where=base_pk)[0]
+                    scheduled_change["base_row"] = base_row
+                    affected_rows.append(base_row)
                 # We don't need to consider the future version of the row when
                 # looking for required signoffs, because it won't exist when
                 # enacted.
--- a/auslib/web/admin/views/scheduled_changes.py
+++ b/auslib/web/admin/views/scheduled_changes.py
@@ -50,9 +50,9 @@ class ScheduledChangesView(AdminView):
                 # We don't need to consider the existing version of a row for
                 # inserts, because it doesn't exist yet!
                 if row["change_type"] != "insert":
-                    base_row = self.table.select(where=base_pk)[0]
-                    scheduled_change["base_row"] = base_row
-                    affected_rows.append(base_row)
+                    original_row = self.table.select(where=base_pk)[0]
+                    scheduled_change["original_row"] = original_row
+                    affected_rows.append(original_row)
                 # We don't need to consider the future version of the row when
                 # looking for required signoffs, because it won't exist when
                 # enacted.
--- a/auslib/web/admin/views/scheduled_changes.py
+++ b/auslib/web/admin/views/scheduled_changes.py
@@ -3,7 +3,7 @@ from sqlalchemy.sql.expression import null
 
 from flask import jsonify, Response
 
-from auslib.web.admin.views.base import AdminView, requirelogin
+from auslib.web.admin.views.base import AdminView, requirelogin, serialize_signoff_requirements
 from auslib.web.admin.views.history import HistoryView
 from auslib.web.admin.views.problem import problem
 from auslib.web.admin.views.validators import is_when_present_and_in_past_validator
@@ -58,9 +58,8 @@ class ScheduledChangesView(AdminView):
                 # enacted.
                 if row["change_type"] != "delete":
                     affected_rows.append(base_row)
-                for rs in self.table.getPotentialRequiredSignoffs(affected_rows):
-                    signoffs_required = max(scheduled_change["required_signoffs"].get(rs["role"], 0), rs["signoffs_required"])
-                    scheduled_change["required_signoffs"][rs["role"]] = signoffs_required
+                signoff_requirements = self.table.getPotentialRequiredSignoffs(affected_rows)
+                scheduled_change["required_signoffs"] = serialize_signoff_requirements(signoff_requirements)
 
             ret["scheduled_changes"].append(scheduled_change)
         return jsonify(ret)
--- a/ui/app/js/services/rules_service.js
+++ b/ui/app/js/services/rules_service.js
@@ -1,4 +1,4 @@
-angular.module("app").factory('Rules', function($http, ScheduledChanges, Helpers) {
+angular.module("app").factory('Rules', function($http, ScheduledChanges, Helpers, ProductRequiredSignoffs) {
   // these routes map to stubbed API endpoints in config/server.js
   var service = {
     getRules: function() {
@@ -122,14 +122,7 @@ angular.module("app").factory('Rules', function($http, ScheduledChanges, Helpers
       }
 
       var relevantRequirements = productSignoffRequirements.filter(matchesRules);
-      // FIXME: Use a Map here instead of a simple object
-      var merged = {};
-      relevantRequirements.map(function(requirement) {
-        if (!merged[requirement.role] || merged[requirement.role] < requirement.signoffs_required) {
-          merged[requirement.role] = requirement.signoffs_required;
-        }
-      });
-      return {length: Object.keys(merged).length, roles: merged};
+      return ProductRequiredSignoffs.convertToMap(relevantRequirements);
     },
   };
   return service;
--- a/ui/app/js/services/rules_service.js
+++ b/ui/app/js/services/rules_service.js
@@ -88,6 +88,49 @@ angular.module("app").factory('Rules', function($http, ScheduledChanges, Helpers
       url += "?csrf_token=" + encodeURIComponent(data["csrf_token"]);
       return $http.delete(url, data);
     },
+    ruleSignoffsRequired: function(oldRule, newRule, productSignoffRequirements) {
+      // Limited support for the only globs that we encounter in
+      // balrog, which are of the form "foo*".
+      function matchGlob(glob, target) {
+        if (glob[glob.length-1] === '*') {
+          return target.slice(0, glob.length-1) === glob.slice(0, glob.length-1);
+        }
+        return target === glob;
+      }
+
+      // Identify signoffs matching oldRule and newRule
+      function matchesRule(rule, signOffRequirement) {
+        if (rule.product && signOffRequirement.product !== rule.product) {
+          return false;
+        }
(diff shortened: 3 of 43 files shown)


ui/app/templates/rules.html
```diff
@@ -69,8 +69,9 @@ <h3 class="panel-title">
         <span ng-show="!rule_id && rule.scheduled_change === 'delete'" class="label label-danger">Pending Delete</span>
         <button ng-show="!rule_id && rule.scheduled_change === null" class="btn btn-warning btn-xs" ng-click="openNewScheduledRuleChangeModal(rule)">Schedule an Update</button>
         <button ng-show="!rule_id && rule.scheduled_change === null" class="btn btn-danger btn-xs" ng-click="openNewScheduledDeleteModal(rule)">Schedule for Deletion</button>
-        <button ng-show="!rule_id" class="btn btn-default btn-xs" ng-click="openUpdateModal(rule)">Update</button>
-        <button ng-show="!rule_id" class="btn btn-default btn-xs" ng-click="openDeleteModal(rule)">Delete</button>
+        <button ng-show="ruleSignoffsRequired(rule).length" class="btn btn-warning btn-xs">Signoffs required</button>
```
glasserc: I think the "signoffs required" thing shouldn't actually be a button, but I'm not sure what UI it should be.

ui/app/js/services/rules_service.js
```diff
@@ -88,6 +88,49 @@ angular.module("app").factory('Rules', function($http, ScheduledChanges, Helpers
       url += "?csrf_token=" + encodeURIComponent(data["csrf_token"]);
       return $http.delete(url, data);
     },
+    ruleSignoffsRequired: function(oldRule, newRule, productSignoffRequirements) {
```
mozbhearsum: What do you think about having the backend return this instead of calculating it here? We already do it for the Scheduled Changes endpoints (eg: https://github.com/mozilla/balrog/blob/master/auslib/web/admin/views/scheduled_changes.py#L59), and it would allow you to re-use the underlying database layer code that already knows how to do this.

ui/app/templates/rules.html
```diff
@@ -69,8 +69,9 @@ <h3 class="panel-title">
...
+        <button ng-show="!rule_id" class="btn btn-default btn-xs {{ruleSignoffsRequired(rule).length && 'disabled'}}" ng-click="openUpdateModal(rule)">Update</button>
+        <button ng-show="!rule_id" class="btn btn-default btn-xs {{ruleSignoffsRequired(rule).length && 'disabled'}}" ng-click="openDeleteModal(rule)">Delete</button>
```
mozbhearsum: I'm also not entirely sure what to do here. A few thoughts:
* If we're disabling the buttons, maybe we can add title text that says "Cannot modify this object directly due to signoff requirements" (or something shorter to the same effect)
* In #394, I started hiding buttons entirely when they weren't useful. That could be an option too. Do you feel strongly one way or the other? We should make sure we do the same thing.
* Either way, I don't *think* we need the "Signoffs required" item -- having the only functional buttons be "Schedule..." should be enough to make that clear.

ui/spec/services/rules_service_spec.js
```diff
@@ -197,4 +197,64 @@ describe("Service: Rules", function() {
...
+  }));
+
+  it("should match only if both product and channel match", inject(function(Rules) {
+    var signoffRequirements = [
+      {product: "Firefox", channel: "nightly", role: "releng", signoffs_required: 2},
+    ];
+    var rule = {product: "Firefox", channel: "aurora", mapping: "Firefox"};
+    var signoffsRequired = Rules.ruleSignoffsRequired(rule, undefined, signoffRequirements);
+    expect(signoffsRequired.length).toBe(0);
+  }));
```
mozbhearsum: OMG tests! You win!

ui/app/templates/rule_modal.html
```diff
@@ -187,6 +187,10 @@ <h3 class="modal-title" ng-show="!is_edit && is_duplicate">Duplicate Existing Ru
     <input type="hidden" name="data_version" ng-model="rule.data_version">
   </form>
 
+  <div ng-show="ruleSignoffsRequired().length">
+    Making this change requires signoffs and therefore must be a scheduled change.
+  </div>
```
mozbhearsum: It looks like this never ends up getting shown, because you can't open modals like this with the buttons disabled. Can you put it in the scheduled change modal instead? I think it would also be good to list out the required signoffs in there.

ui/app/js/services/rules_service.js
glasserc: How would that work? We would send a request on every change? It doesn't seem like a big win to me because the backend code is focused on database entries, whereas a potential change isn't in the database yet.

ui/app/templates/rule_modal.html
glasserc: This can get shown if, for example, there are required signoffs on Firefox Nightly, and you click "Update" on Firefox Aurora, and change the channel to Nightly.

ui/app/js/services/rules_service.js
mozbhearsum: Hmmm, I think I got my use cases mixed up. I was thinking about https://bugzilla.mozilla.org/show_bug.cgi?id=1343915, which requires information about existing objects. I agree that for this use case (signoff requirements for stuff that doesn't exist yet), this makes sense. Sorry!

ui/app/templates/rule_modal.html
mozbhearsum: Ah, indeed! Perhaps we should disable the "Save Changes" button in that case, too, since the request would just return a 400 anyways?

mozbhearsum: And styling the text a little differently would be nice, too. Maybe the same as the error div just beneath it?

ui/app/templates/rules.html
glasserc: I don't feel like there's a difference between buttons being present but disabled without explanation and buttons being completely absent. I feel like having a disabled button with information explaining why that button is disabled is the best UI. I'm worried less about the release managers who probably know why it's disabled than for the next developer who has to try to figure out what the interface is doing and why. I'm worried less about the release managers who probably know why it's disabled than for the next developer who has to try to figure out what the interface is doing and why.

I like the idea of "title text", but I can't figure out a way to implement it in Angular. Button `title` attributes aren't displayed on hover by default. Bootstrap offers a "tooltip" mechanism but I can't get it to work easily. (It's implemented in Bootstrap JS, and it's opt-in, and it seems like the Bootstrap JS isn't included. Then, ideally, I'd have a clear-enough way to say "show this version of the button if it requires signoffs, and this version otherwise", but there's nothing like "else" in Angular as far as I can tell.)

I guess the right thing to do might be to hide the buttons and punt on tooltips till a later day?

ui/app/templates/rule_modal.html
glasserc: That sounds good, thanks. Do you still want the signoffs listed? Per https://bugzilla.mozilla.org/show_bug.cgi?id=1355074, you wrote "I don't think we need to show the full details of who needs to signoff, just some sort of "signoff required" indicator."

ui/app/templates/rules.html
mozbhearsum: That sounds reasonable to me - we already do it for other things, too (eg: we hide "Schedule an Update" when a change is already scheduled for the Rule).

ui/app/templates/rule_modal.html
mozbhearsum: Heh, apparently I'm indecisive. We may as well show them if we have them - final answer!

glasserc:
I've now added signoff requirements to all the places I've found that it made sense to. I think this might be a good time for another review. Outstanding questions:

- Is it possible to test that buttons are/aren't shown depending on the controller state?
- Is there too much duplication? I felt it was hard to abstract some things because properties had different names in different controllers, etc.
- Are there other places where signoff requirements are necessary?

mozbhearsum:
I ran out of time to give this a full fledged review this week, but I did some toying around with it locally and it appears to be in good shape functionally - I tried various different scheduled changes of various objects, and the correct required signoffs were shown in all all cases.

To your open questions:
* I don't know if we can test the state of buttons. Maybe you can if you get a handle on the DOM? This is still an area of Angular that I know little about.
* Technically, the Delete Rule modal can still be used....but only for things that don't require Signoff. So, there's no reason to add signoff information there. Same goes for direct updates of Rules. I _think_ you should merge the latest upstream master to get #394, which made some changes to how these buttons are shown/hidden.

auslib/web/admin/views/base.py
```diff
@@ -83,3 +83,14 @@ def delete(self, *args, **kwargs):
         self.log.debug("processing DELETE request to %s" % request.path)
         with dbo.begin() as trans:
             return self._delete(*args, transaction=trans, **kwargs)
+
+
+def serialize_signoff_requirements(requirements):
```
mozbhearsum: Hooray! It's good to see this get centralized. There might be an issue with this function though, or maybe one of the callers. Did you see the test failures like:
```
E       -                          u'required_signoffs': {u'releng': 1},
E       ?                         --                      -
E       
E       +                         'required_signoffs': {'releng': 1, 'relman': 1},
E       ?                                                         +++++++++++++```

I haven't seen any changes to the test data, so I _think_ these are legitimate failures.

File: auslib/web/admin/views/base.py [hunk 7]
```diff
@@ -83,3 +83,14 @@ def delete(self, *args, **kwargs):
...
+    dct = {}
```
mozbhearsum: I might just be a bit dense today, but what does `dct` mean?

auslib/web/admin/views/scheduled_changes.py
```diff
@@ -50,15 +50,16 @@ def get(self):
                 # We don't need to consider the existing version of a row for
                 # inserts, because it doesn't exist yet!
                 if row["change_type"] != "insert":
-                    affected_rows.append(self.table.select(where=base_pk)[0])
+                    base_row = self.table.select(where=base_pk)[0]
+                    scheduled_change["base_row"] = base_row
```
mozbhearsum: It looks like this is changing the schema of the responses from from /scheduled_changes/rules et. al. You should update https://github.com/mozilla/balrog/blob/master/auslib/web/admin/swagger/api.yaml#L1815 to reflect this.

auslib/web/admin/views/base.py
glasserc: It's a stand-in for any kind of `dict` (because `dict` itself is a built-in type name). It's not really a semantically meaningful name..

auslib/web/admin/views/base.py
glasserc: I believe the test failures are fixed. The biggest changes had to do with adding `original_row` to all the scheduled changes endpoints.

auslib/web/admin/views/scheduled_changes.py
glasserc: Yep, thanks!

mozbhearsum:
Thanks for all the hard work here - I know this ended up being more work than it seemed, but this is really nice to have!

PR #409, opened 2017-09-19
Smarter partial merging

mozbhearsum:
This is a replacement for the current merging done by "dictdiffer" that handles our lists of completes and partials better. It probably has some bugs still, but it passes all tests. It's not ready for proper review yet, but feedback is welcome.

https://bugzilla.mozilla.org/show_bug.cgi?id=1400141#c15 has additional background on this.


auslib/blobs/base.py
```diff
@@ -55,6 +55,42 @@ def createBlob(data):
...
+def merge_dicts(ancestor, left, right):
+    result = {}
+    for key in ancestor.keys() + left.keys() + right.keys():
+        key_types = set([type(ancestor.get(key)), type(left.get(key)), type(right.get(key))])
+        key_types.discard(type(None))
+        if len(key_types) > 1 and not key_types.issubset([str, unicode]):
+            raise ValueError("Cannot merge blobs: type mismatch for '{}'".format(key.encode('ascii', 'replace')))
+
+        if isinstance(ancestor.get(key), dict) or isinstance(left.get(key), dict) or isinstance(right.get(key), dict):
+            result[key] = {}
```
allan-silva: This initialization to an empty dict seems unnecessary.

auslib/blobs/base.py
```diff
@@ -55,6 +55,42 @@ def createBlob(data):
...
+        else:
+            if key in ancestor:
+                if key in left and key in right and ancestor[key] != left[key] and ancestor[key] != right[key]:
+                    raise ValueError("Cannot merge blobs: left and right are both changing '{}'".format(key.encode('ascii', 'replace')))
+                if key in left and ancestor[key] != left.get(key):
+                    result[key] = left[key]
+                else:
+                    result[key] = right[key]
+            else:
+                if key in left and key in right and left[key] != right[key]:
```
allan-silva: Maybe this validation and validation at line 77 can be abstracted in another function.

auslib/db.py
```diff
@@ -2064,27 +2061,20 @@ def update(self, where, what, changed_by, old_data_version, transaction=None, dr
...
-                        super(Releases, self).update(where={"name": name}, what=what, changed_by=changed_by, old_data_version=tip_data_version,
-                                                     transaction=transaction, dryrun=dryrun)
-                        # cache will have a data_version of one plus the tip
-                        # data_version
-                        new_data_version = tip_data_version + 1
-                    except dictdiffer.merge.UnresolvedConflictsException:
-                        self.log.debug("latest version of release %s cannot be merged with new blob" % name)
+                        what['data'] = createBlob(merge_dicts(ancestor_blob, tip_blob, blob))
+                    except ValueError:
+                        self.log.exception("Couldn't merge blobs.")
```
allan-silva: How merge_dicts raises `ValueError("Cannot merge blobs: left and right are both changing '{}'".format(key.encode('ascii', 'replace')))`, here can log the message.

auslib/blobs/base.py
mozbhearsum: Good point. I think it may have been necessary in an earlier iteration...I'll remove it.

auslib/blobs/base.py
mozbhearsum: Hmmm, I'm not seeing how. Line 77 deals with ancestor, left, and right while this one only deals with left and right.

auslib/db.py
mozbhearsum: I'm not exactly sure what you're suggesting here - can you clarify? If you're saying it should log the message from merge_dicts, that's already happening by log.exception() (it includes the traceback + message).

allan-silva: Ah OK, that's it. I didn't know.

auslib/blobs/base.py
allan-silva: Do you think a function with ancestor parameter none by default is good? So check ancestor only if it is not null.

mozbhearsum: I might be missing something here, but I'm not sure how it will help. The two lines you mentioned test different things (ancestor vs left and ancestor vs right on line 77, left vs right on line 84).

mozbhearsum:
I'm still doing some final testing with jamun nightlies+releases, but I think it's about time I get some review on this. Between the existing merging tests and the new merge_dicts tests that I've added, I'm reasonably confident that this works.

@catlee, I know you've looked and thought about this a bit already, so I'd appreciate any other thoughts you have. @nthomas-mozilla - you're probably the best person to review this from top to bottom.

auslib/blobs/base.py
```diff
@@ -55,6 +55,41 @@ def createBlob(data):
     return blob_map[schema_version](**data)
 
 
+def merge_dicts(ancestor, left, right):
+    result = {}
+    for key in ancestor.keys() + left.keys() + right.keys():
```
catlee: IIRC, keys() returns a list, so you may end up handling the same key twice. Maybe put this into a set?

auslib/blobs/base.py
```diff
@@ -55,6 +55,41 @@ def createBlob(data):
     return blob_map[schema_version](**data)
 
 
+def merge_dicts(ancestor, left, right):
```
catlee: Reading through this function, there are lots of places where you're looking at all ancestor, left and right.

Would having these 3 in a variable help clean the code up?

`
dicts = (ancestor, left, right)
for key in set(key for d in dicts for key in d.keys()):
...
`

and

`
if any(isinstance(d.get(key), dict) for d in dicts):
...
`

auslib/blobs/base.py
```diff
@@ -55,6 +55,41 @@ def createBlob(data):
...
+            raise ValueError("Cannot merge blobs: type mismatch for '{}'".format(key.encode('ascii', 'replace')))
+
+        if isinstance(ancestor.get(key), dict) or isinstance(left.get(key), dict) or isinstance(right.get(key), dict):
+            result[key] = merge_dicts(ancestor.get(key, {}), left.get(key, {}), right.get(key, {}))
+        elif isinstance(ancestor.get(key), list) or isinstance(left.get(key), list) or isinstance(right.get(key), list):
+            result[key] = ancestor.get(key, [])[:]
+            for side in (left, right):
+                for i in side.get(key, []):
+                    if i not in result[key] or not isinstance(i, type(result[key][result[key].index(i)])):
+                        result[key].append(i)
```
catlee: it would be easier to test this if it was in a merge_lists function

auslib/blobs/base.py
```diff
@@ -55,6 +55,41 @@ def createBlob(data):
...
+        else:
+            if key in ancestor:
+                if key in left and key in right and ancestor[key] != left[key] and ancestor[key] != right[key]:
+                    raise ValueError("Cannot merge blobs: left and right are both changing '{}'".format(key.encode('ascii', 'replace')))
+                if key in left and ancestor[key] != left.get(key):
+                    result[key] = left[key]
+                else:
```
catlee: I think you need an `if key in right` here too.

auslib/blobs/base.py
```diff
@@ -55,6 +55,41 @@ def createBlob(data):
...
+                    result[key] = right[key]
+            else:
+                if key in left and key in right and left[key] != right[key]:
+                    raise ValueError("Cannot merge blobs: left and right are both changing '{}'".format(key.encode('ascii', 'replace')))
+                if left.get(key) is not None:
```
catlee: Is None ever a valid value? If so, `if key in left` would be better here.

auslib/blobs/base.py
mozbhearsum: Seems like it should - I'll give it a shot.

auslib/blobs/base.py
mozbhearsum: Good idea.

auslib/blobs/base.py
mozbhearsum: Yes....it would.

auslib/blobs/base.py
mozbhearsum: It's not, but that's a better way to write it anyways!

auslib/blobs/base.py
mozbhearsum: I think this works either way (line 76 checks that left or right is changing the key, and line 78 checks if it's left, so the only option left is right) -- but maybe explicit is better here. I think we'd need if/elif/else to do this, otherwise result[key] might end up empty.

mozbhearsum:
While testing this patch with release promotion I was surprised to see a few 400s. After a lot of investigating, I've discovered that weren't even using the _old_ merge logic on most release updates :(. Most of those end up going through Releases.addLocaleToRelease, but the merging logic happens in Releases.update. Releases.addLocaleToRelease ends up calling the parent class' update (https://github.com/mozilla/balrog/blob/3b4a2fd72fa4ab1f015e8d1d5fabf754a015d5e3/auslib/db.py#L2132), which bypasses the merging. It looks like this is a regression from about a year ago - the previous version of addLocaleToRelease correctly calls self.update() instead of super(Releases, self).update() (per https://github.com/mozilla/balrog/commit/cffc810c6f8f5b205fd956602a89dacf97036ded#diff-7c2c8a646f34fec41c773940931bf694L1237)

My latest commit adds a test + fix for this, and I've fired off another staging release to see how that goes.

mozbhearsum:
My staging release with the latest version of this PR seemed to go better. There were 5 instances of 400s still (compared to at least 20 with the previous attempt), and 17 log messages indicating that merging was successful (compared to 0 with the previous attempt).

The 400s that remain all have "History.getChange#752: Found 2 changes, should have been 1" messages associated with them. I dug into that a bit, and discovered that we sometimes have duplicate data versions for the same PK in the releases_history table. For example, there are two changes with data_version 227 on http://54.90.211.22:8080/releases/Firefox-58.0b1-build3. I haven't dug into that any further yet, but I suspect https://bugzilla.mozilla.org/show_bug.cgi?id=1246993 may be related.

mozbhearsum:
I also filed https://bugzilla.mozilla.org/show_bug.cgi?id=1402928 with some ideas to help make server side merging more debuggable in the future.

auslib/blobs/base.py
```diff
@@ -55,6 +55,51 @@ def createBlob(data):
     return blob_map[schema_version](**data)
 
 
+def merge_lists(*lists):
+    result = []
+    for l in lists:
+        for i in l:
+            if i not in result or not isinstance(i, type(result[result.index(i)])):
+                result.append(i)
```
nthomas-mozilla: This is where the lists of partials and completes get built up, I think, so what does the isinstance check do ? It looks kinda like a str vs unicode special case for hypothesis.

auslib/test/blobs/test_base.py
```diff
@@ -63,3 +67,132 @@ def testSchemaCaching(self):
...
+# ends up failing tests.
+useful_values = st.none() | st.floats(allow_nan=False) | st.text()
+
+
+@given(st.lists(useful_values), st.lists(useful_values), st.lists(useful_values))
+def test_merge_lists_unique_items_present(base, left_additional, right_additional):
+    left = deepcopy(base) + left_additional
+    right = deepcopy(base) + right_additional
+    expected = sorted(list(set(deepcopy(base) + left_additional + right_additional)))
+    got = sorted(merge_lists(base, left, right))
```
nthomas-mozilla: [Purist mode on] The use of sorting in the test ignores the convention that lists are ordered. I don't think that actually matters for partials/completes down at the locale level, but could you go
   ``expected = deepcopy(base) + left_additional + [l for l in right_additional if l not in left_additional]``
and make the test stronger ?

auslib/test/blobs/test_base.py
```diff
@@ -63,3 +67,132 @@ def testSchemaCaching(self):
...
+    assert got == expected
+
+
+def unique_items_only(i):
+    """Returns a hashable value for various types to be used for testing
+    uniqueness when generating test data."""
+    if isinstance(i, dict):
+        return tuple(i.keys())
+    if isinstance(i, list):
+        return tuple(i)
```
nthomas-mozilla: Not technically unique values in the list case, unless you cast it into a set first ?

auslib/test/blobs/test_base.py
```diff
@@ -63,3 +67,132 @@ def testSchemaCaching(self):
...
+    got = sorted(merge_lists(base, left, right))
+    assert got == expected
+
+
+@given(st.lists(useful_values), st.lists(useful_values))
+def test_merge_lists_no_dupes(base, additional):
+    left = deepcopy(base) + additional
+    right = deepcopy(base) + additional
+    expected = sorted(list(set(deepcopy(base) + additional)))
+    got = sorted(merge_lists(base, left, right))
```
nthomas-mozilla: ``expected = left; got = merge_lists(base, left, right)`` or is there some subtlety I'm missing ?

auslib/blobs/base.py
mozbhearsum: Yep, pretty much. merge_lists and merge_dicts are written more generically than is purely necessary. Do you think it would be preferable to target them more, and adjust the tests to remove more cases that we know we won't encounter?

auslib/test/blobs/test_base.py
mozbhearsum: Good idea, let's do that.

In an ideal world, we'd be treating partials/completes as sets (or maybe dicts) everywhere and this wouldn't be an issue.

auslib/test/blobs/test_base.py
mozbhearsum: Unfortunately, we need to de-dupe `expected` because we may get tests like:
`Falsifying example: test_merge_lists_no_dupes(base=[None, None], additional=[])`

And once we're de-duping, casting back to a sorted list is required to avoid failures due to ordering differences.

This can be simplified to use `left` as the base for expected, but not any further as far as I can tell.

auslib/test/blobs/test_base.py
mozbhearsum: It depends on how you define unique :). In this case, we're looking at whether the list as a whole is unique compared to another list (aka: do we have the same items in the same places).

mozbhearsum:
I did another staging release overnight. Functionally, it should have used the same code as the previous one (the only non-test changes were https://github.com/mozilla/balrog/pull/409/commits/3baee2e93ec1c65268dce204710b11e3fad7f8fb, which shouldn't affect behaviour). For whatever reason, it did both better and worse at merging. There were 15 failures to merge - again all accompanied by the getChange error, and 23 successes.

Looking at the database, I found 16 duplicate data_versions for the previous release and 37 for the latest one -- so it doesn't surprise me that we had more merge errors with it.

@relud ran a query to get some idea of how many duplicate data_versions we tend to have in production. Unfortunately, this number appears to be very high (in the hundreds) since 56.0b6. Curiously, this is about 5 days prior to us enabling 2 nightlies per day, but perhaps it was high in 56.0b6 & 7 for other reasons (eg: submitting at the same day as the old nightlies). The full data for that is on https://gist.github.com/mozbhearsum/436fa1fd4f0434b46bc9c95755bb43af.

It's hard to be certain, but based on the evidence of the most recent two staging releases, this PR should still improve the situation, so I'd like to avoid any more scope creep on it and get it merged when everyone is happy with it. @nthomas-mozilla, @catlee - how do you feel about this, and the current state of this PR?

Given how many we have in production, I think it's also important to pursue a fix for the duplicate data_versions (or at the very least, a workaround that makes merging possible still). I'll get started on that ASAP.

auslib/blobs/base.py
nthomas-mozilla: I'd be happy if there was a comment to provide some explanation.

auslib/test/blobs/test_base.py
nthomas-mozilla: OK, so ``base`` may have dupes and merge_lists will remove them, and the test needs to account for that. Does that reflect a useful situation for Balrog ? Would it make sense to dedupe base ?

auslib/blobs/base.py
mozbhearsum: I added docstrings to merge_lists and merge_dicts in https://github.com/mozilla/balrog/commit/70e07aae4d39854c0d5b1323cf4b5e75c5f0c680.

PR #410, opened 2017-09-20
patch 1380874: Remove signoff text from history

Alweezy:
#### What does this PR do?
Removes text about ` signoffs` from history pages.
#### How should this be manually tested?
Once `balrog` is running on `http://127.0.0.1:8080` navigate to `Rules` and schedule a change.
Click on the `history` button on that scheduled change, there should be no text about `signoffs`.
#### What are the relevant bugzilla stories?
https://bugzilla.mozilla.org/show_bug.cgi?id=1380874


ui/app/templates/rule_scheduled_changes.html
```diff
@@ -54,13 +54,13 @@
     <h3 class="panel-title">
       <div style="float: right">
         <i ng-if=" scheduled_changes_rules_count && $first && (currentPage == 1)">Current</i>
-        <button class="btn btn-xs btn-primary" ng-click="signoff(sc)"
-                ng-show="! isEmpty(sc['required_signoffs']) && ! sc['signoffs'].hasOwnProperty(current_user)">
-            Signoff as...
+        <button ng-show="!scheduled_changes_rules_count && (! isEmpty(sc['required_signoffs']) &&
```
mozbhearsum: It would be better to use sc_id here -- that's the variable that actually determines whether or not we're on the history page.

ui/app/templates/permission_scheduled_changes.html
```diff
@@ -42,13 +42,13 @@
     <h3 class="panel-title">
       <div style="float: right">
         <i ng-if=" scheduled_changes_rules_count && $first && (currentPage == 1)">Current</i>
-        <button class="btn btn-xs btn-primary" ng-click="signoff(sc)"
-                ng-show="! isEmpty(sc['required_signoffs']) && ! sc['signoffs'].hasOwnProperty(current_user)">
-            Signoff as...
+        <button ng-show="!scheduled_changes_rules_count && (! isEmpty(sc['required_signoffs']) &&
```
mozbhearsum: I think you've got a copy/paste error here and on the releases page - scheduled_changes_rules_count doesn't exist. See below about changing this to use a different variable, too.

Alweezy: @mozbhearsum The copy/paste error, yes. I have realized that permissions have no history. I have also changed to detect page using `sc.sc_id` please check if its okay now.

mozbhearsum: Did you forget to push an update? This seems to be unchanged since the last time I looked at it.

mozbhearsum:
Are you going to finish this up @Alweezy?

Alweezy:
@bhearsum yes, I'll resume working on it.

collin5:
Hi @mozbhearsum, PR should now be ready for another review. Thanks

PR #411, opened 2017-09-21
Bug 1341622: autofill/hints for permissions UI

collin5:
PR adds autofill/hints for `options` on permissions UI.

[https://github.com/mozilla/balrog/pull/387](https://github.com/mozilla/balrog/pull/387) fixes adding hints for `permission_input` as a drop down


ui/app/js/controllers/user_edit_controller.js
```diff
@@ -5,6 +5,88 @@ function ($scope, $modalInstance, CSRF, Permissions, users, is_edit, user) {
...
+      // clear  all previous selections 
+      [
+          $scope.selected_options,
+          $scope.selected_actions,
+          $scope.selected_products
+      ] = [[], [], []];
+  };
+  $scope.selected_options = [];
+
+  $scope.actions = ["create", "modify", "enact"];
```
mozbhearsum: Keep in mind that actions are different depending on what the Permission is.

ui/app/js/controllers/user_edit_controller.js
```diff
@@ -5,6 +5,88 @@ function ($scope, $modalInstance, CSRF, Permissions, users, is_edit, user) {
...
+  $scope.products = [];
+  $scope.selected_products = [];
+  $scope.product = null;
+
+  // get products in releases
+  Permissions.getProducts().success(function(response) {
+    $scope.products = response.product;
+  });
+
+  $scope.$watch(function(){
```
mozbhearsum: I don't think this is how $scope.$watch is supposed to work...the first argument to it is a watchExpression, which is thing that determines whether or not to run some code when something changes, _not_ something that makes changes. You can see an example in Balrog's codebase here: https://github.com/mozilla/balrog/blob/master/ui/app/js/controllers/rules_controller.js#L115

It's possible you may want multiple $watch expressions to handle options, products, and actions.

ui/app/js/controllers/user_edit_controller.js
```diff
@@ -5,6 +5,88 @@ function ($scope, $modalInstance, CSRF, Permissions, users, is_edit, user) {
...
+      // set actions & products to [] when undefined, always happens when running tests
+      $scope.actions = (typeof $scope.actions === 'undefined')? []: $scope.actions;
+      $scope.products= (typeof $scope.products === 'undefined')? []: $scope.products;
```
mozbhearsum: When would they be set to undefined? They're initialized just a few lines up, so this should never be the case, even in tests....

ui/app/js/controllers/user_edit_controller.js
```diff
@@ -5,6 +5,88 @@ function ($scope, $modalInstance, CSRF, Permissions, users, is_edit, user) {
...
+
+  $scope.onPermission = function(){
+      $scope.available_options = allPermissions[$scope.permission.permission];
+
+      // clear  all previous selections 
+      [
+          $scope.selected_options,
+          $scope.selected_actions,
+          $scope.selected_products
+      ] = [[], [], []];
```
mozbhearsum: This would be easier to read if it was written more plainly as:
$scope.selected_options = [];
etc.

ui/app/js/services/permissions_service.js
```diff
@@ -102,6 +102,9 @@ angular.module("app").factory('Permissions', function($http, $q, ScheduledChange
       url += "?csrf_token=" + encodeURIComponent(data["csrf_token"]);
       return $http.delete(url, data);
     },
+    getProducts: function(){
+      return $http.get('/api/releases/columns/product');
+    }
```
mozbhearsum: No need to do this - you can just pass the Releases service to the permissions controller and use https://github.com/mozilla/balrog/blob/master/ui/app/js/services/releases_service.js#L18 instead.

ui/app/templates/permissions_modal.html
```diff
@@ -23,7 +23,7 @@ <h3 class="modal-title" ng-show="showRow()"> Username: {{ user.username }}</h3>
         <form role="form" ng-submit="is_edit ? updatePermission(permission) : addPermission()">
           <div class="form-group" ng-class="{'has-error': errors.permission}">
             <label for="id_permission">Permission</label><br>
-              <select name="singleSelect" class="form-control" id="id_permission" ng-model="permission.permission">
+              <select name="singleSelect" class="form-control" id="id_permission" ng-model="permission.permission" ng-change="onPermission()">
```
mozbhearsum: Have you considered using $scope.$watch on permission.permission instead? It would be more consistent with what we do elsewhere.

collin5: This is actually better and cleaner, thanks

ui/app/js/controllers/user_edit_controller.js
collin5: @mozbhearsum Thanks a lot for the resource, you just saved my day. Had spent lots hours trying to figure out all the possible options. 👍

collin5:
@mozbhearsum Implemented all requested changes. Also implemented the `options autofill` as a directive to avoid code repetition. Thanks for feedback :)

ui/app/js/directives/permission_options_directive.js
```diff
@@ -0,0 +1,127 @@
...
+            $scope.products = [];
+            $scope.selected_products = [];
+            $scope.product = null;
+
+            // get products in releases
+            Releases.getProducts().success(function(response) {
+                $scope.products = response.product;
+            });
+
+            $scope.$watch('[ action, product ] | json',function(){
```
mozbhearsum: What does the `| json` do here?

ui/app/js/directives/permission_options_directive.js
```diff
@@ -0,0 +1,127 @@
...
+                // add tag on action input
+                if ($scope.actions[$scope.permission.permission].indexOf($scope.action) !== -1 && $scope.selected_actions.indexOf($scope.action) === -1){
+                    $scope.selected_actions.push($scope.action);
+                    $scope.action = null;
```
mozbhearsum: Have you considered using sets (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) for the `selected` variables? They should let you avoid doing duplication checks manually.

ui/app/js/directives/permission_options_directive.js
```diff
@@ -0,0 +1,127 @@
+angular.module('app').directive("permissionOptions", [function(){
+    return {
+        restrict: 'E',
+        scope: {
+            permission: '='
+        },
+        template: ' <div class="form-group" ng-class="{\'has-error\': errors.options_as_json}"> '+
```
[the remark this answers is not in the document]
collin5: Nice Idea, will do that ASAP

ui/app/js/directives/permission_options_directive.js
collin5: Tried `$scope.$watch('[ action, product ]', function() {}` and wasn't working. Got the idea from SO  https://stackoverflow.com/questions/11952579/watch-multiple-scope-attributes#answer-14410060

ui/app/js/directives/permission_options_directive.js
collin5: Yeah, will also do that ASAP

collin5: Tried using sets but I keep on getting `Error: Can't find variable: Set` when running tests

ui/app/js/controllers/permission_scheduled_change_new_controller.js
```diff
@@ -42,6 +42,7 @@ function ($scope, $modalInstance, CSRF, Permissions, scheduled_changes, sc) {
 
 
   $scope.addScheduledPermission = function() {
+    $scope.permission.options = $scope.permission.options_as_json;
```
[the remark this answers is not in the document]
collin5: Yeah, I was getting unexpected responses when adding scheduled permissions till I checked the request content and figured that It was sending the options twice, as `options_as_json` and as `options`. 
screen shot 2017-09-27 at 4 46 20 pm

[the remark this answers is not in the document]
collin5: And the `options` are not sent coz they are empty on update

collin5:
@mozbhearsum Thanks again for the feedback, I've implemented some of the requested changes, though still having a little trouble with some.

ui/app/js/directives/permission_options_directive.js
mozbhearsum: OK, so it's clear that it's performing some useful function - but it would be good to understand exactly what.

ui/app/js/directives/permission_options_directive.js
mozbhearsum: I think we talked about this elsewhere...sets probably aren't available in the version of phantomjs that we use, we'll have to do without them :(

mozbhearsum:
Good progress - let me know when you think this is ready for another review.

ui/app/js/directives/permission_options_directive.js
collin5: @mozbhearsum Is it ok if I just add a comment for that ?

ui/app/js/directives/permission_options_directive.js
collin5: Yeah 👍

collin5:
@bhearsum Thanks, I guess PR is ready for another review 👍

ui/app/js/directives/permission_options_directive.js
mozbhearsum: Yes...but the current comment doesn't really answer the question. Is the `[...] | json` structure needed because you're watching multiple variables instead of just one? That's the part I don't understand.

ui/app/templates/permission_options.html
```diff
@@ -0,0 +1,30 @@
+<div class="form-group" ng-class="{'has-error': errors.options_as_json}">
```
mozbhearsum: Much cleaner - thank you!

ui/app/js/directives/permission_options_directive.js
collin5: @mozbhearsum  Yeah exactly, It is because I'm watching both `action` and `product` at the same time.

collin5:
@mozbhearsum Thanks, this is fixed now 👍 . Products or actions which are not in autofill can now be added by using a comma `,` as the token separator.

collin5:
@mozbhearsum 👍 All done :)

PR #413, opened 2017-09-27
Bug 1391012 make it clear that fallbackMapping is unused when backgroundRate is 100

hemant1612:
Fixes the issue, with some alert text.

### DIFF
--- a/ui/app/templates/rules.html
+++ b/ui/app/templates/rules.html
@@ -205,6 +205,11 @@
             </a>
           </span>
       </h5>
+      <span ng-show="(rule.fallbackMapping || (!rule_id && show_sc && rule.scheduled_change !== null && rule.scheduled_change.fallbackMapping !== null )) && (rule.backgroundRate === 100 || rule.scheduled_change.backgroundRate === 100)" >
+        <span class="alert-danger">
+          <strong>Note!</strong> Fallback Mapping is unused when Background Rate is 100
+        </span>        
+     </span>
       <rulefield showdiff="!rule_id" showsc="show_sc" rule="rule" fieldname="data_version" fieldtitle="Data Version"></rulefield>
       <rulefield showdiff="!rule_id" showsc="show_sc" rule="rule" fieldname="rule_id" fieldtitle="Rule ID"></rulefield>
       <rulefield showdiff="!rule_id" showsc="show_sc" rule="rule" fieldname="comment" fieldtitle="Comment"></rulefield>
@@ -229,11 +234,6 @@
                  help='Whether or not to check a client&#39;s 32 -> 64-bit migration opt-in status.'></rulefield>
       <rulefield showdiff="!rule_id" showsc="show_sc" rule="rule" fieldname="update_type" fieldtitle="Update type"></rulefield>
       <rulefield showdiff="!rule_id" showsc="show_sc" rule="rule" fieldname="headerArchitecture" fieldtitle="Header Architecture"></rulefield>
-      <span ng-show="rule.fallbackMapping || (!rule_id && show_sc && rule.scheduled_change !== null && rule.fallbackMapping !== rule.scheduled_change.fallbackMapping && rule.scheduled_change.fallbackMapping !== null) && rule.backgroundRate === 100" > 
-       <span class="alert-danger">
-         <strong>Note!</strong> Fallback Mapping is unused when Background Rate is 100             
-       </span>
-      </span>
     </div>
 
   </div>
--- a/ui/app/templates/rules.html
+++ b/ui/app/templates/rules.html
@@ -205,7 +205,7 @@
             </a>
           </span>
       </h5>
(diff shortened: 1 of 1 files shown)


ui/app/templates/rules.html
```diff
@@ -229,6 +229,11 @@ <h5 title="Fallback Mapping" ng-show="rule.fallbackMapping || (show_sc && rule.s
                  help='Whether or not to check a client&#39;s 32 -> 64-bit migration opt-in status.'></rulefield>
       <rulefield showdiff="!rule_id" showsc="show_sc" rule="rule" fieldname="update_type" fieldtitle="Update type"></rulefield>
       <rulefield showdiff="!rule_id" showsc="show_sc" rule="rule" fieldname="headerArchitecture" fieldtitle="Header Architecture"></rulefield>
+      <span ng-show="rule.fallbackMapping || (!rule_id && show_sc && rule.scheduled_change !== null && rule.fallbackMapping !== rule.scheduled_change.fallbackMapping && rule.scheduled_change.fallbackMapping !== null) && rule.backgroundRate === 100" >
```
mozbhearsum: This condition doesn't seem to be fully working. For example, when backgroundRate is set to 0 (or even 99), the text still shows up: https://screenshots.firefoxusercontent.com/images/12650353-b623-4e76-ad64-a20f4245512d.png

You also need to account for a possible scheduled change here. The logic for showing this should be something like:
* If the Rule's backgroundRate is not 100 OR the scheduled change's backgroundRate is not 100, show the warning

Please ensure to test the following scenarios:
* Rule at 100 with no scheduled change
* Rule < 100 with no scheduled change
* Rule and scheduled change both at 100
* Rule and scheduled change both < 100
* Rule < 100, scheduled change at 100
* Rule at 100, scheduled change < 100
* Just a scheduled change (use the "Schedule a new Rule" button) at 100
* Just a scheduled change (use the "Schedule a new Rule" button) < 100

hemant1612: Ahh ! thats probably because a closed parenthesis at the start of  `rule_id &&` is missing.

Does it boils down to fact that if fallbackMapping is present(either from rule or scheduled change) if  backgroundRate( either from rule or scheduled change ) is 100 then throw an error.

mozbhearsum: Yep, that's another way to phrase it!

ui/app/templates/rules.html
```diff
@@ -205,6 +205,11 @@ <h5 title="Fallback Mapping" ng-show="rule.fallbackMapping || (show_sc && rule.s
             </a>
           </span>
       </h5>
+      <span ng-show="(rule.fallbackMapping || (!rule_id && show_sc && rule.scheduled_change !== null && rule.scheduled_change.fallbackMapping !== null )) && (rule.backgroundRate === 100 || rule.scheduled_change.backgroundRate === 100)" >
```
mozbhearsum: Closer, but this still isn't working in at least a couple of cases. For example, the note should be shown in this case: https://screenshots.firefoxusercontent.com/images/92e2a8b5-8969-4234-a9e1-32787b97daf7.png

Part of the problem is that the last part of this condition uses "=== 100" -- that's actually the opposite of what we want, it should _never_ unless the backgroundRate of the Rule or Scheduled Change is _not_ 100.

Please make sure you do some of your own testing locally to ensure the cases I identified earlier are working.

hemant1612: This was difficult to digest for me, until I went through the bug again, and realized why this is so. Actually I took it other way as if fallback Mapping is already set, but here we  are trying to set it afterwards. 

I tested all cases and it should work now.

ui/app/templates/rules.html
```diff
@@ -205,6 +205,11 @@ <h5 title="Fallback Mapping" ng-show="rule.fallbackMapping || (show_sc && rule.s
             </a>
           </span>
       </h5>
+      <span ng-show="(rule.fallbackMapping || (!rule_id && show_sc && rule.scheduled_change !== null && rule.scheduled_change.fallbackMapping !== null )) && (rule.backgroundRate !== 100 || rule.scheduled_change.backgroundRate !== 100)" >
```
allan-silva: Question from who has no experience with frontend:

Can expressions like bellow, be abstracted as controller functions?
`(rule.fallbackMapping || (!rule_id && show_sc && rule.scheduled_change !== null && rule.scheduled_change.fallbackMapping !== null )) && (rule.backgroundRate !== 100 || rule.scheduled_change.backgroundRate !== 100)`

On this page there is 5 tests like `!rule_id && show_sc && rule.scheduled_change !== null`

mozbhearsum: Yeah, it could be moved to a function defined in $scope, and the function could be called from there. That's not a blocker IMO. Sometimes we avoid doing that because there are not many instances of exactly the same condition (eg: it could look at different fields, or maybe one of the parts is removed). I don't feel strongly about this.

PR #416, opened 2017-10-03
1365658 fix priority background rate validators

dmkitui:
#### What does this PR do?
This PR is to resolve [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1365658)
To supplement the backend validation hack that had been implemented earlier with front-end validation for Priority and Background Rate input fields with specific validation errors.
#### Description of Task to be completed?
Implement integer and range validation for Priority and Background rate inputs on the UI.
#### How should this be manually tested?
1. Once balrog is running on http://127.0.0.1:8080 navigate to `Rules` and click on `Add a New Rule` or `Update` on an existing rule. 
2. Test by inputting invalid data for `Rate`- either negative numbers(eg `-40`), over 100(eg 144) or non-digit characters(eg 'x', or '&').
3. Also, input invalid data for `Priority` - either negative numbers or non-digit characters. 
4. An error label will appear below the respective input with a relevant error message when the user clicks on the save changes button.
5. Repeat 2 - 3 after navigating to 'Rules/Scheduled Changes' and 'Rules/Scheduled Changes/Update'. 
 
#### What are the relevant bugzilla stories?
https://bugzilla.mozilla.org/show_bug.cgi?id=1365658
#### Screenshots (if appropriate)
screen shot 2017-10-03 at 14 35 00

### DIFF
--- a/ui/app/js/controllers/rule_edit_controller.js
+++ b/ui/app/js/controllers/rule_edit_controller.js
@@ -34,9 +34,9 @@ function ($scope, $modalInstance, CSRF, Rules, Releases, rule, signoffRequiremen
     $scope.saving = true;
 
     // Evaluate the values entered for priority and background rate.
-    $scope.integer_validation_errors = Helpers.integerValidator({'priority': $scope.rule.priority, 'rate': $scope.rule.backgroundRate});
+    $scope.errors = Helpers.integerValidator({'priority': $scope.rule.priority, 'rate': $scope.rule.backgroundRate});
     // Stop sending the request if any number validation errors.
-    if($scope.integer_validation_errors.priority || $scope.integer_validation_errors.rate) {
+    if($scope.errors.priority || $scope.errors.rate) {
       $scope.saving = false;
       return;
     }
--- a/ui/app/js/controllers/rule_edit_controller.js
+++ b/ui/app/js/controllers/rule_edit_controller.js
@@ -75,6 +75,7 @@ function ($scope, $modalInstance, CSRF, Rules, Releases, rule, signoffRequiremen
       .error(function(response) {
         if (typeof response === 'object') {
           $scope.errors = response;
+          Helpers.addErrorFields($scope.errors);
           sweetAlert(
             "Form submission error",
             "See fields highlighted in red.",
--- a/ui/app/js/controllers/rule_edit_controller.js
+++ b/ui/app/js/controllers/rule_edit_controller.js
@@ -34,9 +34,9 @@ function ($scope, $modalInstance, CSRF, Rules, Releases, rule, signoffRequiremen
     $scope.saving = true;
 
     // Evaluate the values entered for priority and background rate.
-    $scope.errors = Helpers.integerValidator({'priority': $scope.rule.priority, 'backgroundRate': $scope.rule.backgroundRate});
+    $scope.integer_validation_errors = Helpers.integerValidator({'priority': $scope.rule.priority, 'backgroundRate': $scope.rule.backgroundRate});
     // Stop sending the request if any number validation errors.
-    if($scope.errors.priority || $scope.errors.backgroundRate) {
+    if($scope.integer_validation_errors.priority || $scope.integer_validation_errors.backgroundRate) {
       $scope.saving = false;
       sweetAlert(
         "Form submission error",
@@ -44,10 +44,8 @@ function ($scope, $modalInstance, CSRF, Rules, Releases, rule, signoffRequiremen
         "error"
       );
       return;
-    } else {
-      // Re-initialise the 'error' variable if no validation errors found in UI validation.
-      $scope.errors = {};
     }
+    
     CSRF.getToken()
     .then(function(csrf_token) {
       // The data we need to submit is a tweaked version of just the Rule fields, so
--- a/ui/app/js/controllers/rule_edit_controller.js
+++ b/ui/app/js/controllers/rule_edit_controller.js
@@ -1,7 +1,7 @@
 /*global sweetAlert */
 angular.module('app').controller('RuleEditCtrl',
-function ($scope, $modalInstance, CSRF, Rules, Releases, rule, signoffRequirements, pr_ch_options) {
-
+function ($scope, $modalInstance, CSRF, Rules, Releases, rule, signoffRequirements, pr_ch_options, Helpers) {
+  
   $scope.names = [];
   Releases.getNames().then(function(names) {
     $scope.names = names;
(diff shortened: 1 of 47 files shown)


ui/app/templates/rule_modal.html
```diff
@@ -47,15 +47,31 @@ <h3 class="modal-title" ng-show="!is_edit && is_duplicate">Duplicate Existing Ru
...
-          <input type="text" class="form-control" id="id_rate" ng-model="rule.backgroundRate">
-          <p class="help-block" ng-show="errors.backgroundRate">{{ errors.backgroundRate.join(', ') }}</p>
+          <input type="text"
+                 class="form-control"
+                 ng-class="{'validation_errors': integer_validation_errors.rate}"
+                 name="rate"
+                 id="id_rate"
+                 ng-model="rule.backgroundRate"
+                 required>
+          <div class="form-group" ng-class="{'has-error': integer_validation_errors.rate}" ng-show="integer_validation_errors.rate">
```
mozbhearsum: What's the purpose of the `<div>`? Why not just apply the ng-class to the `<p>` directly?

mozbhearsum: I think there's a subtle bug here by not looking at errors.backgroundRate - if the backend catches an issue with this field that the frontend doesn't, the error won't be displayed. Can you make sure we still show backend errors, too?

ui/app/js/controllers/rule_new_controller.js
```diff
@@ -25,10 +25,20 @@ function($scope, $http, $modalInstance, CSRF, Releases, Rules, rules, rule, pr_c
   $scope.saveChanges = function () {
     $scope.saving = true;
     $scope.errors = {};
+    $scope.rate_error = {};
```
mozbhearsum: This looks unused. Please remove it if that's the case.

ui/app/js/controllers/rule_new_controller.js
```diff
@@ -25,10 +25,20 @@ function($scope, $http, $modalInstance, CSRF, Releases, Rules, rules, rule, pr_c
...
     CSRF.getToken()
     .then(function(csrf_token) {
       rule = angular.copy($scope.rule);
-      // rule.priority = '' + rule.priority;
+      
+      // Evaluate the values entered for priority and background rate.
+      $scope.integer_validation_errors = Helpers.integerValidator({'priority': rule.priority, 'rate': rule.backgroundRate});
```
mozbhearsum: It would be ideal if you could integrate this into $scope.errors instead of introducing another variable for it. See below for an example of a bug that this causes...

ui/app/templates/rule_modal.html
dmkitui: I was having trouble with the CSS Specificity in applying the error styling to the `<p>` elements directly. I have somehow managed to get rid of the `<div>` elements by adding a `p.has-error` css definition that takes care of the specificity issues. Is there a better way?

ui/app/js/controllers/rule_new_controller.js
dmkitui: The reason I decided to define a separate variable for the UI errors(specifically the two in hand) was the `$scope.errors` is populated from the response after making the POST request.  And this is when backend errors are caught and displayed in the 2nd-to-last `<div>`. 

This bug fix specifies to validate the two fields before the POST request.

With the backend hacks back in place, even the *backgroundRate* and *Priority* error messages are displayed in this div.
screen shot 2017-10-04 at 11 26 07
screen shot 2017-10-04 at 11 23 04

And I have actually confirmed that we can use the `$scope.errors` in the meantime, to hold the UI validation errors, so making the change.

ui/app/js/controllers/rule_new_controller.js
dmkitui: Removed. It was a relic from previous attempts at a solution which I neglected to remove. Good catch.

ui/app/templates/rule_modal.html
dmkitui: I have highlighted this in another comment, but gist is the backend validation errors are displayed in the second-to-last error label div. This includes the ones caught for the *backgroundRate* and *Priority*. 
screen shot 2017-10-04 at 11 23 04

Let me see if I can decode the `details` in the response object and have these error labels also light up should an error be caught by the backend validators. But in view, this in as much brings about uniformity in the error display, it will also be duplication of error flagging on the UI.

ui/app/js/controllers/rule_new_controller.js
dmkitui: Having looked further, I believe the format of the response -that in turn populates the $scope.error is not what was intended to be: Currently we have a response of the following format:
 `{detail: "Invalid input for backgroundRate. Not an integer. backgroundRate: u'y' is not a 'backgroundRate'", status: 400, title: "Bad Request", type: "about:blank"}`. I want to format the response to the format:
`{detail: {field1:field1-error, field2:field2-error...}, title:"Bad Request", type:""...}`. The rest of the fields will easily be styled from the **errors.detail.*** without much modifications to current codebase. Currently looking at where/how the response is composed so I can make the necessary changes.

dmkitui:
I have done some work on this and here is what I have come up with. The error object from the backend validations is as shown below on the right. I had to modify it with a helper method so as to include the invalid field:error value pair for easier template checking so as to style all the various fields and the applicable error label below them.
screen shot 2017-10-05 at 12 25 43

I also made the UI validation masseges to be as detailed as the back-end validation errors for the priority and backgroundRate fields.
screen shot 2017-10-05 at 14 24 08
I added a sweetAlert prompt even for the UI validation errors, to call the user to the appropriate field causing hte error.
screen shot 2017-10-05 at 14 33 15

I welcome your feedback.

ui/app/templates/rule_modal.html
mozbhearsum: It's OK not to worry about changing how the backend errors are displayed in this PR.

ui/app/js/controllers/rule_edit_controller.js
```diff
@@ -59,6 +73,7 @@ function ($scope, $modalInstance, CSRF, Rules, Releases, rule, pr_ch_options) {
       .error(function(response) {
         if (typeof response === 'object') {
           $scope.errors = response;
+          Helpers.addErrorFields($scope.errors);
```
mozbhearsum: I'm sorry if I sent you down the wrong path here...but I don't think it's a good to change the overall error handling. It's tightly wound with the response that the backend returns, and I'm concerned we're going to stop showing some errors by hardcoding specific fields. For example: the backend has many cases where "detail" or "exception" will be in the returned error, and it looks like you're removing support for showing them.

dmkitui: I think so too. That being said, I could leave the backend error handling logic and display and for now, go with an additional variable for UI validation errors and display and maybe raise a separate ticket for harmonization of the two?

mozbhearsum: That seems like a good plan.

dmkitui:
Reverted the backend error handling logic and display, and refactored the UI integer validation errors and back-end validation errors to be displayed. If UI errors are caught, the request is stopped and errors displayed. Where they are not caught,  the backend validators will catch them and display them in the error label near the bottom of the modal.

screen shot 2017-10-07 at 15 41 06
screen shot 2017-10-07 at 15 07 46
screen shot 2017-10-07 at 15 07 25
screen shot 2017-10-07 at 15 02 33

mozbhearsum:
The changes seem fine but the diff seems to include a lot of unrelated things. For example, all of #408 is there...can you please fix it to just include your changes? If it's easier, you're free to create a new branch instead.

dmkitui:
I will do make a new branch and do another PR to avoid issues brought about by the rebase.

PR #417, opened 2017-10-04
Bug 1335196: Role-centric view on Permissions page

andela-ookwuolisa:
This PR displays a list of `roles` and a list of `users` under each `role` on the Permissions UI. There is also an option to view a selected `role` from the filter provided.

An endpoint was created to fetch the users for a particular role.

### DIFF
--- a/auslib/db.py
+++ b/auslib/db.py
@@ -2296,20 +2296,15 @@ class Permissions(AUSTable):
 
     def getAllUsers(self, transaction=None):
         res_users = self.select(columns=[self.username], distinct=True, transaction=transaction)
-        users_list = list(set([r['username'] for r in res_users]))
-        users = []
+        users_list = list([r['username'] for r in res_users])
+        users = {}
         for user in users_list:
             res_roles = self.user_roles.select(where=[
                 self.user_roles.username == user],
-                columns=[self.user_roles.role],
+                columns=[self.user_roles.role, self.user_roles.data_version],
                 transaction=transaction)
-            roles_list = list([r['role'] for r in res_roles])
-            roles = {}
-            roles["roles"] = roles_list
-            user_roles = {}
-            user_roles[user] = roles
-            users.append(user_roles)
-        return sorted(users)
+            users[user] = {"roles": res_roles}
+        return users
 
     def getAllPermissions(self, transaction=None):
         ret = defaultdict(dict)
--- a/auslib/db.py
+++ b/auslib/db.py
@@ -2295,10 +2295,21 @@ class Permissions(AUSTable):
                 raise ValueError('Unknown option "%s" for permission "%s"' % (opt, permission))
 
     def getAllUsers(self, transaction=None):
-        res_permissions = self.select(columns=[self.username], distinct=True, transaction=transaction)
-        res_roles = self.user_roles.select(columns=[self.user_roles.username], transaction=transaction)
-        res = res_roles + res_permissions
-        return list(set([r['username'] for r in res]))
+        res_users = self.select(columns=[self.username], distinct=True, transaction=transaction)
+        users_list = list(set([r['username'] for r in res_users]))
+        users = []
+        for user in users_list:
+            res_roles = self.user_roles.select(where=[
+                self.user_roles.username == user],
+                columns=[self.user_roles.role],
+                transaction=transaction)
+            roles_list = list([r['role'] for r in res_roles])
+            roles = {}
+            roles["roles"] = roles_list
+            user_roles = {}
+            user_roles[user] = roles
+            users.append(user_roles)
+        return sorted(users)
 
     def getAllPermissions(self, transaction=None):
         ret = defaultdict(dict)
@@ -2422,10 +2433,6 @@ class Permissions(AUSTable):
                                      columns=[self.user_roles.role, self.user_roles.data_version],
                                      distinct=True, transaction=transaction)
         return [{"role": r["role"], "data_version": r["data_version"]} for r in res]
-
-    def getAllRoles(self, transaction=None):
-        res = self.user_roles.select(columns=[self.user_roles.role], distinct=True, transaction=transaction)
-        return [r["role"] for r in res]
 
     def isAdmin(self, username, transaction=None):
         return bool(self.getPermission(username, "admin", transaction))
--- a/auslib/db.py
+++ b/auslib/db.py
@@ -397,14 +397,15 @@ class AUSTable(object):
         if self.versioned:
             data['data_version'] = 1
         query = self._insertStatement(**data)
-
-        if self.onInsert:
-            self.onInsert(self, "INSERT", changed_by, query, trans)
-
         ret = trans.execute(query)
         if self.history:
             for q in self.history.forInsert(ret.inserted_primary_key, data, changed_by):
                 trans.execute(q)
+        if self.onInsert:
+            pk_columns = self.t.primary_key.columns.keys()
+            pk_values = ret.inserted_primary_key
+            pk_args = dict(zip(pk_columns, pk_values))
+            self.onInsert(self, "INSERT", changed_by, query, trans, **pk_args)
         return ret
 
     def insert(self, changed_by=None, transaction=None, dryrun=False, **columns):
@@ -2533,7 +2534,7 @@ def send_email(relayhost, port, username, password, to_addr, from_addr, table, s
 
 
 def make_change_notifier(relayhost, port, username, password, to_addr, from_addr, use_tls):
-    def bleet(table, type_, changed_by, query, transaction):
+    def bleet(table, type_, changed_by, query, transaction, **pk_args):
         body = ["Changed by: %s" % changed_by]
         if type_ == "UPDATE":
             body.append("Row(s) to be updated as follows:")
@@ -2558,6 +2559,7 @@ def make_change_notifier(relayhost, port, username, password, to_addr, from_addr
                 body.append(UTF8PrettyPrinter().pformat(row))
         elif type_ == "INSERT":
             body.append("Row to be inserted:")
+            query.parameters.update(pk_args)
             body.append(UTF8PrettyPrinter().pformat(query.parameters))
 
         subj = "%s to %s detected %s" % (type_, table.t.name, generate_random_string(6))
--- a/auslib/web/admin/swagger/api.yaml
+++ b/auslib/web/admin/swagger/api.yaml
@@ -1189,7 +1189,7 @@ paths:
         description: "Returns all of Users in Balrog's DB."
       responses:
         '200':
-          description: Get all users inclusive of their roles
+          description: Get all Users and their Roles
           schema:
             type: object
             properties:
@@ -1201,16 +1201,18 @@ paths:
                 items:
                   type: object
                 example: {
-                  "balrogadmin": [
-                    {
-                      "data_version": 1,
-                      "role": "releng"
-                    },
-                    {
-                      "data_version": 1,
-                      "role": "relman"
-                    }]
-                         }
+                  "balrogadmin": {
+                    "roles":[
+                      {
+                        "data_version": 1,
+                        "role": "releng"
+                      },
+                      {
+                        "data_version": 1,
+                        "role": "relman"
+                      }]
+                  }
+                }
 
 
   /users/{username}:
--- a/auslib/web/admin/swagger/api.yaml
+++ b/auslib/web/admin/swagger/api.yaml
@@ -1189,7 +1189,7 @@ paths:
         description: "Returns all of Users in Balrog's DB."
       responses:
         '200':
-          description: Get all Users
+          description: Get all users inclusive of their roles
           schema:
             type: object
             properties:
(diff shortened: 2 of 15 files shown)


auslib/db.py
```diff
@@ -2265,7 +2265,7 @@ def getAllUsers(self, transaction=None):
     def getAllPermissions(self, transaction=None):
         ret = defaultdict(dict)
         for r in self.select(transaction=transaction):
-            ret[r["username"]][r["permission"]] = r["options"]
+            ret[r["username"]][r["permission"]][r["role"]] = r["options"]
```
mozbhearsum: This doesn't look right. Roles are not dependent on a Permission, just on the User.

auslib/db.py
```diff
@@ -2423,6 +2429,10 @@ def hasRole(self, username, role, transaction=None):
         roles_list = [r['role'] for r in self.getUserRoles(username, transaction)]
         return role in roles_list
 
+    def hasUser(self, username, role, transaction=None):
```
mozbhearsum: "roleHasUser" would be a better name for this, since the class is "Permissions".

auslib/db.py
```diff
@@ -2649,6 +2659,9 @@ def hasRole(self, *args, **kwargs):
     def getUserRoles(self, *args, **kwargs):
         return self.permissions.getUserRoles(*args, **kwargs)
 
+    def getRoleUsers(self, *args, **kwargs):
```
mozbhearsum: No need to add this - the only reason getUserRoles exists is because we need a non-table specific way to find which Roles a specific User has. We don't need that for getting the lists of all Roles with membership.

auslib/web/admin/swagger/api.yaml
```diff
@@ -1231,6 +1231,36 @@ paths:
                   type: string
                 example: ["releng", "qa"]
 
+  /roles:
```
mozbhearsum: Have you considered returning the list of users with each Role as part of this, too? It would make things simpler for the frontend, because you'd only need to make one HTTP request.

auslib/web/admin/swagger/api.yaml
```diff
@@ -1231,6 +1231,36 @@ paths:
...
+            type: object
+            properties:
+              users:
+                description: list of roles
+                type: array
+                uniqueItems: true
+                minItems: 0
+                items:
+                  type: string
+                example: ["releng", "qa"]
```
mozbhearsum: Unfortunately, it's not safe to return a JSON list (https://haacked.com/archive/2008/11/20/anatomy-of-a-subtle-json-vulnerability.aspx/)

ui/app/js/controllers/permissions_controller.js
```diff
@@ -1,161 +1,211 @@
 angular.module("app").controller('PermissionsController',
-function($scope, $routeParams, $location, $timeout, Permissions, Search, $modal, Page, PermissionsRequiredSignoffs) {
```
mozbhearsum: Please revert all of the bulk whitespace changes that you've made...it makes reviewing this very difficult.

andela-ookwuolisa:
Hi @mozbhearsum, users who hold each Role show up on click. Did you try clicking on any Role?

mozbhearsum:
I did try clicking - an HTTP request happened, but nothing was shown. It would be better if everything was shown right away though (that's what I thought the mockup on https://bug1335196.bmoattachments.org/attachment.cgi?id=8907749 was demonstrating...)

auslib/db.py
Mercy-Muchai: Cool, will make this change.

auslib/web/admin/swagger/api.yaml
Mercy-Muchai: Serves a better purpose than the two endpoints. Making changes.

andela-ookwuolisa:
@mozbhearsum, I changed to this current design after a suggestion was made to make the roles clickable to display the users. I thought the suggestion was cool. Now that you've pointed out that the original design is better, I will revert to it. Thanks

andela-ookwuolisa:
@mozbhearsum All changes effected, Please review.

auslib/web/admin/swagger/api.yaml
```diff
@@ -1189,7 +1189,7 @@ paths:
         description: "Returns all of Users in Balrog's DB."
       responses:
         '200':
-          description: Get all Users
+          description: Get all users inclusive of their roles
```
mozbhearsum: "Get all Users and their Roles" would be a bit clearer here.

auslib/web/admin/swagger/api.yaml
```diff
@@ -1199,37 +1199,37 @@ paths:
                 uniqueItems: true
                 minItems: 0
                 items:
-                  type: string
-                example: ["balrogadmin"]
+                  type: object
```
mozbhearsum: Please include a full example of the new Response here.

auslib/web/admin/swagger/api.yaml
```diff
@@ -1199,37 +1199,37 @@ paths:
...
+
 
-  /users/roles:
+  /roles:
```
mozbhearsum: I'm not sure I understand why you've modified this endpoint. It appears to return exactly the same data as /users, except in a different format. Is there a reason you're not able to build this UI with your new version of /users (that include the Roles a user holds), and the existing /users/roles endpoint?

auslib/db.py
```diff
@@ -2403,7 +2411,23 @@ def getUserRoles(self, username, transaction=None):
 
     def getAllRoles(self, transaction=None):
         res = self.user_roles.select(columns=[self.user_roles.role], distinct=True, transaction=transaction)
-        return [r["role"] for r in res]
+        roles_list = list(set([r['role'] for r in res]))
```
mozbhearsum: This change to getAllRoles seems unnecessary. getAllUsers already returns all users and the Roles they hold. This view returns exactly the same data, except it makes role the top level key. Are you sure you need this?

auslib/db.py
mozbhearsum: This still needs to be removed - it is completely unused.

auslib/db.py
```diff
@@ -2439,6 +2463,10 @@ def hasRole(self, username, role, transaction=None):
         roles_list = [r['role'] for r in self.getUserRoles(username, transaction)]
         return role in roles_list
 
+    def roleHasUser(self, username, role, transaction=None):
```
mozbhearsum: This function is not used at all - please remove it.

ui/app/js/controllers/permissions_controller.js
```diff
@@ -6,35 +6,73 @@ function($scope, $routeParams, $location, $timeout, Permissions, Search, $modal,
...
+      .then(function (response) {
+        $scope.permissions = response;
+      });
   } else {
     Permissions.getUsers()
-    .success(function(response) {
-      $scope.users = _.map(response.users, function (each) {
-        return {username: each};
+      .success(function (response) {
+        $scope.users = _.map(response.users, function (each) {
```
mozbhearsum: Further to the above point, you should be able to use the data in the response here to build up the entire list of roles and their associated users.

ui/app/js/controllers/permissions_controller.js
```diff
@@ -6,35 +6,73 @@ function($scope, $routeParams, $location, $timeout, Permissions, Search, $modal,
...
+  };
+
+  $scope.isSet = function (tabNum) {
+      return $scope.tab === tabNum;
+  };
+
+  $scope.setTab = function (newTab) {
+    $scope.tab = newTab;
+  };
+  $scope.roles_users = {};
```
mozbhearsum: This is another place where you're duplicating data, and I think it contributes to the problem I mentioned above about newly added Roles not showing up. Instead of creating $scope.users and $scope.roles_users, you can use one data structure that holds them both. You'll need to update the templates a little bit to cope with this.

ui/app/templates/permissions.html
```diff
@@ -1,14 +1,24 @@
-<div loader ng-show="loading"></div>
+<div class="tabbable tabs-below">
+  <ul class="nav nav-pills">
+    <li ng-class="{ active: isSet(1) }">
+      <a href ng-click="setTab(1)">Users</a>
```
mozbhearsum: isSet and setTab actually make this a bit tougher to read, because it requires looking at the template and the controller. Because they are such trivial functions (each contains only a single statement)< this would be a bit more readable if you got rid of them and inlined it instead. Eg:
```
    <li ng-class="{ active: (tab === 1) }">
      <a href ng-click="tab = 1">Users</a>
    </li>
```

ui/app/js/controllers/permissions_controller.js
```diff
@@ -66,18 +104,26 @@ function($scope, $routeParams, $location, $timeout, Permissions, Search, $modal,
...
-        if ((on === '*' || on === 'username') && item.username && item.username.match(regex)) {
-          matches++;
-          return;
+        if ('username' in item) {
+          if ((on === '*' || on === 'username') && item.username && item.username.match(regex)) {
+            matches++;
+            return;
+          }
+        }
+        else {
```
mozbhearsum: It doesn't look like filterBySearch is enabled for Roles - are you sure you need this section?

ui/app/pages/index.us
```diff
@@ -59,5 +59,6 @@
 
 
     <script type="text/javascript" src="/js/app.js"></script>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/popper.js/1.11.0/umd/popper.min.js" integrity="sha384-b/U6ypiBEHpOf/4+1nzFpr53nxSS+GLCkfwBdFNTxtclqqenISfwAzpKaMNFNmj4" crossorigin="anonymous"></script>
```
mozbhearsum: Are you using anything from this library? If so, it needs to go in the vendor directory. We don't want to load 3rd party content.

ui/app/templates/permissions.html
```diff
@@ -46,29 +55,91 @@
...
+      </select>
+    </div>
+  </div>
 </div>
+
+  <div class="pagination-container" ng-show="roles_list.length > pageSize">
+    <pagination class="pagination-sm" total-items="roles_list.length" ng-model="currentPage" items-per-page="pageSize"></pagination>
+  </div>
+
+  <div ng-if="isAllRoles(current_role)" class="panel-group" ng-repeat="(key, role) in roles_users">
```
mozbhearsum: This is another place where inlining would help readability - there's no point in calling out to the controller for this.

ui/app/pages/index.us
andela-ookwuolisa: I made use of it when I used the previous design. But not anymore, so I will do away with it.

ui/app/templates/permissions.html
andela-ookwuolisa: This is way better. I will effect change.

auslib/web/admin/swagger/api.yaml
Mercy-Muchai: I modified it for consistency sake, so there would only be **/users and /roles**
Given your recommendation, we shall have one endpoint users that includes the roles and that is used to build the UI.

ui/app/js/controllers/permissions_controller.js
andela-ookwuolisa: I have used the data from the `api/users` response to build the entire roles and their associated users. But I also made use of $scope.roles_users and $scope.roles_list to hold the respective data after the build. I couldn't figure out a way to do it with less. At least not yet.  I will appreciate more clarification on this. Thanks

mozbhearsum: What I'm suggesting is that `$scope.users` should be a single data structure that contains both user and role information. Something structured like:
```
{
   "bob": {
      "roles": ["releng", "relman"]
   },
   "obinna": {
      "roles": ["relman", "qe"]
   }
}
```

This single data structure contains all of the information that your current `$scope.users`, `$scope.roles_list`, and `$scope.roles_users` currently does, which means when changes are made there's only one data structure to update.

Using this unified data structure means that code that currently uses `$scope.roles_users` (like https://github.com/mozilla/balrog/pull/417/files#diff-5c2e73a618f68d1bf0ac1eabb3d8dff8R107) will need to be adjusted to flip the data structure around. For example, you could have a function called `getRoleUsers` that looks iterates over `$scope.users` and returns a data structure where roles are the top level key.

Does that clarify things at all?

andela-ookwuolisa: Oh. Yeah.  Now I get. I will effect changes.

mozbhearsum:
Are you planning to finish this up @andela-ookwuolisa?

andela-ookwuolisa:
@mozbhearsum I have finished it. I am waiting for feedback if any. I thought you were on leave.

auslib/db.py
```diff
@@ -2295,10 +2295,21 @@ def assertOptionsExist(self, permission, options):
...
+                self.user_roles.username == user],
+                columns=[self.user_roles.role],
+                transaction=transaction)
+            roles_list = list([r['role'] for r in res_roles])
+            roles = {}
+            roles["roles"] = roles_list
+            user_roles = {}
+            user_roles[user] = roles
+            users.append(user_roles)
+        return sorted(users)
```
mozbhearsum: I can see why you made `users` a list (because the current return value is a list), but it's better off being a dictionary, because we don't support multiple users with the same name. The format returned here should be:
```
{
  "user1": {
    "roles": ["foo", "bar"]
  },
  "user2": {
    "roles": ["bar"]
  }
}
```

This change should make all of the code that consumes this data cleaner, too.

ui/app/js/controllers/permissions_controller.js
```diff
@@ -6,35 +6,68 @@ function($scope, $routeParams, $location, $timeout, Permissions, Search, $modal,
...
+    var roles = [];
+    $scope.users.forEach(function (eachUser){
+      eachUser.roles.forEach(function(role){
+        if(roles.indexOf(role)=== -1){
+            roles.push(role);
+        }
+      });
+    });
+    return roles;
+  };
```
mozbhearsum: Much cleaner, nice work!

ui/app/js/controllers/user_edit_controller.js
```diff
@@ -29,29 +34,16 @@ function ($scope, $modalInstance, CSRF, Permissions, users, is_edit, user, permi
...
-        );
-      }
-    })
-    .finally(function() {
+    $scope.getUserRoles = function (username){
+      $scope.users.forEach(function(eachUser){
+        if(eachUser.username === username){
+          $scope.user.roles = eachUser.roles;
+        }
+      });
```
mozbhearsum: There's no reason to define getUserRoles() only to call it immediately afterwards -- just drop the function and run for the forEach directly.

andela-ookwuolisa: Oh yeah. I will take it out. Thanks

andela-ookwuolisa:
Hi, @mozbhearsum, the feedback has been implemented. Kindly review. Thanks.

PR #418, opened 2017-10-04
bug 1246993: admin api doesn't always rollback transactions on failures

mozbhearsum:
There's two distinct cases where we may end up with unwanted rows in the database:
1) If a View _returns_ a 400 or 500 instead of raising an Exception, the transaction will still be committed. If any of the View code made a change to the database (for example, how changeRelease will create a release before actually adding the desired data to it), this will be committed even though it shouldn't be. The fix for this is to check return code and explicitly rollback the transaction instead of relying on the context manager to do it.
2) If we start making changes to the database, then something happens that invalidates them, but we recover from it and continue on - the original changes we made will still be present. This is what's happening when we handled OutdatedDataError in Releases.update right now -- we've already called AUSTable.update, which has added a row the history table, and then when we call it _again_ after handling the merge error, we end up with a duplicate history row. The fix for this is to not update the history table before potentially raising OutdatedDataError. This is not a perfect fix (ideally, we'd rollback the transaction and create a new one), but it fixes the currently known issues, and is not terribly invasive.

With these two fixes applied we've had 3 staging releases with absolutely no failures to merge updates:
```
$ for i in build10 build11 build12; do echo -n "Attempted Merges for $i: " && grep 'Trying to merge' output-oct3.txt | grep $i | wc -l && echo -n "Successful Merges for $i: " && grep 'Successfully merged' output-oct3.txt | grep $i | wc -l; done
Attempted Merges for build10: 24
Successful Merges for build10: 24
Attempted Merges for build11: 134
Successful Merges for build11: 134
Attempted Merges for build12: 112
Successful Merges for build12: 112
```

### DIFF
--- a/auslib/test/admin/views/test_releases.py
+++ b/auslib/test/admin/views/test_releases.py
@@ -319,7 +319,7 @@ class TestReleasesAPI_JSON(ViewTest):
                 }
             }
         }"""
-        ret = self._post('/releases/ee', data=dict(data=blob, hashFunction="sha512", name='ee', product='ee', data_version=1))
+        ret = self._post('/releases/ee', data=dict(data=blob, hashFunction="sha512", name='ee', product='ee'))
         self.assertStatusCode(ret, 201)
 
         # Updating same release
@@ -340,12 +340,16 @@ class TestReleasesAPI_JSON(ViewTest):
         history_rows = dbo.releases.history.t.select().where(dbo.releases.history.name == "ee").execute().fetchall()
         self.assertEqual(len(history_rows), 4)
         self.assertEqual(history_rows[0]["data"], None)
+        self.assertEqual(history_rows[0]["data_version"], None)
         self.assertEqual(history_rows[0]["read_only"], False)
         self.assertEqual(history_rows[1]["data"], {"name": "ee", "schema_version": 1, "hashFunction": "sha512"})
+        self.assertEqual(history_rows[1]["data_version"], 1)
         self.assertEqual(history_rows[1]["read_only"], False)
         self.assertEqual(history_rows[2]["data"], blob)
+        self.assertEqual(history_rows[2]["data_version"], 2)
         self.assertEqual(history_rows[2]["read_only"], False)
         self.assertEqual(history_rows[3]["data"], blob)
+        self.assertEqual(history_rows[3]["data_version"], 3)
         self.assertEqual(history_rows[3]["read_only"], False)
 
     def testReleasePostMismatchedName(self):
--- a/auslib/test/admin/views/test_releases.py
+++ b/auslib/test/admin/views/test_releases.py
@@ -476,6 +476,35 @@ class TestReleasesAPI_JSON(ViewTest):
 }
 """))
 
+    def testLocalePutOutdatedDataError(self):
+        data = json.dumps({
+            "complete": {
+                "filesize": 435,
+                "from": "*",
+                "hashValue": "abc",
+            }
+        })
+        ret = self._put('/releases/ab/builds/p/l', data=dict(data=data, product='a', data_version=1, schema_version=1))
+        self.assertStatusCode(ret, 201)
+
+        history_rows = dbo.releases.history.t.select().where(dbo.releases.history.name == "ab").execute().fetchall()
+        self.assertEquals(len(history_rows), 3)
+
+        data = json.dumps({
+            "complete": {
+                "filesize": 435,
+                "from": "*",
+                "hashValue": "def",
+            }
+        })
+        ret = self._put('/releases/ab/builds/p/l', data=dict(data=data, product='a', data_version=1, schema_version=1))
+        self.assertStatusCode(ret, 400)
+
+        # Ensure that history wasn't created for second request.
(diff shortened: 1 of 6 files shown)


auslib/web/admin/views/base.py
```diff
@@ -69,17 +69,53 @@ def __init__(self, *args, **kwargs):
...
+            raise
+        finally:
+            trans.close()
 
     @handleGeneralExceptions("DELETE")
     def delete(self, *args, **kwargs):
         self.log.debug("processing DELETE request to %s" % request.path)
-        with dbo.begin() as trans:
-            return self._delete(*args, transaction=trans, **kwargs)
+        trans = dbo.begin()
```
allan-silva: Maybe the transaction behavior in POST, PUT and DELETE, can be in a decorator, something like this:

https://pastebin.mozilla.org/9069305

auslib/test/admin/views/test_releases.py
```diff
@@ -557,7 +646,14 @@ def testLocalePutForNewRelease(self):
...
 }
-"""))
+""")
+        self.assertEqual(ret, expected)
+
+        history_rows = dbo.releases.history.t.select().where(dbo.releases.history.name == "e").execute().fetchall()
+        self.assertEqual(len(history_rows), 3)
+        self.assertEqual(history_rows[0]["data"], None)
+        self.assertEqual(history_rows[1]["data"], {"name": "e", "schema_version": 1, "hashFunction": "sha512"})
+        self.assertEqual(history_rows[2]["data"], expected)
```
nthomas-mozilla: This also seems odd to me - a new product post via the /releases/<release>/builds/[platform]/[locale] endpoint ends up with 3 history rows rather than 2. Would make sense if this was product ab, but it's not here. Am I missing something related to how we test or put/post differences ?

auslib/test/admin/views/test_releases.py
```diff
@@ -637,7 +760,14 @@ def testLocalePutForNewReleaseWithAlias(self):
...
```
nthomas-mozilla: Same as last comment.

auslib/test/admin/views/test_releases.py
```diff
@@ -323,6 +336,18 @@ def testReleasePostUpdateOutdatedDataNotBlob(self):
                                                    name='ee', product='ee', data_version=1))
         self.assertStatusCode(ret, 400)
 
+        blob = json.loads(blob)
+        history_rows = dbo.releases.history.t.select().where(dbo.releases.history.name == "ee").execute().fetchall()
+        self.assertEqual(len(history_rows), 4)
+        self.assertEqual(history_rows[0]["data"], None)
+        self.assertEqual(history_rows[0]["read_only"], False)
+        self.assertEqual(history_rows[1]["data"], {"name": "ee", "schema_version": 1, "hashFunction": "sha512"})
+        self.assertEqual(history_rows[1]["read_only"], False)
```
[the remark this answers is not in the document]
mozbhearsum: Items 0 and 1 are created as part of the initial _post. Whenever a new object is added we end up with a row in the history table that is all null except for the PK and history metadata (https://github.com/mozilla/balrog/blob/bc70d0f8c9675f99407b276bb4b449f7fa84f471/auslib/db.py#L693).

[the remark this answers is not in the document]
mozbhearsum: The fact that the initial _post passes data_version is probably part of the confusion here. It shouldn't be doing that since it's postihng a new release...

auslib/test/admin/views/test_releases.py
```diff
@@ -323,6 +336,18 @@ def testReleasePostUpdateOutdatedDataNotBlob(self):
...
+        self.assertEqual(history_rows[2]["data"], blob)
+        self.assertEqual(history_rows[2]["read_only"], False)
+        self.assertEqual(history_rows[3]["data"], blob)
+        self.assertEqual(history_rows[3]["read_only"], False)
```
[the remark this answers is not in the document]
mozbhearsum: It actually creates history_rows[3] (beacuse of the above), with data_version 2.

auslib/test/admin/views/test_releases.py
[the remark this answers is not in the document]
mozbhearsum: And I'd completely forgotten (until further debugging) that creating a release with this type of _post ends up creating it, then updating it -- so the first 3 items of the history_rows are all created by the first _post, and then the final one by the second.

auslib/test/admin/views/test_releases.py
mozbhearsum: I _think_ this is for the same reason as above. Both end up in the changeRelease "helper", and end up at https://github.com/mozilla/balrog/blob/master/auslib/web/admin/views/releases.py#L142 -- which creates the release without the data, and subsequently updates it. So we end up with the empty history row, then data_version 1 & 2 as part of a single http request.

Let me know if there's still any confusion - I know this code/tests are clear as mud.

nthomas-mozilla: Ah hah - that bit of code explains all my questions, thanks.

PR #419, opened 2017-10-05
Improve History.getChange efficiency.

mozbhearsum:
The current version of the query we use to grab the ancestor version of a Release when merging ends up running queries like:
```
SELECT releases_history.change_id, releases_history.changed_by, releases_history.timestamp, releases_history.name, releases_history.product, releases_history.`read_only`, releases_history.data, releases_history.data_version FROM releases_history WHERE releases_history.data_version = %s AND releases_history.name = %s
```

This ends up using an index on releases_history.name. In the nightly case, this means there's still ~15,000 rows to scan through to find the one we want. In my local testing, reducing the number of columns that we select on this big query speeds it up by ~450 times. On a releases_history table with ~13,000 rows, selecting all columns took 18.09 seconds, and selecting only change_id took 0.04. Once we have the change_id, selecting the entire row we want goes very fast!

I'm not actually 100% sure I understand why this helps - I'm hoping to confirm my theory with someone who knows mysql better before landing this.

### DIFF
--- a/auslib/test/test_db.py
+++ b/auslib/test/test_db.py
@@ -460,6 +460,7 @@ class TestHistoryTable(unittest.TestCase, TestTableMixin, MemoryDatabaseMixin):
         self.assertTrue('foo' in columns)
         self.assertTrue('changed_by' in columns)
         self.assertTrue('timestamp' in columns)
+        self.assertTrue('data_version' in columns)
 
     @mock.patch("time.time", mock.MagicMock(return_value=1.0))
     def testHistoryUponInsert(self):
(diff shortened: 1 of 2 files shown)


mozbhearsum:
I did some more investigation into why the query that selects all columns is so slow. I _think_ this explains it (from https://www.safaribooksonline.com/library/view/high-performance-mysql/9780596101718/ch04.html):
> Retrieve rows from the table, then filter nonmatching rows (“Using where” in the Extra column).
> This happens at the server layer and requires the server to read rows from the table before it can filter them.

The `EXPLAIN` plan for the above query indicates that we are "using where", which means we're retrieving the rows _before_ checking the WHERE values. I assume that "retrieving rows" still only means retrieving the columns necessary for the query, which would explain the vastly difference performance when "data" is removed from the list of columns to select.

auslib/test/test_db.py
```diff
@@ -460,6 +460,7 @@ def testHistoryTableHasAllColumns(self):
         self.assertTrue('foo' in columns)
         self.assertTrue('changed_by' in columns)
         self.assertTrue('timestamp' in columns)
+        self.assertTrue('data_version' in columns)
```
mozbhearsum: I noticed that this columns were missing from these tests while trying to write a useful test for this. I couldn't think of any useful tests to add, but we may as well keep this...

mozbhearsum:
@catlee r+'ed in https://bugzilla.mozilla.org/show_bug.cgi?id=1400141#c28.

PR #420, opened 2017-10-09
bug 1396959 Enable removal of scheduled deletes to required signoff

Alweezy:
Currently, there is no way to remove a scheduled delete from a role in required signoffs. This PR leverages on pre-existing API endpoints to enable this.

### DIFF
--- a/ui/app/js/controllers/required_signoffs_controller.js
+++ b/ui/app/js/controllers/required_signoffs_controller.js
@@ -43,8 +43,7 @@ function($scope, $modal, $q, CSRF, ProductRequiredSignoffs, PermissionsRequiredS
     }
   }, true);
 
-  $scope.undoRemoveRole = function(role, rs) {
-    var roleName = role.$key;
+  $scope.undoRemoveRole = function(roleName, rs) {
     var product = Object.keys(rs)[0];
     var channel = Object.keys(rs[product].channels)[0];
 
--- a/ui/app/js/controllers/required_signoffs_controller.js
+++ b/ui/app/js/controllers/required_signoffs_controller.js
@@ -47,8 +47,6 @@ function($scope, $modal, $q, CSRF, ProductRequiredSignoffs, PermissionsRequiredS
     var product = Object.keys(rs)[0];
     var channel = Object.keys(rs[product].channels)[0];
 
-    $scope.required_signoffs[product].channels[channel][roleName].sc.signoffs_required = 1;
-
     CSRF.getToken()
     .then(function(csrf_token) {
       var sc_id = $scope.required_signoffs[product].channels[channel][roleName].sc.sc_id;
--- a/ui/app/js/controllers/required_signoffs_controller.js
+++ b/ui/app/js/controllers/required_signoffs_controller.js
@@ -43,28 +43,23 @@ function($scope, $modal, $q, CSRF, ProductRequiredSignoffs, PermissionsRequiredS
     }
   }, true);
 
-  $scope.undoRemoveRole = function(role, rs, csrf_token) {
+  $scope.undoRemoveRole = function(role, rs) {
     var roleName = role.$key;
     var product = Object.keys(rs)[0];
-    var chans = rs[product].channels;
-    var channel = Object.keys(chans)[0];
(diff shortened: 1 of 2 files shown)


ui/app/templates/required_signoffs.html
```diff
@@ -97,6 +97,11 @@ <h4 class="panel-title">
           <span ng-show="role['sc'] !== null">{{ role["sc"]["signoffs_required"] }} member(s) of {{ role.$key }} (currently: {{ role["signoffs_required"] }})</span>
           <span class="label label-info" ng-show="role['sc'] !== null && role['sc']['change_type'] !== 'delete'">Pending</span>
           <span class="label label-danger" ng-show="role['sc'] !== null && role['sc']['change_type'] === 'delete'">Pending Delete</span>
+          <button class="btn btn-xs btn-primary" ng-show="role['sc'] !== null && role['sc']['change_type'] === 'delete'"
+                  ng-click="undoRemoveRole(role, required_signoffs)">
+            Undo delete
```
mozbhearsum: "Cancel delete" would be better text for this button.

ui/app/js/controllers/required_signoffs_controller.js
```diff
@@ -43,6 +43,33 @@ function($scope, $modal, $q, ProductRequiredSignoffs, PermissionsRequiredSignoff
...
+    $scope.required_signoffs[product].channels[channel][roleName].sc.signoffs_required = 1;
+    $scope.required_signoffs[product].channels[channel][roleName].sc.change_type = "update";
+    var update_sc_id = $scope.required_signoffs[product].channels[channel][roleName].sc.sc_id;
+    var data = {
+      "product": $scope.product, 
+      "role": roleName,
+      "data_version": $scope.required_signoffs[product].channels[channel][roleName].data_version};
+    data["when"] = new Date().getTime() + 5000;
+    data["sc_data_version"] = $scope.required_signoffs[product].channels[channel][roleName].sc.sc_data_version;
+    data["signoffs_required"] = $scope.required_signoffs[product].channels[channel][roleName].sc.signoffs_required;
```
mozbhearsum: I think you've done a bit more work here than is necessary. The only things that you need to include to delete a scheduled change are sc_id, data_version (which comes from sc_data_version), and csrf_token - so you should be able to remove a bunch of these lines of code.

ui/app/js/controllers/required_signoffs_controller.js
```diff
@@ -43,6 +43,33 @@ function($scope, $modal, $q, ProductRequiredSignoffs, PermissionsRequiredSignoff
     }
   }, true);
 
+  $scope.undoRemoveRole = function(role, rs, csrf_token) {
```
mozbhearsum: Does csrf_token actually do anything here?

ui/app/js/controllers/required_signoffs_controller.js
```diff
@@ -43,6 +43,33 @@ function($scope, $modal, $q, ProductRequiredSignoffs, PermissionsRequiredSignoff
...
+    var roleName = role.$key;
```
mozbhearsum: Although $key is accessible here, it's not safe to refer to it. You should pass it in as another argument instead, if you truly need it.

ui/app/js/controllers/required_signoffs_controller.js
```diff
@@ -43,6 +43,27 @@ function($scope, $modal, $q, ProductRequiredSignoffs, PermissionsRequiredSignoff
     }
   }, true);
 
+  $scope.undoRemoveRole = function(roleName, rs) {
+    var product = Object.keys(rs)[0];
+    var channel = Object.keys(rs[product].channels)[0];
+
+    $scope.required_signoffs[product].channels[channel][roleName].sc.signoffs_required = 1;
```
mozbhearsum: Does this serve a purpose? You're setting sc to null if the http request succeeds a few lines down.

Alweezy: @mozbhearsum The line was serving no purpose for sure, I have since removed it. Thanks.

PR #422, opened 2017-10-11
Bug 1396106: Edits to "scheduled changes" are not auto signed-off

andela-iikikin:
**What does this PR do?**
1) Previously when a user creates a schedule change the system automatically adds the required signoff for the user but the same user editing the schedule change also automatically revokes the signoff. This PR makes sure the signoff is not revoked if only it's the same user who created the schedule change that is editing the schedule change.
2) If a different user happens to be the one editing the schedule change then the system revokes all previous signoffs and adds a signoff for the editing user

**How should this be manually tested?**
_CASE 1_
Navigate to the rules page and try to create a schedule change and before the change is completed edit the schedule change. It should not revoke the signoff that was added while creating the schedule change
_CASE 2_
Navigate to the rules page and try to create a schedule change and before the change is completed
set a different remote user and try editing the schedule change. It should remove the previous users signoff but automatically add the current users signoff

### DIFF
--- a/auslib/db.py
+++ b/auslib/db.py
@@ -1114,13 +1114,14 @@ class ScheduledChangeTable(AUSTable):
         #   we don't know which Role we'd want to signoff with. The user will need to signoff
         #   manually in these cases.
         user_roles = self.db.getUserRoles(username=changed_by, transaction=transaction)
-        required_roles = set()
-        required_signoffs = self.baseTable.getPotentialRequiredSignoffs([columns], transaction=transaction)
-        if required_signoffs:
-            required_roles.update([rs["role"] for rs in required_signoffs])
-        if len(user_roles) == 1 and user_roles[0]["role"] in required_roles:
-            self.signoffs.insert(changed_by=changed_by, transaction=transaction, dryrun=dryrun,
-                                 sc_id=sc_id, role=user_roles[0].get("role"))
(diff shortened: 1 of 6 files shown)


auslib/db.py
```diff
@@ -1102,6 +1102,20 @@ def validate(self, base_columns, condition_columns, changed_by, sc_id=None, tran
...
+        # - If the User scheduling a change only holds one Role, record a signoff with it.
+        # - If the User scheduling a change holds more than one Role, we cannot a Signoff, because
+        #   we don't know which Role we'd want to signoff with. The user will need to signoff
+        #   manually in these cases.
+        user_roles = self.db.getUserRoles(username=changed_by, transaction=transaction)
+        required_roles = set()
+        required_signoffs = self.baseTable.getPotentialRequiredSignoffs([columns], transaction=transaction)
+        if required_signoffs:
+            required_roles.update([rs["role"] for rs in required_signoffs])
+        if len(user_roles) == 1 and user_roles[0]["role"] in required_roles:
```
allan-silva: You can consider prevent one hit in database if you check `user_roles` is equals 1 before call `getPotentialRequiredSignoffs`

andela-iikikin:
@mozbhearsum Added the test and made the requested changes. It's now ready for review

PR #426, opened 2017-10-13
Fixed bug 1407392

Avedis777:


auslib/blobs/apprelease.py
```diff
@@ -888,7 +887,7 @@ class ReleaseBlobV7(ReleaseBlobBase, NewStyleVersionsMixin, MultipleUpdatesXMLMi
     # for the benefit of get*XML
     optional_ = ('showPrompt', 'showNeverForVersion',
                  'actions', 'openURL', 'notificationURL', 'alertURL',
-                 'promptWaitTime', 'backgroundInterval')
+                 'promptWaitTime')
```
mozbhearsum: Instead of doing this, please remove ReleaseBlobV7 altogether. As you can see from the comment above, backgroundInterval was the only thing added here! There's no point in keeping the class if we've removed it. Make sure to remove the schema from auslib/blobs/schemas too, and there may be one or more tests to remove.

PR #429, opened 2017-10-17
Bug 1339532[make backend tests faster]

RayNjeri:
#### What does this PR do?
Improves the run time of backend tests by implementation of pytest fixtures.
#### Observations:
1.When tests that depend on the same fixture are run, that operation is executed multiple times off which it can add significantly to the overall testing runtime.
2.When the internet connection is slow the tests take a longer period running as compared to strong internet connection. And the difference of this could be up to 200 secs.
3.The runtime also increases with increase in tests hence pytest fixtures might not be a permanent solution to the problem as backend tests increase with increase in bug fixes.


RayNjeri:
When started out the task, The tests run to upto 5mins when added the fixtures and with good internent connction the tests run to nearly 3mins but will look into it

RayNjeri:
@mozbhearsum i have made a bit of changes to this PR kindly check on them

mozbhearsum:
This still hasn't made any improvements to the test execution time. The average time over 5 executions of the test_db.py tests is nearly identical with and without this patch, even when running the tests directly with "tox" (which doesn't depend on any internet connection at all - so that's not a factor).

I appreciate your attempts here @RayNjeri, but I don't have time to mentor this any further right now...

PR #430, opened 2017-10-17
Histories endpoint

andela-onnenanya:
### What does this PR do:
Create a single endpoint for getting all histories with filters. [bug: 1403917](https://bugzilla.mozilla.org/show_bug.cgi?id=1403917)


andela-onnenanya:
Here are some documentations that may be necessary for this task:

#### Example `http://localhost:8086/api/histories?rules=TRUE&product_required_signoffs_histories=TRUE&page=4&limit=2&releases=TRUE`

#### Result:
```
{
product_required_signoffs_histories: {
count: 1,
required_signoffs: [ ]
},
releases: {
count: 102,
revisions: [
{
_different: [ ],
_time_ago: "1 month and 5 weeks ago",
change_id: 12298204,
changed_by: "balrog-ffxbld",
data_version: 719151,
name: "Firefox-mozilla-central-nightly-latest",
product: "Firefox",
read_only: "False",
timestamp: 1505185760980
},
{
_different: [
"data"
],
_time_ago: "1 month and 5 weeks ago",
change_id: 12298202,
changed_by: "balrog-ffxbld",
data_version: 719150,
name: "Firefox-mozilla-central-nightly-latest",
product: "Firefox",
read_only: "False",
timestamp: 1505185743393
}
]
},
rules: {
count: 9228,
revisions: [
{
alias: "firefox-beta-localtest",
backgroundRate: 100,
buildID: null,
buildTarget: null,
change_id: 9266,
changed_by: "balrog-ffxbld",
channel: "beta-localtest",
comment: null,
data_version: 379,
distVersion: null,
distribution: null,
fallbackMapping: null,
headerArchitecture: null,
instructionSet: null,
locale: null,
mapping: "Firefox-57.0b3-build1",
memory: null,
mig64: null,
osVersion: null,
priority: 50,
product: "Firefox",
rule_id: 25,
timestamp: 1506365466797,
update_type: "minor",
version: null
},
{
alias: "firefox-beta-cdntest",
backgroundRate: 100,
buildID: null,
buildTarget: null,
change_id: 9265,
changed_by: "balrog-ffxbld",
channel: "beta-cdntest",
comment: "Default beta-cdntest rule updated by automation",
data_version: 350,
distVersion: null,
distribution: null,
fallbackMapping: null,
headerArchitecture: null,
instructionSet: null,
locale: null,
mapping: "Firefox-57.0b3-build1",
memory: null,
mig64: null,
osVersion: null,
priority: 90,
product: "Firefox",
rule_id: 45,
timestamp: 1506365466472,
update_type: "minor",
version: null
}
]
}
}
```


We have a list of histories that can be included in the endpoint simply by assigning `TRUE` to the specified history query string. Like in the example above, rules=TRUE etc.

Filtering: 
Each specified history is returned with the specified filter applied to all. Specifying a filter that does not exist in the table means an empty array will be returned any applicable history.

auslib/web/admin/views/history.py
```diff
@@ -76,6 +85,82 @@ def get_revisions(self,
         ret['count'] = total_count
         return jsonify(ret)
 
+    def get_all_revisions(self,
```
mozbhearsum: There is almost no difference between this function and `get_revisions` - are you sure you need this? If it does have unique functionality, are you sure you need it? Can that functionality be rolled into `get_revisions`?

auslib/web/common/history_all.py
```diff
@@ -0,0 +1,134 @@
...
+        'permissions_required_signoffs_histories': get_permissions_required_signoffs_histories(),
+        'product_required_signoffs_histories': get_product_required_signoffs_histories(),
+        'releases_scheduled_change_histories': get_releases_scheduled_change_histories(),
+        'rules_scheduled_change_histories': get_rules_scheduled_change_histories(),
+        'permissions_scheduled_change_histories': get_permissions_scheduled_change_histories(),
+    }
+
+
+def get_filtered_history():
+    requested_histories = json.loads(_get_input_dict().data)['obj_keys']
```
mozbhearsum: The way this is implemented now, it looks like each different object needs to be explicitly specified to have its history returned. This is going to be a bit of a maintenance burden (whenever we add new objects, we'll need to update the UI to change the query it makes). What do you think about retrieving history for all objects _except_ if it's explicitly disabled. For example, passing "rules_history=0" would cause rules history not be returned, but not passing "rules_history" at all would return it.

auslib/web/common/history_all.py
```diff
@@ -0,0 +1,134 @@
...
+        if obj[key] != keys_values[key]:
+            return False
+    return True
+
+
+def method_constants():
+    return {
+        'rules': get_rules_histories(),
+        'releases': get_releases_histories(),
+        'permissions': get_permissions_histories(),
```
mozbhearsum: Is there a reason you're not appending these the way you are the rest of the strings below?

auslib/web/common/history_all.py
```diff
@@ -0,0 +1,134 @@
...
+    return PermissionsRequiredSignoffsHistoryAPIView().get_all()
+
+
+def _get_input_dict():
+    args = request.args
+    obj_keys = []
+    query_keys = []
+    query = {}
+    for key in args:
+        if args.get(key) == 'TRUE':
```
mozbhearsum: The standard convention for true/false in query parameters is 0 and 1. I've also got another suggestion below that might make this simpler.

auslib/web/admin/views/required_signoffs.py
```diff
@@ -74,6 +90,43 @@ def get(self, input_dict):
 
         return jsonify(count=total_count, required_signoffs=revisions)
 
+    def get_all(self):
```
mozbhearsum: Same question here...what does this do that `get()` doesn't?

auslib/web/admin/swagger/api.yaml
```diff
@@ -349,6 +349,132 @@ paths:
               type: string
           examples:
             text/html: "csrf_token: 1491342563##c4e6fef0b978e6c89af9ff1015e67b9ca7c45d14"
+  
+  /histories:
```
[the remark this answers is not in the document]
andela-onnenanya: Yes, I understand very well how the swagger API works. This PR is just to let you have a look at what this task will look like. Nevertheless, I have detailed my thought and concern in the comment below.

andela-onnenanya:
These are my considerations:

1. I intend having a single endpoint `/history` where all histories can be accessible. This single endpoint will call the other endpoints to get the requested history with the appropriate parameter.

2. For this single endpoint to work as desired, it won't require defined parameters, otherwise we will have to have a long list of parameters to define considering the possible filter field in each history object. For example a user who wants to get all histories where a particular user has made some changes, passing `changed_by` in the query can easily return all changes made by the requested user in rules, releases, sc, required_signoffs, etc. But if we consider a case where `rule_id` is specified, this is peculiar to rules. So, if we want all these peculiarity to be included in the parameters, that will really long.

3. There are two ways I considered we may handle the filters:
  - It's either we leave the endpoint without defining the parameters (meaning we have to 
     make some checks on the parameters; which swagger would have helped us to do)

    OR
 
 - We filter by common parameters (we can define in swagger). Those common parameters 
   include `change-id`, `change_by`, `timestamp`, `product` and `data_version`.
  This means the filtering will be limited.

auslib/web/admin/views/history.py
andela-onnenanya: Though I have changed a lot of things in entire code, this suppose to return all history without default limit of 10 when the user queries the endpoint without providing limits and pages. @andela-hngerebara requested that I should not limit the history to 10, instead she needs the entire history for has UI task.

auslib/web/admin/views/required_signoffs.py
andela-onnenanya: Same thing with the case above.

auslib/web/admin/swagger/api.yaml
```diff
@@ -349,6 +405,160 @@ paths:
               type: string
           examples:
             text/html: "csrf_token: 1491342563##c4e6fef0b978e6c89af9ff1015e67b9ca7c45d14"
+  
+  /rrp/history:
```
mozbhearsum: The endpoints look much improved now, but I'd like to clean up the API a bit further. At the moment, you've got 3 endpoints that all grab history, but the objects in each endpoint are only loosely related. I think there's two ways we can go:
1) Use one endpoint for ALL objects (rules, releases, permissions, scheduled changes, required signoffs -- everything).
2) Use individual endpoints for each type of object (eg: rules would have its own endpoint AND rule scheduled changes would have its own endpoint).

I don't have a strong preference either way, but either would be cleaner than the current endpoints.

auslib/web/admin/swagger/api.yaml
```diff
@@ -251,6 +251,62 @@ parameters:
...
+
+  historyDataVersionParam:
+    name: data_version
+    in: query
+    description: data_version field value to filter history by.
+    type: integer
+    x-nullable: true
+    required: false
+
+  timestampParam:
```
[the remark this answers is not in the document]
andela-onnenanya: Yes, that's true. I'll remove it.

auslib/web/admin/swagger/api.yaml
```diff
@@ -251,6 +251,62 @@ parameters:
...
+
+  timestampToParam:
+    name: timestamp_to
+    in: query
+    description: timestamp to/below which history is filtered.
+    type: integer
+    x-nullable: true
+    required: false
+
+  channelParam:
```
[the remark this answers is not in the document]
andela-onnenanya: Ok, cool.

auslib/web/admin/views/required_signoffs.py
```diff
@@ -48,6 +49,23 @@ def __init__(self, table, decisionFields):
...
+            where = [getattr(self.table.history, f) == query.get(f) for f in query]
+            where.append(self.table.history.data_version != null())
+            where.append(self.history_table.product != null())
+            request = connexion.request
+            if request.args.get('timestamp_from'):
+                where.append(self.history_table.timestamp >= int(request.args.get('timestamp_from')))
+            if request.args.get('timestamp_to'):
+                where.append(self.history_table.timestamp <= int(request.args.get('timestamp_to')))
+            return where
+        except AttributeError:
```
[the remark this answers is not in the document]
andela-onnenanya: Yes, now that we have specified the required parameters, I'm removing that part of the code. Swaggar will ensure that no extra parameter is supplied, hence no need catching AttributeError.

auslib/web/admin/views/myapi.txt
```diff
@@ -0,0 +1,58 @@
+parameters:
```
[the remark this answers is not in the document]
andela-onnenanya: It was an oversight, was not meant to be tracked.

auslib/web/admin/swagger/api.yaml
```diff
@@ -350,6 +406,323 @@ paths:
...
+        - $ref: '#/parameters/timestampFromParam'
+        - $ref: '#/parameters/timestampToParam'
+        - $ref: '#/parameters/channelParam'
+      responses:
+        '200':
+          description: Successfully fetched permissions history
+        '400':
+          $ref: '#/responses/invalidFormData'
+
+  /product_required_signoffs/history:
```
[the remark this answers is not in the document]
andela-onnenanya: Oh! I see "/required_signoffs/product/revisions" endpoint already does almost everything "/product_required_signoffs/history" tries to do. Just a few changes may make things better here. Can I add `time_from` and `time_to` params to the `revisions` endpoint; then delete the `history` endpoint?

auslib/web/admin/swagger/api.yaml
```diff
@@ -350,6 +406,323 @@ paths:
...
+      tags:
+        - "History Histories Permissions"
+      operationId: auslib.web.common.history_all.get_permissions_history
+      consumes: []
+      produces: [application/json]
+      externalDocs:
+        url: "http://mozilla-balrog.readthedocs.io/en/latest/admin_api.html#put"
+        description: "Returns a list of rules, releases and permissions histories"
+      parameters:
+        - $ref: '#/parameters/productParam'
```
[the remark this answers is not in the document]
andela-onnenanya: Ok, I'll take care of that

auslib/web/admin/swagger/api.yaml
```diff
@@ -350,6 +406,323 @@ paths:
...
+        - $ref: '#/parameters/limitParam'
+        - $ref: '#/parameters/pageParam'
+        - $ref: '#/parameters/changedByParam'
+        - $ref: '#/parameters/changeIdParam'
+        - $ref: '#/parameters/historyDataVersionParam'
+        - $ref: '#/parameters/timestampParam'
+        - $ref: '#/parameters/timestampFromParam'
+        - $ref: '#/parameters/timestampToParam'
+        - $ref: '#/parameters/channelParam'
```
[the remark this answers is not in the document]
andela-onnenanya: I'm removing it, not needful here.

auslib/web/admin/swagger/api.yaml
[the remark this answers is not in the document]
mozbhearsum: That's a great idea. They currently only work on one row of those tables at a time, which means they have some required query parameters (like `product` and `role`) - so you'll need to tweak them to make it possible to return all history at once.

andela-onnenanya:
Hi Ben, what could be the reason why the `frontend` test fails >>>> Says phantom.js is missing? I didn't do anything on the frontend.

mozbhearsum:
That's a bug from a recent change to a module we depend on - we fixed it on master already. I just pushed that merge to your branch, which should fix things up.

auslib/web/admin/swagger/api.yaml
```diff
@@ -350,6 +406,323 @@ paths:
...
+      tags:
+        - "History Signoff Product"
+      operationId: auslib.web.common.history_all.get_product_required_signoffs_history
+      consumes: []
+      produces: [application/json]
+      externalDocs:
+        url: "http://mozilla-balrog.readthedocs.io/en/latest/admin_api.html#put"
+        description: "Returns a list of product required signoff history"
+      parameters:
+        - $ref: '#/parameters/productParam'
```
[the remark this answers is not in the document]
andela-onnenanya: I've taken care of this too

andela-onnenanya:
Hi, Ben. I learnt from @andela-hngerebara that she has made a lot of changes from what I have here currently. So, a quick one. Are there some other things you want me to fix aside the test?
I'm currently working on the test, though it's very slow and docker also goes bad often. 
OR
If what @andela-hngerebara has done is quite better than what I have, I wouldn't mind if you go ahead and close my PR; she then continues on her own branch.
Thank you!

andela-onnenanya:
Never mind, I have clarified with her.

auslib/web/admin/swagger/api.yaml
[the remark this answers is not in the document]
andela-onnenanya: I tried tweaking the code so we can retain only `revisions` endpoint and make it serve also for `history`, but it was getting dirty; I had to leave them as separate endpoints.

auslib/web/common/history.py
```diff
@@ -53,6 +53,20 @@ def get_history(self, response_key='revisions'):
...
+    args = request.args
+    query_keys = []
+    query = {}
+    for key in args:
+        if key not in reserved_filter_params:
+            query_keys.append(key)
+
+    for key in query_keys:
+        query[key] = request.args.get(key)
+    return jsonify(query_keys=query_keys, query=query)
```
mozbhearsum: `jsonify` should only be used when returning an HTTP Response - there's no point in returning JSON back to other Python functions -- just return a dict or tuple.

auslib/web/common/history_all.py
```diff
@@ -0,0 +1,105 @@
...
+log = logging.getLogger(__name__)
+
+
+def _get_filters(obj, history_table):
+    input_dict = get_input_dict()
+    query = json.loads(input_dict.data)['query']
+    where = [False, False]
+    try:
+        where = [getattr(history_table, f) == query.get(f) for f in query]
+        where.append(history_table.product != null())
```
mozbhearsum: This appears to be the line causing the tracebacks - it's not a safe assumption that product exists for all history tables.

auslib/web/common/history.py
andela-onnenanya: Got that! I've fixed it now!

auslib/web/common/history_all.py
andela-onnenanya: Yea, I've fixed that. Testing the endpoints has actually exposed some bugs I have to fix.
Thank you!

andela-onnenanya:
Hi @mozbhearsum, I don't know why some of the CI Task Groups fail while the front/back end tests pass. 

I have added the required tests. Kindly review. Thanks!

auslib/test/admin/views/test_releases.py
```diff
@@ -1729,6 +1765,71 @@ def testGetRevisions(self):
...
+        ret = self._get(url)
+        self.assertEquals(ret.status_code, 200, msg=ret.data)
+        data = json.loads(ret.data)
+        self.assertEquals(data["count"], 3)
+        expected = {
+            'count': 3,
+            'revisions': [
+                {
+                    'product': 'b', 'data_version': 1, 'changed_by': 'bill', '_different': [], 'read_only': 'False',
+                    'change_id': 6, '_time_ago': '48 years and 583 months ago', 'name': 'b', 'timestamp': 16
```
mozbhearsum: I don't think these _time_ago fields are continue be accurate forever. It looks like they'll be wrong starting in ~1 month. You're not actually comparing them below as far as I can tell, so you may as well just remove them here.

auslib/web/admin/views/required_signoffs.py
```diff
@@ -48,6 +50,19 @@ def __init__(self, table, decisionFields):
         self.decisionFields = decisionFields
         super(RequiredSignoffsHistoryAPIView, self).__init__(table=table)
 
+    def _get_filters(self):
+        query = get_input_dict()
+        where = [getattr(self.table.history, f) == query.get(f) for f in query]
+        where.append(self.table.history.data_version != null())
+        if hasattr(self.history_table, 'product'):
+            where.append(self.history_table.product != null())
```
mozbhearsum: I'm not sure I understand this block. All required signoffs tables have a required column of product, so it will always be in every row of the history table. Are you sure this is useful?

mozbhearsum:
For posterity, the test failures from the latest CI were infrastructure related - reruns of them worked fine.

Very nice work here @andela-onnenanya! This was a pretty complicated feature and it's in a good place now! I know many of us are looking forward to using it.

auslib/web/admin/views/required_signoffs.py
andela-onnenanya: I think it's redundant. I'll remove it. Thank you!

auslib/test/admin/views/test_releases.py
andela-onnenanya: Ok, I'll take a closer look at that and fix as necessary.

andela-onnenanya:
Wao! @mozbhearsum  I'm excited that my contribution is very useful! 

For the two reviews above, I'll fix them.

PR #433, opened 2017-10-17
bug 1402411: add support for JAWS in SYSTEM_CAPABILITIES

mozbhearsum:
There's nothing too new or interesting in this patch - the JAWS field is like instructionSet/memory in that it lives in SYSTEM_CAPABILITIES, but its matching is like mig64 in that it's a boolean.

I also did a tiny bit of renaming and refactoring in the Rules class to improve readability.

@glasserc - Since this is the first Balrog you're doing I'm going to ask @JohanLorenzo for a sanity check as well. It looks like you're not able to give this the official r+ either...

### DIFF
--- a/auslib/web/common/swagger/definitions.yml
+++ b/auslib/web/common/swagger/definitions.yml
@@ -325,7 +325,7 @@ definitions:
         example: true
 
       jaws:
-        description: Whether or not the Rule should apply to queries that indicate an incompatible version of the JAWS screen reader. If set to false or true, the query and the rule must match precisely. If set to null, the jaws value in the updateQuery is ignored.
+        description: Whether or not the Rule should apply to queries that indicate a client that has an incompatible version of the JAWS screen reader installed. If set to false or true, the query and the rule must match precisely. If set to null, the jaws value in the updateQuery is ignored.
         type: ["boolean", "null"]
         example: true
 
--- a/auslib/web/common/swagger/definitions.yml
+++ b/auslib/web/common/swagger/definitions.yml
@@ -325,7 +325,7 @@ definitions:
         example: true
 
       jaws:
-        description: Whether or not the Rule should apply to queries that specify their JAWS screen reader incompatibility status. If set to false or true, the query and the rule must match precisely. If set to null, the jaws value in the updateQuery is ignored.
+        description: Whether or not the Rule should apply to queries that are incompatible with the JAWS screen reader. If set to false or true, the query and the rule must match precisely. If set to null, the jaws value in the updateQuery is ignored.
         type: ["boolean", "null"]
         example: true
 
--- a/auslib/web/common/swagger/definitions.yml
+++ b/auslib/web/common/swagger/definitions.yml
@@ -324,6 +324,11 @@ definitions:
         type: ["boolean", "null"]
         example: true
 
+      jaws:
+        description: Whether or not the Rule should apply to queries that specify their JAWS screen reader incompatibility status. If set to false or true, the query and the rule must match precisely. If set to null, the jaws value in the updateQuery is ignored.
+        type: ["boolean", "null"]
+        example: true
+
   RuleObject:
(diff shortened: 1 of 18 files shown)


auslib/web/common/swagger/definitions.yml
```diff
@@ -325,7 +325,7 @@ definitions:
         example: true
 
       jaws:
-        description: Whether or not the Rule should apply to queries that are incompatible with the JAWS screen reader. If set to false or true, the query and the rule must match precisely. If set to null, the jaws value in the updateQuery is ignored.
+        description: Whether or not the Rule should apply to queries that indicate an incompatible version of the JAWS screen reader. If set to false or true, the query and the rule must match precisely. If set to null, the jaws value in the updateQuery is ignored.
```
glasserc: I had to read through both linked bugs to understand this correctly. Could we maybe elaborate a little on this, either here or below, to say something like "queries that have JAWS installed but with a version before the fix that makes it compatible with e10s"? Also, where are the docs for arguments to the update function? We could add similar information there.

docs/source/database.rst
```diff
@@ -105,6 +105,11 @@ Following tables show columns according to different Categories:
   |                        |                    | requesting the update has                        | memory to compare the incoming  | ">=8096"                   |
   |                        |                    |                                                  | one against                     |                            |
   |                        +--------------------+--------------------------------------------------+---------------------------------+----------------------------+
+  |                        | jaws               | Whether or not the the Rule should apply to      | Exact match only                | True, False, or NULL       |
+  |                        |                    | queries that indicate an imcompatible version of |                                 |                            |
```
glasserc: Oh, also, `imcompatible` has a typo.

auslib/web/common/swagger/definitions.yml
```diff
@@ -324,6 +324,11 @@ definitions:
         type: ["boolean", "null"]
         example: true
 
+      jaws:
+        description: Whether or not the Rule should apply to queries that indicate an incompatible version of the JAWS screen reader. If set to false or true, the query and the rule must match precisely. If set to null, the jaws value in the updateQuery is ignored.
```
JohanLorenzo: Wow! Up until this point, I though JAWS was some Java-related capability. Thanks for documenting that. TIL JAWS!

ui/app/templates/rule_modal.html
```diff
@@ -90,58 +90,68 @@ <h3 class="modal-title" ng-show="!is_edit && is_duplicate">Duplicate Existing Ru
...
+        <div class="form-group" ng-class="{'has-error': errors.jaws}">
+          <label for="id_jaws">
+            JAWS Screen Reader Compatibility
+            <span class="glyphicon glyphicon-question-sign"
+                  title='Whether or not to match incoming queries that indicate an incompatible version of the JAWS screen reader. "Ignore" is usually what you want.'>
+            </span>
+          </label>
+          <select id="id_jaws" ng-model="rule.jaws" class="form-control"
+                  ng-options="value as key for (key, value) in {'Ignore': null, 'Yes': true, 'No': false}">
+            <!-- Angular creates an extra entry because one of our values is null. This hides it.
```
JohanLorenzo: TIL, thanks!

auslib/web/common/swagger/definitions.yml
mozbhearsum: Good idea - I'll improve on this. What do you mean when you say "update function"?

glasserc: I meant, where do we document the parameters to the /update URLs, and especially what `SYSTEM_CAPABILITIES` are valid and what they mean.

mozbhearsum: Hmm, we don't actually have any update URL centric documentation that I can find (either in Balrog or elsewhere) - that's a very good idea though. For now, I'll do what I can to improve the clarity in the existing Balrog docs.

PR #436, opened 2017-10-18
Docs - README.md - adding extra gnu commands

tiagomoraismorgado88:
## Docs - README.md - adding extra gnu commands

**This PR does basically aim at:**
- *adding extra gnu commands to Readme.md*;

---

**Main motivation behind such does basically lie within**:
- making file more complete

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -9,8 +9,13 @@ Download
 Clone from git using
 
 ::
-
+    $ cd ~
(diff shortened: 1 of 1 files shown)


mozbhearsum:
I appreciate the suggestion, but I don't think this adds much value, but adds a lot of noise. I think it's a fair assumption that developers can create their own workspaces if desired.

PR #438, opened 2017-10-23
Bug 1403975: scheduled change modal hangs when no "when" is set and asap is not clicked

collin5:
Fixes `Error: b.when is undefined` which occurs when adding scheduled changes without setting time/date.

### DIFF
--- a/auslib/test/admin/views/test_rules.py
+++ b/auslib/test/admin/views/test_rules.py
@@ -1,5 +1,7 @@
 import json
 import mock
+import datetime
+import time
 
 from auslib.global_state import dbo
 from auslib.test.admin.views.base import ViewTest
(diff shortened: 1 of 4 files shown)


collin5:
@mozbhearsum Thanks for the feedback.
I also had to check if `data.when` in `null` in rule scheduled changes backend to be consistent with using null as this is the case for release schedule changes

collin5:
@mozbhearsum PR should be now ready for another review. Thanks

auslib/web/admin/views/rules.py
```diff
@@ -200,6 +200,9 @@ def __init__(self):
 
     @requirelogin
     def _post(self, transaction, changed_by):
+        if connexion.request.get_json().get("when", None) is None:
+            return problem(400, "Bad Request", "'when' cannot be set to null when scheduling a new change "
+                                               "for a Release")
```
mozbhearsum: I think this should be "Rule", not "Release"?

auslib/test/admin/views/test_rules.py
```diff
@@ -1206,10 +1208,11 @@ def testGetScheduledChangesWithCompleted(self):
         self.assertEquals(json.loads(ret.data), expected)
 
     def testAddScheduledChangeExistingRule(self):
+        future_time = int(time.mktime((datetime.datetime.now() + datetime.timedelta(minutes=5)).timetuple()) * 1000)
```
mozbhearsum: This works, but please following the existing pattern for this by using mock.patch instead, like: https://github.com/andela/balrog/blob/20ee523439c183b9581812f23afe2a69f5d2b706/auslib/test/admin/views/test_rules.py#L1289.

auslib/web/admin/views/rules.py
collin5: Thank you, I had completely missed that one.

collin5:
@mozbhearsum PR should be ready now, thanks again 👍

PR #439, opened 2017-10-23
 bug-1408178: Sort rules history by data version

andela-ookwuolisa:
This PR sorts the rules' history by the data version


ui/app/js/controllers/rules_controller.js
```diff
@@ -132,7 +132,10 @@ function($scope, $routeParams, $location, $timeout, Rules, Search, $modal, $rout
     // will be inserted based on the priority in the Scheduled Change.
     // Rules that have Scheduled updates or deletes will remain sorted on their current priority
     // because it's more important to make it easy to assess current state than future state.
-    if (rule.priority === null || rule.priority === undefined) {
+    if(rule.timestamp){
```
mozbhearsum: Please base this decision on $scope.rule_id instead - that's what we do everywhere else.

ui/app/js/controllers/rules_controller.js
```diff
@@ -170,12 +170,23 @@ function($scope, $routeParams, $location, $timeout, Rules, Search, $modal, $rout
...
+  };
+
   $scope.orderRules = function(rule) {
     // Rules are sorted by priority. Rules that are pending (ie: still just a Scheduled Change)
     // will be inserted based on the priority in the Scheduled Change.
     // Rules that have Scheduled updates or deletes will remain sorted on their current priority
     // because it's more important to make it easy to assess current state than future state.
-    if (rule.priority === null || rule.priority === undefined) {
+
+    if($scope.returnRuleType() === "Revision"){
```
mozbhearsum: While this works, I think you're overcomplicating it. You can simply check if `$scope.rule_id` is set to know whether or not we're on the Rules History page, like we do at https://github.com/mozilla/balrog/blob/0b0e7e53c6c399aca913b49ecef53966d76dba01/ui/app/js/controllers/rules_controller.js#L60.

ui/app/js/controllers/rules_controller.js
```diff
@@ -170,12 +170,23 @@ function($scope, $routeParams, $location, $timeout, Rules, Search, $modal, $rout
...
 
+  $scope.returnRuleType = function() {
+    var firstRuleId = $scope.rules[0].rule_id;
+    return $scope.rules.every(function(rule) {
+      return rule.rule_id === firstRuleId;
+    }) ? "Revision" : "All Rules";
+  };
+
   $scope.orderRules = function(rule) {
     // Rules are sorted by priority. Rules that are pending (ie: still just a Scheduled Change)
```
mozbhearsum: Please keep this comment just above in the "else" block -- it doesn't relate to the new code you've added.

ui/app/js/controllers/rules_controller.js
andela-ookwuolisa: Hi, Ben, I thought about that initially, but I observed `$scope.rule_id` is always set for both Rules History and All Rules. The reason why I had to check if `$scope.rule_id` has the same value for each rule or not.

mozbhearsum: Are you sure about this? $scope.rule_id comes from $routeParams.id. When we're on /rules, there's no id set as far as I know...

PR #442, opened 2017-10-31
Enable python threads

Rajesh-Veeranki:
Ref: https://bugzilla.mozilla.org/show_bug.cgi?id=1316602
More Info: http://uwsgi-docs.readthedocs.io/en/latest/WSGIquickstart.html#a-note-on-python-threads
http://uwsgi-docs.readthedocs.io/en/latest/ThingsToKnow.html?highlight=threads

### DIFF
--- a/uwsgi/admin.ini
+++ b/uwsgi/admin.ini
@@ -1,6 +1,9 @@
 [uwsgi]
 http = :$(PORT)
 
+# Enable Python threads
+enable-threads = True
+
 # Mount the admin app at /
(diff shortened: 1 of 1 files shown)


mozbhearsum:
Hi, I appreciate the pull request but I want to give you a heads up that I won't be able to look it at for awhile. Doing this requires a lot of testing in our deployed environments, and I'm not sure when we'll be able to do it.

Rajesh-Veeranki:
No problem at all. Consider only when you have time :)

PR #443, opened 2017-11-02
fixes bug 1414031 - neat columns in cards

peterbe:
What it looks like after:
after

Sorry. The diff looks enormous. If only you could do diffs here in GitHub without whitespace. 
[Here's what you get if you diff ignoring whitespace](https://gist.github.com/peterbe/886ef467c5c64835465e3ece2b8ae7eb).
The diff is so big because two divs had to be contained in one and indented 2 spaces to the right.

### DIFF
--- a/ui/app/templates/rules.html
+++ b/ui/app/templates/rules.html
@@ -148,99 +148,102 @@
         </ul>
       </span>
     </div>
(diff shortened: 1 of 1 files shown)


leplatrem:
Protip from @Natim: You can ignore whitespaces with `?w=1` :) https://github.com/mozilla/balrog/pull/443/files?w=1

PR #447, opened 2017-11-10
bug-1355477 View Diff in Scheduled Release updates

Alweezy:


auslib/web/admin/swagger/api.yaml
```diff
@@ -376,6 +376,34 @@ paths:
         '404':
           $ref: '#/responses/resourceNotFound'
 
+  /scheduled_change/diff/release/{sc_id}/{field}:
```
mozbhearsum: There's no other field that's actually useful to diff - it would be better to explicitly only support "data" here.

Please add a test or two for this endpoint, too!

ui/app/js/controllers/scheduled_release_update_controller.js
```diff
@@ -0,0 +1,18 @@
+angular.module('app').controller('ScheduledReleaseUpdateDataCtrl',
```
mozbhearsum: Just a couple of minor nits on this:
1) Please improve the name - the current one makes it sound like it's for updating scheduled releases. "ScheduledReleaseDiffCtrl" and "scheduled_release_diff_controller.js" would be better.
2) Please improve the formatting to make the indentation match the rest of the files.

collin5:
@mozbhearsum Thanks for the feedback. Done implementing the requested changes 👍

ui/app/js/services/releases_service.js
```diff
@@ -46,6 +46,19 @@ angular.module("app").factory('Releases', function($http, $q, ScheduledChanges,
         }
       });
     },
+    getUpDiff: function(sc_id) {
+      var url = '/api/scheduled_change/diff/release/' + sc_id;
+      return $http({
+        url: url,
+        method: 'GET',
+        transformResponse: function(value) {
```
mozbhearsum: Huh, this seems very strange given that the response's Content-Type is text/plain - good catch though.

auslib/web/admin/views/releases.py
```diff
@@ -622,3 +622,48 @@ def get(self, change_id, field):
...
+        if 'data' not in release:
+            return problem(400, "Bad Request", "Bad field")
+
+        previous = json.dumps(release['data'], indent=2, sort_keys=True)
+        value = json.dumps(sc["base_{}".format('data')], indent=2, sort_keys=True)
+        result = difflib.unified_diff(
+            previous.splitlines(),
+            value.splitlines(),
+            fromfile="Data Version {}".format(release["data_version"]),
+            tofile="Data Version {}".format(sc["data_version"]),
```
mozbhearsum: These labels aren't quite right - sc_data_version and data_version aren't really comparable in this way (the former tracks changes to the release, the latter tracks changes to a specific scheduled update to a release). Better labels would be "Current Version (Data Version X)" for fromfile, and "Scheduled Update" for tofile. The sc data_version isn't important to show here.

collin5:
Ur welcome. Thanks too

PR #448, opened 2017-11-12
Bug 1416606 - apply poll interval hack to new location

nthomas-mozilla:
The fix in #140 broke at some point, and this makes it work again.


peterbe:
Will take a look.

PR #450, opened 2017-11-24
bug_1307169 - Emergency shut off backend API

allan-silva:
Emergency shut off API backend implementation.

Main functionality:
- Shut off updates for a given product/channel.
- `POST - /emergency_shutoff`: Enables a product/channel to disable updates. The product should require signoffs, this is an important requirement to reenable updates in a controlled way and by specialized people (e.g. releng, relman).
- `PUT - /emergency_shutoff/{shutoff_id}/disable_updates`: Disables updates for product/channel. Signoffs are not required to disable updates. When updates are disabled, a ASAP change is scheduled to reenable updates, when signoffs are provided for Scheduled Change, updates are automatically reenabled.

At rule evaluation process we looking for emergency shut off for given product/channel, if emergency shut off exists and `updates_disabled` is  `True` then an empty update is provided.

TODO:
- Add UI to manage Emergency Shut off;
- Evaluate a way to allow people that does not have Balrog access disable the updates (possibly a site in mozilla/services).


auslib/AUS.py
```diff
@@ -52,9 +52,23 @@ def __init__(self):
...
+    def get_emergency_shutoff(self, product, channel):
+        where = dict(product=product, channel=channel)
+        emergency_shutoffs = dbo.emergencyShutoff.getEmergencyShutoffs(where=where)
+        return emergency_shutoffs[0] if emergency_shutoffs else None
+
     def evaluateRules(self, updateQuery):
         self.log.debug("Looking for rules that apply to:")
         self.log.debug(updateQuery)
+
+        if 'product' in updateQuery and 'channel' in updateQuery:
```
mozbhearsum: I think it's safe to assume that product and channel are in the update query, since they exist in all update URL styles.

auslib/db.py
```diff
@@ -2490,6 +2491,53 @@ def _putWatchdogValue(self, changed_by, value, where=None, transaction=None, dry
             super(Dockerflow, self).update(where=where, what=value, changed_by=changed_by, transaction=transaction, dryrun=dryrun)
 
 
+class EmergencyShutoff(AUSTable):
+    def __init__(self, db, metadata, dialect):
+        self.table = Table('emergency_shutoff', metadata,
+                           Column('shutoff_id', Integer, primary_key=True, autoincrement=True),
+                           Column('product', String(15), nullable=False),
+                           Column('channel', String(75), nullable=False),
```
mozbhearsum: Have you considered getting rid of shutoff_id and using product+channel as the primary key?

auslib/AUS.py
```diff
@@ -52,9 +52,23 @@ def __init__(self):
         self.rand = functools.partial(randint, 0, 99)
         self.log = logging.getLogger(self.__class__.__name__)
 
+    def get_emergency_shutoff(self, product, channel):
```
mozbhearsum: Will this handle fallback channels as well? Eg: shutoffs for "release" should apply when the updateQuery has a channel of "release-cck-foo".

auslib/db.py
```diff
@@ -2490,6 +2491,53 @@ def _putWatchdogValue(self, changed_by, value, where=None, transaction=None, dry
...
+                           Column('updates_disabled', CompatibleBooleanColumn, default=False, nullable=False),
```
mozbhearsum: "not updates_disabled" is a double negative, and makes any code using this tougher to read. "enabled" would be better here, then we'll have "enabled" and "not enabled", which are more readable IMO.

auslib/db.py
```diff
@@ -2490,6 +2491,53 @@ def _putWatchdogValue(self, changed_by, value, where=None, transaction=None, dry
...
+                           Column('additional_notification_list', String(500)))
+        AUSTable.__init__(self, db, dialect, scheduled_changes=True, scheduled_changes_kwargs={"conditions": ["time"]})
+
+    def getEmergencyShutoff(self, shutoff_id):
```
mozbhearsum: Is this (and the next) method adding any value? Can callers just call select() on their own?

auslib/db.py
```diff
@@ -2490,6 +2491,53 @@ def _putWatchdogValue(self, changed_by, value, where=None, transaction=None, dry
...
+        return emergency_shutoff[0] if emergency_shutoff else None
+
+    def getEmergencyShutoffs(self, where=None):
+        return self.select(where=where)
+
+    def insert(self, changed_by, transaction=None, dryrun=False, signoffs=None, **columns):
+        ret = super(EmergencyShutoff, self).insert(changed_by=changed_by, transaction=transaction, dryrun=dryrun, **columns)
+        return ret.inserted_primary_key[0]
+
+    def update(self, where, what, changed_by, old_data_version, transaction=None, dryrun=False, signoffs=None):
```
mozbhearsum: Out of curiosity, did you consider removing rows altogether instead of updating them? Either is OK, I'm just wondering if you considered both.

auslib/db.py
```diff
@@ -2490,6 +2491,53 @@ def _putWatchdogValue(self, changed_by, value, where=None, transaction=None, dry
             super(Dockerflow, self).update(where=where, what=value, changed_by=changed_by, transaction=transaction, dryrun=dryrun)
 
 
+class EmergencyShutoff(AUSTable):
```
mozbhearsum: Naming nit: this class and the table name should be plural ("EmergencyShutoffs").

auslib/test/web/test_client.py
```diff
@@ -1287,6 +1287,38 @@ def testRuleNullQueryFalse(self):
 """)
 
 
+class ClientTestEmergencyShutoff(ClientTestBase):
+    def setUp(self):
+        super(ClientTestEmergencyShutoff, self).setUp()
+        dbo.emergencyShutoff.t.insert().execute(
+            shutoff_id=1, product='b', channel='a',
+            updates_disabled=False, data_version=1,
+            additional_notification_list='foo@bar.org')
```
mozbhearsum: Please add a test case for "foo-cck-bar" channels, too.

auslib/db.py
```diff
@@ -2490,6 +2491,53 @@ def _putWatchdogValue(self, changed_by, value, where=None, transaction=None, dry
...
+    def __init__(self, db, metadata, dialect):
+        self.table = Table('emergency_shutoff', metadata,
+                           Column('shutoff_id', Integer, primary_key=True, autoincrement=True),
+                           Column('product', String(15), nullable=False),
+                           Column('channel', String(75), nullable=False),
+                           Column('updates_disabled', CompatibleBooleanColumn, default=False, nullable=False),
+                           Column('additional_notification_list', String(500)))
```
mozbhearsum: I see nothing is actually using this yet - do you have an idea for how it will be implemented?

auslib/test/admin/views/test_emergency_shutoff.py
```diff
@@ -0,0 +1,176 @@
...
+        resp = self._post('/emergency_shutoff', data=data.copy())
+        self.assertStatusCode(resp, 201)
+
+        qs = {'channel': 'release', 'product': 'Thunderbird'}
+        resp = self._get('/emergency_shutoff', qs=qs)
+        self.assertStatusCode(resp, 200)
+        resp_data = json.loads(resp.data)
+
+        shutoff = resp_data['shutoffs'][0]
+        for k, v in data.items():
```
mozbhearsum: I think you should also ensure that "updates_disabled" gets the correct value.

auslib/test/admin/views/test_emergency_shutoff.py
```diff
@@ -0,0 +1,176 @@
...
+
+    def test_create_emergency_shutoff(self):
+        data = {'product': 'Thunderbird',
+                'channel': 'release',
+                'additional_notification_list': 'm@m.org'}
+        resp = self._post('/emergency_shutoff', data=data.copy())
+        self.assertStatusCode(resp, 201)
+
+        qs = {'channel': 'release', 'product': 'Thunderbird'}
+        resp = self._get('/emergency_shutoff', qs=qs)
```
mozbhearsum: You're testing the _post() function in this test, but this test will start to fail if _get() breaks. It would be better to pull the new values directly from the database (like we do in other web tests) instead of relying on the HTTP API for them.

auslib/web/admin/emergency_shutoff.py
```diff
@@ -0,0 +1,154 @@
+import time
```
mozbhearsum: Thank you for implementing these directly and not adding more indirection with mapper.py!

auslib/web/admin/swagger/api.yaml
```diff
@@ -3402,6 +3402,232 @@ paths:
...
+              description: data version of the shut off
+              type: integer
+              format: int32
+              minimum: 1
+        '404':
+          $ref: '#/responses/resourceNotFound'
+
+    put:
+      summary: Update an emergency shut off.
+      description: Update an emergency shut off. The product/channel can not be the same than an existing emergency shut off.
```
mozbhearsum: Does this endpoint allow someone to change the product or channel of an existing emergency shutoff? If so, that's probably confusing. It would be better to have users delete the existing shutoff and create an entirely new one. The only "update" that should happen to an existing shutoff is changing whether or not updates are disabled or not.

allan-silva: Sure, I was in doubt by include or not this update endpoint. Will be removed.

auslib/db.py
allan-silva: Just to keeping api module more clean, select() can be called directly.

auslib/db.py
allan-silva: People to be notified in addition to mailing list.
Oh! I forgot to include in mail list!
Any way, I can remove it, if it seems unnecessary for now.

auslib/db.py
allan-silva: I do not understand your question, I think.
The update method changes enable/disable state, more than that, update does not require signoffs for disabling updates.

auslib/db.py
mozbhearsum: Yeah, I think that's best until there's actually code that will send e-mail when something happens. It's a nice idea though.

auslib/db.py
mozbhearsum: From my perspective, there's two ways that shutoffs can be handled:
1) If no row exists in this table, or updates_disabled is False - the product+channel is NOT shutoff. If the row exists in the table, and updates_disabled is True - the produt+channel is shutoff. This means that the first time we shut updates off, a new row is created, and whenever we make a change after that, we update the updates_disabled column.
2) If no row exists in the table, updates are on. If a row does exist in the table, updates are off. This means that when we shut updates off, we insert a row, and when we turn them back on, we remove it.

As far as I can tell, this PR does the first, and I was wondering if you thought about the second.

allan-silva: Ah, actually not! Because I thought about workflow in another way:
1 - User wants to give to product/channel the possibility to disable updates, so he creates a row with 'updates_disabled' False.
1.1 - If updates needs be disabled, just call to change 'updates_disabled' to True and Schedule an ASAP change (that requires signoffs) to go back 'updates_disabled' to False.

So in that way an admin can define what can be shut off and the others user can see what is available to be shut off.

I like that second way, is more simple. Do you think we can change to add/remove approach?

mozbhearsum: Ah, I see what you mean - it gives a bit more control over the types of shutoffs that can be done. I think it's a bit more than is needed though - if we need to be more precise about the shutoffs that can be done we can add support for "product" and/or "channel" as an option on whichever permission we create for making emergency shutoffs.

auslib/AUS.py
allan-silva: Actually not. Can I consider the database documention to implement this?

> The channel has special handling to try "falling back" to a simpler channel, for example a request with release-cck-foo will also consider rules for 'release'. This only applies to channels containing '-cck-'.

In other words, the portion of channel name before '-cck-' sequence is the fallback channel.

mozbhearsum: Yup, that's exactly right.

allan-silva: Oh! Thanks for response, I realized now that the answer for my question is at line 44 in that file!

agent/balrogagent/test/test_cmd.py
```diff
@@ -181,14 +181,16 @@ def side_effect(balrog_api_root, endpoint, auth, loop, method='GET'):
...
+        self.assertEquals(request.call_count, 9)
         called_endpoints = [call[0][1] for call in request.call_args_list]
         self.assertIn('/scheduled_changes/releases', called_endpoints)
         self.assertIn('/scheduled_changes/permissions', called_endpoints)
         self.assertIn('/scheduled_changes/rules', called_endpoints)
+        self.assertIn('/scheduled_changes/emergency_shutoff', called_endpoints)
         self.assertIn('/scheduled_changes/releases/4/enact', called_endpoints)
         self.assertIn('/scheduled_changes/releases/5/enact', called_endpoints)
         self.assertIn('/scheduled_changes/releases/6/enact', called_endpoints)
+        self.assertIn('/scheduled_changes/releases/6/enact', called_endpoints)
```
mozbhearsum: Accidental copy/paste? I don't think this endpoint will be called twice.

auslib/AUS.py
```diff
@@ -52,9 +52,22 @@ def __init__(self):
...
+    def updates_are_disabled(self, product, channel):
+        where = dict(product=product, channel=channel)
+        emergency_shutoffs = dbo.emergencyShutoffs.select(where=where)
+        return bool(emergency_shutoffs)
+
     def evaluateRules(self, updateQuery):
         self.log.debug("Looking for rules that apply to:")
         self.log.debug(updateQuery)
+
+        if self.updates_are_disabled(updateQuery['product'], updateQuery['channel']) or \
```
mozbhearsum: This is the perfect spot to handle this - nice work.

auslib/db.py
```diff
@@ -1156,7 +1176,8 @@ def insert(self, changed_by, transaction=None, dryrun=False, **columns):
             if not self._dataVersionsAreSynced(sc_id, transaction):
                 raise MismatchedDataVersionError("Conditions data version is out of sync with main table for sc_id %s",
                                                  sc_id)
-            self.auto_signoff(changed_by, transaction, sc_id, dryrun, columns)
+            if allow_auto_signoff:
```
mozbhearsum: Can you explain why auto signoff shouldn't be allowed here? If I'm understanding correctly, users in one of the required groups could still sign off manually afterwards...

auslib/test/admin/views/test_emergency_shutoff.py
```diff
@@ -0,0 +1,85 @@
...
+        for key in data.keys():
+            self.assertEquals(data[key], shutoffs[0][key])
+
+    def test_try_create_for_existent_product_channel_shutoff(self):
+        data = {'product': 'Fennec',
+                'channel': 'beta'}
+        resp = self._post('/emergency_shutoff', data=data)
+        self.assertStatusCode(resp, 400)
+
+    def test_try_create_for_product_with_no_required_signoff(self):
```
mozbhearsum: Supporting emergency shutoffs for channels without signoffs is actually something we want to explicitly support -- one of the main use cases for the feature is allowing sheriffs to quickly shutoff Nightly updates when something goes wrong, which aren't protected by signoffs.

auslib/test/test_db.py
```diff
@@ -4955,6 +4963,23 @@ def _add_jaws_test(self, db, upgrade=True):
...
+    def _add_emergency_shutoff_tables(self, db, upgrade=True):
+        metadata = self._get_reflected_metadata(db)
+        shutoff_tables = ["emergency_shutoffs",
+                          "emergency_shutoffs_history",
+                          "emergency_shutoffs_scheduled_changes",
+                          "emergency_shutoffs_scheduled_changes_history",
+                          "emergency_shutoffs_scheduled_changes_conditions",
+                          "emergency_shutoffs_scheduled_changes_conditions_history",
+                          "emergency_shutoffs_scheduled_changes_signoffs",
+                          "emergency_shutoffs_scheduled_changes_signoffs_history"]
```
mozbhearsum: I know it's a pain, but can you also do column-level verification here? One of the main things we try to catch here is column difference betweens db.py and the migration files.

auslib/web/admin/emergency_shutoff.py
```diff
@@ -0,0 +1,104 @@
...
+def post(emergency_shutoff, changed_by, transaction):
+    if shutoff_already_exists(emergency_shutoff['product'], emergency_shutoff['channel']):
+        return problem(
+            400, 'Bad Request', 'Invalid Emergency shutoff data',
+            ext={'data': 'Emergency shutoff for product/channel already exists.'})
+    if not product_requires_signoffs(emergency_shutoff['product'], emergency_shutoff['channel']):
+        return problem(
+            400, 'Bad Request', 'Invalid Emergency shutoff data',
+            ext={'data': 'The given product/channel should requires signoffs.'})
+    emergency_shutoff = process_shutoff_post(emergency_shutoff)
```
mozbhearsum: I think it would be better to get rid of this and explicitly pass product and channel. I don't think there's a reason to be setting data_version here (correct me if I'm wrong though - maybe deleting & recreating rows with the same PK works different), and past that the helper provides little value.

auslib/web/admin/emergency_shutoff.py
```diff
@@ -0,0 +1,104 @@
...
+    dbo.emergencyShutoffs.insert(
+        changed_by=changed_by, transaction=transaction, **emergency_shutoff)
+    schedule_reenable_updates(emergency_shutoff, changed_by, transaction)
```
mozbhearsum: Hm, this is interesting! I wasn't expecting this to be scheduled right at shutoff time - I would've expected a user to need to hit a button to schedule it. I guess this is why you disabled auto signoff for the new table? I'm curious why you chose to do it here instead of in a separate endpoint.

allan-silva: This is to achieve two requirements: 
1) anyone with permission can disable updates;
2) only experienced (releng, relman, etc) people can reenable updates.

So the way that I thought to implement that was:
- User disables the updates by product/channels;
- At the same moment of deactivation, an ASAP change is scheduled;
- When all is OK with product/channel, it is necessary only  perform signoff to reenable the updates. This behaviour will work only if the product requires signoffs.

auslib/test/test_db.py
allan-silva: `testModelIsSameAsRepository` tests the column metadata, and generates a log like bellow:

```
E           AssertionError: Schema diffs:
E             table with differences: emergency_shutoffs_scheduled_changes
E               column with differences: complete
E                    model: Column('complete', Boolean(create_constraint=True, name=None), table=<emergency_shutoffs_scheduled_changes>, default=ColumnDefault(False))
E                 database: Column(u'complete', INTEGER(), table=<emergency_shutoffs_scheduled_changes>)

auslib/test/test_db.py:4824: AssertionError
```

Is this the same thing or you mean other test?

auslib/test/admin/views/test_emergency_shutoff.py
allan-silva: The signoffs will be required only to reenable the updates. It is not necessary signoffs to disable updates.

auslib/web/admin/emergency_shutoff.py
mozbhearsum: Two things stand out to me about this:
1) Can it be made to work for product+channels that don't require signoffs (like Firefox nightly) - that's a hard requirement for feature.
2) How will this be presented in the UI? I'm a bit concerned that people may see the re-enabling waiting on signoff, and assume that it's ready to be turned back on.

auslib/test/test_db.py
mozbhearsum: Oh, heh. Sorry! I forgot how these tests worked while reviewing.

auslib/test/admin/views/test_emergency_shutoff.py
mozbhearsum: What I mean is that we must be able to do emergency shutoff for product+channel combinations that _aren't_ in the product_req_signoffs table -- does that make sense?

auslib/web/admin/emergency_shutoff.py
allan-silva: Hmmm that makes sense.
So, we can simplify more:
-Who has the 'create' permission for emergency shutoff, can disable updates.
-Who has the 'delete' permission for emergency shutoff, can reenable updates.

No need to schedule changes nor ask to sign offs. What do you think?

mozbhearsum: For product+channels that don't have signoffs defined - yes (eg: Firefox nightly). For product+channels that do have signoffs defined, re-enabling updates must bgo through multiple signoffs still.

mozbhearsum:
I played around with this a bunch in final review and the API seems to work great. So excited to have this landed!

PR #452, opened 2017-11-30
bug 1384816: Compress production dumps with xz

mozbhearsum:
xz gets the current production dump down to ~20MB (gzip took it to ~40MB), and extracts in less than 2 seconds on my laptop. Compression can taken awhile, but I don't think that's going to be an issue on the server since we run the cronjob at an off-hour?

@relud - there's some background on this in https://bugzilla.mozilla.org/show_bug.cgi?id=1384816. The idea is that we'd start uploading this new, compressed dump to a different location, and then update the dev set-up script to decompress + use that instead of use the uncompressed dump. Is that OK for you? I imagine it may require a change to the cronjob.


relud:
that works for me

mozbhearsum:
Good idea - I'll add a reminder about that.

PR #456, opened 2018-01-10
Move rule matching logic out of the database layer to make it reusable

mozbhearsum:
This is in support of an upcoming PR for https://bugzilla.mozilla.org/show_bug.cgi?id=1400016. It pulls out the per-column rule matching logic from the database layer, and puts it in a more re-usable location. None of the logic has changed, and it never actually touched database operations - so it should be pretty straightforward.


nthomas-mozilla:
nthomas woz here - f+

PR #457, opened 2018-01-10
Proof of concept for in-release what's new page decision making

mozbhearsum:
This is the best idea I've come up with to fix https://bugzilla.mozilla.org/show_bug.cgi?id=1400016. It makes most <update> line properties configurable through the newly added "updateLine" block. It's built on top of #456 to make it easier to do channel/version comparisons. https://github.com/mozbhearsum/balrog/compare/move-rule-matching-logic...mozbhearsum:wnp?expand=1 might be a more useful diff when reviewing - it will show you use PR without the #456 parts.

My original idea was to move all of the <update> line properties into the "updateLine" block, but I reversed course on that after realizing that it's probably an anti-feature to allow appVersion & displayVersion to become conditional (we do allow them to be overridden at the locale level, but that's an ugly hack for Nightly). The binary transparency properties may still belong at the top level too, but I _think_ all of the other ones should go in "updateLine".

I took the opportunity add "type" into the blob. It's not required at the moment, but if we start specifying it in all of the new blobs, we could eventually drop it from the Rules table (in retrospect, I don't think it ever really belonged there, and it's particularly confusing when working with GMP or SystemAddons).

I've done some local testing with this and was able to replace the current 3 rules that point at 57.0.4 (lzma) with a single Rule that points at unified blob with the following at the top level:
```
    "appVersion": "57.0.4",
    "displayVersion": "57.0.4",
    "updateLine": [
        {
            "for": {},
            "fields": {
                "detailsUrl": "https://www.mozilla.org/%LOCALE%/firefox/57.0.4/releasenotes/",
                "update_type": "minor"
            }
        },
        {
            "for": {
                "locales": ["ast", "bg", "bs", "cak", "cs", "cy", "da", "de", "dsb", "en-GB", "en-US", "eo", "es-AR", "es-CL", "es-ES", "es-MX", "et", "fa", "fr", "fy-NL", "hi-IN", "hsb", "hu", "id", "it", "ja", "ja-JP-mac", "ka", "kab", "ko", "lt", "ms", "nb-NO", "nl", "nn-NO", "pa-IN", "pl", "pt-BR", "pt-PT", "ru", "sk", "sl", "sq", "sv-SE", "ta", "tr", "uk", "zh-CN", "zh-TW"],
                "versions": ["<57.0"]
            },
            "fields": {
                "actions": "showURL",
                "openURL": "https://www.mozilla.org/%LOCALE%/firefox/57.0.4/whatsnew/?oldversion=%OLD_VERSION%"
            }
        }
    ],
```


auslib/blobs/apprelease.py
```diff
@@ -929,6 +936,162 @@ def getReferencedReleases(self):
...
+                    else:
+                        if cond == "channels":
+                            for (value1, value2) in itertools.product(group1["for"][cond], group2["for"][cond]):
+                                # Exact match of concrete channel or two globs
+                                if value1 == value2:
+                                    matches = True
+                                    break
+                                # Logical match in either direction, which takes into account
+                                # globbing and fallback channels
+                                elif matchChannel(value1, value2.rstrip("*"), getFallbackChannel(value2.rstrip("*"))):
```
nthomas-mozilla: This would fail in the case of release-cdntest vs release* ?

auslib/blobs/apprelease.py
```diff
@@ -929,6 +936,162 @@ def getReferencedReleases(self):
...
+                                elif matchVersion(value1, strip_operator(value2)):
+                                    matches = True
+                                    break
+                                elif matchVersion(value2, strip_operator(value1)):
+                                    matches = True
+                                    break
+
+                    condition_results.append(matches)
+
+                if all(condition_results):
```
nthomas-mozilla: Feels like all() should be any(), but I didn't get to the tests yet.

nthomas-mozilla:
Looks like a good approach to me. Would be interesting to know if there's much of an increase in time serving a request, hopefully fairly negligible and caching would help.

auslib/blobs/apprelease.py
mozbhearsum: I think that one works:
```
>>> matchChannel("release*", "release-cck-", "release")
True
```

I'll see about fleshing out the tests more, though.

auslib/blobs/apprelease.py
mozbhearsum: I think you're right, and it looks like a gap in the tests right now.

mozbhearsum: Actually, `all` is correct because we have one entry in condition_results for each of (channels, versions, locales), and if any part of any of them matched, we get a True result into the list (and we need each part to be True to have an overall match).

mozbhearsum:
I also looked at potential performance impacts and I don't think this will affect things in any noticable way. To test this, I made a bunch of requests without this patch. The first one took 76s, and the rest took between 5 and 13ms (~7ms on average). With the patch, the first one took 66s and the rest between 5 and 14ms (~8ms on average). My read of this is that the most expensive part is retrieving the blob from the database (that's why the first requests were so expensive), and that the new release blob takes marginally more time to construct a response. It's probably worth running this test again in dev or stage to get a sense of the impact in a deployed environment instead of Docker.

auslib/test/blobs/test_apprelease.py
```diff
@@ -2847,6 +2849,412 @@ def testSchema8OptionalAttributes(self):
...
+        {"versions": [">49.0"]},
+        {"versions": ["<=50.0"]}
+    ),
+    (
+        {"versions": [">=49.0"]},
+        {"versions": ["<=50.0"]}
+    ),
+    (
+        {"versions": ["<=49.0"]},
+        {"versions": ["<=50.0"]}
```
nthomas-mozilla: I was playing around in this space because `decrement_version()` seems like a good place for unexpected things to happen. For example, if the human is sloppy and puts `<49` instead of `<49.0` then it's decremented to 48 instead of 48.99, but in practice `MozillaVersion()` will raise `BadDataError: Version number 49 is invalid` before any comparison. I'm not sure if the there is anything to prevent malformed input but it could be quite useful.

The code performed nicely with tests like `<49.0` with `<48.0.4`, `<49.0.2` with `<49.0`, and `<49.0.1` with `<49.0`.

nthomas-mozilla:
Also, thank you for looking into the perf impact. I'm really looking forward to squashing several rules with this work.

auslib/test/blobs/test_apprelease.py
mozbhearsum: Good find here. Even though we can't find a way to decrement badly right now, it doesn't mean it can't happen in the future (eg: if MozillaVersion ever gets replacde/removed).

What if we tweak increment/decrement to add a .0 if there's only one part to the version? Considering these are only meant to be used with version strings (not numbers in general) it seems like a reasonable safeguard.

nthomas-mozilla:
I think it would also be OK to insist that version conditions use well
formed versions when submitting a blob, and reject any that don't - less
magic behind the scenes then. Either way, lets make sure rules and WNP
require the same thing. Was looking at the docs to see if I could figure
that out and realized we'll need some doc changes for this work.

auslib/test/blobs/test_apprelease.py
nthomas-mozilla: I think it would also be OK to insist that version conditions use well formed versions when submitting a blob, and reject any that don't - less magic behind the scenes then. Either way, lets make sure rules and WNP require the same thing. Was looking at the docs to see if I could figure that out and realized we'll need some doc changes for this work.

mozbhearsum: On IRC, I believe we said that we want to make sure that Rules can't have versions like "4" or "<4" (single part version numbers) to avoid differences between its matching and what we do for updateLine. Turns out we already do that, via the swagger spec! https://screenshots.firefox.com/R7vxOpgKb3Lv34K5/localhost

`version` has a format defined in https://github.com/mozilla/balrog/blob/master/auslib/web/common/swagger/definitions.yml#L222, which is implemented in https://github.com/mozilla/balrog/blob/master/auslib/web/admin/views/validators.py#L58, which attempts to parse incoming versions with MozillaVersion.

I'm adding tests to make sure this doesn't regress, and I'll update the docs.

nthomas-mozilla: Ooh, that's cool! Can we leverage version_validator() by using `type: version` in apprelease-v9.yml at line 69 ? ie blobs are checked on the way into the app.

mozbhearsum: I think they're already checked because .validate() is called by the database layer before insert/update - but it would be good to have the web layer do this (it would be faster, and we'd get a better error message). I'll look into it.

mozbhearsum: ...except that .validate() doesn't do any manual validation against MozillaVersion. I'm pushing an update that enables that via the jsonschema validator.

PR #458, opened 2018-01-20
Fix Bug 1423381 - 'Schedule a duplication' button is missing on rules…

vikasmahato:
… page for signoff-controlled channels

### DIFF
--- a/ui/app/js/controllers/rules_controller.js
+++ b/ui/app/js/controllers/rules_controller.js
@@ -321,6 +321,50 @@ function($scope, $routeParams, $location, $timeout, Rules, Search, $modal, $rout
     });
   };
 
+  $scope.openNewScheduledDuplicateModal = function(rule) {
+
+    var modalInstance = $modal.open({
+      templateUrl: 'rule_scheduled_change_modal.html',
+      controller: 'NewRuleScheduledChangeCtrl',
+      size: 'lg',
+      backdrop: 'static',
+      resolve: {
+        scheduled_changes: function() {
+          return [];
+        },
+        sc: function() {
+          if(rule.scheduled_change !== null && rule.scheduled_change.change_type === "insert") {
+            sc = angular.copy(rule.scheduled_change);
+            delete sc.sc_id;
+            delete sc.sc_data_version;
+            delete sc.when;
+            delete sc.complete;
+            delete sc.change_type;
+
+          }
+          else {
+            sc = angular.copy(rule);
+          }
+          sc["change_type"] = "insert";
+          delete sc.data_version;
+          delete sc.rule_id;
+          delete sc.alias;
+          return sc;
+        },
+        original_row: function() {
+          return null;
+        },
+        signoffRequirements: function() {
+          return $scope.signoffRequirements;
+        },
+      }
+    });
+    modalInstance.result.then(function(sc) {
+      var rule = {"scheduled_change": sc};
+      $scope.rules.push(rule);
+    });
+  };
+
   $scope.openNewScheduledRuleChangeModal = function(rule) {
 
     var modalInstance = $modal.open({
--- a/ui/app/templates/rules.html
+++ b/ui/app/templates/rules.html
@@ -78,6 +78,10 @@
                 ng-click="openEditScheduledRuleChangeModal(rule)">
             Modify Pending Rule
         </button>
+        <button class="btn btn-info btn-xs"
+                ng-click="openNewScheduledDuplicateModal(rule)">
+            Schedule Duplication
+        </button>
         <button ng-show="show_sc && rule.scheduled_change.change_type === 'update'" class="btn btn-warning btn-xs"
                 ng-click="openEditScheduledRuleChangeModal(rule)">
             Modify Pending Update


ui/app/templates/rules.html
```diff
@@ -90,6 +90,10 @@ <h3 class="panel-title">
                 ng-click="openNewScheduledRuleChangeModal(rule)">
             Schedule an Update
         </button>
+        <button ng-show="rule.scheduled_change === null" class="btn btn-warning btn-xs"
```
mozbhearsum: This button doesn't need the ng-show applied to it. We use this for Updates and Deletes because a Rule may only have a single Scheduled Change associated with it. Scheduling a Duplication doesn't result in a Scheduled Change associated with the Rule, so a user should be able to do this regardless of of whether or not the Rule currently has a change scheduled.

A couple of styling nits too:
* This should be btn-info (we use -info for creation, -warning for modification, and -danger for deletes)
* It should be placed before "Schedule an Update"

ui/app/js/controllers/rules_controller.js
```diff
@@ -321,6 +321,36 @@ function($scope, $routeParams, $location, $timeout, Rules, Search, $modal, $rout
...
+      templateUrl: 'rule_scheduled_change_modal.html',
+      controller: 'NewRuleScheduledChangeCtrl',
+      size: 'lg',
+      backdrop: 'static',
+      resolve: {
+        scheduled_changes: function() {
+          return [];
+        },
+        sc: function() {
+          sc = angular.copy(rule);
```
mozbhearsum: This is largely correct, but you need to remove rule_id, alias and data_version from the copied version (rule_id and alias must be unique across all rules, and data_version shouldn't be provided for newly created rules).

ui/app/js/controllers/rules_controller.js
```diff
@@ -321,6 +321,36 @@ function($scope, $routeParams, $location, $timeout, Rules, Search, $modal, $rout
...
+          sc["change_type"] = "insert";
+          return sc;
+        },
+        original_row: function() {
+          return rule;
```
mozbhearsum: This needs a tweak too. `original_row` is meant to be "previous" version of the row, and is used to help style the Scheduled Change better. Because the created Scheduled Change is an insert, it doesn't have a previous version. This should return null, like openNewScheduledRuleModal does.

FirstSafetyFirst:
I'm on it. I'll let you know as soon as I'm done

vikasmahato:
@mozbhearsum  I have updated the pull request. Sorry for the delay!

ui/app/templates/rules.html
```diff
@@ -86,6 +86,10 @@ <h3 class="panel-title">
                 ng-click="openEditScheduledRuleChangeModal(rule)">
             Modify Pending Delete
         </button>
+        <button class="btn btn-info btn-xs"
```
mozbhearsum: I think this button should appear to the left of the "Modify Pending..." button. It looks odd to have it shown after the orange button in one row, and before it in another: https://screenshotscdn.firefoxusercontent.com/images/2d94cf06-bb49-48a7-a606-ff1aac507546.png

ui/app/js/controllers/rules_controller.js
```diff
@@ -321,6 +321,39 @@ function($scope, $routeParams, $location, $timeout, Rules, Search, $modal, $rout
...
+      templateUrl: 'rule_scheduled_change_modal.html',
+      controller: 'NewRuleScheduledChangeCtrl',
+      size: 'lg',
+      backdrop: 'static',
+      resolve: {
+        scheduled_changes: function() {
+          return [];
+        },
+        sc: function() {
+          sc = angular.copy(rule);
```
mozbhearsum: When I tried to duplicate a Rule that only had a Scheduled Change, this ended up not prefilling any data (because `rule.product` et. al is null). It would be good to be able to duplicate pending inserts as well. Can you update this to copy `rule.scheduled_change` instead of `rule` when `rule.scheduled_change.change_type` is `insert`? Scheduled changes have additionals as well - so you'll need to be sure to strip away `sc_id`, `sc_data_version`, `when`, `complete`, and `change_type` in this case.

vikasmahato: Can you give me the rule id of any rule which has only has a  Scheduled Change so that I can test it??

vikasmahato:
I have updated the PR.

ui/app/js/controllers/rules_controller.js
mozbhearsum: There usually isn't any in the sample data - you can create one with the "Schedule a new Rule" button at the top of the Rules page.

vikasmahato:
I have updated the Pull Request

mozbhearsum:
Looks like those failures were Taskcluster infrastructure issues - not anything to do with your code.

ui/app/js/controllers/rules_controller.js
```diff
@@ -321,6 +321,49 @@ function($scope, $routeParams, $location, $timeout, Rules, Search, $modal, $rout
...
+          return [];
+        },
+        sc: function() {
+          if(rule.scheduled_change !== null && rule.scheduled_change.change_type === "insert") {
+            sc = angular.copy(rule.scheduled_change);
+            delete sc.sc_id;
+            delete sc.sc_data_version;
+            delete sc.when;
+            delete sc.complete;
+            delete sc.change_type;
```
mozbhearsum: I think you also need the bits in the "else" block here, to fix the problem noted above.

vikasmahato:
I hv updated the PR

PR #459, opened 2018-01-26
bug 1433448: fix distribution matching logic for update query v6

mozbhearsum:
Looks like we forgot to update the hardcoding in db.py when we added query version 6. This patch reworks that code to futureproof it.

### DIFF
--- a/auslib/test/web/test_client.py
+++ b/auslib/test/web/test_client.py
@@ -229,7 +229,6 @@ class ClientTestBase(ClientTestCommon):
     }
 }
 """))
-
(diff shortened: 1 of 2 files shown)


auslib/test/web/test_client.py
```diff
@@ -656,7 +681,13 @@ def testVersion2Get(self):
 
     def testVersion2GetIgnoresRuleWithDistribution(self):
```
JohanLorenzo: Should we rename the function, now that it doesn't return an empty response anymore?

mozbhearsum: Good idea

PR #462, opened 2018-02-08
Bug 1419590 Slow API endpoint /api/releases

collin5:
**[Before]**

screen shot 2018-02-08 at 13 01 07

**[After]**

screen shot 2018-02-08 at 12 17 52

PR fixes slow endpoint `/api/releases`


auslib/db.py
```diff
@@ -387,7 +387,10 @@ def select(self, where=None, transaction=None, **kwargs):
         if hasattr(where, "keys"):
             where = [getattr(self, k) == v for k, v in where.iteritems()]
 
-        query = self._selectStatement(where=where, **kwargs)
+        if where and isinstance(where[0], list):
```
mozbhearsum: I'm not sure I like this interface change. It's pretty magical to have `select` behave differently based on the type of the first item of the `where` argument. Per the above, this interface change may not be necessary at all.

auslib/db.py
```diff
@@ -1755,21 +1758,44 @@ def __init__(self, db, metadata, dialect):
...
+            # get all rules as a union
+            q_rules = [[self.db.rules.rule_id == rule_id] for row in info for rule_id in row['rule_ids']]
+            all_rules = self.db.rules.select(where=q_rules, transaction=transaction)
+
+            # map rules according to rule_id
+            rules_map = {rule['rule_id']: rule for rule in all_rules}
+
+            rs_cache = {}
+
+            for row in info:
```
mozbhearsum: In my local testing, this block still took about 75% of the execution time (about 600ms). In production, this is likely to mean page load times are still over 10 seconds. Would you mind trying to speed up Rules.getPotentialRequiredSignoffs, too? That can be a done in a follow-up pull request, if you prefer.

auslib/db.py
```diff
@@ -1755,21 +1758,44 @@ def __init__(self, db, metadata, dialect):
...
+            rows.append(row)
+        info = self.getReleaseInfo(name=[row['name'] for row in rows] if batch else rows[0]['name'], transaction=transaction, batch=batch)
+        # Releases do not affect live updates on their own, only the
+        # product+channel combinations specified in Rules that point
+        # to them. We need to find these Rules, and then return _their_
+        # Required Signoffs.
+        if info:
+            # get all rules as a union
+            q_rules = [[self.db.rules.rule_id == rule_id] for row in info for rule_id in row['rule_ids']]
+            all_rules = self.db.rules.select(where=q_rules, transaction=transaction)
```
mozbhearsum: Good idea here - it looks like it reduces the number of times we call `Rules.getPotentialRequiredSignoffs`.

auslib/db.py
```diff
@@ -1802,10 +1828,13 @@ def countReleases(self, transaction=None):
         return count
 
     def getReleaseInfo(self, name=None, product=None, limit=None,
-                       transaction=None, nameOnly=False, name_prefix=None):
+                       transaction=None, nameOnly=False, name_prefix=None, batch=False):
```
mozbhearsum: This would be cleaner/faster if you changed `name` to `names` and used "WHERE IN" instead of a union. You'll need to update a few callers to cope with the change, but I think it's worth it.

auslib/db.py
collin5: Thanks, hadn't thought of this before. Going to try using `WHERE IN` instead 👍

auslib/db.py
collin5: Yeah, I could try fixing this in this PR though I would be more comfortable doing it in a follow-up pull request. Perhaps we could create/file an issue/bug which depends on this PR then have it assigned to me. I could then tackle it once this is merged.

auslib/db.py
collin5: Thanks 👍

auslib/db.py
collin5: Yeah, great Idea. Will do this ASAP. 👍

auslib/db.py
mozbhearsum: Yup, that sounds perfect. If you have access to file bugs, feel free to file it. (If you don't, we should fix that!)

collin5:
@mozbhearsum Thanks for the feedback 👍. I've been able to work on the requested changes.

auslib/db.py
```diff
@@ -1756,21 +1756,40 @@ def __init__(self, db, metadata, dialect):
...
+                for rule_id in row["rule_ids"]:
+                    rule = rules_map[rule_id]
+                    if rule_id in rs_cache:
+                        _rs = rs_cache[rule_id]
+                    else:
+                        _rs = self.db.rules.getPotentialRequiredSignoffs([rule], transaction=transaction)
+                        rs_cache[rule_id] = _rs
+                    rs.extend(_rs)
+                potential_required_signoffs[row['name']] = rs
+        return next(iter(potential_required_signoffs.values())) if len(potential_required_signoffs) is 1 else potential_required_signoffs
```
mozbhearsum: It just clicked with me that we now have two different possible return types for this. One is the list, like all of the other `getPotentialRequiredSignoffs' functions, and the other is a dictionary. This kind of thing makes the code difficult to reason about - if there's no way to improve the query time without changing the return type, this new code is better off in another function. Not needing to support the list return type might let you simplify things a bit too.

Sorry for not catching this part earlier.

auslib/db.py
```diff
@@ -1801,11 +1820,14 @@ def countReleases(self, transaction=None):
...
         return count
 
-    def getReleaseInfo(self, name=None, product=None, limit=None,
+    def getReleaseInfo(self, names=None, product=None, limit=None,
                        transaction=None, nameOnly=False, name_prefix=None):
         where = []
-        if name:
-            where.append(self.name == name)
+        if names:
+            if isinstance(names, list):
```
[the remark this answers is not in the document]
collin5: Great idea, going to do this ASAP

auslib/db.py
```diff
@@ -1756,21 +1756,40 @@ def __init__(self, db, metadata, dialect):
...
+            rows.append(row)
+        info = self.getReleaseInfo(names=[row['name'] for row in rows], transaction=transaction)
+        # Releases do not affect live updates on their own, only the
+        # product+channel combinations specified in Rules that point
+        # to them. We need to find these Rules, and then return _their_
+        # Required Signoffs.
+        if info:
+            # get all rules as a union
+            q_rules = [self.db.rules.rule_id.in_(tuple([rule_id for row in info for rule_id in row['rule_ids']]))]
+            all_rules = self.db.rules.select(where=q_rules, transaction=transaction)
```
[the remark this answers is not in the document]
collin5: Thanks :)

auslib/db.py
collin5: @mozbhearsum No problem.
This is the case coz `getPotentialRequiredSignoffs` is expected to return an array when getting `releaseScheduledChanges` https://github.com/mozilla/balrog/blob/8da414c3762e0e5e87cbe041d3400de76619bd7b/auslib/web/admin/views/scheduled_changes.py#L62 . We could perhaps change this to use dictionaries so we make our return type a dictionary, or create another function. What do you think would be a better approach?

mozbhearsum: It's that, as well as numerous invocations in db.py (like https://github.com/mozilla/balrog/blob/master/auslib/db.py#L1142). Updating them to support the dict return format is a really great idea - let's do that!

collin5: Great! will do that then. Thanks

collin5:
Hi, @mozbhearsum updated PR. Maybe you can take another look. Thanks

auslib/db.py
```diff
@@ -1756,20 +1759,39 @@ def __init__(self, db, metadata, dialect):
         AUSTable.__init__(self, db, dialect, scheduled_changes=True, scheduled_changes_kwargs={"conditions": ["time"]})
 
     def getPotentialRequiredSignoffs(self, affected_rows, transaction=None):
-        potential_required_signoffs = []
+        potential_required_signoffs = {}
```
mozbhearsum: When we talked earlier about changing the return type I was under the impression that you'd be changing it for all instances of `getPotentialRequiredSignoffs`, not just this one. Keeping that return type consistent will avoid the need for ugly hacks like two hunks above this (they can be updated to assume dictionaries instead).

collin5: Oh, now I get it. Sorry about that, I had thought that all we had to do was change `releases.getPotentialRequiredSignoffs` to return dict. Going to work on this ASAP. Thanks.

collin5:
Thanks for the feedback. I've implemented the requested changes. PR should be ready for another review 👍

auslib/db.py
```diff
@@ -1899,7 +1920,7 @@ def insert(self, changed_by, transaction=None, dryrun=False, signoffs=None, **co
             raise PermissionDeniedError("%s is not allowed to create releases for product %s" % (changed_by, columns["product"]))
 
         if not dryrun:
-            potential_required_signoffs = self.getPotentialRequiredSignoffs([columns], transaction=transaction)
+            potential_required_signoffs = next(iter(self.getPotentialRequiredSignoffs([columns], transaction=transaction).values()))
```
mozbhearsum: This class' `getPotentialRequiredSignoffs` may return a dict with multiple values, which means you may miss some of the required signoffs. You should iterate over all of the returned key/value pairs to make sure all of them are taken into account. It would be best to do this for all of the other places where getPotentialRequiredSignoffs is called outside of a loop or comprehension, too.

collin5: Hadn't thought of this. Let me fix this ASAP. Thanks

collin5:
Fixed 👍 . Thanks

mozbhearsum:
Thank you!

mozbhearsum:
Hey @collin5, I just realized this broke a few things on the Releases endpoints. There's a few calls (such as https://github.com/mozilla/balrog/blob/master/auslib/web/admin/views/releases.py#L274) that pass a string for "names", which ends up getting split into a tuple full of one character strings. You can see this in the UI when you try to delete a release, for example.

Can you provide a follow-up PR with a fix?

PR #463, opened 2018-02-14
Fix Bug 1432044 Comma separate fields in rules should strip whitespace from the beginning and end of each entry

vikasmahato:


auslib/web/admin/views/rules.py
```diff
@@ -22,6 +22,11 @@ def process_rule_form(form_data):
     release_names = dbo.releases.getReleaseNames()
 
     mapping_choices = [(item['name'], item['name']) for item in release_names]
+    
+    # Bug 1432044 Remove whitespaces from locale
```
allan-silva: Probably it is not necessary to reference the bug here.

vikasmahato:
@allan-silva  https://bugzilla.mozilla.org/show_bug.cgi?id=1432044 Here @bhearsum  asked to do the changes before calling the database layer in https://github.com/mozilla/balrog/blob/master/auslib/web/admin/views/rules.py

mozbhearsum:
For posterity, the Taskcluster failure is transient - a retriggered version had no issues.

PR #465, opened 2018-03-06
Slow endpoint when looking up Rules.getPotentialRequiredSignoffs

collin5:
PR is a continuation of #462 . Cuts down `/releases` lookup time to about `~ 130ms`

_**[BEFORE]**_
before

_**[AFTER]**_
after


mozbhearsum:
Nice work!

PR #467, opened 2018-03-10
bug_1433146 - add support for lists of distributions

allan-silva:


auslib/test/admin/views/test_rules.py
```diff
@@ -143,20 +143,6 @@ def testVersionMaxFieldLength(self):
         self.assertEquals(ret[0].get("version"), version)
         self.assertEquals(len(ret[0].get("version")), len(version))
 
-    def testNewRulePostJSON(self):
```
allan-silva: I removed this test, because I realize that is the same as `testNewRulePost`.

auslib/test/web/test_client.py
```diff
@@ -731,6 +758,16 @@ def testVersion6GetWithDistribution(self):
         <patch type="complete" URL="http://a.com/y" hashFunction="sha512" hashValue="13" size="12"/>
     </update>
 </updates>
+""")
+
+    def testVersion6GetWithDistributionInList(self):
```
mozbhearsum: Can you add a few more tests? Eg: to make sure substring matching isn't happening, and that a totally different string doesn't match?

auslib/web/common/swagger/definitions.yml
```diff
@@ -274,10 +274,10 @@ definitions:
         example: 'Windows_NT 5.0'
```
mozbhearsum: Please update https://github.com/mozilla/balrog/blob/master/docs/source/database.rst too!

ui/app/templates/rules.html
```diff
@@ -302,7 +302,7 @@ <h5 title="Fallback Mapping" ng-show="rule.fallbackMapping || (show_sc && rule.s
         <rulefield showdiff="!rule_id" showsc="show_sc" rule="rule" fieldname="buildID" fieldtitle="Build ID"></rulefield>
         <rulefield showdiff="!rule_id" showsc="show_sc" rule="rule" fieldname="buildTarget" fieldtitle="Build Target"></rulefield>
         <rulefield showdiff="!rule_id" showsc="show_sc" rule="rule" fieldname="locale" fieldtitle="Locale(s)"></rulefield>
-        <rulefield showdiff="!rule_id" showsc="show_sc" rule="rule" fieldname="distribution" fieldtitle="Distribution"></rulefield>
+        <rulefield showdiff="!rule_id" showsc="show_sc" rule="rule" fieldname="distribution" fieldtitle="Distribution" limittextto="20" help="{{'Distribution list: ' + rule.distribution}}"></rulefield>
```
mozbhearsum: Is it possible to have the entire string shown as a tooltip? Right now, the only way to see it is to open a dialog, which isn't terribly convenient. I think you should do this on rule_scheduled_changes.html, too.

auslib/web/admin/views/rules.py
```diff
@@ -23,10 +23,11 @@ def process_rule_form(form_data):
...
 
-    if 'locale' in form_data:
-        locale = form_data['locale']
-        if locale is not None:
-            form_data['locale'] = ''.join(locale.split())
+    csv_columns = ['locale', 'distribution']
+    for column in csv_columns:
+        column_data = form_data.get(column)
+        if column_data:
+            form_data[column] = ''.join(column_data.split())
```
allan-silva: @mozbhearsum, let me know if you think that we should remove duplicates to save space.

mozbhearsum:
For posterity, the CI failure is a Taskcluster issue, not an issue with this PR.

PR #468, opened 2018-03-13
bug 1431789: fix bug in validation when updateLine is unset in schema 9 blobs

mozbhearsum:
I discovered this while verifying my work to switch submission to V9. When the blob is first submitted, we always end up submitting a version of it with just name, schema_version, and hashFunction before we submit the actual contents -- this makes that possible. The schema doesn't require updateLine to be set, so I think this is a bug regardless of how the submission tools work.


mozbhearsum:
Good idea, will do.

PR #469, opened 2018-03-14
Providing default value to get() to prevent a runtime error

allan-silva:


mozbhearsum:
Thanks for catching this!

PR #535, opened 2018-04-04
Bug 1445236 show an indication when a rule points at a blob with wnp data (updateLine object)

collin5:
Adds an indication to respective rules for `mapping` or `fallbackMapping` with WNP.

screen shot 2018-04-29 at 15 13 50


auslib/blobs/apprelease.py
```diff
@@ -955,6 +971,15 @@ def __init__(self, **kwargs):
         Blob.__init__(self, **kwargs)
         if 'schema_version' not in self.keys():
             self['schema_version'] = 9
+        self['has_wnp'] = self.has_wnp
+
+    @property
+    def has_wnp(self):
```
allan-silva: As a suggestion to not reimplement the entire logic, you can turn `has_wnp` to a simple function with a param "wnp_field":
`has_wnp(self, wnp_field)`

collin5:
@allan-silva Fixed 👍. Thanks

auslib/blobs/apprelease.py
```diff
@@ -75,6 +75,14 @@ def getBuildID(self, platform, locale):
                 raise BadDataError("No buildID for platform '%s'" % (platform))
             return self['platforms'][platform]['buildID']
 
+    def has_wnp(self, field):
+        # check for What's new page for Blobs < v9
+        for group in self.get(field, []):
+            if isinstance(group, dict) and '/whatsnew/' in group['fields'].get('openURL', ''):
```
mozbhearsum: This can be simpler, actually - you just need to look for whether or not "openURL" exists at all. It doesn't matter what the URL is, if it's set, it's a WNP! I think it would be better to have separate implementations of this for V9 and <V9, too. They're different enough that trying to make them all work in one function is pretty ugly.

auslib/blobs/apprelease.py
```diff
@@ -310,6 +318,7 @@ def __init__(self, **kwargs):
         Blob.__init__(self, **kwargs)
         if 'schema_version' not in self.keys():
             self['schema_version'] = 1
+        self['has_wnp'] = self.has_wnp(field='openURL')
```
mozbhearsum: This is _not_ (and shouldn't be) part of the blob data, so it's actually better off as a property or plain function that looks at openURL when it's called. By adding it as part of the blob, there's potential for it to get out of sync with the state of the opneURL field.

auslib/db.py
```diff
@@ -1923,6 +1923,15 @@ def getBlob():
 
         return blob
 
+    def getReleaseBlobs(self, names, transaction=None):
+        dataVersions = self.select(where=[self.name.in_(names)], columns=[self.name, self.data_version], transaction=transaction)
+        blobs = {version['name']: version for version in dataVersions}
+
+        q = self.select(where=[self.name.in_(names)], columns=[self.data], transaction=transaction)
```
mozbhearsum: I _think_ it would be better to use `getReleaseBlob` (https://github.com/mozilla/balrog/blob/99e230f9eef15c29c8cdb9000978eeec46c63a9e/auslib/db.py#L1879) for this, because it will automatically use the in-memory cache.

auslib/web/common/rules.py
```diff
@@ -20,6 +20,10 @@ def get_rules():
             where[field] = request.args[field]
 
     rules = dbo.rules.getOrderedRules(where=where)
+    blobs = dbo.releases.getReleaseBlobs(names=[rule['mapping'] or rule['fallbackMapping'] for rule in rules])
+    for rule in rules:
+        rule.update({'has_wnp': blobs[rule['mapping'] or rule['fallbackMapping']]['data'].get('has_wnp', False)})
```
mozbhearsum: This needs a tweak - we should actually be marking `mapping` and `fallbackMapping` separately (see below for more on that).

ui/app/templates/rules.html
```diff
@@ -287,6 +287,7 @@ <h5 title="Fallback Mapping" ng-show="rule.fallbackMapping || (show_sc && rule.s
             <strong>Note!</strong> Fallback Mapping is unused when Background Rate is 100
           </span>
         </span>
+        <rulefield showdiff="!rule_id" showsc="show_sc" rule="rule" fieldname="has_wnp" fieldtitle="WNP?"></rulefield>
```
mozbhearsum: We need to do this for both `mapping` and `fallbackMapping`. How about some sort of badge that appears next to them? Something like: https://screenshotscdn.firefoxusercontent.com/images/8c7569c2-1e9d-48d0-9d9d-86981d9c13a7.png -- except with a nicer background colour.

collin5: Oops! hadn't noticed that both `mapping` and `fallbackMapping` are shown to the user in cases where also `mapping` exists. Going to fix this. Thanks 👍

collin5: Using badge is also a good idea. What colors do you have in mind? What do you think about blue? I'm not so good with colors 😄

mozbhearsum: Blue seems fine :). We can always change it later if we don't like it!

auslib/blobs/apprelease.py
mozbhearsum: Which part do you think is expensive? Figuring out whether or not a WNP exists on a blob should be cheap. For example, for ReleaseBlobV9, you just need to walk through self.updateLine and look for any "openURL" keys.

The expensive part is retrieving the actual blobs from the database (which we don't currently do when building the Rules page). For now, I think it's OK to just go ahead and do it, and compare the before and after performance. We have an in-memory cache for Releases, and that may end up making this pretty fast. If it causes a significant slow down, we can consider other options.

If you need any other clarification, let's chat on IRC.

auslib/db.py
collin5: Hi @mozbhearsum I've tried using `getReleaseBlob` but this significantly increases response time to `>12000ms`. That's if I loop through a set of both `mappings` and `fallbackMappings` then call `getReleaseBlob` on each. I've also tried getting all blobs in one query and this takes about `~500ms` which is pretty reasonable but this means I would have to modify the `getReleaseBlob` function return type and also omit caching. What do you think about this?

mozbhearsum: Does the response time go up on just the first request, or all subsequent requests, too? Either way, I'm open to grabbing them all in one query. I'd like to chat about the details of that first, though.

Nice investigative work!

collin5: The first request is actually `~3600ms` the rest are `~2000ms`, when testing on local machine.

mozbhearsum: That's still pretty slow, and would be slower in production :(. Sounds like we should pursue your idea for improvement.

collin5: Alright, I'm going to try doing that. Thank you!

collin5:
Hi @mozbhearsum, this should now be ready for another review. Thanks!

auslib/blobs/apprelease.py
```diff
@@ -75,13 +75,22 @@ def getBuildID(self, platform, locale):
                 raise BadDataError("No buildID for platform '%s'" % (platform))
             return self['platforms'][platform]['buildID']
 
+    @property
+    def has_wnp(self):
+        # check for What's new page for Blobs < v9
+        for group in self.get('openURL', []):
```
mozbhearsum: This doesn't appear to be working: https://screenshots.firefox.com/BsG3coeqJOFaphwl/localhost -- most likely because openURL is just a string for <V9, not a list of dicts.

auslib/blobs/schemas/apprelease-v1.yml
```diff
@@ -22,6 +22,9 @@
       description: Blob identifier number
       enum:
         - 1
+    has_wnp:
```
mozbhearsum: This isn't necessary. Remember: has_wnp is not part of the blob schema itself, just part of the JSON response from the web API.

auslib/db.py
```diff
@@ -1876,52 +1876,14 @@ def getReleaseInfo(self, names=None, product=None, limit=None,
...
-            # blob cache ttl, if the blob cache expired prior to retrieving the
-            # data version, the blob version cache would've expired as well.
-            # If we hit one of these cases, we should bring the blob version
-            # cache up to date since we have it.
-            if cached_blob["data_version"] > data_version:
-                cache.put("blob_version", name, data_version)
-            blob = cached_blob["blob"]
-
-        return blob
+    def getReleaseBlobs(self, names, transaction=None):
```
mozbhearsum: The interface of this function is fine, but I'm concerned with the performance implications of completely removing the caching - it will likely cause major issues in production when nightly updates are submitted.

Can you improve this to look in the cache first, and only pull the names that aren't already there? Let me know if you need any guidance here.

auslib/db.py
```diff
@@ -1828,7 +1828,7 @@ def getReleases(self, name=None, product=None, limit=None, transaction=None):
         rows = self.select(columns=[self.name, self.product, self.data_version],
                            where=where, limit=limit, transaction=transaction)
         for row in rows:
-            row["data"] = self.getReleaseBlob(row["name"], transaction)
+            row["data"] = next(iter(self.getReleaseBlobs(names=[row["name"]], transaction=transaction).values()))
```
mozbhearsum: I see this `next(iter(...))` pattern being used a lot where you're replacing `getReleaseBlob` calls. Perhaps it's worth keeping `getReleaseBlob` as a thin wrapper for `getReleaseBlobs` that maintains the old interface?

auslib/web/common/rules.py
```diff
@@ -20,6 +21,25 @@ def get_rules():
...
 
     rules = dbo.rules.getOrderedRules(where=where)
+
+    names = set(chain.from_iterable((rule['mapping'], rule['fallbackMapping']) for rule in rules))
+    blobs = dbo.releases.getReleaseBlobs(names=names)
+
+    for rule in rules:
+        if rule['mapping']:
+            b = blobs.get(rule['mapping'], {})
+            rule['mapping'] = {
```
mozbhearsum: It looks like something weird is happening in the UI because of this change; when I try to edit a Rule I end up with [object Object] for the Mapping and Fallback Mappinig values: https://screenshots.firefox.com/42aqbbyNjv9CN0fc/localhost.

It might be easier if you split this into two properties: has_wnp_mapping, has_wnp_fallbackMapping. I'm open to other solutions, too.

auslib/blobs/apprelease.py
collin5: Could you please send me an example of <V9 blob to test with?

auslib/db.py
collin5: This is a good idea, I think I may have an idea of what you are suggesting. Will reach out if any questions arise.

auslib/blobs/apprelease.py
mozbhearsum: You should have `Firefox-52.0.1esr-build2-WNP` in your local database -- that's a V4 blob with a WNP. `Firefox-57.0.1-build2` is a V4 blob without a WNP.

auslib/web/common/rules.py
mozbhearsum: I'm open to trying it, but please do a lot of testing -- there's a lot of things in the frontend that assume the rule state can be read and written to. In particular, please make sure that creating/updating/delete rules and rule scheduled changes still continues to work, and that the UI displays changes to them correctly with and without a refresh.

collin5: Will do. Thanks!

auslib/blobs/apprelease.py
collin5: Thanks 👍

auslib/db.py
```diff
@@ -1923,6 +1923,25 @@ def getBlob():
...
+            cached_blob = cache.get('blob', name)
+            if cached_blob:
+                blobs.update({name: cached_blob})
+            else:
+                uncached.append(name)
+
+        q = self.select(where=[self.name.in_(tuple(uncached))], columns=[self.name, self.data, self.data_version], transaction=transaction)
+        for row in q:
+# TODO: investigate what's slowing down cache.put
+#            cache.put('blob', row['name'], row['data'])
```
collin5: Putting items in cache seems to be slowing this down, I suspect it's what was slowing down the `getReleaseBlob` when doing it on more than one blobs. Is there a way of putting a `dict/multiple items` in cache at once instead of how we are currently doing it here https://github.com/mozilla/balrog/blob/5e64240e77d50b6e12c8034728b8c71c3974e7e2/auslib/util/cache.py#L75. Tried look at the documentation of `repoze.lru API` (http://docs.repoze.org/lru/api.html) but couldn't find anything helpful.

mozbhearsum: How slow is slow? And is it only slow on the first request, or slow on subsequent requests as well?

I don't see a way to add multiple items in one call :(.

collin5: Yeah it should only be slow on first request, but then retrieving an item from `cache` also tends to be slow if the item exists ( this is not the case if item doesn't exist though ). To give you a clear example. When testing on local, 
 - Request takes `~380ms` when not using cache at all
- `~1371ms` on first request when using cache (putting items in cache)
- Then unexpectedly `~1100ms` on next requests when using cache (retrieving items from cache)
The last call made me do some digging and also found out `cache.get(k,v)` is also slow when entry exists. Not yet sure why though.

collin5: Hmm, ok. Let me think about this

mozbhearsum: Hm, it looks like this was added as footgun protection against a caller putting something into the cache, and then modifying it. I think this is primarily to make sure we don't end up with bad data in the cache when trying to merge blobs together in https://github.com/mozilla/balrog/blob/dd21686a6dbd5b94698f04e518b8d686ad032d95/auslib/db.py#L2012. We could look at disabling the deepcopy for admin if it's truly causing that much performance degradation. Could you try turning it off and see how much of a difference that makes before we go down this road?

Nice detective work!

collin5: Wow! setting this to `False` significantly improves speed from `~2000ms` to `~600ms` on the first request and from `~1300ms` to `~200ms` on subsequent requests with caching.

mozbhearsum: Wow! I'm seeing similar results here. I'd also be curious to know what the difference is in load times of the Rules with and without this patch -- ultimately what we care about is the load time of that page.

It sounds like this is an angle worth pursuing, though. I think we need to dive a bit deeper into whether or not disabling the deepcopys would cause cache poisoning or not. I'm not sure how easy it is, but enabling the cache and verifying its contents in the tests around that code (https://github.com/mozilla/balrog/blob/master/auslib/test/test_db.py#L1399) would be a good start.

collin5:
Hi @mozbhearsum, Sorry I've been unavailable lately. For the issue of `cache poisoning`, I've tried out a few stuff. i.e replacing `deep copy` with something like
```
def deepish_copy(obj):
    if isinstance(obj, list):
        return [deepish_copy(x) for x in obj]
    elif not isinstance(obj, dict):
        return obj
    return {k: deepish_copy(v) for k, v in obj.iteritems()}
```
But this only reduces time to `~600ms` which is still pretty bad if you ask me. One other option I can think of is making the blobs `read-only` then disable deep copy for these. Do you think this would be a good angle to pursue. Thanks!

mozbhearsum:
(I'm sorry this is taking so long to look at - I'm caught up in some other work at the moment.)

collin5:
It's alright. No pressure!

mozbhearsum:
I agree that that's still a pretty slow response time - slow enough that we couldn't live with it in production.

Did you do any tests with the `deepcopy` disabled fully? If that's fast enough, we can think some more about whether or not that's an OK approach (no point in thinking about it if it's still too slow).

I also still think it's worth trying a different LRU cache - the issues we discussed earlier with repoze removing items at unexpected times means that it might be even slower than necessary. We already depend on `functools32`, which has an `lru_cache` available (https://github.com/MiCHiLU/python-functools32, docs for the function are at https://docs.python.org/3.5/library/functools.html#functools.lru_cache). This should be a pretty simple change to https://github.com/mozilla/balrog/blob/master/auslib/util/cache.py - we can drop the timeout for the purposes of this test.

collin5:
I've done some tests,

**With deep copy**
- First request `~1429 ms`
- Subsequent requests `~1088 ms`

**Without deep copy**
- First request `~508 ms`
- Subsequent requests `~111 ms`

Right, going to be changing to a different LRU cache as well 👍

mozbhearsum:
That's a pretty big improvement without deepcopy, that's great! Probably still a bit slower than we would like...hopefully a different LRU cache will perform better.

collin5:
Yeah, going to try out these changes and see.

mozbhearsum:
We're in the process of starting work on a new UI for Balrog. I'm going to close this PR because it's not worth completing at this point. Apologies for any sunk time on this, @collin5.

PR #538, opened 2018-04-08
bug 1452412 - [WIP] Create a new Balrog notification scheme

allan-silva:
This is a WIP bug, the PR were opened to keep informed the interested people and for me to be warned if take a wrong way.

More context in: https://bugzilla.mozilla.org/show_bug.cgi?id=1452412

Bellow the planned model (using rule as example):
mozilla balrog notifier


mozbhearsum:
To make sure I understand, it looks like you're proposing the following:
* We will replace Balrog's direct e-mails with Pulse messages that send the same information (eg: rule changed, release created, scheduled change enacted, etc.)
* Balrog Notifier is a new service that will be written which will listen for all of the Pulse messages that Balrog publishes, and send e-maill for anything we require (how will these be configured? will other types of notifications be supported?)

allan-silva:
Exactly. For this bug we'll support e-mails and amazon mobile messages. 
The main idea is support other notification types. We can call endpoints in others systems, publish in other queues (weird, but we can), there are the possibility to listen events from others systems and tell to balrog that an event ocurred, etc.

For configurations like mail lists, I though to keep in environment variables like balrog.

The information flow will be more clear in my next commit, but basically is:
A Listener will have a list of middlewares, each middleware will have a nofifier. 

E.g.:

Rule Changed <----- Listener calls the middlewares -----> ThrottleMiddleware realizes that the backgroud rate has changed -----> A message is sent via Amazon SNS (bug 1376537).

allan-silva:
Closing for now

PR #564, opened 2018-04-29
[WIP] bug 1303424 - upgrade to latest version of sqlalchemy and sqlalchemy-migrate

allan-silva:
WIP, opened just to follow the Taskcluster tests.

TODO:

- [x] Migrate to latest SA version
- [x] Review SAWarning/UnicodeWarning
- [x] Change mysql driver
- [x] Restablish notifications
- [x] find regression


allan-silva:
@mozbhearsum,
I had no time to cover all endpoints manually yet. But about code, the PR seems ready for review.

auslib/db.py
```diff
@@ -405,7 +406,9 @@ def _insertStatement(self, **columns):
 
            @rtype: sqlalchemy.sql.express.Insert
         """
-        return self.t.insert(values=columns)
+        table_columns = {k: columns[k] for k in columns.keys() if k in self.table.c}
+        unconsumed_columns = {k: columns[k] for k in columns.keys() if k not in table_columns}
```
mozbhearsum: Hm, do we have cases where we end up passing invalid columns for some queries? If we do, we should fix them (in a followup bug, maybe), and eventually raise an error for this.

auslib/db.py
```diff
@@ -720,30 +726,36 @@ def forInsert(self, insertedKeys, columns, changed_by):
...
+        queries.append(query)
         return queries
 
     def forDelete(self, rowData, changed_by):
         """Deletes cause a single row to be created, which only contains the
            primary key data. This represents that the row no longer exists."""
         row = {}
-        for k in rowData:
-            row[str(k)] = rowData[k]
+        table_row_data = {k: rowData[k] for k in rowData.keys() if k in self.table.c}
```
mozbhearsum: Spots like this should probably raise an error eventually too - anytime a caller is trying to do something to a column that doesn't exist, that's a yellow flag.

auslib/test/test_db.py
```diff
@@ -381,17 +382,26 @@ def testUpdateClosesConnectionOnImplicitTransaction(self):
...
-        what = dict(foo=123)
+        what = dict(foo=123, bar=42)
         self.test.update(changed_by='bob', where=where, what=what, old_data_version=1)
         # update adds data_version and id to the query, so we need to add that before comparing
         what["data_version"] = 2
         what["id"] = 1
         self.assertEquals(shared[0], self.test)
         self.assertEquals(shared[1], "UPDATE")
         self.assertEquals(shared[2], "bob")
+        what.pop('bar')
```
mozbhearsum: It's better to avoid deleting things from `what` - the more it changes (especially through a delete), the harder it is to reason about it. Use a different variable, or a literal version of the data structure in the assert below.

requirements.txt
```diff
@@ -21,9 +21,12 @@ jinja2==2.10 \
...
     --hash=sha256:6ff5f3180870836cae40f06fa10419f557208175f13ad7bc26caa77beb1f6e02
-mysql-python==1.2.5 \
-    --hash=sha256:ab22d1322099098730a57fd59d610f60738f95a1cb68dacca2d1c47cb0cbe8ee \
-    --hash=sha256:811040b647e5d5686f84db415efd697e6250008b112b6909ba77ac059e140c74
+mysqlclient==1.3.12 \
+    --hash=sha256:b3b1a7e4468180afb79289b54069d9499242946a4cedf3928cbf6b2a13800016 \
+    --hash=sha256:d56e379c03efad746e84705cbb97401f60d1f98b05e11a27f2d9c2d043936974 \
+    --hash=sha256:371df79d000af56b4e540b7ce2120d1c9afb04b751bfce25a1eb609c50fd10ff \
+    --hash=sha256:1e85e48b167e2af3bb08f273fdbd1ad6401cbe75057fa6513f97387dc7b282dc \
+    --hash=sha256:2d9ec33de39f4d9c64ad7322ede0521d85829ce36a76f9dd3d6ab76a9c8648e5
```
mozbhearsum: Looks like this is a drop-in replacement? That's great!

auslib/db.py
allan-silva: There are columns that targets to other tables. In the last version of SqlAlchemy, these columns results in a warning. After 0.8 the warning was turned to exception.
So I decided to  keep the suport to columns from other tables, to not touch in flows like this:
https://github.com/mozilla/balrog/blob/master/auslib/web/admin/views/scheduled_changes.py#L100
https://github.com/mozilla/balrog/blob/master/auslib/db.py#L1231

`base_what ` has conditions information.

Also, if I try remove these columns, notifications will fail, because `conditions` are returned when we select ScheduledChangesTable.

PR #594, opened 2018-05-19
bug 1457905 - Check if user is known by Balrog on login required operations

allan-silva:
This will affects all operations that requires a logged user.
Even for authorized user that does not keep any permission, this PR prevents request to touch database layer.

### DIFF
--- a/auslib/db.py
+++ b/auslib/db.py
@@ -2398,7 +2398,7 @@ class Permissions(AUSTable):
         roles_list = [r['role'] for r in self.getUserRoles(username, transaction)]
         return role in roles_list
 
-    def isKnowUser(self, username):
+    def isKnownUser(self, username):
         if not username:
             return False
         cache_column = 'username'
@@ -2675,8 +2675,8 @@ class AUSDatabase(object):
                                                                use_tls)
         self.releases.onUpdate = read_only_bleeter
 
-    def isKnowUser(self, username):
-        return self.permissions.isKnowUser(username)
+    def isKnownUser(self, username):
+        return self.permissions.isKnownUser(username)
 
     def isAdmin(self, *args, **kwargs):
         return self.permissions.isAdmin(*args, **kwargs)
--- a/auslib/db.py
+++ b/auslib/db.py
@@ -2401,18 +2401,14 @@ class Permissions(AUSTable):
     def isKnownUser(self, username):
         if not username:
             return False
+
         cache_column = 'username'
-        cache_name = 'users'
-        cache_key = 'usernames'
 
         def user_getter():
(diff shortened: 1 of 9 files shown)


auslib/db.py
```diff
@@ -2395,6 +2398,23 @@ def hasRole(self, username, role, transaction=None):
...
+        cache_column = 'username'
+        cache_name = 'users'
+        cache_key = 'usernames'
+
+        def user_getter():
+            permissions = self.select(columns=[cache_column], distinct=True)
+            return [permission[cache_column] for permission in permissions]
+
+        if cache_name not in cache.caches:
+            cache.make_cache(cache_name, 1, 300)
```
mozbhearsum: It would be better to do this when the app starts up (like the other caches in uwsgi/admin.wsgi) instead of on the fly -- it makes it more clear what caches we have in the app.

auslib/web/admin/views/base.py
```diff
@@ -7,12 +7,18 @@
...
+
 def requirelogin(f):
     def decorated(*args, **kwargs):
         username = request.environ.get('REMOTE_USER', request.environ.get("HTTP_REMOTE_USER"))
         if not username:
-            logging.warning("Login Required")
-            return Response(status=401)
+            log.warning("Login Required")
+            return problem(401, 'Unauthenticated', 'Login Required')
+        elif not dbo.isKnownUser(username):
```
mozbhearsum: I think this is one of those bugs where we may not remember why we added this a year from now - could you add a comment with a brief explanation, and a pointer back to the bug to help us remember? :)

auslib/web/admin/swagger/api.yaml
```diff
@@ -787,6 +787,8 @@ paths:
           description: successfully created a new release
         '400':
           $ref: '#/responses/invalidFormData'
+        '401':
```
mozbhearsum: It seems like we're missing 401s (and 403s for that matter) from a whole bunch of endpoints. Eg:   /releases/{release}/read_only,   /releases/{release}/revisions,   /users/{username}/roles/{role}, and many others - would you mind going through and adding them? No need to add new unit tests for every single endpoint.

allan-silva: Ah, you are right! Indeed I added the new status in endpoints that already had 403.

auslib/web/admin/swagger/api.yaml
```diff
@@ -1384,6 +1388,10 @@ paths:
       responses:
         '400':
           $ref: '#/responses/invalidFormData'
+        '401':
```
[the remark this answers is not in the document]
allan-silva: Should not, these endpoints are not marked with @requirelogin.

PR #597, opened 2018-05-22
Fix rules signoff buttons.

mozbhearsum:
Regression from #584 -- a simple problem with inverted logic :(.

### DIFF
--- a/ui/app/js/controllers/rule_scheduled_changes_controller.js
+++ b/ui/app/js/controllers/rule_scheduled_changes_controller.js
@@ -9,7 +9,6 @@ function($scope, $routeParams, $location, $timeout, Rules, Search, $modal, $rout
   $scope.user_roles = [];
 
   $scope.sc_id = parseInt($routeParams.sc_id, 10);
-  console.log($scope);
 
   Permissions.getCurrentUser()
   .success(function(response) {
--- a/ui/app/js/controllers/rule_scheduled_changes_controller.js
+++ b/ui/app/js/controllers/rule_scheduled_changes_controller.js
@@ -9,6 +9,7 @@ function($scope, $routeParams, $location, $timeout, Rules, Search, $modal, $rout
   $scope.user_roles = [];
 
   $scope.sc_id = parseInt($routeParams.sc_id, 10);
+  console.log($scope);
 
   Permissions.getCurrentUser()
   .success(function(response) {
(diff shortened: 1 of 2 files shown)


ui/app/js/controllers/rule_scheduled_changes_controller.js
```diff
@@ -9,6 +9,7 @@ function($scope, $routeParams, $location, $timeout, Rules, Search, $modal, $rout
   $scope.user_roles = [];
 
   $scope.sc_id = parseInt($routeParams.sc_id, 10);
+  console.log($scope);
```
lundjordan: is this a debug line?

ui/app/templates/rule_scheduled_changes.html
```diff
@@ -54,11 +54,11 @@ <h2>
     <h3 class="panel-title">
       <div style="float: right">
         <i ng-if="scheduled_changes_rules_count && $first && (currentPage == 1)">Current</i>
-        <button ng-show="sc_id && (!isEmpty(sc['required_signoffs']) &&
+        <button ng-show="!sc_id && (!isEmpty(sc['required_signoffs']) &&
```
lundjordan: why would we not want a scheduled change id? Originally we also did not want a [scheduled_changes_rules_count](https://github.com/mozilla/balrog/commit/6c3fa23a9334100ec8804b345f23fcfeb597bae9#diff-6519a85befd711c7aa8c47a82504c865) but I'm not sure why either of those should mean that we want to show a signoff button in that case.

ui/app/js/controllers/rule_scheduled_changes_controller.js
lundjordan: debug line?

ui/app/templates/rule_scheduled_changes.html
mozbhearsum: There's two scheduled change ids in the mix here. `sc_id` comes from $scope, which is null when we're on the history page, and not null when we're on the primary scheduled changes page -- so that's what we want to use to make the decision. `sc.sc_id` is the id of the scheduled change we're currently looking at in the loop - it will always be not null on both pages.

Does that make sense?

lundjordan: makes sense enough to stamp. Thanks for the detailed explanation

mozbhearsum:
The Taskcluster failure is an intermittent issue with npm install. A rerun worked fine.

PR #602, opened 2018-05-24
Add nginx container

mozbhearsum:
While debugging https://bugzilla.mozilla.org/show_bug.cgi?id=1457905 it was very helpful to have a more production-like development environment, one that uses nginx to serve traffic rather than bare uwsgi. We generally try to be as production like as possible in dev, so I think this is a good thing to have going forward. One of the side benefits is that nginx actually prompts for http auth, which makes it much easier to use multiple different usernames in local development (which makes it easier to test anything involving signoffs, where you want to do things as multiple different users).

This PR isn't quite ready to merge yet -- the admin.ini change ends up switching production to using non-minified UI, which is not something we want to do. I'm not sure how to fix this without totally forking admin and dev uwsgi configs.


mozbhearsum:
While beginning to look into Auth0 integration, I discovered that we need SSL support in order to have Auth0 redirect back to a local dev instance, which prompted me to pick this back up. I also managed to solve the minified UI issue with a if-env statement in the uwsgi config.

The nginx config is based on https://github.com/mozilla-services/cloudops-deployment/blob/master/projects/balrog/puppet/modules/balrog/templates/admin.openresty.conf.erb - I pulled in all of the parts that seemed relevant.

This seems to work well for me locally, with the caveat that you need to know to use "balrogadmin" as the username and password (we could possibly hardcode this instead of prompting, although it makes it more difficult to test multiple username scenarios).

docker-compose-py3.yml
```diff
@@ -84,28 +85,46 @@ services:
...
     image: node:0.10
     depends_on:
       balrogadmin-py3:
         condition: service_healthy
-    ports:
-      - "18080:18080"
-    links:
-      - balrogadmin-py3
+    expose:
+      - 16060
```
nthomas-mozilla: Is this `balrogui` still exposed in case it's useful for debugging nginx ?

scripts/nginx.conf
```diff
@@ -0,0 +1,57 @@
+log_format balrog_admin_format
+    'admin $msec "$remote_addr" "$request" $status $request_length $body_bytes_sent '
+    '$remote_user $request_time $upstream_response_time "$http_user_agent" "$gzip_ratio"';
+log_format balrog_agent_format
```
nthomas-mozilla: `balrog_agent_format` doesn't seem to be used, and I don't see a "custom entrypoint for balrog agent" in this file like prod does. Is that because agent is talking directly to balrogadmin-py3 ? Is there a downside to making it talk to nginx to be even more prod-like ?

docker-compose-py3.yml
mozbhearsum: Hm, I can't remember why I did this now (this part was written last year). I suspect it was probably to help me verify some part of the nginx setup - but debugging is definitely enabled by it too. Now that you point it though, it makes me wonder if it's confusing to have both exposed...maybe we should just remove it, and let it stick to background rebuilding?

scripts/nginx.conf
mozbhearsum: Good catch. Let me look into doing that, it seems like it should work OK, and would be beneficial while we're testing the auth0 integration work, come to think of it.

mozbhearsum:
@nthomas-mozilla - I think I've addressed your comments in the latest few commits. I also added a health check to the nginx container, and fixed a spurious error coming from the database container that I noticed while trying to reduce logspam a bit.

docker-compose-py3.yml
```diff
@@ -70,42 +71,68 @@ services:
...
+      - ./scripts/server.crt:/etc/nginx/server.crt
+      - ./scripts/server.key:/etc/nginx/server.key
+      - ./scripts/dhparam.pem:/etc/nginx/dhparam.pem
+    environment:
+      - NGINX_PORT=18080
+      - NGINX_BALROG_AGENT_PORT=18081
+      - BALROG_ADMIN_ROOT=http://balrogadmin-py3:17070
+    command: /bin/bash -c "envsubst '$$NGINX_PORT $$NGINX_BALROG_AGENT_PORT $$BALROG_ADMIN_ROOT' < /etc/nginx/conf.d/http_balrog.conf.template > /etc/nginx/conf.d/http_balrog.conf && nginx -g 'daemon off;'"
+    healthcheck:
+        test: nc -z -v nginx 8080 && nc -z -v nginx 8081
```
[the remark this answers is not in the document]
mozbhearsum: Hm, I thought I caught all of these, but evidently not. I'll fix it up.

PR #608, opened 2018-06-02
Fix build issues

mozbhearsum:
While looking at some PRs I noticed that all Docker builds were failing with errors like:
```
Package libmysqlclient-dev is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source

[91mE: Unable to locate package npm
E: Package 'libmysqlclient-dev' has no installation candidate
```

I'm not sure exactly when (I looked through both the Python and Debian dockerfiles and didn't see anything obvious), but it appears the underlying Debian image we're based on was updated at some point, and caused bustage with npm/node, and the mysql package we require to build our mysql connector. This PR should fix things -- I was able to "docker-compose build --no-cache" to get my local dev enviroment working again, as well as build the production image with a carefully crafted task (https://tools.taskcluster.net/groups/L3rkYJTESYmyhu6p5bKR2A/tasks/L3rkYJTESYmyhu6p5bKR2A/runs/0/artifacts).

The highlights are:
* Switch to libmariadbclient-dev, which now provides `mysql_config`
* Install node from a more official place, which means adding that as a mirror. It's not ideal to be doing this, but. This will go away with the UI rewrite.
* Add a fix for jslint, for reasons I don't fully understand (link to the issue is in comments).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,11 +3,11 @@ FROM python:2.7-slim
 MAINTAINER bhearsum@mozilla.com
 
 # Some versions of the python:2.7 Docker image remove libpcre3, which uwsgi needs for routing support to be enabled.
-# Node and npm are to build the frontend. nodejs-legacy is needed by this version of npm. These will get removed after building.
-# libmysqlclient-dev is required to use SQLAlchemy with MySQL, which we do in production.
+# libmariadbclient-dev is required to use SQLAlchemy with MySQL, which we do in production.
+# mysql-client is needed to import sample data, which we do in dev & stage.
 # xz-utils is needed to compress production database dumps
 RUN apt-get -q update \
-    && apt-get -q --yes install libpcre3 libpcre3-dev libmysqlclient-dev mysql-client xz-utils \
+    && apt-get -q --yes install libpcre3 libpcre3-dev libmariadbclient-dev mysql-client xz-utils \
     && apt-get clean
 
 WORKDIR /app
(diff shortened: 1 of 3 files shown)


nthomas-mozilla:
It looks like python:2.7-slim & 3.6-slim were changed from Debian Jesse to Stretch in the last few days, see also this [docker-library/python's issue](https://github.com/docker-library/python/issues/293). We could change our Dockerfile to use 2.7-slim-jessie/3.6-slim-jessie for an immediate revert of that, although I'm not sure if Jessie is still supported. FWIW, balrogdb's use of mysql:5.7 changed in [February](https://github.com/docker-library/mysql/commit/f279c9cc0d20407a2c4d9465adae6b077dcf1b51).

Does the switch to libmariadbclient-dev open up the possibility of divergence with the mysql packages, even though they are supposed to be broadly equivalent ? It looks like sqlalchemy doesn't depend on the headers files in the package for the three C files it compiles, at least. 

The node change makes me a bit twitchy, can you convince me it's not a sec regression doing the `# curl <script> | bash` trick ? It pulls in a copy of python 2.7.13 too, but I don't think that shadows the 2.7.15 in /usr/local/bin from the original image.

nthomas-mozilla:
Per https://wiki.debian.org/LTS Jessie goes into long term support on June 17, for another couple of years.

mozbhearsum:
Thanks for that digging - I wasn't able to find that myself!

> Does the switch to libmariadbclient-dev open up the possibility of divergence with the mysql packages, even though they are supposed to be broadly equivalent ? It looks like sqlalchemy doesn't depend on the headers files in the package for the three C files it compiles, at least.

It affects "mysqlclient", not SQLAlchemy. That package requires mysql_config to build, and does import headers like "mysql.h" (https://github.com/PyMySQL/mysqlclient-python/blob/master/_mysql.c#L29).

I can imagine that this switch could prevent an issue at some point in the future, it looks like MarioDB strives to be a drop-in replacement though (https://mariadb.com/kb/en/library/mariadb-vs-mysql-compatibility/). Of course that could change!

> The node change makes me a bit twitchy, can you convince me it's not a sec regression doing the # curl <script> | bash trick ? It pulls in a copy of python 2.7.13 too, but I don't think that shadows the 2.7.15 in /usr/local/bin from the original image.

It made me twitchy, too. The truth is that we're pulling in all sorts of stuff from the internet when building these Docker images though - mostly from official Debian mirrors, Pypi, and npm. Doing this means we're adding a new place that we pull from, so I suppose the question is: do we trust the Node Debian mirror any less than the existing things?

> Per https://wiki.debian.org/LTS Jessie goes into long term support on June 17, for another couple of years.

OK, this has convinced me that pinning to Jessie is the right way to go -- at least for now. We need to upgrade at some point, of course, but if we wait until after the new UI is ready we may be able to drop the need for npm in this image.

mozbhearsum:
(I'm going to try closing and re-opening this PR to kick Taskcluster.)

PR #629, opened 2018-06-13
bug 1457894: Re-enable CSRF protection.

mozbhearsum:
After some investigation I discovered a middleware packaged with flask_wtf that makes it pretty easy to re-enable our existing CSRF protection. The only complication is that it only looks in `request.form` for the incoming tokens (which is only used for multipart data), so I had to subclass it to look at `request.json`.

CSRF protection is explicitly disabled in unit tests. I verified this by:
* Double checking that CSRF protection is currently broken by tweaking https://github.com/mozilla/balrog/blob/b8265d01e6b3a633d9d5ccdda2f87c900064f54b/ui/app/js/services/rules_service.js#L21 to set a fake CSRF token, and ensuring I could still update a Rule without an error.
* Applying this code change, and trying another Rule update -- which gave me an error about the CSRF token being invalid
* Removing the UI hack, and verifying that I could update Rules again.


nthomas-mozilla:
I've rerun the failed test [here](https://tools.taskcluster.net/groups/WJC43tctRPeNmJNb7y1GHA/tasks/WJC43tctRPeNmJNb7y1GHA/runs/1) (infra fail rather than tests) but I don't know if it will update the status appropriately on success.

PR #631, opened 2018-06-14
bug 1457895: Use secure, strict samesite, and httponly cookies.

mozbhearsum:
In addition to what was recommended by the audit, it looks like we should set the samesite and httponly flags as well -- I've commented this pretty well, let me know if you have any questions.

@psiinon - can you have a look at this as well? You probably have a deeper understanding of the cookie flags than either of us.


mozbhearsum:
The test failure were was a Taskcluster issue, and succeeded after a rerun.

PR #634, opened 2018-06-15
bug 1457893: stop accepting unicode input to fields that don't support it

mozbhearsum:
This patch should prevent unicode input to all fields, enforced via new custom format that connexion validates. I don't really think this is the ideal solution (we should probably support unicode, because I strongly suspect we'll have a use case for unicode in the update URL at some point), but doing this for now prevents the footgun/attack of DoS by putting unicode into the database.

I think it's mostly straightforward, but I did hit a couple of notable issues:
* Some of our error handling code was using `str()` when constructing some error messages. This causes `UnicodeEncodeError`s when the string contains unicode characters. I switched these to explicit UTF-8 encodes instead.
* Some of `connexion`'s error handling code also has a similar issue. The best way I could come up with to workaround this was to catch the errors and return a proper 400 (without it, it's an ISE 500). I filed https://github.com/zalando/connexion/issues/604 for this, too. I think this can be removed after we upgrade to Python 3, because the default encoding is smarter:
```
Python 2.7.15rc1 (default, Apr 15 2018, 21:51:34) 
[GCC 7.3.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import sys
>>> sys.getdefaultencoding()
'ascii'
```
```
Python 3.6.5 (default, Apr  1 2018, 05:46:30) 
[GCC 7.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import sys
>>> sys.getdefaultencoding()
'utf-8'
```

I didn't write too many tests for this because this is primarily going through connexion code paths. I did a bunch of manual testing though, and got reasonable error messages in those.

I also discovered that we accept non-integer buildids, which I figured I'd fix.

### DIFF
--- a/auslib/util/jsonschema_validators.py
+++ b/auslib/util/jsonschema_validators.py
@@ -4,7 +4,7 @@ from jsonschema.compat import str_types
 import logging
 import operator
 
-from auslib.util.comparison import get_op
+from auslib.util.comparison import get_op, strip_operator
 from auslib.util.versions import MozillaVersion
 
 logger = logging.getLogger(__name__)
@@ -24,6 +24,10 @@ def operator_validator(field_value):
     except TypeError:
         # get_op field returns None if no operator or no match, can't be unpacked
         raise jsonschema.ValidationError("Invalid input for buildID : %s. No Operator or Match found." % field_value)
+    try:
+        int(strip_operator(field_value))
+    except ValueError:
+        raise jsonschema.ValidationError("Invalid input for buildID: must be an integer.")
     return True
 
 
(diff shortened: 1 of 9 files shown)


auslib/util/jsonschema_validators.py
```diff
@@ -24,6 +24,10 @@ def operator_validator(field_value):
     except TypeError:
         # get_op field returns None if no operator or no match, can't be unpacked
         raise jsonschema.ValidationError("Invalid input for buildID : %s. No Operator or Match found." % field_value)
+    try:
+        int(strip_operator(field_value))
+    except ValueError:
+        raise jsonschema.ValidationError("Invalid input for buildID: must be an integer.")
```
nthomas-mozilla: It looks like the the very similar auslib.web.admin.views.forms.operator_validator() is now superceded by connexion, so a cleanup could be in order.

PR #647, opened 2018-06-25
Add hash for hypothesis 3.61.0's py2 wheel

nthomas-mozilla:
This is a followup fix for #645. I think what happened is pyup created a PR and the tests passed based on the tar.gz hash. Some time later a wheel was added to pypi, which future test runs preferred over the tar.gz, but we didn't have the hash included.


nthomas-mozilla:
Tests actually passed on rerun. We're hitting a lot of this at the moment:
`[taskcluster:error] Task was aborted because states could not be created successfully. Error: Error calling 'link' for dind : Error: Failed to initialize dind-service waiting for socket`

Callek:
Related: https://github.com/pyupio/pyup/issues/250

nthomas-mozilla:
Also filed HypothesisWorks/hypothesis#1360

PR #683, opened 2018-07-18
bug 1468638: Mock time in a few more places to ensure consistency.

mozbhearsum:
In #635 I mocked time for any tests using `/releases/history`, which ends up calling `timesince`. However, I missed that `/releases/:release/revisions` also uses it (both of them end up calling `annotateRevisionDifferences` (from https://github.com/mozilla/balrog/blob/8eed08e41b710b92d325438ea76b371e37c688f3/auslib/web/common/history.py#L73), which eventually calls `timesince`).

It's hard to be 100% sure, but I _think_ this will fix the last of the the non-determinism.


nthomas-mozilla:
Perhaps some specific tests for timesince() to avoid the coverage decrease ? I wonder if the whole thing can be replaced with arrow, eg `firstdate.humanize(seconddate)`

mozbhearsum:
> Perhaps some specific tests for timesince() to avoid the coverage decrease ? I wonder if the whole thing can be replaced with arrow, eg firstdate.humanize(seconddate)

This is a great idea. I'll look into it.

mozbhearsum:
That was easy :)

nthomas-mozilla:
Cool! Did you see the `_time_ago` references in auslib/test/admin/views/test_releases.py, which have those strange values like `48 years and 585 months ago`. They're not actually tested but as examples could get updated to `48 years ago`. Otherwise land away.

PR #686, opened 2018-07-19
Force insecure cookies in the agent

mozbhearsum:
I tested this locally by disabling insecure cookies, verifying that the agent didn't work, then applying this patch and verifying that it did.

### DIFF
--- a/agent/balrogagent/client.py
+++ b/agent/balrogagent/client.py
@@ -28,6 +28,11 @@ async def request(api_root, path, method="GET", data={}, headers=default_headers
                 resp.raise_for_status()
                 data["csrf_token"] = resp.headers["X-CSRF-Token"]
 
+        # Workaround for https://bugzilla.mozilla.org/show_bug.cgi?id=1471109
+        # In our deployed environments, the Agent doesn't connect to admin over
+        # https, which means it won't send back the session token by default,
+        # which breaks csrf token validation. Changing the cookies to insecure
+        # will let them be sent back, but it's a horrible back.
         for c in client.cookie_jar:
             c["secure"] = False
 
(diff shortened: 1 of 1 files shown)


mozbhearsum:
That Taskcluster failure looks like an intermittent in the frontend tests - worked fine in a rerun: https://tools.taskcluster.net/groups/ZgFN5j89QfevQ2ttaGHdEg/tasks/H_p5TlZ7QQ2j88Nix9c2Yg/runs/1

agent/balrogagent/client.py
```diff
@@ -28,6 +28,14 @@ def get_url(api_root, path):
                 resp.raise_for_status()
                 data["csrf_token"] = resp.headers["X-CSRF-Token"]
 
+        # Workaround for https://bugzilla.mozilla.org/show_bug.cgi?id=1471109
+        # In our deployed environments, the Agent doesn't connect to admin over
```
allan-silva: @mozbhearsum, for curiosity, in these environments, agent calls admin using ip address or DNS name?

nthomas-mozilla: It's using localhost. We actually run into this check https://github.com/aio-libs/aiohttp/blob/c3c1e3a/aiohttp/cookiejar.py#L197 and never send the cookie back.

agent/balrogagent/client.py
```diff
@@ -28,6 +28,14 @@ def get_url(api_root, path):
...
+        # https, which means it won't send back the session token by default,
+        # which breaks csrf token validation. Changing the cookies to insecure
+        # will let them be sent back, but it's a horrible back.
+        for c in client.cookie_jar:
```
nthomas-mozilla: Perhaps we could feel better about this if we only did it in the production case, where other design provides protection. If someone runs a local agent against dev admin then secure cookies could work normally; also when/if we fix up ssl for local dev. Does that add real value or is it just marginal ?

mozbhearsum: I think it's a good idea. Isolating the awful/insecure-in-some-contexts hack to where we know it's OK to use makes sense.

agent/balrogagent/client.py
mozbhearsum: Yeah, the key here is that it's not over https, so `unsafe` doesn't help us.

PR #696, opened 2018-07-25
Stop using docker-in-docker for test jobs.

mozbhearsum:
This turned out to be pretty simple - I just needed to:
- replace the python-test-runner images with the appropriate one for each test
- install some dependencies with apt-get
- run tox/npm instead of run-tests.sh

It doesn't get rid of the docker-in-docker usage for the image building tasks, because we actually do need Docker to do it!

I tested this with some manually tweaked jobs (I'm not sure the PR will use the new yml or not, we'll see!). They are:
https://tools.taskcluster.net/groups/NgUBXgnrSe2dMTxOgaFqZg/tasks/NgUBXgnrSe2dMTxOgaFqZg/details
https://tools.taskcluster.net/groups/a41RD1BLT36wNpMUMqSqrw/tasks/a41RD1BLT36wNpMUMqSqrw/details
https://tools.taskcluster.net/groups/ZsuCYyD5SQiZP5TN7-4x0w/tasks/ZsuCYyD5SQiZP5TN7-4x0w/details
https://tools.taskcluster.net/groups/GZfc-asiRaKXvFUMBSSXBw/tasks/GZfc-asiRaKXvFUMBSSXBw/details

There seems to be a bit of traction on https://bugzilla.mozilla.org/show_bug.cgi?id=1472913 now, so I'm not sure we want to do this. Or maybe we should do it, and then back it out once dind is more stable.

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -42,7 +42,7 @@ tasks:
         - "-c"
         - >
           apt-get -q --yes update && apt-get -q --yes install git libmysqlclient-dev gcc &&
-          git clone $GITHUB_HEAD_REPO_URL balrog && cd balrog && git checkout $GITHUB_HEAD_BRANCH && pip install -r requirements-tox.ini && tox
+          git clone $GITHUB_HEAD_REPO_URL balrog && cd balrog && git checkout $GITHUB_HEAD_BRANCH && pip install -r requirements-tox.txt && tox
       features:
         taskclusterProxy: true
     extra:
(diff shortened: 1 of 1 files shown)


mozbhearsum:
Looks like the PR ended up using the new .taskcluster.yml, which obviously needed a few fixes.

nthomas-mozilla:
lgtm. We'll just have to be careful to update .taskcluster.yml and Dockerfiles if we need to change deps. As we discussed lets wait and see how long it takes https://github.com/taskcluster/docker-worker/pull/407 to make it prod, and if it helps. That'll probably update docker (via https://github.com/taskcluster/docker-worker/commit/353239ac6cddd16db4c379bae5eaacaa787ebd1d) as well, hopefully for the good but we'll see!

mozbhearsum:
The Taskcluster issue appears to have fixed the problem. I'll close this out, but it's something we can come back to later if necessary.

PR #709, opened 2018-07-31
Parallelise tests 4x since we run on c3/m3/r3.xlarge with 4 vCPU

nthomas-mozilla:
Testing to see if this really is quicker than 10-14 minutes with -n2.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -11,7 +11,7 @@ deps = -rrequirements-test.txt
 
 commands=
     flake8 auslib scripts uwsgi
-    py.test -n2 --cov=auslib --doctest-modules {posargs:auslib}
+    py.test -n4 --cov=auslib --doctest-modules {posargs:auslib}
     coverage run -a scripts/test-rules.py
 
 [testenv:py27]


nthomas-mozilla:
Some rough numbers - on taskcluster it saves ~100s in tox, starting from ~450s, in a total job time of 600-900s. Lots of instances pull a bunch of Debian updates in, and there's some variation depending on the instance family too, so it's no slam dunk.  Locally it drops the tox time from ~520s to ~385s. So probably worth taking.

tox.ini
```diff
@@ -11,7 +11,7 @@ deps = -rrequirements-test.txt
 
 commands=
     flake8 auslib scripts uwsgi
-    py.test -n2 --cov=auslib --doctest-modules {posargs:auslib}
+    py.test -n4 --cov=auslib --doctest-modules {posargs:auslib}
```
allan-silva: There are an option to auto detect the number of cores, can be useful for this case:
https://github.com/pytest-dev/pytest-xdist#speed-up-test-runs-by-sending-tests-to-multiple-cpus

mozbhearsum: It looks like all of the possible instances that these jobs can run on have 4 vCPUs, so `-nauto` would be functionally equivalent. I have a vague memory of -n2 being better (even with 4 CPUs) when we first parallelized, but that could've been on different machinery or something....it's clearly not the case now.

I don't feel strongly about this, I'll leave it up to you, @nthomas-mozilla

nthomas-mozilla: `-nauto` sounds great for automation, but possibly painful for local development if it makes your machine crawl while you wait for tests to run.

And I noticed that I started to get these warnings with -n4:
auslib/test/web/test_client.py::ClientTest::testUnknownQueryStringParametersAreAllowedV4
  hypothesis_temporary_module_3723fa5305bedc3db726bc8b02b3bb0a606e0fbf:1069: HypothesisDeprecationWarning: Test took 248.11ms to run. In future the default deadline setting will be 200ms, which will make this an error. You can set deadline to an explicit value of e.g. 300 to turn tests slower than this into an error, or you can set it to None to disable this check entirely.

So I'm going to close this in favour of [bug 1339532](https://bugzilla.mozilla.org/show_bug.cgi?id=1339532).

PR #735, opened 2018-08-08
bug 1481317: Don't allow beta tags for 5.0+ versions.

mozbhearsum:
This hasn't been tested well enough yet, but it seems to work.


nthomas-mozilla:
f+, although I'd need to go check everywhere we use MozillaVersion to see what it'd expect/allow.

mozbhearsum:
I did some additional testing/reading, and I think this is good. We use MozillaVersion in a few places:
* forms.py (which is dead code AFAICT)
* apprelease.py - for ensuring we don't serve updates older than the incoming version + a few special cases in ReleaseBlobV1
* the "version" jsonschema validator (which I verified is working fine for both normal rules and scheduled changes
* comparison.py - I did some ad-hoc testing around this and everything seems good (no surprise - we're using the same regex groups as the other classes).

PR #761, opened 2018-08-25
Bug 1461406: Show aliases for scheduled changes to rules.

collin5:
Resolves https://bugzilla.mozilla.org/show_bug.cgi?id=1461406


nthomas-mozilla:
That test failure is fallout from #375 and the newly added Python 3.6 tests, and not to do with your patch.

collin5:
Oh! makes sense now. Thank you

collin5:
@mozbhearsum Feel free to take another look at this. Thanks!

PR #792, opened 2018-09-05
Disable response validation for the admin app.

mozbhearsum:
Rail discovered that after https://bugzilla.mozilla.org/show_bug.cgi?id=1481317 landed, the Rule Scheduled Changes page was returning 500s like:
```
{
  "detail": "u'<56.0b8' is not a 'version'\n\nFailed validating 'format' in schema['allOf'][1]['properties']['scheduled_changes']['items']['allOf'][1]['allOf'][4]['properties']['version']:\n    {'description': 'The version requirements that the client must meet in order for the rule to match. Exact matches, comma separated lists (eg: \"36.0,36.0.1\") and comparisons (eg: \"<45.0\") are allowed.\\n',\n     'example': '36.0',\n     'format': 'version',\n     'maxLength': 75,\n     'minLength': 0,\n     'type': ['string', 'null']}\n\nOn instance['scheduled_changes'][233]['version']:\n    u'<56.0b8'",
  "status": 500,
  "title": "Response body does not conform to specification",
  "type": "about:blank"
}
```

This is because we're loading Scheduled Changes history that includes versions with beta numbers in them, and we have response validation enabled on the admin app.

This is definitely the quick fix, and possibly the correct one - I don't think response validation can be disabled for specific endpoints...

### DIFF
--- a/auslib/web/admin/base.py
+++ b/auslib/web/admin/base.py
@@ -32,7 +32,7 @@ validator_map = {
 }
 
 connexion_app = connexion.App(__name__, validator_map=validator_map, debug=False)
-connexion_app.add_api(spec, validate_responses=True, strict_validation=True)
+connexion_app.add_api(spec, strict_validation=True)
 app = connexion_app.app
 sentry = Sentry()
 


allan-silva:
@mozbhearsum, this is not only the ResponseValidator, this is about field format validation. Take a look at the demonstration code bellow:

Here is the current version validator:
```
      version:
        description: >
          The version requirements ...
        type: ["string", "null"]
        format: version   ###
        minLength: 0
        maxLength: 75
        example: '36.0'
```
https://github.com/allan-silva/balrog/blob/847ceefd3feb030f16727136136ce56540ddbef4/auslib/util/jsonschema_validators.py#L34 :
```
@jsonschema.draft4_format_checker.checks(format="version", raises=jsonschema.ValidationError)
def version_validator(field_value):
             ... 
            version = MozillaVersion(operand)
             ...
    return True
```

___
For cases where it necessary support all versions format, other format can be specified:
https://github.com/allan-silva/balrog/blob/847ceefd3feb030f16727136136ce56540ddbef4/auslib/util/jsonschema_validators.py#L65 :

```
@jsonschema.draft4_format_checker.checks(format="mozillaAllVersions", raises=jsonschema.ValidationError)
def mozilla_all_versions(field_value):
        ...
            try:
                version = PostModernMozillaVersion(operand)
                continue
            except ValueError:
                pass
            try:
                version = ModernMozillaVersion(operand)
                continue
            except ValueError:
                pass
            version = AncientMozillaVersion(operand)
        ...
    return True
```

Then changing the format in yaml:
```
      version:
        description: >
          The version requirements ...
        type: ["string", "null"]
        format: mozillaAllVersions    ###
        minLength: 0
        maxLength: 75
        example: '36.0'
```

mozbhearsum:
I think the tricky part there is that we'd need separate version definitions for requests vs responses (whereas right now version is defined in RulesBase, and used for both).

There's no technical reason we couldn't split them, but there may not be value. I still think I'd like to disable Response validation if for no other reason than preventing future issues like this where we change the format of something, but still have data in the old format in the database, and it causes an error.

nthomas-mozilla:
I took a quick look at having different formats for POST/PUT vs GET and it is a bit fiddly with the current schema. Probably could be done if there's enough motivation.

What's the down side of disabling validation on responses ? That we could regress the response format without realizing it ? I see that's the default for connexion.

mozbhearsum:
Yeah, I think that's the basic risk. That's lessened by the fact that we have tests for the APIs (although I don't think there's tests whose primary purpose is adherence to a response format).

PR #798, opened 2018-09-13
bug_1447315 - Fix warnings

allan-silva:
Big part of the warnings from https://taskcluster-artifacts.net/BW9QlMpERfC91iaIEvimmg/0/public/logs/live_backing.log, was fixed in #621.

### DIFF
--- a/auslib/test/admin/views/test_releases.py
+++ b/auslib/test/admin/views/test_releases.py
@@ -948,7 +948,7 @@ class TestReleasesAPI_JSON(ViewTest):
                 "hashValue": "abc",
             }
         }
-        self.assertEquals(got, expected)
+        self.assertEqual(got, expected)
         self.assertEqual(ret.headers['X-Data-Version'], '1')
 
     def testLocalePutNotAllowed(self):
@@ -1040,12 +1040,12 @@ class TestReleasesAPI_JSON(ViewTest):
 }
 """))
 
-        self.assertEquals(ret.status_code, 201, "Status Code: %d, Data: %s" % (ret.status_code, ret.get_data()))
+        self.assertEqual(ret.status_code, 201, "Status Code: %d, Data: %s" % (ret.status_code, ret.get_data()))
         r = dbo.releases.t.select().where(dbo.releases.name == 'new_release').execute().fetchall()
-        self.assertEquals(len(r), 1)
-        self.assertEquals(r[0]['name'], 'new_release')
-        self.assertEquals(r[0]['product'], 'Firefox')
-        self.assertEquals(r[0]['data'], createBlob("""
+        self.assertEqual(len(r), 1)
+        self.assertEqual(r[0]['name'], 'new_release')
+        self.assertEqual(r[0]['product'], 'Firefox')
+        self.assertEqual(r[0]['data'], createBlob("""
 {
     "name": "new_release",
     "hashFunction": "sha512",
@@ -1113,12 +1113,12 @@ class TestReleasesAPI_JSON(ViewTest):
     "actions": "doit"
 }
 """))
-        self.assertEquals(ret.status_code, 200, "Status Code: %d, Data: %s" % (ret.status_code, ret.get_data()))
+        self.assertEqual(ret.status_code, 200, "Status Code: %d, Data: %s" % (ret.status_code, ret.get_data()))
         r = dbo.releases.t.select().where(dbo.releases.name == 'd').execute().fetchall()
-        self.assertEquals(len(r), 1)
-        self.assertEquals(r[0]['name'], 'd')
-        self.assertEquals(r[0]['product'], 'Firefox')
-        self.assertEquals(r[0]['data'], createBlob("""
+        self.assertEqual(len(r), 1)
+        self.assertEqual(r[0]['name'], 'd')
+        self.assertEqual(r[0]['product'], 'Firefox')
+        self.assertEqual(r[0]['data'], createBlob("""
 {
     "name": "d",
     "schema_version": 3,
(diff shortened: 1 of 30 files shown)


tox.ini
```diff
@@ -15,7 +15,7 @@ commands=
 basepython=python2.7
 deps = -rrequirements-test.txt
 commands=
-    py.test -n2 --cov=auslib --doctest-modules {posargs:auslib}
+	py.test -W default::DeprecationWarning -W default::PendingDeprecationWarning -n2 --cov=auslib --doctest-modules {posargs:auslib}
```
mozbhearsum: Can you adjust agent/ and client/ for this, too?

auslib/test/admin/views/test_releases.py
```diff
@@ -438,7 +438,7 @@ def testReleasePostRejectedURL(self):
     def testDeleteRelease(self):
         ret = self._delete("/releases/d", qs=dict(data_version=1))
         self.assertStatusCode(ret, 200)
-        ret = dbo.releases.t.count().where(dbo.releases.name == 'd').execute().first()[0]
+        ret = dbo.releases.count(where=[dbo.releases.name == 'd'])
```
mozbhearsum: I take it this change is to fix warnings such as:
`  /app/.tox/py27/lib/python2.7/site-packages/sqlalchemy/sql/expression.py:1783: SAWarning: The IN-predicate on "rules.rule_id" was invoked with an empty sequence. This results in a contradiction, which nonetheless can be expensive to evaluate.  Consider alternative strategies for improved performance.`

And I'm guessing that using plain `t.count()` is source of them?

allan-silva: This one is not deprecated anymore: https://docs.sqlalchemy.org/en/latest/changelog/migration_12.html?highlight=predicate#the-in-not-in-operator-s-empty-collection-behavior-is-now-configurable-default-expression-simplified

The count fix is about https://docs.sqlalchemy.org/en/rel_1_1/core/selectable.html?highlight=fromclause%20count#sqlalchemy.sql.expression.FromClause.count

tox.ini
allan-silva: Sure

allan-silva:
@mozbhearsum, in my first commit I completely forgot py3 task, the warnings were fixed for our code, but a bunch of warnings that come from sqlalchemy-migrate and other packages still printed in the warning summary: 

https://tools.taskcluster.net/groups/DZ8mS1Z6RRG2k7n8O8vkXQ/tasks/GXdavCVURAunRZJ_KQ5fQQ/runs/0/logs/public%2Flogs%2Flive.log#L3291

I do not touched in client test config, once it does not use py.test, also, I do not found an explicity option to enable deprecation warnings in nosetests. Let me know if you want to change test suite of cliente to py.test.

nthomas-mozilla:
I'll take over the review here. Planning to take a look first thing next week.

nthomas-mozilla:
The balrog changes all look good to me, there's just the py36 warnings from the deps to suppress if we can. It's somewhat confusing that py27 does not have this behaviour so it might have been a change in py36 itself. Probably we need to set up `filterwarnings` in the tox.ini to fix this up.

Of note is that pytest 3.8 has a [changelog entry](https://docs.pytest.org/en/latest/changelog.html#features)
`#2908: DeprecationWarning and PendingDeprecationWarning are now shown by default if no other warning filter is configured. This makes pytest more compliant with PEP-0506. See the docs for more info.`
so the tox.ini changes may no longer be necessary. Configuring `filterwarnings` replaces that, I think, based on [this pytest issue](https://github.com/pytest-dev/pytest/issues/4013). A fix for that has landed for 3.9 but currently unshipped. Perhaps there's a config that'll work both now and after that ships, which suppresses DeprecationWarning and PendingDeprecationWarning in general, but not for Balrog code.

nthomas-mozilla:
FYI, we just updated pytest from 3.7.3 to 3.82 today.

allan-silva:
@nthomas-mozilla, 
Until the next release of pytest, we need set filter for `DeprecationWarning` and `PendingDeprecationWarning` (https://github.com/pytest-dev/pytest/blob/features/changelog/4013.feature.rst).

The only way I found to hide warnings from other libs is using the regex filter:
`ignore:.*inspect.getargspec.* is deprecated.*`.

Here the test log, in which I forced a warning in balrog code:
py27 - https://tools.taskcluster.net/groups/bYoNawiNQD28zGpCgfGXXw/tasks/bYoNawiNQD28zGpCgfGXXw/runs/0/logs/public%2Flogs%2Flive.log

py36 - https://tools.taskcluster.net/groups/bYoNawiNQD28zGpCgfGXXw/tasks/QppSMfIrQFCeoabCw7WWpw/runs/0/logs/public%2Flogs%2Flive.log

tox.ini
```diff
@@ -33,3 +33,8 @@ ignore = E402
 [pytest]
 norecursedirs = .tox .git .hg sandbox build
 python_files = test_*.py
+filterwarnings =
+    default::DeprecationWarning
+    default::PendingDeprecationWarning
+    ignore:.*inspect.getargspec.* is deprecated.*
+    ignore:invalid escape sequence.*
```
nthomas-mozilla: @allan-silva sorry for the delay in getting back to this. We've taken the pytest 3.9 upgrade via pyup so it would be good to rerun the tests and make sure the filters are still set correctly. 

I also wondered if this approach might be helpful in ./tox.ini:
```
filterwarnings =
    ignore
    error:::auslib
    default::DeprecationWarning:auslib
    default::PendingDeprecationWarning:auslib
```
The idea is to ignore all warnings in deps, while in auslib warn about deprecations, and upgrade to error anything else. It would avoid suppressing the named errors in Balrog code, I think.

The other tox.ini would need equivalent changes too ?

allan-silva:
@nthomas-mozilla I updated the tox files to use `filterWarnings` as you suggested, actually, I had no much known the `filterWarnings` match rules.
It was necessary include `ignore::ImportWarning:auslib.db`, once the sql-alchemy-migrate module conflicts with `auslib/migrate` folder (other option is rename the folder). Also, your work on #833 is already applied here.

PR #832, opened 2018-10-31
Scheduled weekly dependency update for week 43

nthomas-mozilla:
Manually created PR as we had a branch without one, because of the recent github outage (presumably).


nthomas-mozilla:
The failing frontend test passed on rerun, it's an intermittent issue.

nthomas-mozilla:
Flake8 and pycodestyle need to be updated in lockstep because of their dependency.

PR #838, opened 2018-11-16
Update python package metadata.

tomprince:


client/setup.py
```diff
@@ -1,13 +1,15 @@
...
 
 setup(
     name="balrogclient",
     version="0.0.4",
     description="Balrog Admin API Client",
-    author="Release Engineers",
-    author_email="release@mozilla.com",
+    author="Mozilla Release Engineers",
+    author_email="release+python@mozilla.com",
+    url="https://github.com/mozilla-releng/balrog",
```
collin5: Shouldn't the url be `https://github.com/mozilla/balrog`?

PR #840, opened 2018-11-23
Use fixtures for current DB metadata

catlee:
This avoids having to run the DB schema migration for each test
fix typo.

also:
* silence coverage warnings

* let pytest choose how many processes to run

### DIFF
--- a/auslib/test/conftest.py
+++ b/auslib/test/conftest.py
@@ -1,0 +1,30 @@
+import pytest
+
+from auslib.global_state import dbo
+
+
+@pytest.fixture(scope='session')
+def db_schema():
+    """This fixture creates a fresh in-memory database, and then runs all the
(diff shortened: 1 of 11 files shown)


auslib/test/conftest.py
```diff
@@ -0,0 +1,15 @@
+import pytest
+
+from auslib.global_state import dbo
+
+
+@pytest.fixture(scope='session')
```
nthomas-mozilla: It might be useful to add some comments here which describe how db_schema caches once for all the tests, avoiding doing a migration each time, while current_db_schema is used in each class setUp call to reuse it on a fresh DB.

nthomas-mozilla:
x-reference https://bugzilla.mozilla.org/show_bug.cgi?id=1339532

PR #843, opened 2018-11-27
Fix exception logging

catlee:
Handle case where exception arguments are not a simple tuple of strings,
e.g. on DB failures we get exceptions with args containing nested
tuples.


catlee:
I hit this locally quite often when my local DB is in a bad state, e.g. if the tables are missing. The exceptions thrown somewhere in the DB stack have arguments that are not simple tuples of strings.

catlee:
The quoting change is to handle differences between python2 and 3. cgi.escape doesn't escape quote characters by default, while html.escape does. Also, with quote=True, cgi.escape only escapes double quotes, not single quotes, while html.escape escapes both types.

When writing the tests for the exceptions with complex arguments, the python3 tests were failing because the output was e.g. "one, (&#x27;two&#x27;, &#x27;three&#x27;)"

PR #854, opened 2019-01-07
Bug 1517743 - support compact xml a couple of Firefox nightlies

nthomas-mozilla:
This is a recovery fix for about half our nightly population after a code change in Firefox broke parsing Balrog's response to update requests. Newlines and 4-space indents cause parsing errors. This would be something we remove again after enough population has migrated onwards, say a month or so.


JohanLorenzo:
I'm sorry, this request fell off my radar. Thanks both for pushing this fix quickly!

PR #858, opened 2019-01-16
Bug 1520540: Build and publish python3 docker images

catlee:


scripts/push-dockerimage.sh
```diff
@@ -2,6 +2,9 @@
 
 set -e
 
+DOCKERFILE="$0"
```
srfraser: Won't `$0` be the name of the script, not the first argument?

catlee: yes! thanks, will fix that up.

Any suggestions for how to test this code?

srfraser: There's https://github.com/sstephenson/bats for bash unit tests, but I'm not sure how we'd go about mocking the external calls it makes.
Do we have a test area to upload the images to?

.taskcluster.yml
```diff
@@ -287,7 +287,117 @@ tasks:
...
+        - "-c"
+        - "git clone -b {{ event.version }} {{ event.head.repo.url }} balrog && cd balrog && scripts/push-dockerimage.sh Dockerfile.py3 {{ event.version }}-py3"
+    extra:
+      github:
+        env: true
+        events:
+          - release
+    routes:
+      - index.project.balrog.releases.latest.app
+      - index.project.balrog.releases.{{ event.version }}.app
```
nthomas-mozilla: Looks like these will stomp on the routes for the py2 release event. I'm not sure that will cause any problems but lets pre-emptively dodge.

Dockerfile.py3
```diff
@@ -0,0 +1,50 @@
...
+COPY auslib/ /app/auslib/
+COPY ui/ /app/ui/
+COPY uwsgi/ /app/uwsgi/
+COPY scripts/manage-db.py scripts/run-batch-deletes.sh scripts/run.sh scripts/reset-stage-db.sh scripts/get-prod-db-dump.py /app/scripts/
+COPY version.json /app/
+
+WORKDIR /app
+
+RUN cd ui && \
+    apt-get -q --yes install nodejs nodejs-legacy npm && \
```
nthomas-mozilla: This looks to be a s/2\.7/3.6/g copy of Dockerfile. I tried to see if it will need any package adjustments by comparing  Dockerfile.dev with Dockerfile.py3.dev but they're structured a little differently from this (eg all apt installs in one go). That said, npm, libfontconfig1, and netcat are removed, while gcc is added. So it's possible we'll no longer need npm here; gcc is probably for compiled python deps, seems unlikely to affect nodejs.

catlee: Don't we still need to build the UI in the py3 image?

nthomas-mozilla: Absolutely.  I think what's happening is that dev is slightly different to prod - it's using a node docker image which will have npm there already, and not having it in Dockerfile.py3.dev was just cleanup. Meanwhile we do have to install npm in our prod image to make sure we can build the UI, so carry on!

nthomas-mozilla:
The front-end test failure on 9e112af is an intermittent fault which reran green at [FVMJGWD_Ryqb5c_WaAIjSg](https://tools.taskcluster.net/groups/FVMJGWD_Ryqb5c_WaAIjSg/tasks/FVMJGWD_Ryqb5c_WaAIjSg/details)

nthomas-mozilla:
Closing & reopening to rerun tests.

nthomas-mozilla:
I think this should be good to merge without setting up a full testing scenario, as it existing tags (and routes) are only extended.

nthomas-mozilla:
Image generation was green on the merge, and dev admin's [__version__ endpoint](https://balrog-admin.dev.mozaws.net/__version__) points to the 'Balrog Docker Image Creation (py2)' [task](https://tools.taskcluster.net/groups/D4wcEQU2SiGUfNJ2VEjT-g/tasks/PZ9FZMnETSGWgplWTsRrwQ/details). The public side is working too. Next step is to look at the py36 images.

PR #875, opened 2019-03-22
Disable hypothesis deadlines

catlee:


mozbhearsum:
Thanks for fixing this!

PR #884, opened 2019-03-28
Stop using Basic Auth

rail:


client/balrogclient/api.py
```diff
@@ -109,18 +97,16 @@ class API(object):
...
     """
     verify = False
-    auth = None
     url_template = None
     prerequest_url_template = None
     url_template_vars = None
 
-    def __init__(self, api_root='https://aus4-admin-dev.allizom.org/api',
-                 auth=None, ca_certs=True, timeout=60, auth0_secrets=None,
+    def __init__(self, auth0_secrets, api_root='https://aus4-admin-dev.allizom.org/api',
```
mozbhearsum: This will be a good way to find out if there are non-balrogscript clients :).

rail:
I tested this with https://github.com/mozilla-releng/balrogscript/pull/52 and it worked as expected. \o/

PR #885, opened 2019-03-28
Remove X-Authorization support from backend and agent.

mozbhearsum:


mozbhearsum:
The agent is being handled in #884; I backed it out of here.

PR #886, opened 2019-04-01
Add button to show access token, and an easy way to copy it.

mozbhearsum:
While @rehandalal was testing Auth0 in Stage it came up that there are some scripts that he and @mythmon run on their laptops to make changes to SystemAddons. @gdestuynder suggested using https://auth0.com/docs/api-auth/tutorials/authorization-code-grant-pkce to support this use case, but I'm not certain that's going to work for our use case (I looked into it, and couldn't see how it would).

In the short term at least, adding an easy way to copy one's access token should be OK, which is what this PR does.


mozbhearsum:
That's a great idea. I think I managed to coax it out of Moment:
image

nthomas-mozilla:
👍

PR #888, opened 2019-04-01
Insert a few permissions, roles, and required signoffs into the database to provide better sample data

mozbhearsum:
@helfi92 requested this when doing some early work on the new Balrog UI, I don't think there's any harm in having these? The two permissions aren't ever going to be used, since nobody can log in as those users (presumably), but they're useful for testing UI around permissions.


mozbhearsum:
They're meant to be different. I figured I'd add a couple to make it easier to eg: test a scheduled update and scheduled delete more easily.

nthomas-mozilla:
I probably wasn't very clear. Bob and Janet have the same permissions, and it sounds like you meant to give them different actions.

mozbhearsum:
Ah, I wasn't too concerned with that actually, just that there was a couple of users to look at if you happen to working on the Permissions UI. But giving them different permissions or actions isn't a bad idea.

PR #890, opened 2019-04-05
bug 1501167: Print extra information about successful merges in certain conditions.

mozbhearsum:
This adds some extra logging messages during merges under very specific conditions. I expect this will cause some log spam with every release, but I think it will be helpful in debugging https://bugzilla.mozilla.org/show_bug.cgi?id=1501167 the next time it happens.

We could also consider limiting by product (ie: to exclude Thunderbird and Devedition) to further limit the spam.

### DIFF
--- a/auslib/blobs/base.py
+++ b/auslib/blobs/base.py
@@ -78,9 +78,17 @@ def merge_dicts(ancestor, left, right):
      * A type mismatch of unicode vs string is OK as long as the text is the same
        (Any other type mismatches result in a failure to merge.)
     """
+    # We can't use "logging" directly, because it ignores our custom code in log.py
+    log = logging.getLogger(__name__)
     result = {}
     dicts = (ancestor, left, right)
     for key in set(key for d in dicts for key in d.keys()):
+        # Extra logging information to help debug https://bugzilla.mozilla.org/show_bug.cgi?id=1501167
+        # This is a very large message, so we limit it as much as possible to reduce spam.
+        if key == "completes" and ancestor.get("appVersion") and "a1" not in ancestor["appVersion"]:
+            log.warning("Ancestor is: %s", ancestor.get(key))
+            log.warning("Left is: %s", left.get(key))
+            log.warning("Right is: %s", right.get(key))
         key_types = set([type(d.get(key)) for d in dicts])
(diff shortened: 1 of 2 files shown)


auslib/blobs/base.py
```diff
@@ -78,9 +78,17 @@ def merge_dicts(ancestor, left, right):
...
        (Any other type mismatches result in a failure to merge.)
     """
+    # We can't use "logging" directly, because it ignores our custom code in log.py
+    log = logging.getLogger(__name__)
     result = {}
     dicts = (ancestor, left, right)
     for key in set(key for d in dicts for key in d.keys()):
+        # Extra logging information to help debug https://bugzilla.mozilla.org/show_bug.cgi?id=1501167
+        # This is a very large message, so we limit it as much as possible to reduce spam.
+        if key == "completes" and ancestor.get("appVersion") and "a1" not in ancestor["appVersion"]:
```
nthomas-mozilla: Could this be further limited by testing for `key in left and key in right` ?

auslib/blobs/base.py
```diff
@@ -78,9 +78,17 @@ def merge_dicts(ancestor, left, right):
...
+            log.warning("Ancestor is: %s", ancestor.get(key))
+            log.warning("Left is: %s", left.get(key))
+            log.warning("Right is: %s", right.get(key))
```
nthomas-mozilla: For brevity, perhaps the left and right could be expressed as diffs against the ancestor.

auslib/blobs/base.py
mozbhearsum: I would think that it would need to be both in left and right for this to happen, but I'm also unclear about the STR :(

auslib/blobs/base.py
mozbhearsum: Hmmm, I think we can get most of what we need from earlier messages (which will contain the same `requestid`. Specifically, the messages around https://github.com/mozilla/balrog/blob/03976a24c3bc3d7ad696e846311356f80e1bdde6/auslib/db.py#L2015 give us the ancestor data version. I think it would be good to print the last two messages (ancestor change id, tip data version) for the success case for now, though. That should give us all the information we need to decode which is ancestor/left/right.

PR #891, opened 2019-04-05
bug 1541581: Loading /rules while logged out causes a 500

mozbhearsum:


mozbhearsum:
Ah, good idea. Let's disable those buttons...

PR #895, opened 2019-04-09
Reformat Balrog code with isort and black

mozbhearsum:
This PR reformats the Balrog code base with isort and black, and sets up CI checks for black. I wanted to also run isort checks in CI, but I couldn't find an isort configuration that was compatible with black. Ideally, black will start to support this natively eventually (https://github.com/ambv/black/issues/333).

I'd also like to add a pre-commit hook (like https://github.com/ambv/black/blob/master/.pre-commit-hooks.yaml), but I don't see any reason why this can't land in the mean time.

Finally, I removed some dead looking scripts rather than reformatting them. @nthomas-mozilla - I'm not sure this patch needs an entire review from you (although feel free if you like), but I wouldn't mind a sanity check on this part.


auslib/log.py
```diff
@@ -53,28 +52,43 @@ class JsonLogFormatter(logging.Formatter):
...
 
     # Map from Python logging to Syslog severity levels
-    SYSLOG_LEVEL_MAP = {
-        50: 2,  # CRITICAL
-        40: 3,  # ERROR
-        30: 4,  # WARNING
-        20: 6,  # INFO
-        10: 7,  # DEBUG
-    }
+    SYSLOG_LEVEL_MAP = {50: 2, 40: 3, 30: 4, 20: 6, 10: 7}  # CRITICAL  # ERROR  # WARNING  # INFO  # DEBUG
```
catlee: this is weird

nthomas-mozilla:
Script removal looks good.

agent/balrogagent/log.py
```diff
@@ -20,28 +20,43 @@ class JsonLogFormatter(logging.Formatter):
...
 
     # Map from Python logging to Syslog severity levels
-    SYSLOG_LEVEL_MAP = {
-        50: 2,  # CRITICAL
-        40: 3,  # ERROR
-        30: 4,  # WARNING
-        20: 6,  # INFO
-        10: 7,  # DEBUG
-    }
+    SYSLOG_LEVEL_MAP = {50: 2, 40: 3, 30: 4, 20: 6, 10: 7}  # CRITICAL  # ERROR  # WARNING  # INFO  # DEBUG
```
catlee: This is an unfortunate transformation. I wonder if using the # fmt: off pragma here would be appropriate? Or introduce constants for CRITICAL, ERROR, etc.? Or just leave it alone? :)

mozbhearsum: Yeah, this is awful. It looks like Black doesn't handle this correctly even with a shorter line length. I filed https://github.com/ambv/black/issues/800 for it. We need to do something in the meantime though - this is horribly unreadable with this change.

catlee: Or use the constants from the logging module?

auslib/migrate/versions/026_rename_system_capabilities.py
```diff
@@ -3,6 +3,8 @@ def upgrade(migrate_engine):
     # Migration disabled prior to code going to production. This was replaced by
     # a two part migration that added instructionSet in one migration, and
     # removed systemRequirements in another.
+
```
catlee: This is a bit strange too. Maybe because the # below are on the first column vs. indented?

auslib/log.py
mozbhearsum: The Agent is a separate app in the repo; it doesn't share any code with Balrog. (Does that make Balrog a monorepo? 🤣)

auslib/migrate/versions/026_rename_system_capabilities.py
mozbhearsum: Yeah, seems likely.

nthomas-mozilla:
I've discovered the changes to `scripts/manage-db.py` result in a runtime error for balrogadmin because auslib can't be imported 
```python scripts/manage-db.py -d mysql://balrogadmin:balrogadmin@balrogdb-py3/balrog upgrade
Traceback (most recent call last):
  File "scripts/manage-db.py", line 11, in <module>
    from auslib.blobs.base import createBlob
ModuleNotFoundError: No module named 'auslib'
```
The import was moved earlier than the path munging needed to find auslib. At face value `scripts/test-rules.py` would also be affected but seems fine (we only run these in py27 backend anyway).

The logging may have also been altered in `uwsgi/admin.wsgi` and `public.wsgi` - see [this comment](https://github.com/mozilla/balrog/commit/e0ef0fc9d4cdb907e3559c93eed4133fcbc6145c#diff-f3b6e69f67d80760299f743d1d4c2c91L28) for why the imports were later.

PR #897, opened 2019-04-11
Fix issues with reordered imports.

mozbhearsum:


mozbhearsum:
D'oh, I missed that part...

PR #901, opened 2019-04-18
[wip] rip out python 2 from Balrog app/agent

mozbhearsum:
I'm leaving Python 2 support in the client for the moment, until we can verify that there are no Python 2 consumers left.

This passes all backend tests locally, but still needs a few more things dealt with:
* Frontend tests no longer work, because the Python 3 dev image doesn't have a useful version of npm. I think we should probably split this out into its own Dockerfile, now that we don't need the ui in the same one as admin for local development.
* Figure out what to do with https://github.com/mozilla/balrog/blob/master/auslib/web/public/client.py#L65

I'm also removing `auslib/web/admin/views/forms.py` which has been long replaced by our jsonschemas. wtforms/flask-wtforms is still needed for CSRF support, but I think we can rip that out in the near future, too.


mozbhearsum:
> Frontend tests no longer work, because the Python 3 dev image doesn't have a useful version of npm. I think we should probably split this out into its own Dockerfile, now that we don't need the ui in the same one as admin for local development.

I ended up moving these tests to the same Node image we use to run the UI in local dev.

> Figure out what to do with https://github.com/mozilla/balrog/blob/master/auslib/web/public/client.py#L65

And for this, I ended up continuing to convert to ascii, which is the only reasonable solution at the moment AFAICT. We need to do unscoped work on the db layer to avoid doing this.

PR #912, opened 2019-05-06
Script to sync releases history to GCS

mozbhearsum:
I've been using this to sync limited amounts of data from https://aus4-admin.mozilla.org/api to a test bucket in GCS, and it appears to work well.

It's not particularly fast on the first pass (my most recent test took ~7min to upload ~500 revisions), but because it can skip over uploaded revisions it should work well enough. A second pass where every upload was skipped took 3min. If it still ends up too slow we can add a bunch more ignores to skip really old releases. I'm open to any other speed up suggestions, too.


mozbhearsum:
(Closing and re-opening to kick Taskcluster.)

scripts/releases-history-to-gcs.py
```diff
@@ -0,0 +1,147 @@
...
+
+    for r in releases:
+        revs_in_gcs = uploads[r]["uploaded"] + uploads[r]["existing"]
+        print("INFO: {}: Found {} existing revisions, uploaded {} new ones".format(r, uploads[r]["existing"], uploads[r]["uploaded"]))
+        if r not in uploads:
+            print("WARNING: {} was found in the Balrog API but does not exist in GCS".format(r))
+        elif releases[r] != revs_in_gcs:
+            print("WARNING: {} has a data version of {} in the Balrog API, but {} revisions exist in GCS".format(r, releases[r], revs_in_gcs))
+
+
```
catlee: put this inside an `if __name__ == '__main__'` block?

PR #913, opened 2019-05-07
Implement disableBITS flag

mozbhearsum:
This flag allows us to disable BITS background downloading remotely, on the off chance that something goes wrong when it rolls with 68.0b1. https://bugzilla.mozilla.org/show_bug.cgi?id=1540191 and https://bugzilla.mozilla.org/show_bug.cgi?id=1546106 have additional background.

### DIFF
--- a/auslib/blobs/schemas/apprelease-v9.yml
+++ b/auslib/blobs/schemas/apprelease-v9.yml
@@ -101,6 +101,9 @@
               promptWaitTime:
                 type: number
                 description: The number of seconds to wait before showing the dialog which prompts users to restart, overrides the in-app preference
+              disableBITS:
+                type: boolean
+                description: Whether or not to disable background downloads with BITS on the client
     # Top level fileUrls are useful primarily for release style builds,
     # where the URLs are predictable and only vary by locale and platform.
     # It"s worth noting that while we normally serve different channels
--- a/auslib/blobs/schemas/apprelease-v9.yml
+++ b/auslib/blobs/schemas/apprelease-v9.yml
@@ -103,6 +103,8 @@
                 description: The number of seconds to wait before showing the dialog which prompts users to restart, overrides the in-app preference
               disableBITS:
                 type: boolean
+                enum:
+                  - true
                 description: Whether or not to disable background downloads with BITS on the client
     # Top level fileUrls are useful primarily for release style builds,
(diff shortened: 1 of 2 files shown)


auslib/blobs/schemas/apprelease-v9.yml
```diff
@@ -101,6 +101,9 @@
               promptWaitTime:
                 type: number
                 description: The number of seconds to wait before showing the dialog which prompts users to restart, overrides the in-app preference
+              disableBITS:
```
MihaiTabara: Nit: is it worth mentioning that despite this being `boolean`, it produces any effects should it's set to `True` only?

auslib/test/blobs/test_apprelease.py
```diff
@@ -3503,6 +3503,82 @@ def testWithActions(self):
...
+    "schema_version": 9,
+    "hashFunction": "sha512",
+    "appVersion": "68.0",
+    "displayVersion": "68.0",
+    "updateLine": [
+        {
+            "for": {},
+            "fields": {
+                "detailsURL": "http://example.org/bitsdetails/%LOCALE%",
+                "disableBITS": true,
```
MihaiTabara: https://bugzilla.mozilla.org/show_bug.cgi?id=1546106#c0 says we should only allow `true`as it doesn't respect a different value of `false`. Based on that I'd assume any other value (including `False`) should produce no effect right? If so, could we add a similar test to this where there's either a different value or `False` to double-check the expected headers stay the same?

auslib/blobs/schemas/apprelease-v9.yml
mozbhearsum: That's a good point actually. I'll add a note.

auslib/test/blobs/test_apprelease.py
mozbhearsum: @MihaiTabara - I ended up only allowing `true` for disableBITS - how's that?

MihaiTabara: LGTM, sorry, I meant to approve it in the first place and have these comments as nits. Didn't mean to block it.

mozbhearsum: No worries. The important part is that this gets in before 68.0b1 ships. Still a bit of time :)

PR #914, opened 2019-05-09
Fix release dumps and cleanup

mozbhearsum:
Release dumps been broken since #902 landed, because we included quotes around an argument that didn't need them (which meant the argument didn't work properly).

While investigating this, I also noticed that our cleanup script sometimes ends up with errors like `"/app/scripts/run-batch-deletes.sh: 34: [: -eq: unexpected operator"`, which appear to be caused by $DELETED not being an integer (for reasons I haven't been able to figure out). I _think_ this will fix them, or at the very least provide some debugging info.

### DIFF
--- a/scripts/run-batch-deletes.sh
+++ b/scripts/run-batch-deletes.sh
@@ -26,7 +26,12 @@ while true; do
 
     printf "Deleting..."
     # DBURI and MAX_AGE should be in the environment
-    DELETED=$(python ./manage-db.py -d "$DBURI" cleanup "$MAX_AGE" 2>/dev/null | awk '/Total/ {print $3}')
+    DELETED=$(python ./manage-db.py -d "$DBURI" cleanup "$MAX_AGE" 2>/dev/null)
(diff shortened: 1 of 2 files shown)


scripts/run-batch-deletes.sh
```diff
@@ -27,10 +27,17 @@ while true; do
     printf "Deleting..."
     # DBURI and MAX_AGE should be in the environment
     DELETED=$(python ./manage-db.py -d "$DBURI" cleanup "$MAX_AGE" 2>/dev/null | awk '/Total/ {print $3}')
+    ORIG_DELETED=$DELETED
+    # Sometimes DELETED ends up not being set for some reason, which breaks
+    # the `expr` expressions below, and gets the script stuck in a loop.
```
catlee: Maybe because manage-db.py failed, or failed to produce the right output?
Should we check $? here?

mozbhearsum: That's possible...the logs didn't make it look that way, but it's hard to be certain. Let's do this!

PR #916, opened 2019-05-10
Add proper Content-Security-Policy on Balrog admin's index page

mozbhearsum:
This adds what I believe to be the tightest manageable CSP for the current Balrog codebase.

@psiinon - I know it was quite awhile ago that we talked about this, but would you mind having a look? I don't think any of the usual Balrog reviewers have as firm of a grasp on this as you do.


ui/app/pages/index.us
```diff
@@ -5,6 +5,18 @@
...
+        * img-src: gravatar & githubusercontent for user avatars
+        * frame-src: *.auth0.com to allow bearer token renewals in the background
+        * connect-src: *.auth0.com to allow us to pull public keys from auth0 to validate the login
+        * script-src: unsafe-{eval,inline} because some of our dependencies require it
+        * style-src: unsafe-inline because some of our dependencies require it. fonts.googleapis.com, because we load a font from it (in a weird way that injects it to the app at runtime)
+        * font-src: fonts.googleapis.com because we load a font from it
+        * frame-ancestors: none because there's no valid reason to embed Balrog in another page
+    -->
+    <meta http-equiv="Content-Security-Policy"
+          content="default-src 'self'; img-src 'self' *.gravatar.com *.githubusercontent.com; frame-src https://*.auth0.com/; connect-src 'self' https://*.auth0.com/; script-src 'self' 'unsafe-eval' 'unsafe-inline'; style-src 'self' 'unsafe-inline' https://fonts.googleapis.com/; font-src 'self' https://fonts.gstatic.com; frame-ancestors: none">
```
tomprince: Looking at [infosec's guideline's](https://infosec.mozilla.org/guidelines/web_security#content-security-policy), suggests using `default-src 'none'`.

psiinon:
I was going to point out the missing frame-ancestors but the PR comment explains that.
However the comment in the code still mentions it, so maybe that should be updated to explain why its not present?
We also recommend a a report-uri pointing to the service's own `/__cspreport__` endpoint
Other than that it looks good.

ui/app/pages/index.us
```diff
@@ -5,6 +5,18 @@
...
+        * img-src: gravatar & githubusercontent for user avatars
+        * frame-src: *.auth0.com to allow bearer token renewals in the background
+        * connect-src: *.auth0.com to allow us to pull public keys from auth0 to validate the login
+        * script-src: unsafe-{eval,inline} because some of our dependencies require it
+        * style-src: unsafe-inline because some of our dependencies require it. fonts.googleapis.com, because we load a font from it (in a weird way that injects it to the app at runtime)
+        * font-src: fonts.googleapis.com because we load a font from it
+        * frame-ancestors: none because there's no valid reason to embed Balrog in another page
+    -->
+    <meta http-equiv="Content-Security-Policy"
+          content="default-src 'none'; img-src 'self' *.gravatar.com *.githubusercontent.com; frame-src https://*.auth0.com/; connect-src 'self' https://*.auth0.com/; script-src 'self' 'unsafe-eval' 'unsafe-inline'; style-src 'self' 'unsafe-inline' https://fonts.googleapis.com/; font-src 'self' https://fonts.gstatic.com">
```
g-k: Assuming there are TLS versions of the image sources I think we could do: `https://*.gravatar.com https://*.githubusercontent.com`

Also, I'd be curious about which dependencies need unsafe scripts sources for future deprecation.

Overall looks good!

g-k:
Also, we're considering deprecating the CSP report recommendation and I landed support for Fx CSP reports in Sentry, so depending on which version @bqbn has deployed we could use that or skip that recommendation for now.

bqbn:
@g-k next version is going to be v9.1.1 unless Sentry releases a newer version before we upgrade.

ui/app/pages/index.us
mozbhearsum: > Assuming there are TLS versions of the image sources I think we could do: https://*.gravatar.com https://*.githubusercontent.com

Hmmm, let me see if this will work for us...I think it's the auth0 code that pulls them, so it may not...

> Also, I'd be curious about which dependencies need unsafe scripts sources for future deprecation.

To be honest, I haven't been able to figure out how to trace it down. However, because we run a super ancient angular and node stack, we have some super ancient dependencies which may be part of it. This UI is going away before EOY though, and I expect we'll be able to lock this down further when this one dies.

tomprince: > > Assuming there are TLS versions of the image sources I think we could do: https://_.gravatar.com https://_.githubusercontent.com
> 
> Hmmm, let me see if this will work for us...I think it's the auth0 code that pulls them, so it may not...

Setting https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/upgrade-insecure-requests might handle that; I'm not sure if that is applied before checking against the allowlist here.

PR #917, opened 2019-05-10
Add release sync script to production dockerfile.

mozbhearsum:
Forgot to do this in #912.


mozbhearsum:
Going to roll this in with a few other enhancements to the script.

PR #921, opened 2019-05-17
Allow a list of allowed CORS origins to be passed to the admin app

mozbhearsum:
We haven't needed this historically because the API & UI have always been on the same origin. While testing the new UI, we realized that we need to set this to alow it to talk to the API. If we end up with it on the same origin later, we can remove this.

### DIFF
--- a/auslib/test/admin/views/base.py
+++ b/auslib/test/admin/views/base.py
@@ -49,6 +49,7 @@ class ViewTest(unittest.TestCase):
         app.config["AUTH_DOMAIN"] = "balrog.test.dev"
         app.config["AUTH_AUDIENCE"] = "balrog test"
         app.config["M2M_ACCOUNT_MAPPING"] = {}
+        app.config["CORS_ORIGINS"] = "*"
         with open(self.version_file, "w+") as f:
             f.write(
                 """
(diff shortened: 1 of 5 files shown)


auslib/test/admin/views/base.py
```diff
@@ -49,6 +49,7 @@ def my_userinfo(*args, **kwargs):
         app.config["AUTH_DOMAIN"] = "balrog.test.dev"
         app.config["AUTH_AUDIENCE"] = "balrog test"
         app.config["M2M_ACCOUNT_MAPPING"] = {}
+        app.config["CORS_ORIGINS"] = "*"
```
JohanLorenzo: Agreed to enable all domains on the test instance 👍

auslib/web/admin/base.py
```diff
@@ -81,6 +81,11 @@ def add_security_headers(response):
     response.headers["X-Frame-Options"] = "DENY"
     response.headers["X-Content-Type-Options"] = "nosniff"
     response.headers["Strict-Transport-Security"] = app.config.get("STRICT_TRANSPORT_SECURITY", "max-age=31536000;")
+    response.headers["Access-Control-Allow-Headers"] = "Authorization"
+    if "*" in app.config["CORS_ORIGINS"]:
+        response.headers["Access-Control-Allow-Origin"] = "*"
```
JohanLorenzo: 👍

PR #922, opened 2019-05-17
Allow filtering required signoffs & scheduled changes on product/channel

mozbhearsum:
This is in service of the new UI, where we'd like to be able to make more specific queries.


auslib/web/admin/views/required_signoffs.py
```diff
@@ -113,6 +113,13 @@ class ProductRequiredSignoffsView(RequiredSignoffsView):
     def __init__(self):
         super(ProductRequiredSignoffsView, self).__init__(dbo.productRequiredSignoffs, ["product", "channel", "role"])
 
+    def get(self):
+        where = {}
```
JohanLorenzo: Nit: A dict comprehension could be used here

```suggestion
        where = {
            param: request.args[param] for param in ("product", "channel") if param in request.args
        }
```

Similar changes can be made below.

mozbhearsum:
No problem, thanks for the review!

PR #924, opened 2019-05-23
Run coverage for python 3 again.

mozbhearsum:
I was surprised to see that we were still submitting to coveralls, it just wasn't reporting here. It looks like coveralls doesn't update PRs/commits unless it actually receives coverage data.


mozbhearsum:
> Coverage has gone up in the meantime, woot.

A great argument for removing more code :)

PR #930, opened 2019-06-03
Print docker digest and change build URL

rail:
According to
https://github.com/mozilla-services/cloudops-infra-deploylib/blob/7d668384e41fbf936af3b706a9a3cce2d10499ff/deploylib/docker.py#L152
the digest of the docker image should be in the log. The digest printed
using the locally built image doesn't produce the same sha256 with the
one pushed to Docker Hub, because Taskcluster uses an older version of
docker.

### DIFF
--- a/scripts/push-dockerimage.sh
+++ b/scripts/push-dockerimage.sh
@@ -31,7 +31,7 @@ cat > version.json <<EOF
     "commit": "${commit}",
     "version": "${version}",
     "source": "https://github.com/mozilla/balrog",
-    "build": "https://tools.taskcluster.net/task-inspector/#${TASK_ID}"
+    "build": "https://tools.taskcluster.net/tasks/${TASK_ID}"
 }
 EOF
 
@@ -45,6 +45,9 @@ for tag in ${tags[*]}; do
     docker tag buildtemp "mozilla/balrog:${tag}"
     echo "Pushing Docker image tagged with ${tag}"
     docker push mozilla/balrog:${tag}
+    # Pull the image to print its digest. cloudops-deploylib verifies the
+    # images comparing their digests to the digests in the logs
+    docker pull mozilla/balrog:${tag}
 done
 
 sha256=$(docker images --no-trunc mozilla/balrog | grep "${tags[0]}" | awk '/^mozilla/ {print $3}')


scripts/push-dockerimage.sh
```diff
@@ -45,6 +45,9 @@ for tag in ${tags[*]}; do
     docker tag buildtemp "mozilla/balrog:${tag}"
     echo "Pushing Docker image tagged with ${tag}"
     docker push mozilla/balrog:${tag}
+    # Pull the image to print its digest. cloudops-deploylib verifies the
+    # images comparing their digests to the digests in the logs
```
mozbhearsum: Can you revise this comment? I think the `pull` is necessary because the digest changes when you push it (?), but this comment doesn't make it clear.

tomprince: Does pulling defeat the purpose of printing the digest in the log? My assumption is that it is printed so we *don't* have to trust docker.

rail: > Does pulling defeat the purpose of printing the digest in the log? My assumption is that it is printed so we _don't_ have to trust docker.

It does. So far I haven't found any better way to find the digest using this old image format... :/

mozbhearsum:
It's worthing noting the current state is that images are not verified, so this patch doesn't regress in any way. We should try to find a way to get it working though.

rail:
Let me investigate it more and keep this PR open as a reminder.

scripts/push-dockerimage.sh
rail: After some investigation it turns out that the digest is something registry specific, so you have to pull or push the image to get the digest. Since we push in the same task, there is no need to pull again, `docker push` prints the digest.

mozbhearsum:
Without any changes to `push-dockerimage.sh`, I'm not sure how this is going to fix validation? If `docker push` prints the digest already, image verification should already work?

rail:
Yes, image validation should work, just need to fix the URL to the new format, it's used in parsing the task ID.

mozbhearsum:
> Yes, image validation should work, just need to fix the URL to the new format, it's used in parsing the task ID.

The old URL still works though AFAICT, so how will this patch help? Deploylib pulls the task id out with:
>     job_key = build_url.split("/")[-1][1:]

Which seems to pull the taskId out OK:
> $ python -c "print('https://tools.taskcluster.net/task-inspector/#GB3uq1QpQiSaOkUdFq779A'.split('/')[-1][1:])"
> GB3uq1QpQiSaOkUdFq779A

And then it generates the log with:
> log_url = "https://taskcluster-artifacts.net/{}/0/public/logs/live_backing.log".format(job_key)

Which works fine:
> $ curl -vL https://taskcluster-artifacts.net/eKZSvGRER-q5wo1LJg4hHg/0/public/logs/live_backing.log
> < HTTP/2 200

PR #934, opened 2019-06-07
Remove releases history table

mozbhearsum:
We ended up disabling the migration for the inital landing to avoid the need to do a restore of the table if something went wrong. We're planning to do the migration on Monday, barring any issues that come up between now and then.


mozbhearsum:
> _**takes deep breath**_ r+

🙃

PR #936, opened 2019-06-08
bug_1492603 - require signoff to change releases from read only -> read write

allan-silva:


auslib/db.py
```diff
@@ -1908,6 +1908,11 @@ def update(self, where, what, changed_by, old_data_version, transaction=None, dr
         current_releases = self.select(where=where, columns=[self.name, self.product, self.read_only], transaction=transaction)
         for current_release in current_releases:
             name = current_release["name"]
+
+            if "read_only" in what and current_release["read_only"] != what["read_only"]:
+                self.change_readonly(where, what["read_only"], changed_by, old_data_version, transaction, dryrun, signoffs)
```
allan-silva: For readonly update, this flow is called when change will be enacted.

auslib/db.py
```diff
@@ -1908,6 +1908,11 @@ def update(self, where, what, changed_by, old_data_version, transaction=None, dr
...
+                continue
```
allan-silva: For readonly update, this flow is called when change will be enacted. So calling the functions that changes the state and ignoring the flow below.

allan-silva:
@mozbhearsum ,The single difference from the master is that to pass a release from rw -> ro, it will not ask by signoffs.

https://github.com/mozilla/balrog/blob/c074a603f96a5f001da09d67f83350b18b474d86/auslib/db.py#L2097-L2100

I assumed the rules to asking by signoffs were the same as for the release data.

So, I think we need a function like `getPotentialRequiredSignoffs`, but for the ro -> rw change.
What do you think if we use the regular `getPotentialRequiredSignoffs` function when the release is pointed by a Rule and another  `getPotentialRequiredSignoffs` like function based in others constraints (just product, maybe?) ?

mozbhearsum:
Sorry for the delay in response.

> 

So, I think we need a function like getPotentialRequiredSignoffs, but for the ro -> rw change. What do you think if we use the regular getPotentialRequiredSignoffs function when the release is pointed by a Rule and another  getPotentialRequiredSignoffs like function based in others constraints (just product, maybe?) ?
--

As we talked about elsewhere, we want signoffs going from rw -> ro, but not the other way around.

Using the same `getPotentialRequiredSignoffs` method makes sense to me though. I believe you were suggesting using the max(role) for each required signoff for the given product, which I think would be OK.

If you need more to move forward for now, just let me know.

allan-silva:
UI SC proposal when the Release is readonly:

image
_____
image
_____
image

allan-silva:
TO-DO:

- [ ] (Backend) Return required signoffs for product.
- [ ] (Front) Add support for required signoffs for product.
- [ ] (Front) Add an indication about RO->RW change in SC page.
- [ ] Functional tests.

allan-silva:
@mozbhearsum, finally this is ready for review again

auslib/web/admin/swagger/api.yaml
```diff
@@ -895,6 +895,33 @@ paths:
         '404':
           $ref: '#/responses/resourceNotFound'
 
+  /releases/{release}/read_only/product/required_signoffs:
```
mozbhearsum: Just to be clear: the reason we want this endpoint (instead of just using /required_signoffs/product), is because we don't want the client to have to parse the entire list of product required signoffs itself?

auslib/web/admin/views/releases.py
```diff
@@ -435,7 +445,11 @@ def _post(self, transaction, changed_by):
             if not what.get("data_version", None):
                 return problem(400, "Bad Request", "Missing field", ext={"exception": "data_version is missing"})
 
-        return super(ReleaseScheduledChangesView, self)._post(what, transaction, changed_by, change_type)
+        try:
+            return super(ReleaseScheduledChangesView, self)._post(what, transaction, changed_by, change_type)
+        except ReadOnlyError as e:
+            msg = f"Failed to schedule change - {e}"
```
mozbhearsum: f strings!!!! :+1:

auslib/web/admin/swagger/api.yaml
allan-silva: The main reason is because this endpoint holds the very specific logic for products that is no associated with a rule, where the "MAX(ROLE_REQUIRED) + all different roles" for product will be considered.

auslib/db.py
```diff
@@ -2098,6 +2096,30 @@ def _proceedIfNotReadOnly(self, name, limit=None, transaction=None):
...
+
+        if not dryrun and not is_readonly:
+
+            def _getPotentialRequiredSignoffs(required_signoffs):
+                return [obj for v in required_signoffs for obj in v]
+
+            potential_required_signoffs = _getPotentialRequiredSignoffs(self.getPotentialRequiredSignoffs([release], transaction=transaction).values())
+            if not potential_required_signoffs:
+                potential_required_signoffs = _getPotentialRequiredSignoffs(
+                    self.getPotentialRequiredSignoffsForProduct(release["product"], transaction=transaction).values()
```
mozbhearsum: I want to make sure I fully understand this section. It looks to me like the process here is:
* We check for regular Required Signoffs (ie: is the Release pointed at by a Rule that requires signoff)
* If it is, we use those signoffs
* Otherwise, use the sum of all of the signoffs for the Release's product

Is that right? Either way, this section could use an explanatory comment, as it's not obvious why would we potentially check for two different types of Required Signoffs.

auslib/db.py
```diff
@@ -2098,6 +2096,30 @@ def _proceedIfNotReadOnly(self, name, limit=None, transaction=None):
...
+        if is_readonly:
+            if not self.db.hasPermission(changed_by, "release_read_only", "set", product, transaction):
+                raise PermissionDeniedError(f"{changed_by} is not allowed to mark {product} products read only")
+        else:
+            if not self.db.hasPermission(changed_by, "release_read_only", "unset", product, transaction):
+                raise PermissionDeniedError(f"{changed_by} is not allowed to mark {product} products read write")
+
+        if not dryrun and not is_readonly:
+
+            def _getPotentialRequiredSignoffs(required_signoffs):
```
mozbhearsum: AFAICT this function just transforms the data in `required_signoffs`. It could use a better name, perhaps relating to what it actually returns.

auslib/db.py
allan-silva: It is a sum of all roles and MAX(signoffs) for them, i.e.:

Product|Channel|Role|Signoffs Required
----------|----------|------|-----------------------
Firefox|Release|relman|1
Firefox|Release|releng|1
Firefox|Beta|releng|2
Firefox|Beta|other_role|1

Signoffs required for product:
relman: 1
other_role: 1
releng: 2

mozbhearsum: Right, yes - sorry. The max for each role, not the sum. But my first bullet (about using regular Required Signoffs) is correct, right? That's what it looks like from reading the code and local testing.

allan-silva: Yeah, that is right. First we look for associated rule, if it is null, we get signoffs for product.

PR #945, opened 2019-06-27
Add more required CORS headers.

mozbhearsum:
While working on the new UI locally I discovered that the CORS work that was done previously was not quite complete enough for the new UI to work. Specifically, the UI uses a value for Content-Type that is not whitelisted, and we also need to allow specific methods. https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#Preflighted_requests has more details on the requirements here.

### DIFF
--- a/auslib/web/admin/base.py
+++ b/auslib/web/admin/base.py
@@ -81,7 +81,8 @@ def add_security_headers(response):
     response.headers["X-Frame-Options"] = "DENY"
     response.headers["X-Content-Type-Options"] = "nosniff"
     response.headers["Strict-Transport-Security"] = app.config.get("STRICT_TRANSPORT_SECURITY", "max-age=31536000;")
-    response.headers["Access-Control-Allow-Headers"] = "Authorization"
+    response.headers["Access-Control-Allow-Headers"] = "Authorization, Content-Type"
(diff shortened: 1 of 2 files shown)


auslib/web/admin/base.py
```diff
@@ -81,7 +81,8 @@ def add_security_headers(response):
     response.headers["X-Frame-Options"] = "DENY"
     response.headers["X-Content-Type-Options"] = "nosniff"
     response.headers["Strict-Transport-Security"] = app.config.get("STRICT_TRANSPORT_SECURITY", "max-age=31536000;")
-    response.headers["Access-Control-Allow-Headers"] = "Authorization"
+    response.headers["Access-Control-Allow-Headers"] = "Authorization, Content-Type"
+    response.headers["Access-Control-Allow-Methods"] = "OPTIONS, GET, POST, PUT, DELETE"
```
nthomas-mozilla: Is it OK to set `Access-Control-Allow-Headers` and `Access-Control-Allow-Methods` but maybe not set `Access-Control-Allow-Origin` ?

mozbhearsum: I believe it's valid, but any response that doesn't contain `Access-Control-Allow-Origin` will fail the CORS preflight check (which is what we want if it's coming from an invalid origin, of course).

PR #946, opened 2019-06-27
Make CSRF tokens optional.

mozbhearsum:
I spoke to @gdestuynder about whether or not we still need CSRF protection awhile back, and he told me that as long as we have no deep links in the UI that can change things on the backend, that we're OK to remove it. This is true for both the current UI, and the new UI.

Step 1 is to make it optional, and once we know that all clients have stopped sending it, we can finish ripping it out.


gdestuynder:
For reference in case this merge comes up:

This is my official stance on this: https://infosec.mozilla.org/guidelines/iam/openid_connect#other-important-security-considerations (I wrote this section after we talked with @bhearsum and @jvehent  a while back :)

TLDR: have to be careful though this can be done safely, in general I'd just go through that link and make sure we understand the trade-offs.

PR #950, opened 2019-06-28
Add support for querying Rule & Release scheduled changes by rule_id/release name

mozbhearsum:
This is in aid of the new UI work. Because editing happens on a separate page, which doesn't have access to the full list of rules/releases/scheduled changes from the primary list view, it needs to be able to find any active scheduled change itself. Rather then grabbing the entire list, this endpoint lets it check for just what it needs.

I don't think we'll need similar enhancements for permissions or required signoffs, because they rarely have any changes scheduled, so loading all of the active scheduled changes is comparatively quick

cc @helfi92


auslib/web/admin/views/releases.py
```diff
@@ -395,6 +395,14 @@ class ReleaseScheduledChangesView(ScheduledChangesView):
     def __init__(self):
         super(ReleaseScheduledChangesView, self).__init__("releases", dbo.releases)
 
+    def get(self):
+        where = {}
+        name = connexion.request.args.get("name")
```
JohanLorenzo: I love seeing this endpoint exposed \o/

auslib/web/admin/views/rules.py
```diff
@@ -198,6 +198,14 @@ class RuleScheduledChangesView(ScheduledChangesView):
     def __init__(self):
         super(RuleScheduledChangesView, self).__init__("rules", dbo.rules)
 
+    def get(self):
+        where = {}
+        rule_id = connexion.request.args.get("rule_id")
```
JohanLorenzo: Same here 😃

PR #952, opened 2019-07-02
Unpin jsonschema, connexion, swagger-spec-validator now that spec-synthase doesn't pin them.

mozbhearsum:
I tested this a bit locally and everything seems in order.


mozbhearsum:
(Not sure why I didn't hit that dependency failure locally....)

mozbhearsum:
OK, this is more complicated than I expected...I'll come back to this later.

mozbhearsum:
Alright, I think I've got this sorted out. There were a few issues that I had to fix:
* Update some `connexion` API usage
* Fix `required` in a ton of endpoints
* Drop basic auth validation in admin spec (it's long gone).

I believe that most of the Swagger spec issues were caused (noticed?) by Connexion switching to a new validation library, so they were probably real issues anyways.

Additionally, I included a couple of other small changes:
* Fix the volume name for tox tests (it was `tox-`, after we dropped python 2.7 support).
* Drop `csrf_token` being included in admin tests, because it is now optional, and deprecated.

I did some sanity checking with the admin interface locally and everything seems to be OK, but I'd also like to run a staging release with this before it hits production.

mozbhearsum:
Thanks! I'm going to hold off merging this until I'm back from PTO -- I don't want this to get in the way if we have to do any follow-up deployments for the Autograph/Guardian work.

PR #953, opened 2019-07-03
bug 1563323: Fix bustage from auth0-python library release

mozbhearsum:
Grabbing extra userinfo was busted by https://github.com/auth0/auth0-python/issues/200.

I reworked how we are detecting machine-to-machine tokens at the same time, because just removing `json.loads` caused a new exception to show up. I think the new way is more reliable, anyways.

I also removed the general `Exception` catching, because it eats the real error, and makes debugging more difficult.

### DIFF
--- a/auslib/util/auth.py
+++ b/auslib/util/auth.py
@@ -51,7 +51,6 @@ def get_jwks(auth_domain):
 
 @lru_cache(2048)
 def get_additional_userinfo(auth_domain, access_token):
-    print(auth0_Users(auth_domain).userinfo(access_token))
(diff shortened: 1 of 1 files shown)


auslib/util/auth.py
```diff
@@ -53,7 +51,8 @@ def get_jwks(auth_domain):
 
 @lru_cache(2048)
 def get_additional_userinfo(auth_domain, access_token):
-    return json.loads(auth0_Users(auth_domain).userinfo(access_token))
+    print(auth0_Users(auth_domain).userinfo(access_token))
```
allan-silva: Probably this is for debug purposes.

PR #954, opened 2019-07-05
Elided long log lines in the client.

tomprince:
This was triggering mozilla-releng/scriptworker#362, when rerunning a submit-toplevel task after all the per-build tasks had run and populated the blog.


client/balrogclient/test/test_balrog_api.py
```diff
@@ -0,0 +1,42 @@
+# -*- coding: utf-8 -*-
```
mozbhearsum: Thanks for these!

PR #958, opened 2019-07-13
Bug 1433466 - Added better validation for memory field

vikasmahato:
Hi @bhearsum ,

I have added integer validation to the memory field but it doesn't support the '+' operator yet. Any suggestions on the best way to handle it?

### DIFF
--- a/auslib/web/common/swagger/definitions.yml
+++ b/auslib/web/common/swagger/definitions.yml
@@ -338,6 +338,7 @@ definitions:
         format: ascii
         minLength: 0
         maxLength: 100
+        pattern: "(^[<>0-9][=]?[0-9]*\\d$)"
         example: '<8000'
 
       fallbackMapping:
--- a/auslib/test/admin/views/test_rules.py
+++ b/auslib/test/admin/views/test_rules.py
@@ -149,6 +149,13 @@ class TestRulesAPI_JSON(ViewTest):
         self.assertEqual(r[0]["memory"], "<7373")
         self.assertEqual(r[0]["data_version"], 1)
 
+    def testCreateRuleWithInvalidMemory(self):
+        for invalid in ("<", "=", "=12", "=>896", "=<896", "abc"):
+            ret = self._post(
+                "/rules", data=dict(backgroundRate=33, mapping="c", priority=0, memory=invalid, product="Firefox", update_type="minor", channel="nightly")
+            )
+        self.assertEqual(ret.status_code, 400, "does not match")
+
     def testCreateRuleWithMig64(self):
         ret = self._post("/rules", data=dict(backgroundRate=33, mapping="c", priority=0, mig64=True, product="Firefox", update_type="minor", channel="nightly"))
         rule_id = int(ret.get_data())


mozbhearsum:
> I have added integer validation to the memory field but it doesn't support the '+' operator yet. Any suggestions on the best way to handle it?

I think there's a bit of confusion based on this comment from the bug:
> We should only accept integers, or integer + operator in this field.

We don't want to support a `+` operator. We want to support either an integer or an integer and an operator. For example, these are all valid values for the memory field:
```
4096
<2048
<=2048
>3184
>=3184
```

And we need this validation to happen in the backend, which is managed by the Swagger specification here: https://github.com/mozilla/balrog/blob/master/auslib/web/common/swagger/definitions.yml#L335

You'll need to add a `pattern` attribute to memory that only allows the above types of strings.

vikasmahato:
Hi,

I wanted to know from where is https://github.com/mozilla/balrog/blob/master/auslib/util/jsonschema_validators.py#36 is called. I could not find any other references to it other than the function definition.

allan-silva:
@vikasmahato 
Internally, jsonschema make calls to validators. So you should specify your own validator function and anotate it with `@jsonschema.draft4_format_checker.checks`. The format name (on the swagger file) will be the same defined by you on annotation.

However, it is possible to define a Regex pattern like as defined on `alias` field (https://github.com/mozilla/balrog/blob/0f16b4e1463e6875519601b39e20c31fea2ef490/auslib/web/common/swagger/definitions.yml#L323).

auslib/web/common/swagger/definitions.yml
```diff
@@ -338,6 +338,7 @@ definitions:
         format: ascii
         minLength: 0
         maxLength: 100
+        pattern: "(^[<>]?[=]?[0-9]*$)"
```
mozbhearsum: This is close, but it still allows a few values it shouldn't, such as:
* =12
* <
* =

We should only allow plain numbers, or plain numbers prefixed by `<`, `>`, `<=`, or `>=`. A test such as:
```
diff --git a/auslib/test/admin/views/test_rules.py b/auslib/test/admin/views/test_rules.py
index 74643e05..25047abe 100644
--- a/auslib/test/admin/views/test_rules.py
+++ b/auslib/test/admin/views/test_rules.py
@@ -144,16 +144,23 @@ class TestRulesAPI_JSON(ViewTest):
         r = dbo.rules.t.select().where(dbo.rules.rule_id == rule_id).execute().fetchall()
         self.assertEqual(len(r), 1)
         self.assertEqual(r[0]["mapping"], "c")
         self.assertEqual(r[0]["backgroundRate"], 33)
         self.assertEqual(r[0]["priority"], 0)
         self.assertEqual(r[0]["memory"], "<7373")
         self.assertEqual(r[0]["data_version"], 1)
 
+    def testCreateRuleWithInvalidMemory(self):
+        for invalid in ("<", "=", "=12", "abc"):
+            ret = self._post(
+                "/rules", data=dict(backgroundRate=33, mapping="c", priority=0, memory=invalid, product="Firefox", update_type="minor", channel="nightly")
+            )
+            self.assertEqual(ret.status_code, 400, "does not match")
+
     def testCreateRuleWithMig64(self):
         ret = self._post("/rules", data=dict(backgroundRate=33, mapping="c", priority=0, mig64=True, product="Firefox", update_type="minor", channel="nightly"))
         rule_id = int(ret.get_data())
         self.assertEqual(ret.status_code, 200, "Status Code: %d, Data: %s" % (ret.status_code, rule_id))
         r = dbo.rules.t.select().where(dbo.rules.rule_id == rule_id).execute().fetchall()
         self.assertEqual(len(r), 1)
         self.assertEqual(r[0]["mapping"], "c")
         self.assertEqual(r[0]["backgroundRate"], 33)
```

...will help you ensure that you don't match any invalid values.

PR #963, opened 2019-07-16
Add support for getting scheduled changes by sc_id

mozbhearsum:
Similar to #950, but for querying by `sc_id` to support frontend development.


auslib/web/admin/views/scheduled_changes.py
```diff
@@ -9,6 +9,44 @@
 from auslib.web.common.history import get_input_dict
 
 
+def add_signoff_information(row, table, sc_table):
```
JohanLorenzo: I like this function being taken out!

PR #971, opened 2019-07-30
admin.usgi: allow use of metadata server credentials

oremj:
We want to be able to use credentials provided by the Google metadata server. The google storage library should use them by default if GOOGLE_APPLICATION_CREDENTIALS is not set.


oremj:
For that, I think it's better to do a write test to the buckets. Even if the credentials are set, we aren't sure that they are working correctly.

mozbhearsum:
> For that, I think it's better to do a write test to the buckets. Even if the credentials are set, we aren't sure that they are working correctly.

That's a fair point. Let's do this for now, and I'll file a bug for that.

PR #978, opened 2019-08-07
Allow emergency_shutoff to be used in permission scheduled changes.

mozbhearsum:
@helfi92 discovered this while working on the new UI. Looks like it was missed in #450.


mozbhearsum:
Oops, @JohanLorenzo is away, moving review.

PR #982, opened 2019-08-16
Revert patch bug 1492603, because it can empty out release data in some situations

mozbhearsum:
I noticed this issue today, which wasn't caught by any of the tests (or by me in review of #936). Steps to reproduce:
1) Find a Release that requires signoff to change, and mark it read-only
2) Schedule a change to mark it read-write
3) Fulfill the signoff requirements
4) Wait for the agent to enact it

It will end up setting the `data` column to null. This happens because:
* We don't set `data` in the scheduled change to mark it read-write
* When we enact the change, we copy the entire release row from the scheduled changes table to the releases table (including the `null` for `data`).

cc @allan-silva 

We need to get this backed out ASAP.

### DIFF
--- a/auslib/db.py
+++ b/auslib/db.py
@@ -1805,18 +1805,6 @@ class Releases(AUSTable):
             potential_required_signoffs["rs"] = []
         return potential_required_signoffs
 
-    def getPotentialRequiredSignoffsForProduct(self, product, transaction=None):
-        potential_required_signoffs = {"rs": []}
-        where = [self.db.productRequiredSignoffs.product == product]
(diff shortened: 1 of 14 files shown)


allan-silva:
That is weird, because `what` to update readonly field does not include `data`field.
What was the specific release that you tried change?

https://github.com/mozilla/balrog/blob/67a0b0b64b8e6e38da08a3ecf297565054c24ac6/auslib/db.py#L1961-L1963

https://github.com/mozilla/balrog/blob/67a0b0b64b8e6e38da08a3ecf297565054c24ac6/auslib/db.py#L2136-L2160

PR #983, opened 2019-08-18
Bug 1492603 - require signoff to change releases from read only -> read write

allan-silva:
@mozbhearsum I'm opening this PR to address the problem found on #936 that was reverted on #982.

So, how we talk on IRC, the SC row must represent the entire new state, not just part of it. I really did not realized this concept.

Even though `what` is ```{"read_only": false}```, scenarios where an `OutdatedDataError` occurs, a null blob leads to a null data in cache:
https://github.com/mozilla/balrog/blob/5766a383a3cf001b0a94f5cbdb35ff9cdf20e98d/auslib/db.py#L2000-L2039

I can't have a method that changes only the readonly flag like bellow, I must update the entire release.
https://github.com/allan-silva/balrog/blob/e1d681415f7113a218e0b70b1072edcae5d9e4bb/auslib/db.py#L2136-L2160

I intent to get the `data` from current release when schedule a change from `ro` to `r+w` and use the regular update method to update the entire release.


mozbhearsum:
> I intent to get the data from current release when schedule a change from ro to r+w and use the regular update method to update the entire release.

I think this makes sense, yep. And because this only happens when the release is in read-only state, we don't need to worry about the possibility of an update to it potentially changing the scheduled change - so this should be pretty simple and safe.

mozbhearsum:
Could you also include https://github.com/mozbhearsum/balrog/pull/new/fix-release-required-signoffs as part of this? It's something I noticed when I was working on implementing UI for read only functionality in the new UI - we need required signoffs information about releases to set up the UI properly.

allan-silva:
@mozbhearsum in current PR, I resolved this using a new endpoint, instead of reference the `releases` back.

Do you prefer the full payload?

https://github.com/mozilla/balrog/blob/06c39df64b61bc674dfec7c4cfe246f47f3ea784/auslib/web/admin/views/releases.py#L329-L339

allan-silva:
@mozbhearsum this is ready for review again, hope no regressions this time.

mozbhearsum:
> @mozbhearsum in current PR, I resolved this using a new endpoint, instead of reference the releases back.

Ah. It's obviously not needed it we're doing that. I'll let you know as part of the full review :)

auslib/db.py
```diff
@@ -1943,7 +1955,9 @@ def update(self, where, what, changed_by, old_data_version, transaction=None, dr
         current_releases = self.select(where=where, columns=[self.name, self.product, self.read_only], transaction=transaction)
         for current_release in current_releases:
             name = current_release["name"]
-            if "product" in what or "data" in what:
+            is_readonly_change = "read_only" in what and current_release["read_only"] != what["read_only"]
+
+            if not is_readonly_change and ("product" in what or "data" in what):
```
mozbhearsum: I realize there's no way easy way to do it at this time, but I think there's a potential bug here. If both read only _and_ data are changing, we'll bypass the `_proceedIfNotReadOnly` check.

auslib/db.py
```diff
@@ -1961,24 +1975,11 @@ def update(self, where, what, changed_by, old_data_version, transaction=None, dr
...
-                    if not self.db.hasPermission(changed_by, "release_read_only", "unset", product, transaction):
-                        raise PermissionDeniedError("%s is not allowed to mark %s products read write" % (changed_by, what.get("product")))
-                elif what["read_only"] is True:
-                    if not self.db.hasPermission(changed_by, "release_read_only", "set", product, transaction):
-                        raise PermissionDeniedError("%s is not allowed to mark %s products read only" % (changed_by, what.get("product")))
+            if is_readonly_change:
+                self.validate_readonly_change(
+                    where, what["read_only"], changed_by, release=current_release, transaction=transaction, dryrun=dryrun, signoffs=signoffs
+                )
+                continue
```
mozbhearsum: This would read better as an if/else instead of using `continue`, since the remainder of the block below is relatively small.

auslib/db.py
allan-silva: Yeah, this is a problem once `data` can't be null in SC table, so I assume that readonly will always be changed independently. 

Do you think that it is possible to identify only the readonly change using other approach ?

I don't remember well (because I was away for a few weeks), probably it is the one of reasons I tried manage readonly in other table than release table, in the other PR.

mozbhearsum: Can we just leave the conditional statement as `if not is_readonly_change` or is there some reason we can't run `_proceedIfNotReadOnly` if `product` or `data` is in `what`? That way, if for some reason `product` or `data` change alongside it in the future, we won't accidentally skip the `_proceedIfNotReadOnly` check.

mozbhearsum:
As mentioned earlier, we're going to need to update the new UI before this can land. This will be done in https://github.com/mozilla-frontend-infra/balrog-ui. Here's some pointers:
* To run the new UI, bring up Balrog as usual, but comment out `balrogreactui` from the `docker-compose.yml`. Then, use `docker-compose up` in the root of the `balrog-ui` repo. The new UI will be available on http://localhost:9000.
* All of the read only modifications are handled in https://github.com/mozilla-frontend-infra/balrog-ui/blob/master/src/views/Releases/ListReleases/index.jsx. There's a few handler functions for the modal dialog that pops up, but the most important one is `handleReadOnlySubmit`. This calls `setReadOnlyFlag`, which ultimately talks to the backend (via https://github.com/mozilla-frontend-infra/balrog-ui/blob/4a785fb895d97951d5379de47458152a4f23dbb2/src/services/releases.js#L9). This logic will need some adjustment now that marking read-write will sometimes require signoff.
* You'll need to address this TODO about amending required signoff information to Releases: https://github.com/mozilla-frontend-infra/balrog-ui/blob/4a785fb895d97951d5379de47458152a4f23dbb2/src/views/Releases/ListReleases/index.jsx#L137. We do something similar for Rules already: https://github.com/mozilla-frontend-infra/balrog-ui/blob/4a785fb895d97951d5379de47458152a4f23dbb2/src/views/Rules/ListRules/index.jsx#L289
* When we schedule a change to re-enable updates, you'll need to make sure that the `releases` object gets updated with the scheduled change information. This is how we update it right now: https://github.com/mozilla-frontend-infra/balrog-ui/blob/4a785fb895d97951d5379de47458152a4f23dbb2/src/views/Releases/ListReleases/index.jsx#L234 - so that will need to be adjusted for the scheduled change case. You can see something similar with Rules here as well: https://github.com/mozilla-frontend-infra/balrog-ui/blob/4a785fb895d97951d5379de47458152a4f23dbb2/src/views/Rules/ListRules/index.jsx#L431. Once that's taken care of, the UI *should* automatically show the scheduled change/signoff information like this:
image
* You'll need to tweak https://github.com/mozilla-frontend-infra/balrog-ui/blob/master/src/components/ReleaseCard/index.jsx#L259 to show some different content for a read-only scheduled change, otherwise it will continue to show "View Diff".
* The `Signoff` and `Revoke` buttons _should_ just work. I'm not 100% certain of this, but I can't find a reason why they won't.
* I'm not sure if you'll need to add any new ones, but backend interactions are wrapped in methods in the `services` subdirectory, like https://github.com/mozilla-frontend-infra/balrog-ui/blob/master/src/services/releases.js.

allan-silva:
@mozbhearsum yeah, was the only new change. 

I realized now that I never asked you: Do you prefer `merge` or `rebase` in PRs? :smile:

mozbhearsum:
> @mozbhearsum yeah, was the only new change.
> 
> I realized now that I never asked you: Do you prefer `merge` or `rebase` in PRs? smile

:). Merge is preferable until it's been approved to make it easier to see follow-ups. You're welcome to rebase after that, of course.

allan-silva:
@mozbhearsum the commit 0900977 , was added after your approved.
This fetch the required signoffs for product on single sc GET endpoint.

mozbhearsum:
(CI is broken due to some migration work - I'll try to get it fixed this morning.)

mozbhearsum:
(Closing and re-opening to trigger Taskcluster.)

mozbhearsum:
@allan-silva - it looks like the latest failure is a real issue, can you take a look?

allan-silva:
> @allan-silva - it looks like the latest failure is a real issue, can you take a look?

Ahh, completely forgot run `tox` before push.
Fixed!

mozbhearsum:
I think we're ready to land this!

PR #990, opened 2019-08-27
Return more information about Rules with Releases

bhearsum:
We currently return a list of rule ids that point at a Release when we return a list of Releases. This was added to enable us to show rule ids that point at a Release way back in https://bugzilla.mozilla.org/show_bug.cgi?id=1113689. In the new UI, we'd like to backlink to the Rules page with a product and channel filter applied. To enable that, I'm enhancing the rule information we return to include both rule ids as well as their product and channel.

I've added a new property for this, and intend to remove the existing `rule_ids` property once the old UI goes away.

### DIFF
--- a/auslib/db.py
+++ b/auslib/db.py
@@ -1783,10 +1783,7 @@ class Releases(AUSTable):
         # to them. We need to find these Rules, and then return _their_
         # Required Signoffs.
         if info:
-            relevant_rules = []
-            for row in info:
-                for rule_info in row["rule_info"].values():
-                    relevant_rules.append(rule_info)
+            relevant_rules = [rule_info for row in info for rule_info in row["rule_info"].values()]
 
             # get all rs as one query
             all_rs = self.db.rules.getPotentialRequiredSignoffs(relevant_rules, transaction=transaction)
@@ -1857,12 +1854,8 @@ class Releases(AUSTable):
             for row in rows:
                 refs = [ref for ref in ref_list if ref[0] == row["name"]]
                 ref_list = [ref for ref in ref_list if ref[0] != row["name"]]
-                if len(refs) > 0:
-                    row["rule_ids"] = [ref[1] for ref in refs]
-                    row["rule_info"] = {str(ref[1]): {"product": ref[2], "channel": ref[3]} for ref in refs}
-                else:
-                    row["rule_ids"] = []
-                    row["rule_info"] = {}
+                row["rule_ids"] = [ref[1] for ref in refs]
+                row["rule_info"] = {str(ref[1]): {"product": ref[2], "channel": ref[3]} for ref in refs}
 
         return rows
 
--- a/auslib/test/admin/views/test_releases.py
+++ b/auslib/test/admin/views/test_releases.py
@@ -2153,7 +2153,6 @@ class TestRuleIdsReturned(ViewTest):
 
         releases = self._get("/releases")
         releases_data = json.loads(releases.data)
-        print(releases_data)
         mapped_rel = next(rel for rel in releases_data["releases"] if rel["name"] == rel_name)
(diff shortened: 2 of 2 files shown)


auslib/db.py
```diff
@@ -1783,21 +1783,18 @@ def getPotentialRequiredSignoffs(self, affected_rows, transaction=None):
         # to them. We need to find these Rules, and then return _their_
         # Required Signoffs.
         if info:
-            # get all rules as one query
-            q_rules = [self.db.rules.rule_id.in_(tuple([rule_id for row in info for rule_id in row["rule_ids"]]))]
-            all_rules = self.db.rules.select(where=q_rules, transaction=transaction)
-
-            # map rules according to rule_id
-            rules_map = {rule["rule_id"]: rule for rule in all_rules}
+            relevant_rules = []
```
JohanLorenzo: Nit: This chunk can be simplified by a list-comprehension.

```suggestion
            relevant_rules = [rule_info for row in info for rule_info in row["rule_info"].values()]
```

auslib/db.py
```diff
@@ -1847,17 +1844,25 @@ def getReleaseInfo(self, names=None, product=None, limit=None, transaction=None,
...
 
             for row in rows:
                 refs = [ref for ref in ref_list if ref[0] == row["name"]]
                 ref_list = [ref for ref in ref_list if ref[0] != row["name"]]
                 if len(refs) > 0:
                     row["rule_ids"] = [ref[1] for ref in refs]
+                    row["rule_info"] = {str(ref[1]): {"product": ref[2], "channel": ref[3]} for ref in refs}
                 else:
                     row["rule_ids"] = []
+                    row["rule_info"] = {}
```
JohanLorenzo: Nit: Lines 

```py
row["rule_ids"] = [ref[1] for ref in refs]
row["rule_info"] = {str(ref[1]): {"product": ref[2], "channel": ref[3]} for ref in refs}
```

already handle the case where there's no refs. The `if` statement is not needed 😃

auslib/db.py
```diff
@@ -1847,17 +1844,25 @@ def getReleaseInfo(self, names=None, product=None, limit=None, transaction=None,
                 ((self.db.releases.name == self.db.rules.mapping) | (self.db.releases.name == self.db.rules.fallbackMapping)),
             )
             if transaction:
-                ref_list = transaction.execute(select([self.db.releases.name, self.db.rules.rule_id]).select_from(j)).fetchall()
+                ref_list = transaction.execute(
```
JohanLorenzo: Nit: I remember the current pattern is used in the code base, but I just realized we can simplify it with:

```py
transaction = transaction if transaction else self.getEngine()
ref_list = transaction.execute(
    select([
        self.db.releases.name, self.db.rules.rule_id, self.db.rules.product, self.db.rules.channel
    ]).select_from(j)
).fetchall()
```

What do you think?

auslib/test/admin/views/test_releases.py
```diff
@@ -2131,14 +2143,19 @@ def testMappingIncluded(self):
...
         self.assertFalse(rule_id in not_mapped_rel["rule_ids"])
+        self.assertFalse(rule_id in not_mapped_rel["rule_info"].keys())
 
         dbo.rules.t.insert().execute(
             id=rule_id, priority=100, version="3.5", buildTarget="d", backgroundRate=100, mapping=rel_name, update_type="minor", data_version=1
         )
 
         releases = self._get("/releases")
         releases_data = json.loads(releases.data)
+        print(releases_data)
```
JohanLorenzo: Nit: Debug leftover?

auslib/db.py
JohanLorenzo: SGTM!

PR #992, opened 2019-08-29
Upgrade .taskcluster.yml to v1; add deploy to prod task

mozbhearsum:
I tested this the best I could on https://github.com/testbhearsum/balrog. Everything seems to work OK, except I get some errors about a missing `secrets:get:repo:github.com/testbhearsum/balrog:coveralls` scope that don't make sense.

I'm going to try to sort that out, but in the meantime I wouldn't mind some feedback/review. The intent here is to behave the same way as the current `.taskcluster.yml`, except for:
* Restricting stage deploys to version tags
* Adding production deploys


mozbhearsum:
@mitchhentges, do you feel comfortable looking at this?

mitchhentges:
No worries, I'll look at this later today

mozbhearsum:
Curiously, the scope issues I saw on `testbhearsum` didn't happen here, which makes me think there's something funky about something in the taskcluster roles or scopes, but I can't figure out what. Could also have to do with the fact that `testbhearsum` isn't part of the Mozilla org? Unless I see issues here I'm not going to dig into it further.

.taskcluster.yml
```diff
@@ -1,304 +1,282 @@
...
+    release_tag: {
+      $if: 'tasks_for == "github-release"',
+      then: '${event.release.tag_name}',
+      else: 'null',
+    }
+  in:
+    $flattenDeep:
+      - $if: '(tasks_for == "github-pull-request" && event["action"] in ["opened", "reopened", "synchronize"]) || tasks_for == "github-push"'
+        then:
+          - taskId: {$eval: as_slugid("backend-tests")}
```
mitchhentges: Why do you need to provide a name/reference for this task?

.taskcluster.yml
```diff
@@ -1,304 +1,282 @@
...
-          - push
-        branches:
-          - master
-    metadata:
-      name: Balrog React UI tests
-      description: Balrog React UI tests
-      owner: "{{ event.head.user.email }}"
-      source: "{{ event.head.repo.url }}"
+      # Deploy the UI to stage whenever we create a versioned release tag
+      - $if: 'tasks_for == "github-release" && release_tag[0] == "v"'
```
mitchhentges: Heads up that this means github releases where the tag mistakenly misses the prefixed "v" will silently fail to create the task

docker-compose.yml
```diff
@@ -109,25 +109,25 @@ services:
...
-      - PORT=9000
-    volumes:
-      - ./react-ui:/app
-      - node_modules:/app/node_modules
-    entrypoint:
-      - /bin/bash
-      - --login
-      - -c
-      - yarn install && yarn start
+#  balrogreactui:
```
mitchhentges: When you land this PR, if these aren't used, it'd be :ok_hand: if you deleted these rather than just commenting them out (since we have VCS etc etc)

.taskcluster.yml
```diff
@@ -1,304 +1,282 @@
...
+    # tasks, so we need to create this variable to make sure it always
+    # has a value instead of using ${event.release.tag_name} directly.
+    release_tag: {
+      $if: 'tasks_for == "github-release"',
+      then: '${event.release.tag_name}',
+      else: 'null',
+    }
+  in:
+    $flattenDeep:
+      - $if: '(tasks_for == "github-pull-request" && event["action"] in ["opened", "reopened", "synchronize"]) || tasks_for == "github-push"'
```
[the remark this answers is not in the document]
mozbhearsum: Ah, I've been using close/reopen for that. I can't decide if editing the title is cleaner or not, but I think it's worthwhile to give us the option. I'll add it.

.taskcluster.yml
mozbhearsum: I didn't know it could be omitted :).

.taskcluster.yml
mozbhearsum: Is there any way to make this fail not-silent? It's not ideal, I agree, but I don't want to create tasks for other types of tags (CloudOps' infrastructure will just ignore them, so it's wasteful).

docker-compose.yml
mozbhearsum: Whoops, didn't mean to include this!

.taskcluster.yml
```diff
@@ -1,304 +1,282 @@
...
-        branches:
-          - master
-    metadata:
-      name: Balrog back-end tests
-      description: Balrog Python tests
-      owner: "{{ event.head.user.email }}"
-      source: "{{ event.head.repo.url }}"
+  $let:
+    # Github events have this stuff in different places...
+    repo_url: {
```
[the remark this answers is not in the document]
mozbhearsum: Thanks, that's good to know. I think I prefer the look with the curly braces, so I'll keep them.

mozbhearsum:
It looks like there's no blocking issues so I'm going to merge this, and I can follow-up with any potential improvements.

.taskcluster.yml
mitchhentges: Unfortunately, the only real way to provide communication from "automation validation" -> "github release-ee" is to create a task on every release that informs the release-ee if validation fails (by logging/failing the task/leaving a comment on the release's commit). I agree that creating such a task on all releases could be wasteful,but it's worth considering the balance of "how much does it cost maintaining this task and wasting machine hours on every release" vs "how much will it cost if someone can't figure out why their release isn't triggering automation".

As I write this, I doubt the extra task is worthwhile, but it's still food for thought.

PR #1000, opened 2019-09-05
Add simple redirect links to new UI

mozbhearsum:
This will result in links such as the following on the Dashboard & Rules pages:
image
image
image
image

We'll need https://github.com/mozilla-services/cloudops-infra/pull/1349 merged before it will work for stage/prod.


ui/app/js/controllers/rules_controller.js
```diff
@@ -20,16 +20,20 @@ function($scope, $routeParams, $location, $timeout, Rules, Search, $modal, $rout
   $scope.current_emergency_shutoff = null;
   $scope.current_user = localStorage.getItem("username");
   $scope.auth0 = Auth0;
+  $scope.new_ui_url = NewBalrogURL + "/rules";
```
JohanLorenzo: Nit: I'm a little bit on the fence about the name of `NewBalrogURL`. To me, an URL represents a complete string. I feel the fact that it's just the prefix make is ambiguous. Not a blocker.

mozbhearsum: Because this is going to be short lived I'm just going to merge as is.

PR #1015, opened 2019-10-04
Autograph signing support for /json/1 endpoint

mozbhearsum:
This patch adds optional support for signing with Autograph to the /json/1 endpoint. It's disabled if `AUTOGRAPH_URL` is unset, so we can land the code without the infrastructure being ready, and turn it on later.

I've added a small cache for signatures for now, which will keep them for an entire day. The certificates are rotated often, so we probably want to resign regularly. Realistically, we'll resign more often than once a day because each webhead will have its own in-memory cache, and those get recycled fairly often as we scale up/down. There's a lot about this we'll want to revisit when we start signing at a bigger scale (in-memory cache vs. shared cache, cache size, etc.) -- but I think this is fine for now.

This patch is built on top of #1012, so until that lands, https://github.com/mozbhearsum/balrog/compare/guardian...mozbhearsum:autograph2 is a useful way to see only the changes being added here.

As best I can tell, this works. I'm still trying to manually verify a signature myself as another sanity check, but I haven't had any luck there yet (probably issues with me, not with the signatures).

cc'ing @oremj and @jvehent, in case they have any comments or concerns.


mozbhearsum:
> I'm curious if there a timeout for Balrog to respond to a client, and how that compares to typical autograph response time and timeouts. And does the `raise_for_status()` in `sign_hash()` mean any autograph errors are just forwarded to the client ? Not hugely concerned about this, just interested in what happens when things go wrong on the external system.

If errors happen, we'll see something on the client, yeah. For example, if I change the autograph URL to something invalid I end up with:
> HTTPConnectionPool(host='autograph', port=8001): Max retries exceeded with url: /sign/hash (Caused by NewConnectionError('&lt;urllib3.connection.HTTPConnection object at 0x7f677e96acc0&gt;: Failed to establish a new connection: [Errno 111] Connection refused'))

400s would probably end up showing whatever autograph returned. Perhaps we should be eating and generalizing these errors (after sending them to Sentry?) to avoid any potential leakage.

mozbhearsum:
@jvehent - are you concerned at all about error responses from Autograph showing up in Balrog responses? I'll note that we already show them on the signing workers.

jvehent:
Nothing private leaks in autograph responses, so that should be ok. Just make sure not to leak the `Authorization` header from the signing requests.

PR #1028, opened 2019-10-26
Fix api.yaml spec

allan-silva:
@mozbhearsum, the 76f1a7f broke the swagger spec on balrog/master, because the `required` property is a property of `Schema Object` instead `allOf`array, so the Balrog Admin fails to start.

I have no idea why on the `test_client` the requests/responses works well (maybe bypassing some connexion's flow).

The `SpecBuilder.validate` is able to validate jsonschema, but how our spec is not valid, we can't use it.

Let me know if I missed something.

### DIFF
--- a/auslib/web/admin/swagger/api.yaml
+++ b/auslib/web/admin/swagger/api.yaml
@@ -251,9 +251,9 @@ parameters:
     required: true
     schema:
       type: object
+      required:
+        - role
       allOf:
-        - required:
-          - role
         - $ref: '#/definitions/CSRFModel'
         - $ref: '#/definitions/RoleModel'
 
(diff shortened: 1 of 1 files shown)


allan-silva:
Sorry, I was wrong,  after tests failure on Task Cluster, I researched and the current syntax is correct and the https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#schemaObject says nothing about use `required` as `allOf` item.  
The `balrogtest` on my env was not updated. When the `balrogtest` old image was removed the tests woks well locally. So, no `test_client` problem...

mozbhearsum:
No worries, this happened to me too!

PR #1033, opened 2019-11-04
Fix build issues with production Dockerfiles + taskcluster; add CI tests for docker builds

mozbhearsum:
The Agent build broke because taskcluster's version of Docker doesn't support `ARG`. The backend build broke because we were missing some files from the `COPY` instructions.


mozbhearsum:
The failure here is a bad worker -- a rerun succeeded.

PR #1034, opened 2019-11-04
Switch line length to 100.

mozbhearsum:
Since I just talked about this in #1030, I may as well go ahead and do it now....


mozbhearsum:
Turns out Black can't reformat a bunch of stuff in a flake8 compatible way. It's probably going to take some time for me to manually reformat everything it missed.

PR #1035, opened 2019-11-05
Include MANIFEST.in and pyproject.toml in production Docker image.

mozbhearsum:
This should fix an issue that has come up where the migration files and JSON schemas are not available (because they weren't installed when `python setup.py install` was run).


mozbhearsum:
For whatever reason, I wasn't able to reproduce the issue locally -- even without this patch my local Docker builds include `migrate/` and `swagger/` directories. But you can see a clear difference in the Docker build log of this PR and #1033 -- this one copies them into `site-packages`, which should fix the issues the latest deploy is seeing.

PR #1036, opened 2019-11-08
Switch to aws-provider worker types.

tomprince:


mozbhearsum:
Do we need to add and/or remove an integration here?

mozbhearsum:
> Do we need to add and/or remove an integration here?

From elsewhere:
> Don't need the new integration to land my patch, though we do need to enable it at some point.

PR #1038, opened 2019-11-13
Bring in line with best practices

mozbhearsum:
This includes a ton of changes, most of which are straightforward. The most significant change is to stop pinning dependencies for `balrogclient` (it's a library, it shouldn't be!).


client/tox.ini
```diff
@@ -34,6 +38,12 @@ depends = py37
 parallel_show_output = true
 
 [testenv:check]
+deps =
+    black
+    flake8
+    flake8-black
+    isort
+    pip-compile-multi
```
nthomas-mozilla: Do you still need this call to pip-compile-multi if there are no longer any requirements files ??

mozbhearsum: Good catch, we certainly don't.

PR #1042, opened 2019-11-26
Add support for FirefoxVPN as the new name for Guardian.

mozbhearsum:
In talking with the Firefox VPN developers, we decided that we should use a new product name in updates, in case `Guardian` becomes less specific in the future.

This patch allows both to be used, and I will be following up to remove `Guardian` references later.


uwsgi/admin.wsgi
```diff
@@ -22,15 +22,15 @@ DOMAIN_WHITELIST = {
...
+    "fpn.firefox.com": ("FirefoxVPN", "Guardian"),
 }
 if STAGING or LOCALDEV:
     SYSTEM_ACCOUNTS.extend(["balrog-stage-ffxbld", "balrog-stage-tbirdbld"])
     DOMAIN_WHITELIST.update(
         {
             "ftp.stage.mozaws.net": ("Firefox", "Fennec", "Devedition", "SeaMonkey", "Thunderbird"),
             "bouncer-bouncer-releng.stage.mozaws.net": ("Firefox", "Fennec", "Devedition", "SeaMonkey", "Thunderbird"),
-            "stage.guardian.nonprod.cloudops.mozgcp.net": ("Guardian",),
+            "stage.guardian.nonprod.cloudops.mozgcp.net": ("FirefoxVPN", "Guardian"),
```
MihaiTabara: I wonder if the URLs will change as well at some point to encompass the new name

mozbhearsum: Yeah, that's a good question. I think that's @jbuck's department. Doesn't matter to me though, especially for stage.

PR #1045, opened 2019-12-02
Add dirschema config

mozbhearsum:
This PR adds a config that the new `dirschema` app and will add continuous checking to make sure this repository conforms to the necessary schemas.

### DIFF
--- a/.dirschema.yml
+++ b/.dirschema.yml
@@ -1,0 +1,24 @@
+repos:
+  schema: https://raw.githubusercontent.com/mozilla-releng/releng-dirschemas/master/repo.yaml
+  projects:
+    - /
+pyup:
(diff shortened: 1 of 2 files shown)


nthomas-mozilla:
BTW, does this status on the dirschema check have a 'Details' when there are failures ?
Screen Shot 2019-12-03 at 9 30 40 AM

mozbhearsum:
> BTW, does this status on the dirschema check have a 'Details' when there are failures ?
> Screen Shot 2019-12-03 at 9 30 40 AM

There's no details link (I don't have any sort of UI around the app), but it will add a comment with a failure. Here's an example from some of my testing in a fork: https://github.com/testbhearsum/balrog/pull/103

nthomas-mozilla:
Ah I see, looks good.

PR #1126, opened 2020-01-06
Bump all Python dependencies

mozbhearsum:
Pyup is broken, doing this by hand.


mozbhearsum:
Test failures are just for the backend, and appear to be caused by a connexion upgrade.

mozbhearsum:
The `connexion` bump ended up breaking a bunch of tests. Most of them were resolved in https://github.com/mozilla-releng/balrog/pull/1126/commits/5d4604a1457eaf42a5e5787fac5629d83c0ab666 and https://github.com/mozilla-releng/balrog/pull/1126/commits/4e45e04bde93f702e7b897a7672ce26e720cb916, which have comments with details about the issues and fixes.

.taskcluster.yml
```diff
@@ -85,7 +85,7 @@ tasks:
                   git clone ${repo_url} balrog &&
                   cd balrog &&
                   git checkout ${head_sha} &&
-                  pip install coveralls==1.9.2 tox &&
+                  pip install coveralls tox &&
```
rail: Shall we move coveralls to `test.in`? It doesn't belong there logically (you don't necessarily submit to coveralls as a   part of every test run), but at least it's pinned.

mozbhearsum: Yeah, that's probably a good idea. I'll try to do it here.

mozbhearsum:
@rail - It's a bit ugly to do this, because we have to pass through a lot of information to tox to do it, but it works :).

PR #1152, opened 2020-01-10
Fix issues with bad data being passed in systemCapabilities field

mozbhearsum:
This should fix https://sentry.prod.mozaws.net/operations/prod-public/issues/6445338/?query=is:unresolved, https://sentry.prod.mozaws.net/operations/prod-public/issues/6440645/?query=is:unresolved, and https://sentry.prod.mozaws.net/operations/prod-public/issues/6445339/?query=is:unresolved -- all of which are around unexpected data being sent in `systemCapabilities`.

### DIFF
--- a/src/auslib/web/public/client.py
+++ b/src/auslib/web/public/client.py
@@ -36,11 +36,18 @@ def getSystemCapabilities(systemCapabilities):
     # New-style SYSTEM_CAPABILITIES, as implemented in https://bugzilla.mozilla.org/show_bug.cgi?id=1373367
     if systemCapabilities.startswith("ISET:"):
         for part in systemCapabilities.split(","):
+            # Skip fields with an unparseable format, which we see often. Eg:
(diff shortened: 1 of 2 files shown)


src/auslib/web/public/client.py
```diff
@@ -36,11 +36,18 @@ def getSystemCapabilities(systemCapabilities):
...
+            if ":" not in part:
+                continue
             key, value = part.split(":", 1)
             if key == "ISET":
                 caps["instructionSet"] = value
             elif key == "MEM":
-                caps["memory"] = int(value)
+                try:
+                    caps["memory"] = int(value)
+                except ValueError:
```
rail: `int()` can also raise `TypeError`, but probably not in this case, where use already ensure that the value passed is a string.

mozbhearsum: TIL! Let's address this if it comes up.

PR #1153, opened 2020-01-14
Add information about latest aus5 and aus4 certificates.

mozbhearsum:
While digging into https://bugzilla.mozilla.org/show_bug.cgi?id=1609192 I noticed that we were missing information about the current aus5 and aus4 certs.


docs/source/client_domains.rst
```diff
@@ -11,7 +11,9 @@ SSL Certificates
...
+| aus5.mozilla.org | DigiCert | 02:39:74:46:BB:F4:1C:48:6B:98:63:A8:54:0B:19:DD | Primary                    | June 16, 2021      |                                                                       |                                                                                                  |
++                  +          +-------------------------------------------------+----------------------------+--------------------+-----------------------------------------------------------------------+--------------------------------------------------------------------------------------------------+
+|                  |          | 07:10:8B:20:9E:D3:45:6C:EE:88:94:91:44:C4:56:0C | Primary                    | August 13, 2019    |                                                                       |                                                                                                  |
```
nthomas-mozilla: There's something a little odd here:
* `07:10:...` and `0D:23:...` both expired Aug 13 ?
* `07:10:...` is no longer a Primary cert

Possibly something incorrect on [wiki history](https://wiki.mozilla.org/index.php?title=Balrog/Client_Domains&action=history). Could you pls fix up before merge.

mozbhearsum: Whoops. The first one appears to be bad info on the wiki. The second one is just bad migration into the source docs.

I'm actually not sure what the difference betwen `07:10:...` and `0D:23:...` is. It's possible that one of them was a backup -- but I don't appear to have any confirmation in e-mail, bugs, or chat history.

nthomas-mozilla:
lgtm, plus bonus points for ether reference.

PR #1155, opened 2020-01-16
Add debug for sentry dsn

mozbhearsum:


mozbhearsum:
No longer needed - we managed to verify that the Agent is indeed working.

Issue #1063, opened 2019-12-20
publish blobs as a separate package?


mozbhearsum:
Balrog blobs have two main jobs:
1) They store and verify the schema of Releases.
2) They create responses based on the Release data.

#2 is something that generally only the server cares about, but #1 is something that would be extremely useful to clients as well as the server.

I think it would be a good idea to investigate if it would be possible to publish auslib.blobs as a separate package that the server and clients could depend on. This might be trickier now that we have multifile updates...we may need to look at refactoring the code such that #1 and #2 are isolated. It might not even be viable, but I think it's worth looking into further.

(Imported from https://bugzilla.mozilla.org/show_bug.cgi?id=1312868)

Issue #1049, opened 2019-12-20
Rebuild Release blob submission


mozbhearsum:
Submitting data to Balrog Release blobs has become increasingly problematic over the years. Much of this boils down to the fact that we hit tons of data races trying to update a Release, because the API requires that clients submit the data version that they're basing their update on, and the database layer rejects that update if the current data version does not match the given one. This has become aggravated in recent times by adding more `balrogworker` instances.

Whatever solution we end up choosing should get us to a place where a submission with valid data works 99.99% of the time. We do not want a solution that still leaves us subject to data races or other scaling issues.

This work is likely to involve the API (I'd like to take this opportunity to get rid of or fix up https://github.com/mozilla-releng/balrog/blob/master/src/auslib/web/admin/views/releases.py, which is really, really ugly and hacky), `balrogscript`, and possibly the database layer.

mozbhearsum:
#1056 may also be relevant here

mozbhearsum:
See also: https://bugzilla.mozilla.org/show_bug.cgi?id=1501167

mozbhearsum:
https://docs.google.com/document/d/1f1aH0puYlpt7Hf-lmH5iXRpP9Pi6X7-ZvZj3TWlRUOI/edit#heading=h.3ds0irlb3wch is being used for planning this.

https://github.com/mozilla-releng/balrog/projects/1 will be used to track the work.

mozbhearsum:
We're pretty far into this project now, and I'm trying to hammer out landing/migration plan for the most tricky bits. Here's what I have in mind (ignoring the parts that are already in production):
* Finish new APIs. Most of this is in #1248, and the remainder will be #1265.
* Add support in the UI for the new Releases API. When this first lands, it will support both the old and the new APIs (the former will be removed later). This is #1266.
* Add support for serving release data from the new tables. This is happening in #1267.
* Add Balrogscript support for the new APIs. This is happening in https://github.com/mozilla-releng/scriptworker-scripts/pull/193
* Test in staging, both nightly and Release submission scenarios, as well as update verify.
* Start submitting Nightlies with the new APIs. This should probably bake for a few days to a week to help uncover any functional, performance, race, or other issues that may be difficult to find ahead of time - particularly with submission itself.
* Start submitting Betas/DevEdition with the new APIs. I'd like this to bake for at least 2 Betas.
[...]

PR #1178, opened 2020-01-31
Prototype of split-table Releases, service layer, and OpenAPI 3.0 API

mozbhearsum:
I do not intend for this to be merged, but I took some time to explore the idea of using a separate service layer to implement the split-table Release logic that we may use to implement #1049. The idea is that the "business logic" of verifying signoffs, checking permissions, splitting releases, etc. would live in its own layer instead of being embedded in the database layer. I'm not necessarily set on doing this but I thought it was interesting enough to share and get some feedback on.

Other than the new structure/architecture itself, the other interesting parts of this patch are:
* The Release splitting logic in `src/auslib/services/releases.py`, which was extremely tricky to get right.
* The new API, using OpenAPI 3.0.

Notably, the Release splitting logic is not part of `auslib.blobs`. I didn't put it there to avoid having that module gain further knowledge of the database layer. I could go either way on this, but if we ever want to do #1063 (which is debatable), this module can't depend on other parts of `auslib`.


mozbhearsum:
@JohanLorenzo - you may be interested in looking at this as you originally suggested the service layer.

tests/admin/views/test_releases_v2.py
```diff
@@ -0,0 +1,178 @@
...
+    for pname, pdata in firefox_56_0_build1["platforms"].items():
+        for lname in pdata.get("locales", {}):
+            old_data_versions["platforms"][pname]["locales"][lname] = 1
+
+    ret = api.put("/v2/releases/Firefox-56.0-build1", json={"blob": firefox_56_0_build1, "old_data_versions": old_data_versions})
+    assert ret.status_code == 400
+
+
+@pytest.mark.usefixtures("releases_db", "mock_verified_userinfo")
+def test_overwrite_succeeds(api, firefox_60_0b3_build1):
```
JohanLorenzo: Potential improvement: We may make 3 different test cases to ensure the whole thing works:

1. A unit test that verifies the api level delegates the right call to the mocked (or stubbed) out service layer.
2. A unit test that makes sure the service level delegates the right call to the database
3. An integration test that shows the API correctly stored the change by `get`ting the resource after the `put`.

This way, the current test doesn't have to deal with both the API and the database and will be lighter to setup/write. What do you think?

src/auslib/db.py
```diff
@@ -691,28 +691,28 @@ def _getBucket(self, identifier):
...
         bname = "{}/{}-{}-{}.json".format(identifier, rowData.get("data_version"), getMillisecondTimestamp(), changed_by)
         blob = self._getBucket(identifier).blob(bname)
         blob.upload_from_string(json.dumps(rowData[self.data_column]), content_type="application/json")
 
     def getChange(self, change_id=None, column_values=None, data_version=None, transaction=None):
-        if self.identifier_column not in column_values or not data_version:
-            raise ValueError("Cannot find GCS changes without {} and data_version".format(self.identifier_column))
-        identifier = column_values[self.identifier_column]
+        if not set(self.identifier_columns).issubset(column_values.keys()) or not data_version:
+            raise ValueError("Cannot find GCS changes without {} and data_version".format(self.identifier_columns))
```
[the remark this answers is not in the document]
mozbhearsum: Ah, Google Cloud Storage (like S3, but in GCP).

tests/admin/views/test_releases_v2.py
mozbhearsum: Interesting that you mention this! The main reason I wrote the tests in a way where all of the entry points were the API was to avoid coupling the implementation and the tests together so tightly, as they have been in the past. This was inspired by https://blog.cleancoder.com/uncle-bob/2017/10/03/TestContravariance.html, the summary of which is:
> The structure of the tests must not reflect the structure of the production code, because that much coupling makes the system fragile and obstructs refactoring. Rather, the structure of the tests must be independently designed so as to minimize the coupling to the production code.

That doesn't mean more tests might not be useful, just wanted to add some background/explanation.

mozbhearsum:
I'm going to close this. Split table Releases was merged in #1199, and #1212 is a more polished version of the service layer/new API.

Thank you for the feedback!

PR #1197, opened 2020-02-14
Allow longer lines before eliding

mozbhearsum:
The current eliding that happens as part of Balrog submission ends up not showing any useful parts of the request, which makes debugging submission issues next to impossible.

This patch should elide long enough to show the entire body of regular requests, while still eliding enough to avoid https://github.com/mozilla-releng/scriptworker/issues/362.

### DIFF
--- a/client/src/balrogclient/api.py
+++ b/client/src/balrogclient/api.py
@@ -9,6 +9,7 @@ import time
 import requests
 import requests.auth
 
+MAX_LOG_LINE_LENGTH = 10000
 # Refresh the tokens 5 minutes before they expire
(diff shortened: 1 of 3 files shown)


mozbhearsum:
@tomprince should probably look at this too.

client/src/balrogclient/api.py
```diff
@@ -16,8 +16,8 @@
 
 def _json_log_data(data):
     log = json.dumps(data)
-    if len(log) > 100:
```
tomprince: I guess maybe this should be a constant?

mozbhearsum: Good idea.

PR #1199, opened 2020-02-14
Add tables for split Releases

mozbhearsum:
Adds the new tables needed to start implementing split table Releases. The most notable part of this is enhancing `GCSHistory` to handle multiple part keys.

Eventually I intend to migrate all of the `Releases` table into the new tables (TBD is whether or not we'll split them, or just let them fully live in `ReleasesJSON`.

### DIFF
--- a/tests/test_db.py
+++ b/tests/test_db.py
@@ -5619,6 +5619,20 @@ class TestDBModel(unittest.TestCase, NamedFileDatabaseMixin):
                 "releases_scheduled_changes_history",
                 "releases_scheduled_changes_signoffs",
                 "releases_scheduled_changes_signoffs_history",
+                "releases_json",
+                "releases_json_scheduled_changes",
+                "releases_json_scheduled_changes_conditions",
+                "releases_json_scheduled_changes_conditions_history",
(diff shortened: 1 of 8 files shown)


tests/test_db.py
```diff
@@ -5899,6 +6054,31 @@ def _add_emergency_shutoff_tables(self, db, upgrade=True):
             for table in shutoff_tables:
                 self.assertNotIn(table, metadata.tables)
 
+    def _add_release_json_tables(self, db, upgrade=True):
+        metadata = self._get_reflected_metadata(db)
+        shutoff_tables = [
```
nthomas-mozilla: Nit, `shutoff` carried over from the copied function, please update.

tests/test_db.py
```diff
@@ -5966,6 +6146,7 @@ def _noop(*args, **kwargs):
 
         versions_migrate_tests_dict = {
             # This version removes the releases_history table, which is verified by other tests
+            33: self._add_release_json_tables,
```
nthomas-mozilla: Nit, the new test should go above the comment, so the latter stays attached to the `32: _noop,`.

PR #1205, opened 2020-02-19
Add bug link to production push section of docs

mozbhearsum:
Forgot to add this when the docs were moved from the wiki.


mozbhearsum:
cc'ing is the right thing to do, I think. We can update it later if needed!

PR #1263, opened 2020-03-31
Fix asyncio loop issues

mozbhearsum:
This should fix the issues we found in https://github.com/mozilla-releng/balrog/issues/1261. I was able to reproduce locally by hacking the gcloud-aio code to refresh tokens more often (which caused the old code to use an old ClientSession (that held a closed EventLoop) to try to get a new token.

The fix I've chosen to go with here is to explicitly set `session` when instantiated the `Storage` objects. To do this, I had to start instantiating them at runtime rather than startup time, because it's extremely painful to try to override them later.

I also did a little bit of cleanup in `services/releases.py` to reuse code better, and fix the errantly named `futures` variable (they're actually coroutines).


mozbhearsum:
@escapewindow - if you'd like someone more familiar with the Balrog codebase to look at this as well, just let me know.

src/auslib/services/releases.py
```diff
@@ -18,6 +18,27 @@
...
+    # a coroutine.
+    async def coroutine_gather(*args, **kwargs):
+        return await asyncio.gather(*args, **kwargs)
+
+    results = asyncio.run(coroutine_gather(*coros, return_exceptions=True))
+    for r in results:
+        if isinstance(r, Exception):
+            # aiohttp exceptions indicate a failure uploading to GCS, which don't warrant
+            # sending an error back to the client.
+            if isinstance(r, ClientError):
```
escapewindow: If you want to reuse this function with different constraints, you could add a kwarg tuple that contains the exception types that are logged rather than raised.

mozbhearsum: That's not a bad idea; I think I'll defer it for now since there's no use case for other exceptions at the moment.

PR #1150, opened 2020-01-10
Update all of the JS packages that we can

mozbhearsum:
This updates nearly all of the Javascript dependencies. The only thing held back was @date-io/date-fns, due to https://github.com/mui-org/material-ui-pickers/issues/1440.

Unfortunately, the React upgrade has broken `react-virtualized`'s `scrollToPosition` function, which appears to be a known issue (https://github.com/bvaughn/react-virtualized/issues/1226, https://github.com/bvaughn/react-virtualized/issues/1179#issuecomment-572854217). I _think_ it's probably better to live with this regression for now than to keep React out of date.

Other notable changes here:
* Disable `prop-spreading` and `exhaustive-deps` eslint rules due to our extensive use of them. I'm open to re-enabling them if that's preferable.
* Use SSL for the local web server.

PR #1266, opened 2020-04-01
Add support for new releases API in UI

mozbhearsum:
This patch adds support for the new V2 Releases API in the UI. While the V1 API still exists, we have to support both, so I've forked and branched code where necessary to do that. The V1 API is expected to be removed after we've migrated all old data to the new API (the new API uses new tables, so the two APIs return different sets of data).

The most notable parts of this are:
* The new API is backed by a datastore that spreads data across multiple database tables, and writes history to multiple places (on per part of the Release). For the UI, this only affects history fetching/display. I've created a new `ListReleaseRevisionsV2` view to cope with this.part of a Release, and diffs those against the 
* The new API is simpler (I hope), and unifies direct and scheduled change modifications into a single endpoint. Because this API is so different, I also created a new `ReleaseV2` view rather than try to support both in the existing view.
* New Releases will always be created with the V2 API.
* I think there's a race or dependency issue in `ListReleaseRevisionsV2`, but I haven't been able to get to the bottom of it. It results in incorrect diffs being shown sometimes. I was hoping you may be able to spot something obvious that I've missed in the relevant `useEffect`.

This patch is built on #1265, which has a few additional API changes not on master that are needed to support the UI. For the purposes of this review, they can be ignored. I'll rebase this once that lands.


helfi92:
Planning to review this sometime this week.

ui/src/views/Rules/ListRules/index.jsx
```diff
@@ -958,24 +961,39 @@ function ListRules(props) {
...
+            setDrawerState({
+              ...drawerState,
+              open: true,
+              item,
+            });
+          });
+        }
       });
     }
   }, [drawerReleaseName]);
```
helfi92, line 999: STR:

1. Navigate to https://localhost:9000/rules while being logged out
2. Click the "Disable Updates" speed dial button
3. Open the developer tools and see error: `TypeError: productChannelQueries is null`

ui/src/views/Releases/routes.js
```diff
@@ -11,17 +16,28 @@ const ListReleaseRevisions = lazy(() =>
 const Release = lazy(() =>
   import(/* webpackChunkName: 'Releases.Release' */ './Release')
 );
+const ReleaseV2 = lazy(() =>
+  import(/* webpackChunkName: 'Releases.Release' */ './ReleaseV2')
```
helfi92, line 20:
```suggestion
  import(/* webpackChunkName: 'Releases.ReleaseV2' */ './ReleaseV2')
```

ui/src/views/Releases/routes.js
```diff
@@ -3,6 +3,11 @@ import lazy from '../../utils/lazy';
 const ListReleases = lazy(() =>
   import(/* webpackChunkName: 'Releases.ListReleases' */ './ListReleases')
 );
+const ListReleaseRevisionsV2 = lazy(() =>
+  import(
+    /* webpackChunkName: 'Releases.ListReleaseRevisions' */ './ListReleaseRevisionsV2'
```
helfi92, line 8:
```suggestion
    /* webpackChunkName: 'Releases.ListReleaseRevisionsV2' */ './ListReleaseRevisionsV2'
```

ui/src/views/Releases/ReleaseV2/index.jsx
```diff
@@ -0,0 +1,346 @@
...
+            )}
+          </Fragment>
+          <Snackbar onClose={handleSnackbarClose} {...snackbarState} />
+        </Fragment>
+      )}
+    </Dashboard>
+  );
+}
+
+export default withUser(ReleaseV2);
```
helfi92, line 346: Will this be renamed to `Releases/Release/index.jsx` once v2 is stable? Also, do we know how long before that would happen? Should we delete the content of `Releases/Release/index.jsx` if it's no longer used?

ui/src/views/Releases/ListReleases/index.jsx
```diff
@@ -533,18 +653,33 @@ function ListReleases(props) {
...
-      ...drawerState,
-      item: {
-        firstRelease: result.data.data,
-        secondRelease: release.scheduledChange.data,
-        firstFilename: `Data Version: ${release.data_version}`,
-        secondFilename: 'Scheduled Change',
-      },
-      open: true,
-    });
+    if (release.api_version === 1) {
```
helfi92, line 656: In one of the `useEffect` methods in `ListRules/index.jsx` we make a request to the v2 endpoint and if there is an error, we fallback to v1. Would it make sense to check the version first before firing up the request?

ui/src/views/Releases/ListReleases/index.jsx
```diff
@@ -337,11 +431,19 @@ function ListReleases(props) {
...
+            ret.read_only = !r.read_only;
+            ret.data_version = result.new_data_version;
+          }
+        } else if (result['.'].sc_id) {
+          ret.scheduledChange = buildScheduledChangeV2(ret, [result['.']]);
+          ret.scheduledChange.read_only = !r.read_only;
         } else {
           ret.read_only = !r.read_only;
-          ret.data_version = result.new_data_version;
+          ret.data_version = result['.'];
```
helfi92, line 446: What does a property field `.` mean?

mozbhearsum:
>  together affects readability a bit but I understand that this is a temporary thing. When can we expect v1 to be fully migrated to v2?

The plan is sometime in Q2. It's possible it may be a bit longer than that - but definitely this year.

> > I think there's a race or dependency issue in ListReleaseRevisionsV2, but I haven't been able to get to the bottom of it. It results in incorrect diffs being shown sometimes. I was hoping you may be able to spot something obvious that I've missed in the relevant useEffect.
> 
> Any chance you could give me some STR to replicate?

I can; you'll need to set up your own GCS buckets and write some data though. If you'd prefer, I can set up a public facing instance with everything needed.

ui/src/views/Rules/ListRules/index.jsx
mozbhearsum: Hm, we shouldn't be showing this button if you're logged out. I'm not sure if that's a new issue or not, but it's easy enough to fix here.

ui/src/views/Releases/ReleaseV2/index.jsx
mozbhearsum: Yeah, I'm going to remove all of the `V2` references after `V1` Release stuff goes away.

And yes! We can kill `Releases/Release/index.jsx` already.

ui/src/views/Releases/ListReleases/index.jsx
mozbhearsum: I believe you're talking about https://github.com/mozilla-releng/balrog/blob/012237085dee260f85da7192389088c82c41d445/ui/src/views/Rules/ListRules/index.jsx#L963? I could go either way on that -- checking the version first would require querying the list of releases for both API versions, which is data that is otherwise unused on that view.

ui/src/views/Releases/ListReleases/index.jsx
mozbhearsum: Ah, the new API returns responses that have either data versions or scheduled change information for 1 or more parts of the Release. `.` refers to the top level of the Release. Perhaps this should be a constant?

mozbhearsum:
I ended up setting up https://23.236.56.207:9000/ for you to play with. Unfortunately, I can't repro the race yet...which leads me to think it's only happening on my laptop because it's quite slow in comparsion.

ui/src/views/Releases/ListReleases/index.jsx
helfi92: Sure, I think that works. A descriptive field name would be more readable than a dot :)

ui/src/views/Releases/ListReleases/index.jsx
helfi92: Should be okay the way it is right now.

ui/src/views/Releases/ReleaseV2/index.jsx
helfi92: Awesome. Thanks for clarifying.

ui/src/views/Rules/ListRules/index.jsx
helfi92: Sounds good. It's possible that it was from before. Disabling it when logged out makes sense to me.

ui/src/views/Releases/ListReleaseRevisionsV2/index.jsx
```diff
@@ -0,0 +1,258 @@
...
+        axios.get(latestRevision.data_url).then(result => {
+          // FIXME: sometimes the page doesn't rerender the diff after this
+          // for some reason...
+          // I think there's some sort of race condition, or issue with the
+          // depedencies on this useEffect.
+          setRightRevisionData(result.data || {});
+        });
+      }
+    }
+  }, [leftRadioCheckedIndex, fetchedRevisions.data]);
```
helfi92, line 156: `olderRevisions`, `latestRevisions` are variables defined outside `useEffect` but are not listed in the list of dependencies in `useEffect`. Maybe that would fix the issue you were seeing?

mozbhearsum: I'll give that a shot!

ui/src/views/Rules/ListRules/index.jsx
mozbhearsum: I wasn't able to reproduce this myself, but there's clearly an unguarded access of `productChannelQueries` in this file, which I'm fixing.

mozbhearsum: Looks like this was actually caused by an invalid use of `SpeedDialAction` -- `disabled` needs to be in `FabProps`, not its own attribute. Looks like this one was missed during #1150.

ui/src/views/Releases/ListReleaseRevisionsV2/index.jsx
mozbhearsum: That was definitely part of the problem. After some further digging, I think this is actually caused by `DiffRelease`. Specifically, anytime any of the dependencies in https://github.com/mozilla-releng/balrog/blob/master/ui/src/components/DiffRelease/index.jsx#L87 change, it fires a new worker. When I change the left radio button, I see 3 workers run (one for `firstFilename`, one for `firstRelease`, and one for `secondRelease`) - and the final one ends up winning.

I'm guessing that we're getting a rerender of that component each time one of these changes, although I'm not sure how to prevent that.

helfi92: What do you think about the following. I think it might prevent the rerendering.

```diff
diff --git a/ui/src/components/DiffRelease/diff.worker.js b/ui/src/components/DiffRelease/diff.worker.js
index 4192ce8c..59e2fbbc 100644
--- a/ui/src/components/DiffRelease/diff.worker.js
+++ b/ui/src/components/DiffRelease/diff.worker.js
@@ -2,14 +2,12 @@ import { createTwoFilesPatch } from 'diff';
 
 onmessage = e => {
   const [
-    firstFilename,
-    secondFilename,
     firstReleaseString,
     secondReleaseString,
   ] = e.data;
   const releaseDiff = createTwoFilesPatch(
-    firstFilename,
-    secondFilename,
+    'Revision (left)',
+    'Revision (right)',
     firstReleaseString,
     secondReleaseString
   );
diff --git a/ui/src/components/DiffRelease/index.jsx b/ui/src/components/DiffRelease/index.jsx
index 7eb4ec47..dc3de26b 100644
--- a/ui/src/components/DiffRelease/index.jsx
+++ b/ui/src/components/DiffRelease/index.jsx
@@ -49,8 +49,6 @@ function DiffRelease(props) {
   const {
     firstRelease,
     secondRelease,
-    firstFilename,
-    secondFilename,
     className,
   } = props;
   const classes = useStyles();
@@ -79,12 +77,10 @@ function DiffRelease(props) {
 
   useEffect(() => {
     diffWorker.postMessage([
-      firstFilename,
-      secondFilename,
       JSON.stringify(deepSortObject(firstRelease), null, 2),
       JSON.stringify(deepSortObject(secondRelease), null, 2),
     ]);
-  }, [firstFilename, secondFilename, firstRelease, secondRelease]);
+  }, [firstRelease, secondRelease]);
 
   const handleRowRender = ({ index, key, style }) => {
     const line = releaseLinesDiff[index];
@@ -139,8 +135,6 @@ function DiffRelease(props) {
 DiffRelease.propTypes = {
   firstRelease: object.isRequired,
   secondRelease: object.isRequired,
-  firstFilename: string.isRequired,
-  secondFilename: string.isRequired,
   className: string,
 };
```

mozbhearsum: Hmmmm, I think that will help but it may not entirely fix the issue? We'll still get two renders, because `firstRelease` and `secondRelease` may both change, so if the first worker finishes after the second one, that will be wrong. I could be misunderstanding though.

helfi92: Ah, I think I misunderstood the issue. Correct me if I'm wrong, but you're looking for a way to not use the result of the worker if that diff belongs to a different pair of releases. If that's what you want to do, then maybe (let me know if it helps) we could either:
* add a cleanup function to useEffect that would terminate any ongoing worker before that useEffect runs again. (preferred)

```jsx
useEffect(() => {
 // ...
 return function cleanup() {
    diffWorker.terminate();
  };
});
```

* append another property (e.g., revision date) to the worker result and then check that it belongs to what the user selected and if it matches.

mozbhearsum: I didn't know about cleanup functions - that's really cool! I suspect there may be a tiny possibility of a race condition still (if it completes before the termination does), but it'll probably be perfectly fine for our use case, especially since I've been unable to repro anywhere but my laptop.

Thanks!

mozbhearsum: After some additional reading, I think I'm wrong about the race condition.

mozbhearsum:
The test failure from the most recent push is a known intermittent. I think this is ready to go now, @helfi92?

ui/src/views/Releases/ListReleaseRevisions/index.jsx
```diff
@@ -64,7 +64,7 @@ function ListReleaseRevisions(props) {
 
   useEffect(() => {
     if (fetchedRelease.data) {
-      fetchRevisions(releaseName, fetchedRelease.data.data.product);
+      fetchRevisions(releaseName, 1);
```
helfi92, line 67: I assume this was intentional but just in case, doesn't hurt to bring it up :)

mozbhearsum: Yup. The interface to `getRevisions` now requires the api version to be passed, and doesn't need product: https://github.com/mozilla-releng/balrog/blob/5324a2fb49e0e650ca1e2fe26785449a87a2df6f/ui/src/services/releases.js#L40

PR #1274, opened 2020-04-07
Use synchronous GCS library for old Releases API

mozbhearsum:
When we pushed the latest Balrog code to prod today we discovered that the old Releases API was unable to write to GCS due to errors such as `errorThere is no current event loop in thread 'uWSGIWorker2Core5'.`

This is happening because when the `gcloud-aio` `Storage` class is instantiated, it requires a `aiohttp.ClientSession` (if you don't give it one, it creates one). That class calls `get_event_loop`, which is what's hitting this error. I tried a few things to fix this while still having `GCSHistory` use the async code, but nothing obvious worked, and the truth is that there's no reason we _have_ to use `gcloud-aio` for this class.

Because of this, I modified the BucketFactory to be able to return the sync or async Bucket class. This adds some temporary complexity, but all of the the sync GCS code is dying with the old Releases table -- so that's temporary.

### DIFF
--- a/src/auslib/db.py
+++ b/src/auslib/db.py
@@ -1,4 +1,3 @@
-import asyncio
 import itertools
 import json
 import logging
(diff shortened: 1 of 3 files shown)


mozbhearsum:
I should also note that I did explicit local tests of:
* insert, update, delete, and merge (which calls `getChange`) with the old releases API
* insert, update, and delete with the new releases API

I'll be running a staging release after this heads to stage before considering a prod deploy (something I should've done with the last attempt :( )

PR #1333, opened 2020-05-13
Add tests to make sure old releases aren't included in new tables and vice versa

mozbhearsum:
This is currently causing 500s in stage, because requests to `/api/releases` are trying to find signoff requirements for v2 releases.


mozbhearsum:
> Seems OK if we fully migrate old releases to the new tables. If not, the v2 API should at least read the old tables, even if not vice versa.

Fully migrating (and dropping the old tables/API) has always been the intention. Let's assume that for now, and we can adjust course if we end up not proceeding with that (or delaying for too long).

PR #1338, opened 2020-05-15
Don't query new releases scheduled changes tables in public app

### LINKED ISSUE #1337: Fix performance issues on public app with new releases tables enabled
When we enabled the new releases tables for the public app we saw roughly a ~3x spike in database load, and a smaller spike in CPU load. Two possible theories (although there may be other explanations:
1) caching of empty results from the new releases tables are broken
2) simply querying the new releases tables is causing this increase in load

I would be surprised if it was 2), as the caching should largely negate this. It's still possible though.

mozbhearsum:
I think this will fix #1337 based on some local analysis. I still want to do some verification with query logging enabled on stage before I feel confident, though.


mozbhearsum:
I've done the verification I can without this in stage. I'll do some more after it lands.

PR #1345, opened 2020-05-21
Log data about requests on BigQuery

### LINKED ISSUE #1048: log data about requests
We've long wanted to log information about what each request to Balrog is served (see https://bugzilla.mozilla.org/show_bug.cgi?id=758373). At the very least, what version we served the request, and whether or not the request got served the primary mapping, or the fallback mapping.

At this point in time, the most sensible place to put it would be BigQuery.

It's worth noting that even if/when we start doing this in the Balrog app, we still have both an nginx and a Cloudfront cache in front of it, so the data won't actually contain information about _all_ updates served, just those that make it to the app.

allan-silva:
Fixes #1048

- *updatetrack.configure_app* registers the public app to be tracked on request app context.
- *UpdateTrack* has helper methods to capture information into request app context.

### DIFF
--- a/src/auslib/AUS.py
+++ b/src/auslib/AUS.py
@@ -106,7 +106,7 @@ class AUS:
             if updateQuery["force"] == FORCE_FALLBACK_MAPPING or self.rand() >= rule["backgroundRate"]:
                 fallbackReleaseName = rule["fallbackMapping"]
                 if fallbackReleaseName:
-                    release = releases.get_release(fallbackReleaseName, transaction)
+                    release = releases.get_release(fallbackReleaseName, transaction, include_sc=False)
                     blob = None
                     if release:
                         blob = createBlob(release["blob"])
@@ -125,7 +125,7 @@ class AUS:
         # 3) Incoming release is older than the one in the mapping, defined as one of:
         #    * version decreases
         #    * version is the same and buildID doesn't increase
-        release = releases.get_release(rule["mapping"], transaction)
+        release = releases.get_release(rule["mapping"], transaction, include_sc=False)
         blob = None
         if release:
             blob = createBlob(release["blob"])
--- a/src/auslib/blobs/apprelease.py
+++ b/src/auslib/blobs/apprelease.py
@@ -88,7 +88,7 @@ class ReleaseBlobBase(XMLBlob):
         # even attempt to look it up.
         if patch["from"] != "*":
             try:
-                release = releases.get_release(patch["from"], None)
+                release = releases.get_release(patch["from"], None, include_sc=False)
                 if release:
                     return createBlob(release["blob"])
                 else:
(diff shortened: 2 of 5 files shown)


src/auslib/web/public/updatetrack.py
```diff
@@ -0,0 +1,36 @@
...
+
+    @staticmethod
+    def set_update_xml(xml):
+        g.update_track.update_xml = xml
+
+
+def configure_app(app):
+    @app.before_request
+    def before_request():
+        g.update_track = UpdateTrack(request.full_path)
```
mozbhearsum, line 30: Hmm, I'm not sure I understand the need for the `UpdateTrack` object. As far as I know, everything we need to send to BigQuery is available in `after_request`, so we should just be able to do everything there...

...however, `after_request` blocks the request from returning, which means our update responses will be slower, and will potentially break if the submission to BigQuery does not work.

Can you look into whether it's possible to do the BigQuery submission after the request returns?

allan-silva: If you are interested only in url parameters and the result XML, it is OK to collect this data on `after_request`.

But if do you want to include intermediate values, evaluated on request process, UpdateQuery object can keep this data until `after_request` reached.

Example here:

https://github.com/mozilla-releng/balrog/blob/86c29f36e2045f085c472eefe0d59306d04027fa/tests/web/test_updatetrack.py#L19-L24

The lines 21 and 22, captures values evaluated in the view function (between `before_request` and `after_request`).


> Can you look into whether it's possible to do the BigQuery submission after the request returns?

Yeah, I realized this problem too, I'm looking for if it is possible trigger an event outside Request context.


Let me know if I was clear about `UpdateTrack`, if the intermediate values are not necessary, no point to keep the object.

mozbhearsum: > Let me know if I was clear about UpdateTrack, if the intermediate values are not necessary, no point to keep the object.

I think you're right about the intermediate values; mapping & fallback mapping are not available after the request has been handled, so we will need to store them partway through handling the request. Do you have any sense which parts of the existing code will end up calling `UpdateTrack`s setters?

> Yeah, I realized this problem too, I'm looking for if it is possible trigger an event outside Request context.

Let me know if you hit an impass here and we can look at it together.

allan-silva: > Do you have any sense which parts of the existing code will end up calling UpdateTracks setters?

Probably in some place at `client.get_update_blob` or `AUS.evaluateRules` when the flow is in a request context.

> Let me know if you hit an impass here and we can look at it together.

I'm looking for a pure flask way, before look into other solutions. My next step will be test flask signals like `request_tearing_down` and `appcontext_tearing_down`.

mozbhearsum: > Probably in some place at client.get_update_blob or AUS.evaluateRules when the flow is in a request context.

Sounds good. Let's keep it in `get_update_blob` if we can, since that's the view code.

PR #1378, opened 2020-06-25
Update most dependencies

mozbhearsum:
`pip-tools` has to be held back due to weirdness with `gcloud-aio` and `typing`. `gcloud-aio` ends up installing `typing` even though it's part of the standard library now, and the wrong version gets used. There's an upstream bug on `gcloud-aio` to stop installing it.


src/auslib/blobs/base.py
```diff
@@ -66,8 +66,8 @@ def merge_lists(*lists):
     lists in blobs as sets. In an ideal world, that's what they'd be, but
     because we use jsonschema for validation, we cannot use proper sets."""
     result = []
-    for l in lists:
-        for i in l:
+    for list in lists:
+        for i in list:
```
rail, line 70: I wish there was something like `flatten` in stdlib...

PR #1382, opened 2020-07-02
Use new domains for VPN

Callek:


rlr:
LGTM. I guess we are stuck with `FirefoxVPN` or could that change to `MozillaVPN`?

mozbhearsum:
> LGTM. I guess we are stuck with `FirefoxVPN` or could that change to `MozillaVPN`?

If we changed what the client sends to `MozillaVPN` we can change it. We'll need to support `FirefoxVPN` for awhile afterwards though. From a balrog standpoint, it's no big deal.

Callek:
I discussed with @rlr - the product still uses `FirefoxVPN` for update requests. This is changeable but not intended to be changed at this time, as its not user facing. Ricky does know that should that change we'll need to update Balrog to support it.

This can land as-is.

PR #1470, opened 2020-09-09
[wip] make it possible to rewind the rules UI

bhearsum:
This is a proof of concept I put together when originally working on the new UI. Surprisingly, the backend bits still apply cleanly and work. The frontend bits will probably apply with a bit of work - I've included them as diff files for now because I don't have time to deal with that at the moment. When it was written, it was working (although ugly and had a few bugs).

This would be super helpful for _any_ case where we're trying to find out what was served in the past -- something we've done a few times in the past few weeks.


bhearsum:
Superceded by the great work that @michellemounde did in #3079 !

PR #1472, opened 2020-09-09
Fix bad comparison of cached data versions with data versions of cached objects

mozbhearsum:
We were previously comparing objects like: `[1, 1, 1, 1]` to `[{'data_version': 1, ...}]` - which resulted in invalidating the cache every single request.

I tried to add a test for this but I couldn't figure out a way to both count the number of calls to `dbo.release_assets.select` _and_ still have it perform the query. I really want a test for this, but it's a big enough footgun that I think it's a bad idea to block on it for long. @escapewindow - maybe you have some idea of how to do this?

### DIFF
--- a/src/auslib/services/releases.py
+++ b/src/auslib/services/releases.py
@@ -299,35 +299,38 @@ def get_releases(trans):
     return {"releases": sorted(releases, key=lambda r: r["name"])}
 
 
+def get_base_row(name, trans):
+    row = dbo.releases_json.select(where={"name": name}, transaction=trans)
+    if row:
+        return row[0]
+
+    return None
+
+
+def get_base_data_version(name, trans):
+    row = dbo.releases_json.select(where={"name": name}, columns=[dbo.releases_json.data_version], transaction=trans)
+    if row:
+        return row[0]["data_version"]
+
+    return None
+
+
+def get_asset_rows(name, trans):
+    return dbo.release_assets.select(where={"name": name}, order_by=[dbo.release_assets.path], transaction=trans) or []
+
+
+def get_asset_data_versions(name, trans):
+    return (
+        dbo.release_assets.select(where={"name": name}, columns=[dbo.release_assets.data_version], order_by=[dbo.release_assets.path], transaction=trans) or []
+    )
+
+
 def get_release(name, trans, include_sc=True):
-    def get_base_row():
-        row = dbo.releases_json.select(where={"name": name}, transaction=trans)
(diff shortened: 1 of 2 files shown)


escapewindow:
> We were previously comparing objects like: `[1, 1, 1, 1]` to `[{'data_version': 1, ...}]` - which resulted in invalidating the cache every single request.

Ouch.

> I tried to add a test for this but I couldn't figure out a way to both count the number of calls to `dbo.release_assets.select` _and_ still have it perform the query. I really want a test for this, but it's a big enough footgun that I think it's a bad idea to block on it for long. @escapewindow - maybe you have some idea of how to do this?

A few thoughts. First off, agreed. I'm ok with landing a fix and leaving tests as a followup. However, there should be some ways we can do this.

1. If we were able to somehow shadow or mock `get_asset_rows` in the tests, the mocking function could both count and perform the query. This is more difficult because `get_asset_rows` is a function inside of a function. We could make it more easily mockable by moving the function to the top level of the module.

2. We could run two tests: one to test the behavior of `get_asset_rows`, and the other to count how many times we call `get_asset_rows`. That may mean we only mock the function in one of the two, but it's still dependent on mocking.

3. Mocking is just one way. We could also have a kwarg that points at the function to call. This is something I do that Tom sees as an antipattern, so you may not want to do that, but, for example,

    ```
    def get_release_asset_rows():
        ...

    def get_release(name, trans, include_sc=True, get_asset_rows=get_release_asset_rows):
        ...
        get_asset_rows()
        # To test with different behavior, e.g. counting, you can define a new function and pass it in as `get_asset_rows`
    ```

4. We can add the counter to the production code:

    ```
    def get_release(name, trans, include_sc=True, counter=None):
        ...
        get_asset_rows()
        if counter is not None:
            # increment counter int or counter obj
    ```

escapewindow (approved):

escapewindow:
>     1. If we were able to somehow shadow or mock `get_asset_rows` in the tests, the mocking function could both count and perform the query. This is more difficult because `get_asset_rows` is a function inside of a function. We could make it more easily mockable by moving the function to the top level of the module.

By this I mean something like

```
def test_foo(mocker):
     def count_and_perform_get_asset_rows():
         # increment counter
         # call real get_asset_rows, and/or do the same thing, and return the value
    mocker.patch(...)
```

escapewindow:
We could also mock `dbo`, maybe. To something that counts and also returns the value from the real dbo.

PR #1483, opened 2020-09-12
add visibility to pending rule changes

hngerebara:
Fix to this [issue](https://github.com/mozilla-releng/balrog/issues/1224)

Previously when a change is made from release to release* it doesnt show up in the **localtest and cdntest** channels as shown below.
Screenshot 2020-09-12 at 2 17 43 PM

Now fixed as show below

Screenshot 2020-09-12 at 2 13 14 PM

Screenshot 2020-09-12 at 2 11 56 PM

### DIFF
--- a/ui/src/views/Rules/ListRules/index.jsx
+++ b/ui/src/views/Rules/ListRules/index.jsx
@@ -460,11 +460,13 @@ function ListRules(props) {
       }
 
       if (channelFilter) {
-        if (ruleChannel.indexOf('*') === -1) {
-          if (ruleChannel !== channelFilter) {
-            return false;
-          }
-        } else if (!channelFilter.startsWith(ruleChannel.split('*')[0])) {
+        let channelRoot = channelFilter.split('-')[0];
+        if (ruleChannel === channelFilter
+          || ruleChannel.startsWith(channelRoot) 
+          && !ruleChannel.includes('-')) {
+            return true;
+        }
+        else {
           return false;
         }
       }
--- a/ui/src/views/Rules/ListRules/index.jsx
+++ b/ui/src/views/Rules/ListRules/index.jsx
@@ -452,20 +452,30 @@ function ListRules(props) {
       const [productFilter, channelFilter] = productChannelQueries;
       const ruleProduct =
         rule.product || (rule.scheduledChange && rule.scheduledChange.product);
-      const ruleChannel =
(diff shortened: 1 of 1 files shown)


mozbhearsum:
Hi @hngerebara - thank you for this! I will look at it as soon as I can. It may not be until next week though.

mozbhearsum (changes requested):
While this works for the very specific bug, I don't think it is the right fix. The root issue here is that the comparisons on line 464 and 467 only look at both `rule.scheduledChange` and `rule.scheduledChange.channel` (if scheduledChange is not null) instead of just one of them (ruleChannel is defined on line 456 as one or the other).

hngerebara:
Ooh I see. I have made some modifications to that logic which fixes the issue

ui/src/views/Rules/ListRules/index.jsx
```diff
@@ -452,13 +452,24 @@ function ListRules(props) {
...
+      const ruleScheduledChange =
+        rule.scheduledChange && rule.scheduledChange.channel;
+      let ruleChannel;
 
       if (ruleProduct !== productFilter) {
         return false;
       }
 
+      if (ruleScheduledChange) {
+        if (ruleScheduledChange.includes('*')) {
```
mozbhearsum, line 464: I don't think is right -- you want to check `ruleScheduledChange.channel` for this, not the entire rule.

mozbhearsum (changes requested):
This still needs a bit of work. Here's how to make sure it's working correctly:
* Modify https://localhost:9000/rules/628 to change `release-localtest` to `release*`, and make the scheduled time at least a day in the future.
* Make sure the scheduled change is visibile on https://localhost:9000/rules?product=Firefox&channel=release-localtest, https://localhost:9000/rules?product=Firefox&channel=release-cdntest, and https://localhost:9000/rules?product=Firefox&channel=release, and not visible on https://localhost:9000/rules?product=Firefox&channel=beta

ui/src/views/Rules/ListRules/index.jsx
hngerebara: @bhearsum `ruleScheduledChange` returns just the channel not the entire rule. I have also tested it out as you indicated and it works as it should

mozbhearsum (approved):

PR #1510, opened 2020-09-23
Elide response logs from v2 api responses when they're too long

mozbhearsum:


client/src/balrogclient/api.py
```diff
@@ -107,7 +107,7 @@ def balrog_request(session, method, url, *args, **kwargs):
         resp.raise_for_status()
         if resp.content:
             recieved_data = resp.json()
-            log.info("Data recieved: %s", recieved_data)
+            log.info("Data recieved: %s", _json_log_data(recieved_data))
```
escapewindow, line 110: Non-blocking nit SP: `received` :)

escapewindow (approved):

client/src/balrogclient/api.py
mozbhearsum: oops :). let's leave that for later

PR #1516, opened 2020-09-25
Log GCS upload duration

mozbhearsum:
The actual log entries I'm 99% sure will work - I was able to verify them locally. I'm less certain about the level changes, but everything is logged as error right now, so this probably can't make things worse...


escapewindow (approved):
Non-blocking: There may be a module that does something like this:
- https://medium.com/pythonhive/python-decorator-to-measure-the-execution-time-of-methods-fa04cb6bb36d
- https://realpython.com/lessons/timing-functions-decorators/
- https://stackoverflow.com/questions/1622943/timeit-versus-timing-decorator
- etc.

Blocking, I think: test bustage

PR #1523, opened 2020-09-28
Look through platform alias' from current release when evaluating pre…

mozbhearsum:
…vious releases for a partial update eligibility

Ideally, we'd fix the client submit alias' for dated blobs, but that's a
much trickier fix, and probably not a worthwhile one given that after we
complete nightly shippable/release convergence work, that code will just
go away.

In theory, this fix could cause us to attempt to serve a partial that's
not valid for a request. This can only happen if we change the meaning a
build target at some point (so that the alias' from the current blob
don't match expectations of the earlier versions). This has never
happened in my time here -- we always add a new build target instead. If
this does happen at some point, the failure mode would be the partial
failing to apply...so I think it's an OK risk to take here.


mozbhearsum:
This is a fix for https://bugzilla.mozilla.org/show_bug.cgi?id=1667667

escapewindow (approved):
Nice, thanks for adding the test!

PR #1699, opened 2021-01-04
Update python deps

bhearsum:


jmaher (approved):
a bit of rubberstamping

PR #1738, opened 2021-01-19
Improve public API queries for releases when names_only is true

bhearsum:
Locally, this cut the time for `names_only=1` queries in half.

### DIFF
--- a/src/auslib/services/releases.py
+++ b/src/auslib/services/releases.py
@@ -244,6 +244,14 @@ def get_assets(name, trans):
     return assets
 
 
+def get_release_names(trans, product=None):
+    where = {}
+    if product:
+        where["product"] = product
+
+    return sorted([r["name"] for r in dbo.releases_json.select(columns=[dbo.releases_json.name], where=where)])
+
+
 def get_releases(trans):
     releases = dbo.releases_json.select(
         columns=[dbo.releases_json.name, dbo.releases_json.product, dbo.releases_json.data_version, dbo.releases_json.read_only], transaction=trans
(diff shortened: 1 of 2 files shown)


src/auslib/services/releases.py
```diff
@@ -244,6 +244,14 @@ def get_assets(name, trans):
     return assets
 
 
+def get_release_names(trans, product=None):
```
escapewindow, line 247: Hm, do we want to cache the result? I imagine the answer might be "no" if the DB changed, and "maybe" or "yes" if the DB hasn't, depending on how expensive it is.

escapewindow (approved):
I'm not terribly familiar with this, so up to you if you want a deeper code review.

src/auslib/services/releases.py
bhearsum: I thought about this, but I think ultimately it won't win much, if anything. We have so many webheads (and they recycle) and so few requests (once per nightly decision task AFAICT), that the hit rate will be zero, or close enough to it.

src/auslib/services/releases.py
```diff
@@ -244,6 +244,14 @@ def get_assets(name, trans):
...
+    where = {}
+    if product:
+        where["product"] = product
+
+    return sorted([r["name"] for r in dbo.releases_json.select(columns=[dbo.releases_json.name], where=where)])
```
jcristau, line 252: does this intentionally not pass `transaction=trans` down?

bhearsum: No :). Thank your for catching that.

PR #1790, opened 2021-02-12
Fix logic for showing/hiding scheduled changes with a channel filter applied

bhearsum:
@rehandalal discovered a crash in the UI yesterday that's triggered when you have a scheduled insert to the Rules. This is a regression from #1483 where we tried to make some enhancements to make sure scheduled changes were always shown in the right places.

This PR fixes the crash without regressing the new feature. I'm not terribly happy with the code, but I had trouble getting it working in a more compact form.

### DIFF
--- a/ui/src/utils/rules.js
+++ b/ui/src/utils/rules.js
@@ -5,7 +5,7 @@ const ruleMatchesChannel = (rule, channel) => {
   // similarly, if the rule or scheduled rule has a wildcard, and the
   // selected channel does not match either, this rule does not match
   const ruleChannel = rule.channel;
-  const ruleScheduledChangeChannel =
+  const scChannel =
     rule.scheduledChange && rule.scheduledChange.channel;
   let ruleChannelMatches = false;
   let scChannelMatches = false;
@@ -18,23 +18,29 @@ const ruleMatchesChannel = (rule, channel) => {
     } else if (channel.startsWith(ruleChannel.split('*')[0])) {
       ruleChannelMatches = true;
     }
-  }
+  } else {
+      ruleChannelMatches = true;
+    }
 
-  if (ruleScheduledChangeChannel) {
-    if (ruleScheduledChangeChannel.indexOf('*') === -1) {
-      if (ruleScheduledChangeChannel === channel) {
+  if (scChannel) {
+    if (scChannel.indexOf('*') === -1) {
+      if (scChannel === channel) {
         scChannelMatches = true;
       }
-    } else if (channel.startsWith(ruleScheduledChangeChannel.split('*')[0])) {
+    } else if (channel.startsWith(scChannel.split('*')[0])) {
       scChannelMatches = true;
     }
+  } else {
+      if (rule.scheduledChange) {
+      scChannelMatches = true;
+      }
   }
 
-  if (!ruleChannelMatches && !scChannelMatches) {
-    return false;
+  if (ruleChannelMatches || scChannelMatches) {
+    return true;
   }
 
-  return true;
+  return false;
 };
 
 export {
--- a/ui/src/utils/rules.js
+++ b/ui/src/utils/rules.js
@@ -6,12 +6,13 @@ const ruleMatchesChannel = (rule, channel) => {
     r && r.includes('*') && c.startsWith(r.split('*')[0]);
   const ruleChannelMatches =
     // empty or absent channel matches anything
-    // however, a rule could also be non-existent (if a scheduled change is an insert)
+    // however, a rule could also be non-existent
+    // (if a scheduled change is an insert)
     // in this case, channel will be undefined, and we should _never_
     // match on that, otherwise non-existent rules would show up
     // on all filters.
     rule.channel === null ||
-    rule.channel === "" ||
+    rule.channel === '' ||
     rule.channel === channel ||
     matchesGlob(rule.channel, channel);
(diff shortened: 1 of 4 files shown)


ui/src/views/Rules/ListRules/index.jsx
```diff
@@ -454,28 +454,44 @@ function ListRules(props) {
...
+        let scChannelMatches = false;
+
+        if (ruleChannel) {
+          if (ruleChannel.indexOf('*') === -1) {
+            if (ruleChannel === channelFilter) {
+              ruleChannelMatches = true;
+            }
+          } else if (channelFilter.startsWith(ruleChannel.split('*')[0])) {
+            ruleChannelMatches = true;
+          }
```
sarah-clements, line 479: Could this be simplified to this?
`const ruleChannelMatches = ruleChannel && ruleChannel.includes(channelFilter);`

sarah-clements: And then:
`const scChannelMatches = ruleScheduledChangeChannel && ruleScheduledChangeChannel.includes(channelFilter)`

ui/src/views/Rules/ListRules/index.jsx
```diff
@@ -454,28 +454,44 @@ function ListRules(props) {
...
       if (ruleProduct !== productFilter) {
         return false;
       }
 
-      if (ruleScheduledChangeChannel) {
-        if (ruleScheduledChangeChannel.includes('*')) {
-          ruleChannel = rule.channel && ruleScheduledChangeChannel;
-        } else {
-          ruleChannel = ruleScheduledChangeChannel && rule.channel;
+      if (channelFilter) {
```
sarah-clements, line 463: I think you could write this as `if (!channelFilter) return false;` here so that on the last line of this block you change it to `return ruleChannelMatches && scChannelMatches;` (and remove the condition on line 494). :)

sarah-clements:
I didn't actually try out my suggestions in the codebase so I might be missing something :)

ui/src/views/Rules/ListRules/index.jsx
sarah-clements: Or would it be `return ruleChannelMatches || scChannelMatches;`? Not entirely clear on that - it sounded like in the relounge channel maybe it should be OR logic??

ui/src/views/Rules/ListRules/index.jsx
rehandalal, line 479: Also a nice shorthand is to use [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) (`?.`):
```js
const ruleChannelMatches = ruleChannel?.includes(channelFilter);
```

ui/src/views/Rules/ListRules/index.jsx
bhearsum: Definitely `||` (we want the rule to be displayed if it's _current_ channel matches OR it matches what's in a scheduled change, if it exists).

ui/src/views/Rules/ListRules/index.jsx
bhearsum: TIL! Thank you both for your suggestions; I'll take a crack at improving this.

rehandalal: I did realize that just using includes may produce false matches for things like filtering by `release` but seeing `release-sysaddon` as well. i think this is probably a better solution:

```js
const ruleChannelMatches = ruleChannel?.split("*").includes(channelFilter);
```

bhearsum: This looks like, but unfortunately our build doesn't work with modern node versions right now so I had to forego it =\

bhearsum:
I want to test the latest code a bit more by hand, but the test coverage is decent, and helped me simplify the code a lot.

ui/src/utils/rules.js
```diff
@@ -0,0 +1,38 @@
...
+      rule.scheduledChange.channel === '' ||
+      rule.scheduledChange.channel === channel ||
+      matchesGlob(rule.scheduledChange.channel, channel)
+    : false;
+
+  if (ruleChannelMatches || scChannelMatches) {
+    return true;
+  }
+
+  return false;
```
sarah-clements, line 32: Couldn't you just do `return ruleChannelMatches || scChannelMatches`?

ui/src/utils/rules.js
```diff
@@ -0,0 +1,38 @@
...
+    rule.channel === null ||
+    rule.channel === '' ||
+    rule.channel === channel ||
+    matchesGlob(rule.channel, channel);
+  // if a scheduled change does not exist at all
+  // we never want this to match, otherwise all rules
+  // without scheduled changes will always match any filter
+  const scChannelMatches = rule.scheduledChange
+    ? rule.scheduledChange.channel === null ||
+      rule.scheduledChange.channel === '' ||
```
sarah-clements, line 23: I think `!rule.scheduledChange.channel` should cover all bases (empty strings are falsey). :)

ui/src/utils/rules.js
bhearsum: Good point!

bhearsum: :)

sarah-clements (approved):

PR #1826, opened 2021-03-03
Return evaluated Rule-ID and Rule-Data-Version to construct response

### LINKED ISSUE #1092: return http header with release name(s) used to construct response
This idea came up today and shortly afterwards I was helping debug why nothing was being returned for a particular update URL. It turned out that a different rule was being matched than we thought. Had we had the release name somewhere in the response (like a header!), this would've been much more obvious.

(Imported from https://bugzilla.mozilla.org/show_bug.cgi?id=1425594)

allan-silva:
Closes #1092

### DIFF
--- a/src/auslib/AUS.py
+++ b/src/auslib/AUS.py
@@ -73,21 +73,27 @@ class AUS:
         self.log.debug("Looking for rules that apply to:")
         self.log.debug(updateQuery)
 
+        eval_metadata = {}
+
         if self.updates_are_disabled(updateQuery["product"], updateQuery["channel"], transaction) or self.updates_are_disabled(
             updateQuery["product"], getFallbackChannel(updateQuery["channel"]), transaction
         ):
             log_message = "Updates are disabled for {}/{}.".format(updateQuery["product"], updateQuery["channel"])
             self.log.debug(log_message)
-            return None, None
+            return None, None, eval_metadata
 
         rules = dbo.rules.getRulesMatchingQuery(updateQuery, fallbackChannel=getFallbackChannel(updateQuery["channel"]), transaction=transaction)
 
         # TODO: throw any N->N update rules and keep the highest priority remaining one?
         if len(rules) < 1:
-            return None, None
+            return None, None, eval_metadata
 
         rules = sorted(rules, key=lambda rule: rule["priority"], reverse=True)
         rule = rules[0]
+
+        eval_metadata["rule"] = rule["rule_id"]
+        eval_metadata["rule_data_version"] = rule["data_version"]
+
         self.log.debug("Matching rule: %s" % rule)
 
         # There's a few cases where we have a matching rule but don't want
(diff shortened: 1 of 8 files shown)


src/auslib/web/public/json.py
```diff
@@ -9,15 +9,22 @@
 from auslib.web.public.helpers import AUS, with_transaction
 
 
+def get_aus_metadata_headers(eval_metadata):
+    header_name_metadata_map = {"rule": "Rule-ID", "rule_data_version": "Rule-Data-Version"}
+    headers = {header_name_metadata_map.get(name, name): value for name, value in eval_metadata.items()}
+    return headers
+
+
```
allan-silva, line 17: This is the same as `client.py` `get_aus_metadata_headers` function. Let me know if is better move to a commom place.

src/auslib/AUS.py
```diff
@@ -73,29 +73,35 @@ def evaluateRules(self, updateQuery, transaction=None):
         self.log.debug("Looking for rules that apply to:")
         self.log.debug(updateQuery)
 
+        eval_metadata = {}
```
bhearsum, line 76: I think it would be good to set defaults for rule_id & rule_data_version here, for a couple of reasons:
1) To make it more obvious what's expected in this data structure
2) To guard against a caller that may try to look for one of the above even if they're unset.

Let's default them to `unknown`

src/auslib/AUS.py
```diff
@@ -73,29 +73,35 @@ def evaluateRules(self, updateQuery, transaction=None):
...
 
         # TODO: throw any N->N update rules and keep the highest priority remaining one?
         if len(rules) < 1:
-            return None, None
+            return None, None, eval_metadata
 
         rules = sorted(rules, key=lambda rule: rule["priority"], reverse=True)
         rule = rules[0]
+
+        eval_metadata["rule"] = rule["rule_id"]
```
bhearsum, line 94: Let's use `rule_id` for the key here (rule already has a meaning, let's not overload it).

src/auslib/web/public/json.py
bhearsum: https://github.com/mozilla-releng/balrog/blob/main/src/auslib/web/public/helpers.py is probably a good place for it.

src/auslib/web/public/json.py
```diff
@@ -9,15 +9,22 @@
 from auslib.web.public.helpers import AUS, with_transaction
 
 
+def get_aus_metadata_headers(eval_metadata):
+    header_name_metadata_map = {"rule": "Rule-ID", "rule_data_version": "Rule-Data-Version"}
```
bhearsum, line 13: I think this would be easier to read if it was simply written as:
```
return {
    "Rule-ID": eval_metadata["rule_id"]
...
}
```

I don't see any benefit to the indirection - but please let me know if there's a reason to keep it that I'm missing.

bhearsum (changes requested):

src/auslib/web/public/json.py
allan-silva: A map was used to handle new evaluation metadata entries that is not mapped to a header, so the header name fallsback to the key. But how we will make `eval_metadata` values explicity set to `unknown`, it is safe to directly acces the key.

bhearsum (approved):
This looks great; thank you!

PR #1843, opened 2021-03-09
Include releases_json table on database dump

allan-silva:


bhearsum (approved):
Sorry for the delay here. This largely looks fine, but I got this error when trying to run it locally:
```
mysqldump: Error: 'Access denied; you need (at least one of) the PROCESS privilege(s) for this operation' when trying to dump tablespaces
```

I think the `--no-tablespaces` option to mysqldump should fix it. I'll make that edit myself, and then we can merge this.

allan-silva:
Weird, I was able to generate dump, but I completely deleting the db and admin images before run.

bhearsum:
> Weird, I was able to generate dump, but I completely deleting the db and admin images before run.

I suspect you're running an older mysql - this restriction is relatively recent.

PR #1924, opened 2021-04-14
add GlobVersion to support version globs.

escapewindow:
For the System Addons in Balrog work, we currently find ourselves
needing to specify a `<next_version` rule for every Firefox version,
just to limit the rules to a single Firefox major version, even if many
of the rules are noop.

We can target our rules more selectively if we're able to use a
`version: 84.*` rather than rely on having both a `version: <84` and a
`version: <85` rule.


src/auslib/util/comparison.py
```diff
@@ -15,8 +15,8 @@ def has_operator(value):
 
 
 def get_op(pattern):
-    # only alphanumeric characters means no operator
-    if re.match(r"\w+", pattern):
+    # only alphanumeric or glob characters means no operator
+    if re.match(r"[\w\*]+", pattern):
```
bhearsum, line 19: I'm wondering if it might be cleaner to treat `*` as an operator, and return a function that implements the comparison, rather than special casing `*` down in `MozillaVersion`. What do you think?

escapewindow:
Superseded by #1928.

escapewindow:
Reopening to see if this approach ends up being cleaner.

escapewindow:
@bhearsum I'm able to update rules to, say, `version: 51.*` locally with no ISE500. The backend tests are hitting ISE500 but I don't get any logging why. Do you have a trick to debug this?

bhearsum:
> @bhearsum I'm able to update rules to, say, `version: 51.*` locally with no ISE500. The backend tests are hitting ISE500 but I don't get any logging why. Do you have a trick to debug this?

Yes! Try removing the exception eater: https://github.com/mozilla-releng/balrog/blob/main/src/auslib/web/admin/base.py#L134

escapewindow (commented):
Oh, I didn't notice these comments were pending?

src/auslib/util/versions.py
```diff
@@ -43,10 +43,46 @@ class AncientMozillaVersion(StrictVersion):
...
+
+    def __eq__(self, value):
+        return str(value).startswith(self.vstring.rstrip("*"))
+
+
 def MozillaVersion(version):
     try:
         if version.count(".") in (1, 2):
-            if int(version[0]) > 4:
+            if int(version.split(".")[0]) > 4:
```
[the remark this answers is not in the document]
bhearsum: Yeah, I think you're right.

src/auslib/util/versions.py
```diff
@@ -43,10 +43,49 @@ class AncientMozillaVersion(StrictVersion):
     )
 
 
+class GlobVersionTuple(tuple):
```
bhearsum, line 46: I'm probably being dense here, but why does this need to exist? AFAICT it behaves the same a regular tuple:
```
>>> (1, 3, 5) == (1, 3, 5)
True
>>> 
>>> class GlobVersionTuple(tuple):
    def __eq__(self, value):
        for i, j in enumerate(self):
            if j != value[i]:
                return False
        return True

    def __ne__(self, value):
        return not self.__eq__(value)
>>> 
>>> 
>>> gvt = GlobVersionTuple((1,3,5))
>>> gvt2 = GlobVersionTuple((1,3,5))
>>> gvt == gvt2
True
```

src/auslib/util/versions.py
```diff
@@ -43,10 +43,49 @@ class AncientMozillaVersion(StrictVersion):
...
+        return not self.__eq__(value)
+
+
+class GlobVersion(StrictVersion):
+    """A version class that supports Firefox versions 5.0 and up, which ends
+    with a glob `*`. Not really a StrictVersion at all, but it needs to be
+    to compare with other MozillaVersions."""
+
+    version_re = re.compile(
+        r"""^(\d+) \. (\d+\.\*|\*)$""",
```
bhearsum, line 63: Feel free to shoot me done if you think this is YAGNI, but I'm surprised we're only supporting globbing two part versions when we could easily support three part versions as well.

bhearsum (commented):
Couple small questions below, but this looks largely fine.

src/auslib/util/versions.py
escapewindow: We're comparing a version against a glob tuple. So

`GlobVersion("89.8.*")` will get a `GlobVersionTuple` of `(89, 8)`. `(89, 8) != (89, 8, 1)` even though `89.8.*` should match `89.8.1`. So basically `GlobVersionTuple` makes sure `version_tuple[:len(glob_tuple)] == glob_tuple`, but that leads to infinite recursion in the `__eq__` function, so instead I enumerate the `glob_tuple` and make sure all of its version numbers match the initial version numbers in the `version_tuple`.

src/auslib/util/versions.py
escapewindow: The 2nd parens is either a minor version + `.*` for the mini version, or just a `.*` for the minor version.
We don't use the regex groups in the `GlobVersion.parse` function; we just use the regex to make sure the schema of the version string is valid.

src/auslib/util/versions.py
bhearsum: Oh, I see. So the left and right side are not treated the same -- we ignore any items from the right that don't have an equivalent on the left. I find this a bit weird (AFAICT it means that foo == bar may not be equivalent to bar == foo), but I can live with this for now as long as it's documented. (If I'm wrong about any of this, please let me know.)

src/auslib/util/versions.py
bhearsum: Right, what I'm saying is that 89.89.5* wouldn't work, for example. I can live with this, it just feels like a weird limitation.

escapewindow: I think we covered this [here](https://github.com/mozilla-releng/balrog/pull/1928#discussion_r614832974). `89.89.5*` would match `89.89.5`, `89.89.51`, and `89.89.509`; that seems more like a footgun than a feature.

src/auslib/util/versions.py
escapewindow: I think it works both ways?

```
#!/usr/bin/env python

import re
from distutils.version import LooseVersion, StrictVersion

class GlobVersionTuple(tuple):
    def __eq__(self, value):
        # If we're 78.8.*, don't match 78.8. Also avoid a KeyError below
        if len(value) < len(self):
            return False
        for i, j in enumerate(self):
            if j != value[i]:
                return False
        return True

    def __ne__(self, value):
        return not self.__eq__(value)


class GlobVersion(StrictVersion):
    """A version class that supports Firefox versions 5.0 and up, which ends
    with a glob `*`. Not really a StrictVersion at all, but it needs to be
    to compare with other MozillaVersions."""

    version_re = re.compile(
        r"""^(\d+) \. (\d+\.\*|\*)$""",
        re.VERBOSE,
    )
    prerelease = None

    def parse(self, vstring):
        self.vstring = vstring
        match = self.version_re.match(vstring)
        if not match:
            raise ValueError("invalid version number '%s'" % vstring)
        parts = [int(i) for i in vstring.split(".") if i != "*"]
        self.version = GlobVersionTuple(parts)

    def __eq__(self, value):
        return str(value).startswith(self.vstring.rstrip("*"))

g = GlobVersionTuple([7, 9])

print(g == (7, 9, 2))
print(g == (7, 9))
print((7,9,2) == g)
print((7,9) == g)
```

prints `True` four times.

src/auslib/util/versions.py
```diff
@@ -45,6 +45,8 @@ class AncientMozillaVersion(StrictVersion):
 
 class GlobVersionTuple(tuple):
     def __eq__(self, value):
+        if len(value) < len(self):
+            return False
```
escapewindow, line 49: I noticed that my test script had this, and the PR didn't. This avoids a KeyError.

escapewindow: (I realized we don't have to have this, since StrictVersion will always have 3+ len...)

src/auslib/util/versions.py
escapewindow: Just to be sure, I tested

```
print(g == StrictVersion("7.9.2").version)
print(g == StrictVersion("7.9.0").version)
print(StrictVersion("7.9.2").version == g)
print(StrictVersion("7.9.0").version == g)
```

and those all printed `True` as well, and

```
print(StrictVersion("7.90.2").version == g)
print(g == StrictVersion("7.90.2").version)
```

returns False. Want me to add some parametrized tests here?

escapewindow: Added more tests in https://github.com/mozilla-releng/balrog/pull/1924/commits/f13bba07e065044ac8aad7194ee93899ffe85f52 .

bhearsum: ok, clearly i'm misreading the code! ty for adding those tests

src/auslib/util/versions.py
bhearsum: ok, ty

bhearsum (approved):

PR #1928, opened 2021-04-16
[RELENG-486] Support globbed version matching.

escapewindow:
In SystemAddons, we want to target specific version ranges, e.g. `80.*`.
Previous to this patch, that meant we needed both a `<80` rule to
prevent sending the same System Addon(s) to previous versions, and a
`<81` rule to target the `80.*` range. This worked while we maintained
SystemAddons rules manually, but will make it difficult to rewrite rules
automatically.

This patch adds support for `version: 80.*` or `version: 80.1.*` type
globs.

### DIFF
--- a/src/auslib/util/comparison.py
+++ b/src/auslib/util/comparison.py
@@ -14,7 +14,15 @@ def has_operator(value):
     return value.startswith(("<", ">"))
 
 
+def glob_op(value, operand):
+    """Compare a string or StrictVersion against a glob stripped of its `*`.
+    Currently only supported in version_compare."""
+    return value.startswith(operand)
+
+
 def get_op(pattern):
+    if pattern.endswith(".*"):
+        return glob_op, pattern.rstrip("*")
     # only alphanumeric characters means no operator
     if re.match(r"\w+", pattern):
         return operator.eq, pattern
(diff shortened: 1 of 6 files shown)


escapewindow:
I think this is cleaner.

@bhearsum, some questions:

- I'm not sure what's failing in the `test_client` tests in #1924; do you know what's going on? I can add tests like that here, once we can get them green.
- Do you know if we might hit any schema issues by allowing the `*`? Should I test adding those types of rules in the UI before we land?
- Semi-unrelated: can we delete the `master` branch from this repo? I keep on muscle-memory `git diff master` or `git rebase -i master` and having to clean up. In shipit, there is no `master` branch, so it gives me an error rather than happily performing the operation against an ancient revision.

bhearsum:
> I think this is cleaner.
> 
> @bhearsum, some questions:
> 
>     * I'm not sure what's failing in the `test_client` tests in #1924; do you know what's going on? I can add tests like that here, once we can get them green.

It looks to me like a Rule pointing to a Release that doesn't exist. And indeed, I don't see a `superblobaddon-with-multiple-response-blob-glob` being added, or existing already.

>     * Do you know if we might hit any schema issues by allowing the `*`? Should I test adding those types of rules in the UI before we land?

Definitely worth a test. You may want to add a test or two to https://github.com/mozilla-releng/balrog/blob/main/tests/admin/views/test_rules.py as well, which would run through the swagger schema validations as well.

>     * Semi-unrelated: can we delete the `master` branch from this repo? I keep on muscle-memory `git diff master` or `git rebase -i master` and having to clean up. In shipit, there is no `master` branch, so it gives me an error rather than happily performing the operation against an ancient revision.

Yes, I will do that!

src/auslib/util/comparison.py
```diff
@@ -14,7 +14,15 @@ def has_operator(value):
...
 
 
+def glob_op(value, operand):
+    """Compare a string or StrictVersion against a glob stripped of its `*`.
+    Currently only supported in version_compare."""
+    return value.startswith(operand)
+
+
 def get_op(pattern):
+    if pattern.endswith(".*"):
```
bhearsum, line 24: For posterity, I was going to suggest that we don't require the preceeding `.` to glob, which would let us do things like `8*` to match 80-89. After some thought I think it's an anti-feature, since it would also match things like `800`, which feels like a footgun.

bhearsum (changes requested):
I agree this is much cleaner, thank you! r=me with some tests

src/auslib/util/comparison.py
escapewindow: +1. That reminds me, I'll grab the [version100](https://github.com/mozilla-releng/balrog/pull/1924/files#diff-70e892168832c6627bc1b314d2a287db1f4ac842e66c89e6a8f8525a2c765ce7R86) bug fix from the previous PR :)

escapewindow:
So when I try to get this working in the schemas, it looks like I have to special case all over the place. The GlobVersion solution may end up being cleaner; I'll try that again.

escapewindow:
I think I'm going to close in favor of #1924.

PR #2065, opened 2021-06-16
bug 1694236: rename whitelist -> allowlist; add test for vpn allowlist

bhearsum:
The rename was done with sed, and sanity checked by hand + through tests. Notably, there's still a few `whitelist` references in schema migration code and tests that cannot be changed.


escapewindow (approved):

bhearsum:
If someone wants to land this feel free, but I'm not going to merge it myself yet since I'm going to be out for a couple of weeks.

PR #2178, opened 2021-09-07
Allow FirefoxVPN to be served from archive.mozilla.org

bhearsum:
We're moving away from publishing these on their own domain in https://bugzilla.mozilla.org/show_bug.cgi?id=1723049. We'll be updating old links in Balrog, but we need to update this first.


jmaher (approved):

MihaiTabara:
@bhearsum is this in preparation for when we'll have beetmover logic in VPN or has something changed in the meantime?

bhearsum:
This is because we're moving away from hosting MozillaVPN on vpn.mozilla.org, and putting it on archive.mozilla.org instead.

PR #2179, opened 2021-09-07
Add content signature to GMP response

gbrownmozilla:


src/auslib/web/public/client.py
```diff
@@ -230,6 +230,8 @@ def get_update_blob(transaction, **url):
     response = make_response(xml)
     response.headers["Cache-Control"] = app.cacheControl
     response.headers.extend(get_aus_metadata_headers(eval_metadata))
+    if query["product"] == "GMP":
```
bhearsum, line 233: Let's move this to a list of products we enable content signatures for in https://github.com/mozilla-releng/balrog/blob/main/uwsgi/public.wsgi (you can then find it in `app.config`).

src/auslib/web/public/helpers.py
```diff
@@ -21,4 +24,19 @@ def get_aus_metadata_headers(eval_metadata):
     return headers
 
 
+def get_content_signature_headers(content):
+    headers = {}
+    if app.config.get("AUTOGRAPH_URL"):
+        hash_ = make_hash(content)
+
+        def sign():
+            return sign_hash(
```
bhearsum, line 33: Now seems like a good time to add a log message here, when we're actually calling out. It should help us gain some insight if we end up with a load spike related to this.

bhearsum (changes requested):
This is great overall! There's a couple of minor things below, but we also need some tests for this. They'll need to go in https://github.com/mozilla-releng/balrog/blob/main/tests/web/test_client.py, and you can see what we're doing for existing tests around this in https://github.com/mozilla-releng/balrog/blob/main/tests/web/test_json.py. Let me know if you have any questions.

src/auslib/web/public/helpers.py
gbrownmozilla: I thought it would be nice to see the result also, so logged a little later -- good enough?

src/auslib/web/public/client.py
gbrownmozilla: I like that idea - thanks. Wasn't sure if there was any reason to add the new symbol in admin.wsgi also, so left that out.

bhearsum: Nope - the content signatures only apply to the public app, so no need to add it there.

src/auslib/web/public/helpers.py
bhearsum: Yeah, that sounds good too.

bhearsum (approved):

src/auslib/web/public/client.py
```diff
@@ -230,6 +230,8 @@ def get_update_blob(transaction, **url):
     response = make_response(xml)
     response.headers["Cache-Control"] = app.cacheControl
     response.headers.extend(get_aus_metadata_headers(eval_metadata))
+    if query["product"] in app.config.get("CONTENT_SIGNATURE_PRODUCTS", []):
+        response.headers.extend(get_content_signature_headers(xml, True))
```
gbrownmozilla, line 234: There's a bit of a disconnect here, where I say, "if this is any of the products that require signing, sign with the GMP autograph". Having a per-product table of autograph servers seems unwarranted at this time, but I'm open to making this more flexible if you think otherwise.

bhearsum: The other way to do this would be to pass a product name instead of a `is_gmp` bool -- and always look up by that product name. This might be a bit more future proof without needing to maintain a table of all of the products. I'm okay either way though - up to you.

gbrownmozilla: Yes, let's do that.

bhearsum (approved):
Looks good!

PR #2285, opened 2021-10-28
Bump python deps

jcristau:
Not quite sure the right granularity for this, happy to split it if preferred.  There's more to come to get us up-to-date, but this is a start...


sarah-clements (commented):
hrm, that's a lot of changes. I'm assuming you checked for any breaking changes in the upgrades to major versions (say in flask, for example?) and tested locally?

jcristau:
I did check all of the changelogs for anything that sounded like it could affect us.

sarah-clements (approved):

PR #2307, opened 2021-11-08
Add Pinebuild to the public endpoint's domain allowlist

jcristau:
We had previously added it to the admin endpoint, but missed public.


gbrownmozilla (approved):
Not sure if it's needed, but it seems safe.

PR #2313, opened 2021-11-10
[RELENG-517] add xpibld auth0 clients

escapewindow:
These are to automate System Addons releases, superblobs, and rules changes.
We may also extend this to Widevine and other plugins, since those are fairly straightforward compared to Firefox rules.


jcristau:
Can we describe what this is for, somewhere that's not the internal jira?

escapewindow:
> Can we describe what this is for, somewhere that's not the internal jira?

I updated the PR description; does that work?

jcristau (approved):
Thanks!

PR #2336, opened 2021-11-19
[RELENG-607] - try to address a balrog fx-version-100 bug

escapewindow:


escapewindow:
@bhearsum is this what you were thinking? Or do we need a "real" fix?

gbrownmozilla (approved):
Simple and essential - nice.

For the larger question of supporting version 100, would we get some piece of mind by adding unit tests with version=100 data?

bhearsum (approved):
For context, this is only used by the code that validates the `updateLine` parts of App Release blobs. Specifically, to ensure that they don't have any version ranges that overlap.

I think this works, but the risk of it causing issues is low if for some reason it doesn't. The most likely scenario is we accidentally set up `updateLine` entries that have version overlaps, and are therefore non-deterministic.

escapewindow:
> For the larger question of supporting version 100, would we get some piece of mind by adding unit tests with version=100 data?

I think so, yes. I'm debating whether I want to add those tests or merge and remove this ticket off our list sooner :)

escapewindow:
Attempting a test in https://github.com/mozilla-releng/balrog/compare/main...escapewindow:balrog100-test?expand=1 ; running locally before I push to this PR.

I'm a bit less worried now that I realized that we'll probably see version 100 bugs in nightly and beta before release, which means we'll get 4-8 weeks' notice before we hit m-r. Though we'll likely have to fix quickly since blocking nightlies isn't great.

jcristau:
As far as I can tell this is not actually an issue for Firefox version 100.  This would affect versions such as 91.0.100 because `decrement_version(91.1.0)` would be `91.0.99` instead.  But we're nowhere near that amount of dot releases so still only a theoretical issue. We may want to assert that `i > 0` when we set `parts[i]` to 99 (or 999) though; that can only happen if all version parts are 0.

escapewindow:
> As far as I can tell this is not actually an issue for Firefox version 100. This would affect versions such as 91.0.100 because `decrement_version(91.1.0)` would be `91.0.99` instead. But we're nowhere near that amount of dot releases so still only a theoretical issue. We may want to assert that `i > 0` when we set `parts[i]` to 99 (or 999) though; that can only happen if all version parts are 0.

6a2a776 ?

escapewindow:
Alternately we can close/punt since it's not an fx100 bug. I'm open :)

escapewindow:
Meh, let's merge and close the story+epic.

PR #2374, opened 2022-01-04
Bump dependency versions for the balrog agent

jcristau:


agent/requirements/base.txt
```diff
@@ -252,22 +252,17 @@ multidict==5.2.0 \
...
-# via -r requirements/base.in
-typing-extensions==3.10.0.2 \
-    --hash=sha256:49f75d16ff11f1cd258e1b988ccff82a3ca5570217d7ad8c5f48205dd99a677e \
-    --hash=sha256:d8226d10bc02a29bcc81df19a26e56a9647f8b0a6d4a83924139f4a8b01f17b7 \
-    --hash=sha256:f1d25edafde516b146ecd0613dabcc61409817af4766fbbcfb8d1ad4ec441a34
-# via async-timeout
+    # via
+    #   aiohttp
+    #   yarl
+sentry-sdk==1.5.1 \
```
jmaher, line 258: I supposed 1.4.3 -> 1.5.1 differences are not too bad:
https://github.com/getsentry/sentry-python/blob/master/CHANGELOG.md

agent/requirements/test.txt
```diff
@@ -9,268 +9,202 @@
...
-# via -r requirements/test.in
-# via
-#   aiohttp
-#   pytest
-black==21.10b0 \
-    --hash=sha256:6eb7448da9143ee65b856a5f3676b7dda98ad9abe0f87fce8c59291f15e82a5b \
-    --hash=sha256:a9952229092e325fe5f3dae56d81f639b23f7131eb840781947e4b2886030f33
-# via -r requirements/test.in
+    # via -r requirements/test.in
+black==21.12b0 \
```
jmaher, line 13: odd that we are using b0 versions of black, not a formal release.

agent/requirements/test.txt
```diff
@@ -9,268 +9,202 @@
...
 pyflakes==2.4.0 \
     --hash=sha256:05a85c2872edf37a4ed30b0cce2f6093e1d0581f8c19d7393122da7e25b2b24c \
     --hash=sha256:3bb3a3f256f4b7968c9c788781e4ff07dce46bdf12339dcda61053375426ee2e
-# via flake8
-pyparsing==2.4.7 \
-    --hash=sha256:c203ec8783bf771a155b207279b9bccb8dea02d8f0c9e5f8ead507bc3246ecc1 \
-    --hash=sha256:ef9d7589ef3c200abe66653d3f1ab1033c3c419ae9b9bdb1240a85b024efc88b
-# via packaging
+    # via flake8
+pyparsing==3.0.6 \
```
jmaher, line 153: another larger version jump, any concerns: 
https://github.com/pyparsing/pyparsing/blob/master/CHANGES

jmaher (approved):
big update!

agent/requirements/test.txt
jcristau: https://github.com/psf/black/tags looks like all black versions are bX for some reason.

agent/requirements/test.txt
jcristau: We're not using this directly, it's pulled in by dependencies, and is only for tests, so I don't think we should be too concerned as long as CI remains green.

jcristau:
Big update, but reasonably limited for base.txt i.e. non-tests.  Thanks for the review!

PR #2376, opened 2022-01-05
Retry any autograph failures in sign_hash

### LINKED ISSUE #2295: retry content signature requests on error from autograph
Sometimes when trying to get a content signature from autograph we get a 502 (or other error?), we should retry them a couple of times to try and reduce the errors that propagate to clients.

gbrownmozilla:
Copy sync retry support from scriptworkers, and use it to retry signing failures, similar to:
https://github.com/mozilla-releng/scriptworker-scripts/blob/5a1ef390a2727f9d7c28bc089fb2a890c90b2c3f/signingscript/src/signingscript/sign.py#L960

Closes #2295

### DIFF
--- a/src/auslib/util/autograph.py
+++ b/src/auslib/util/autograph.py
@@ -4,6 +4,8 @@ from hashlib import sha384
 import requests
 from requests_hawk import HawkAuth
 
+from auslib.util.retry import retry_sync
+
 SIGNATURE_PREFIX = "Content-Signature:\x00"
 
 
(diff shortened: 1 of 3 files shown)


jcristau (commented):
Can we maybe add a test, where the mocked autograph returns an exception on the first call and a proper result the second time?

(As a side note, https://github.com/jd/tenacity may be another option for the retry stuff, if we wanted to avoid copying code.)

gbrownmozilla:
Thanks for the tenacity link, but I prefer to keep it the way it is - one less dependency, etc.

I added simple tests for temporary and permanent autograph failures. Use of the global is obviously sub-optimal, but gets the job done.

jcristau (approved):
:+1: thank you!

PR #2382, opened 2022-01-11
Version bump to 3.11

gbrownmozilla:
I tagged 3.10 with version.txt still at 3.09, so let's bump to 3.11 and then I will re-tag and re-release @ v3.11.


jcristau (approved):
:+1:

PR #2415, opened 2022-01-27
Replace sqlalchemy.interfaces usage

jcristau:
The sqlalchemy.interfaces module was deprecated in 0.7 and removed in 1.4.

Use the replacement event API (https://docs.sqlalchemy.org/en/14/core/event.html)

### DIFF
--- a/src/auslib/db.py
+++ b/src/auslib/db.py
@@ -10,11 +10,11 @@ from os import path
 
 import migrate.versioning.api
 import migrate.versioning.schema
+import sqlalchemy.event
 import sqlalchemy.types
 from aiohttp import ClientSession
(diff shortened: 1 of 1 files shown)


src/auslib/db.py
```diff
@@ -3059,10 +3059,9 @@ def bleet(table, type_, changed_by, query, transaction, additional_columns=None)
 # lets us put the database in a stricter mode that will disallow things like
 # automatic data truncation.
 # From http://www.enricozini.org/2012/tips/sa-sqlmode-traditional/
-class SetSqlMode(PoolListener):
-    def connect(self, dbapi_con, connection_record):
-        cur = dbapi_con.cursor()
-        cur.execute("SET SESSION sql_mode='TRADITIONAL'")
+def my_on_connect(dbapi_con, connection_record):
```
gbrownmozilla, line 3062: How will my_on_connect() be called?

PR #2428, opened 2022-02-07
Fix local dockerfiles

gabrielBusta:
I ran into a few issues running/building balrog locally. This PR contains my fixes.

* The first issue was a missing `appstream` repo when installing nmap-ncat in the `Dockerfile.nginx` image. This issue is related to CentOS 8 reaching End Of Life at the end of 2021. [The `openresty/openresty:centos` image we are using is based on `centos:8`](https://github.com/openresty/docker-openresty/blob/master/centos/Dockerfile#L5)
* The second issue was a missing C++ compiler in the `python:python:3.8-slim` image we are basing the `Dockerfile.test` image on. This was causing an error when building the `brotli` wheel on `pip install`
* The third issue was caused with the mysql image not supporting ARM based architectures. To build an image based on mysql on a computer with an ARM chip (such as my M1 MacBook) we need to specify the platform we are building the image for. On my machine the resulting image it's run in a AMD64 emulator (with some performance degradation). We could potentially move to a mariadb image as a drop in replacement for mysql. Those images support ARM architectures. [Using a mariadb image is the recommended solution in Docker's docs](https://docs.docker.com/desktop/mac/apple-silicon/#known-issues) (not sure if it truly would be a drop in replacement)

### DIFF
--- a/uwsgi/admin.wsgi
+++ b/uwsgi/admin.wsgi
@@ -34,7 +34,7 @@ if STAGING or LOCALDEV:
     SYSTEM_ACCOUNTS.extend(["balrog-stage-ffxbld", "balrog-stage-tbirdbld"])
     DOMAIN_ALLOWLIST.update(
         {
-            "ftp.stage.mozaws.net": ("Firefox", "Fennec", "Devedition", "SeaMonkey", "Thunderbird", "Pinebuild"),
+            "ftp.stage.mozaws.net": ("Firefox", "Fennec", "Devedition", "SeaMonkey", "Thunderbird", "Pinebuild", "SystemAddons"),
             "bouncer-bouncer-releng.stage.mozaws.net": ("Firefox", "Fennec", "Devedition", "SeaMonkey", "Thunderbird", "Pinebuild"),
             "stage.guardian.nonprod.cloudops.mozgcp.net": ("FirefoxVPN", "Guardian"),
             "stage-vpn.guardian.nonprod.cloudops.mozgcp.net": ("FirefoxVPN", "Guardian"),
--- a/Dockerfile.nginx
+++ b/Dockerfile.nginx
@@ -1,5 +1,8 @@
 FROM openresty/openresty:centos
 
+RUN sed -i 's/mirrorlist/#mirrorlist/g' /etc/yum.repos.d/CentOS-Linux-* &&\
+    sed -i 's|#baseurl=http://mirror.centos.org|baseurl=http://vault.centos.org|g' /etc/yum.repos.d/CentOS-Linux-*
+
 # netcat is used to listen to the ports.
 
 RUN yum install -y nmap-ncat
--- a/Dockerfile.test
+++ b/Dockerfile.test
@@ -13,7 +13,7 @@ MAINTAINER bhearsum@mozilla.com
 # xz-utils is needed to unpack sampled ata
 # git is needed to send coverage reports
 RUN apt-get -q update \
-    && apt-get -q --yes install netcat libpcre3 libpcre3-dev default-libmysqlclient-dev mariadb-client curl gcc xz-utils git \
+    && apt-get -q --yes install g++ netcat libpcre3 libpcre3-dev default-libmysqlclient-dev mariadb-client curl gcc xz-utils git \
     && apt-get clean
 
 WORKDIR /app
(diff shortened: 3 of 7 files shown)


gbrownmozilla (changes requested):
This doesn't work for me, on Ubuntu:
```
~/balrog2$ docker-compose up
ERROR: The Compose file './docker-compose.yml' is invalid because:
Unsupported config option for services.balrogdb: 'platform'

```

gabrielBusta:
> This doesn't work for me, on Ubuntu:
> 
> ```
> ~/balrog2$ docker-compose up
> ERROR: The Compose file './docker-compose.yml' is invalid because:
> Unsupported config option for services.balrogdb: 'platform'
> ```

@gbrownmozilla it might be the `version` in `docker-compose.yml`. The `platform` config option was added in [compose file version 2.4](https://docs.docker.com/compose/release-notes/#new-features-3)

What is the output of `docker-compose version` & `docker version` (Server.Engine.Version) on your machine?

gbrownmozilla:
```
$ docker version
Client:
 Version:           20.10.7
 API version:       1.41
 Go version:        go1.13.8
 Git commit:        20.10.7-0ubuntu5~20.04.2
 Built:             Mon Nov  1 00:34:17 2021
 OS/Arch:           linux/amd64
 Context:           default
 Experimental:      true

Server:
 Engine:
  Version:          20.10.7
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.13.8
  Git commit:       20.10.7-0ubuntu5~20.04.2
  Built:            Fri Oct 22 00:45:53 2021
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.5.5-0ubuntu3~20.04.1
  GitCommit:        
 runc:
  Version:          1.0.1-0ubuntu2~20.04.1
  GitCommit:        
 docker-init:
  Version:          0.19.0
  GitCommit:        
$ docker-compose version
docker-compose version 1.25.0, build unknown
docker-py version: 4.1.0
CPython version: 3.8.10
OpenSSL version: OpenSSL 1.1.1f  31 Mar 2020
```
I'll look into an upgrade tomorrow.

gabrielBusta:
Doesn't seem like the versions are far behind... Hopefully the file `version`  attribute helps.

gabrielBusta:
It's probably a good idea to update to a newer `docker-compose` version. I am on v2.x

```zsh
$ docker-compose version
Docker Compose version v2.2.3
```

My `docker` engine version is just a few minor revisions ahead.

gbrownmozilla:
I upgraded to docker-compose 2.2.3, and that works fine. I had to do that upgrade manually, and I noticed the site said 1.x is still considered the latest stable release, so others might trip over the same issue, but if we need it...

gbrownmozilla (approved):

gabrielBusta:
@gbrownmozilla I see. We need it to run Balrog locally on M1/ARM chips. I could add a `docker-compose-arm.yml` file. Users on ARM could then specify `docker-compose-arm.yml` using the `docker-compose --file` option.

gbrownmozilla:
I agree we need to support arm. docker-compose-arm.yml is an interesting alternative. I don't have a strong feeling one way or the other: I trust your best judgment to either go ahead with this patch as-is, or use docker-compose-arm.yml.

gabrielBusta:
@gbrownmozilla it turns out I can override the keys in the `docker-compose.yml` file by passing in a second file to the `-f` option. I added `docker-compose.arm.yml` & added the command to build the images on ARM to the README. This way `docker-compose.yml` will remain the same.

uwsgi/admin.wsgi
```diff
@@ -34,7 +34,7 @@ if STAGING or LOCALDEV:
     SYSTEM_ACCOUNTS.extend(["balrog-stage-ffxbld", "balrog-stage-tbirdbld"])
     DOMAIN_ALLOWLIST.update(
         {
-            "ftp.stage.mozaws.net": ("Firefox", "Fennec", "Devedition", "SeaMonkey", "Thunderbird", "Pinebuild"),
+            "ftp.stage.mozaws.net": ("Firefox", "Fennec", "Devedition", "SeaMonkey", "Thunderbird", "Pinebuild", "SystemAddons"),
```
gabrielBusta, line 37: Sorry about the unrelated change @gbrownmozilla. This allows system addon releases in staging & locally to point to the staging archive `ftp.stage.mozaws.net`

Issue #2479, opened 2022-03-25
Internal Service Errors after deploying Balrog


ahal:
SRE Services deployed Balrog around 17:00 EDT. At 17:45 EDT we started seeing balrogscript task failures and the queue grew to 1000+ tasks. Sheriffs alerted #firefox-ci and #taskcluster-cloudops at 21:00 EDT. Balrog web was also hitting network errors on all pages.

JBuck rolled back the deploy and everything cleared up immediately. He was able to grab some stack traces first:
```
"Uncaught exception:
  File "/usr/local/lib/python3.9/site-packages/flask/app.py", line 2073, in wsgi_app
    response = self.full_dispatch_request()
  File "/usr/local/lib/python3.9/site-packages/flask/app.py", line 1519, in full_dispatch_request
    return self.finalize_request(rv)
  File "/usr/local/lib/python3.9/site-packages/flask/app.py", line 1540, in finalize_request
    response = self.process_response(response)
  File "/usr/local/lib/python3.9/site-packages/flask/app.py", line 1885, in process_response
    response = self.ensure_sync(func)(response)
  File "/usr/local/lib/python3.9/site-packages/balrog-3.15-py3.9.egg/auslib/web/admin/base.py", line 93, in complete_request
    request.transaction.rollback()
<class 'AttributeError'>
AttributeError("'Request' object has no attribute 'transaction'")
"
```

and

```
"Uncaught exception:
  File "/usr/local/lib/python3.9/site-packages/flask/app.py", line 1540, in finalize_request
    response = self.process_response(response)
  File "/usr/local/lib/python3.9/site-packages/flask/app.py", line 1885, in process_response
    response = self.ensure_sync(func)(response)
  File "/usr/local/lib/python3.9/site-packages/balrog-3.15-py3.9.egg/auslib/web/admin/base.py", line 93, in complete_request
    request.transaction.rollback()
<class 'AttributeError'>
AttributeError("'Request' object has no attribute 'transaction'")
"
```

Maybe something in the recent dependency updates / or upgrade to Python 3.9 is the cause here?

/cc @jcristau

ahal:
See also https://bugzilla.mozilla.org/show_bug.cgi?id=1761398

ahal:
The exception appears to be happening in a failure path, so it's possibly hiding the true problem.

jbuck:
We deployed v3.15 today, and rolling back the admin to v3.14 fixed things up

[...]

PR #2481, opened 2022-03-25
Guard against intermittent AttributeError on request.transaction

gbrownmozilla:
Consider, what if flask updates request.full_path between before/after? (I can't find any docs that say it will, or won't!)

Or, what if dbo.begin() raises, in setup_request()?

Otherwise, is it possible that @app.before_request is not always called??

### DIFF
--- a/src/auslib/web/admin/base.py
+++ b/src/auslib/web/admin/base.py
@@ -134,7 +134,7 @@ def unicode(error):
 @app.errorhandler(Exception)
 def ise(error):
     capture_exception(error)
-    log.error("Caught ISE 500 error: %r", error)
+    log.exception("Caught ISE 500 error: %r", error)
     log.debug("Request path is: %s", request.path)
     log.debug("Request environment is: %s", request.environ)
     log.debug("Request headers are: %s", request.headers)
--- a/src/auslib/web/admin/base.py
+++ b/src/auslib/web/admin/base.py
@@ -134,7 +134,7 @@ def unicode(error):
 @app.errorhandler(Exception)
 def ise(error):
     capture_exception(error)
-    log.error("Caught ISE 500 error.")
+    log.error("Caught ISE 500 error: %r", error)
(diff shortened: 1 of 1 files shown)


gbrownmozilla:
I don't think this will address the recent outage in version 3.15, but it should avoid the AttributeError noted in #2479 and seen intermittently in the past.

gbrownmozilla:
I expect the logging change to produce a log message like:

`Caught ISE 500 error: AttributeError("'Request' object has no attribute 'transaction'")`

This might help us better understand ISEs.

src/auslib/web/admin/base.py
```diff
@@ -88,7 +88,7 @@ def setup_request():
 
 @app.after_request
 def complete_request(response):
-    if request.full_path.startswith("/v2"):
+    if hasattr(request, "transaction"):
```
jcristau, line 91: I wonder if it's worth keeping the `full_path` check, for symmetry, but maybe not.

src/auslib/web/admin/base.py
```diff
@@ -134,7 +134,7 @@ def unicode(error):
 @app.errorhandler(Exception)
 def ise(error):
     capture_exception(error)
-    log.error("Caught ISE 500 error.")
+    log.error("Caught ISE 500 error: %r", error)
```
jcristau, line 137: Would `log.exception` instead work?

jcristau (approved):

src/auslib/web/admin/base.py
gbrownmozilla: I think the hasattr way is better: We know it's possible to have "/v2" and not have request.transaction; throwing an AttributeError here doesn't help us at all.

src/auslib/web/admin/base.py
gbrownmozilla: Good idea.

PR #2487, opened 2022-03-28
Add requests retries to balrogclient

gbrownmozilla:
In case some failures are temporary, enable balrogclient (used by balrogscript) to retry requests.

https://docs.python-requests.org/en/latest/api/?highlight=max_retries#requests.adapters.HTTPAdapter
https://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html

### DIFF
--- a/client/src/balrogclient/api.py
+++ b/client/src/balrogclient/api.py
@@ -71,6 +71,10 @@ class BearerAuth(requests.auth.AuthBase):
 def get_balrog_session(auth0_secrets, session=None):
     if not session:
         session = requests.Session()
+        retry = Retry(total=5, backoff_factor=0.1, status_forcelist=[429, 500, 502, 503, 504])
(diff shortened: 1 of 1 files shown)


gbrownmozilla:
I don't feel strongly about the exact Retry parameters, but they seem reasonable, and are very similar to some other uses in the wider code base:
https://searchfox.org/mozilla-central/rev/66b43502bdbe4b62cbc8b0c8ff10efc5707352e4/build/upload_generated_sources.py#99

gabrielBusta (approved):

jcristau (commented):
I guess my question would be whether we want this here or in consumers (i.e. balrogscript), so they can control the parameters.  This is probably fine though.

gbrownmozilla:
Good point. I think I prefer keeping it here, as a basic capability of balrogclients.

PR #2489, opened 2022-04-01
Add adhoc-tests.py

gbrownmozilla:


gbrownmozilla:
I don't want to spend *too* much time on this. I am finding it useful (but in a limited way). There are still a lot of TODO comments -- lots of things that could be added. I'd like to check it in so we can all share it and see if it's useful and how much we want to expand it.

jcristau (approved):

PR #2497, opened 2022-04-07
admin: return the sqlalchemy connection to the pool on complete_request

jcristau:


jcristau:
I'm not sure how to test this.  Haven't been able to break things locally yet, even with `pool_size=1, max_overflow=1`.  I'll try some more...

gbrownmozilla (approved):

src/auslib/web/admin/base.py
```diff
@@ -93,6 +93,7 @@ def complete_request(response):
             request.transaction.rollback()
         else:
             request.transaction.commit()
+        request.transaction.close()
```
jcristau, line 96: Might need to do this in a try/finally block so we still close the connection when commit fails?

jcristau: Updated.

gbrownmozilla (approved):

PR #2499, opened 2022-04-07
AUSTransaction: don't issue a redundant rollback

jcristau:
When AUSTransaction.commit raises, it rolls back the transaction
already, so no need to do it again in __exit__.


gbrownmozilla (approved):
Nice find.

Issue #2194, opened 2021-09-16
Require 2 `admin` signoffs for changes to permissions


bhearsum:
Currently, we calculate the signoffs required for changes to permissions by taking the full set of all product signoffs (https://github.com/mozilla-releng/balrog/blob/3ed6b8036caecd58cdd629d4194de6250757d9c1/src/auslib/db.py#L2658). This often means that we need 2, 3, or sometimes even 5 signoffs to make a simple change to permissions.

We should adjust this to require just two signoffs in total. Ideally, it would require them from a full fledged `admin` (ie: has `admin` permission without any conditions attached). If this ends up being too difficult (it my - because it moves us from looking at "roles" when checking signoffs to looking at "permissions" -- it also may complicate the UI) -- we could consider hardcoding 2 "releng" signoffs instead. We'd have to evaluate potential security issues before going that route - most importantly, we need to make sure that nobody currently has that role that shouldn't, and cannot be granted it by anybody other than a full fledged admin.

PR #2528, opened 2022-05-16
Require two releng signoffs for changes to permissions

gabrielbusta:
#2194


gabrielBusta:
I tried to implement two admin sign-offs. Because that moves us from looking at "roles" when checking sign-offs to looking at "permissions" it required some refactoring in the back-end and UI. I ended up hard-coding 2 "releng" sign-offs. Before this lands in production we need to make sure that nobody currently has that role that shouldn't.

jcristau (changes requested):
I'm a bit unhappy about hardcoding a specific role here (iirc we don't do that anywhere else). More importantly, this will break local setups where there may not be 2 users with that role.

Finally, before we merge anything we'll need to add tests.

gabrielBusta:
Yea. It is not ideal. We could add configuration parameter for the number of "releng" sign-offs and set it to 1 when running locally. Is it worth it landing this hacky fix so the VPN team can self-serve updates?

PR #2529, opened 2022-05-16
Deprecate repoze.lru

gabrielbusta:
#2530


gabrielBusta:
It looks like it does use a [lock](https://github.com/tkem/cachetools/blob/bc998ba4278e06f19ab0270c42b8ff7af6448cab/src/cachetools/func.py#L49) in the [decorator](https://github.com/tkem/cachetools/blob/bc998ba4278e06f19ab0270c42b8ff7af6448cab/src/cachetools/func.py#L161) version of the caches. I can wrap the mutations in `MaybeCacher` with a lock to be on the safe side

gabrielBusta:
It looks like it does use a [lock](https://github.com/tkem/cachetools/blob/bc998ba4278e06f19ab0270c42b8ff7af6448cab/src/cachetools/func.py#L49) in the [decorator](https://github.com/tkem/cachetools/blob/bc998ba4278e06f19ab0270c42b8ff7af6448cab/src/cachetools/func.py#L161) version of the caches. I can wrap the mutations in `MaybeCacher` with a lock to be on the safe side

jcristau:
Looks like our tests access stats on the cache, which cachetools doesn't maintain.  Dunno if that means we should add that back for the test, or something else.

jcristau:
> Looks like our tests access stats on the cache, which cachetools doesn't maintain. Dunno if that means we should add that back for the test, or something else.

Uh, not sure what I was thinking here.  cachetools wraps the stdlib's functools which has a cache_info method which returns the necessary stats, so it looks like we should be able to use this.

PR #2532, opened 2022-05-18
Fix pyjwt dependency issues

gabrielBusta:
Pining pyjwt to 2.4.0 fixed the issue.
When I run `pip-compile-multi` for some reason it updates `repoze.lru` to `repoze-lru` in the generated `.txt` requirements... I fixed it manually by replacing `-` with `.` and now it works


gbrownmozilla (approved):
I think this is fine and gets us unstuck. 

We should follow-up on the larger pip-compile-multi / repoze.lru issue, perhaps with a pin.sh like in scriptworker-scripts.

jcristau:
Do we understand why the pyjwt dep started breaking?

gabrielBusta:
A dependency clash? Hard to tell. What if we switched to poetry?

gabrielBusta:
Maybe we can standardize on a `pip-compile-multi` version?

I could follow through with the deprecation of repoze.lru

It will take a while because the tests broke all over the place and cachetools doesn't keep stats (used in the tests)

PR #2534, opened 2022-05-19
Update Pinning

bytesized:


bytesized:
@jcristau @gbrownmozilla @gabrielBusta I see that you all are marked as the reviewers for this. I'm really sorry to try to rush this along (especially since there is so much code here), but I'm really hoping to have this merged in time for the 102 soft freeze (2022-05-26). Please let me know if there is anything I can do to help make that happen.

src/auslib/db.py
```diff
@@ -2944,6 +2945,73 @@ def delete(self, where, changed_by=None, old_data_version=None, transaction=None
...
+        )
+        AUSTable.__init__(self, db, dialect, scheduled_changes=True, scheduled_changes_kwargs={"conditions": ["time"]}, historyClass=HistoryTable)
+
+    def getPotentialRequiredSignoffs(self, affected_rows, transaction=None):
+        # Implementing this is required to schedule changes to this table
+        return None
+
+    def insert(self, changed_by, transaction=None, dryrun=False, **columns):
+        # XXX check the release exists?
+        # XXX signoffs?
```
gbrownmozilla, line 2966: What's the plan for resolving these?

bytesized: Perhaps @jcristau can speak more to these comments, since he wrote them. It's not clear to me that we need to check if the release exists. If we do, it seems reasonable enough to either make the mapping a foreign key or to simply add a check that that mapping exists.

I'm going to completely defer to @jcristau on the signoffs issue. I believe that we discussed this during the creation of this patch, and IIRC, it was still an open question as to whether signoffs are necessary for this. And I don't feel like I have a deep enough understanding of the issue to have a particularly good opinion about it.

src/auslib/web/admin/swagger/api_v2.yml
```diff
@@ -414,3 +414,44 @@ paths:
...
+                  type: string
+                # Creates a scheduled change if passed
+                when:
+                  type: integer
+                  nullable: true
+      responses:
+        '200':
+          description: OK
+          content:
+            # TODO: check
```
gbrownmozilla, line 450: This looks right to me. Any reason to keep the TODO?

src/auslib/util/versions.py
```diff
@@ -135,3 +135,67 @@ def get_version_class(product):
...
+
+    def _cmp(self, other):
+        if not isinstance(other, StrictVersion):
+            return NotImplemented
+        if isinstance(other, GlobVersion):
+            # This would require some extra handling. Since this type of
+            # comparison isn't currently needed, explicitly do not handle it.
+            return NotImplemented
+
+        if len(other.version) < len(self.version):
```
gbrownmozilla, line 187: Could/should we check for <= here?

gbrownmozilla (changes requested):
Well, this is exciting! This seems like the biggest balrog change I've seen during my tenure. I feel like there are subtleties to the implementation that I don't fully appreciate. That said, it looks comprehensive and I don't see anything obviously wrong. Requesting changes for clarification or quick follow-up on my minor concerns.

tests/web/test_client.py
```diff
@@ -2352,3 +2352,420 @@ def testBrokenNightly(self):
             '<?xml version="1.0"?><updates><update type="minor" version="1.0" extensionVersion="1.0" buildID="30000101000000">'
             '<patch type="complete" URL="http://a.com/z" hashFunction="sha512" hashValue="4" size="3"/></update></updates>',
         )
+
+
+class ClientTestPinning(ClientTestCommon):
```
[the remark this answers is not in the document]
bytesized: I had thought that it was ok, but I'm happy to change it if you would like. Just let me know.

src/auslib/web/admin/swagger/api_v2.yml
bytesized: @jcristau wrote that, I'll let him weigh in on if there is still something needed to be done there.

src/auslib/util/versions.py
bytesized: We need `len(other.verson) >= len(self.version)` or else the simple trimmed comparison won't work. As things are `len(other.version)` should always be `3` and `len(self.version)` should always be either `1` or `2`.

src/auslib/web/admin/swagger/api_v2.yml
jcristau: Maybe just drop the comment if things works as-is, I think I put this in there because I hadn't actually tested the API so wasn't sure what the response content should be.

src/auslib/db.py
jcristau: At this point I'm reasonably confident we don't need/want signoffs for pins.  They'd add extra friction for each release, require admin UI work, and I don't think there's that much value.

bytesized:
Updated to remove the two TODO comments that jcristau has indicated are no longer needed.

src/auslib/web/public/swagger/api.yml
```diff
@@ -161,6 +165,7 @@ paths:
         - $ref: '#/parameters/avast'
         - $ref: '#/parameters/force'
         - $ref: '#/parameters/mig64'
+        - $ref: '#/parameters/pin'
```
jcristau, line 168: We talked about this elsewhere, but just noting here we don't actually need this param in endpoints other than `/update/6`.  It doesn't hurt elsewhere though, so I think it's fine to keep it.

tests/test_db.py
```diff
@@ -6104,6 +6143,22 @@ def _add_release_json_tables(self, db, upgrade=True):
...
                 self.assertNotIn(table, metadata.tables)
 
+    def _add_pinnable_releases_tables(self, db, upgrade=True):
+        metadata = self._get_reflected_metadata(db)
+        pin_tables = [
+            "pinnable_releases",
+            "pinnable_releases_scheduled_changes",
+            "pinnable_releases_scheduled_changes_history",
+            "pinnable_releases_scheduled_changes_conditions",
+            "pinnable_releases_scheduled_changes_conditions_history",
```
jcristau, line 6153: Add the signoffs tables here?

src/auslib/AUS.py
```diff
@@ -131,16 +133,62 @@ def evaluateRules(self, updateQuery, transaction=None):
...
+                blob_version = MozillaVersion(blob_version)
+                # We should only pay attention to the pin version if we would otherwise return
+                # something newer. If we unconditionally returned the pinned version, we would do
+                # incorrect things like skipping watersheds.
+                if blob_version > version_pin:
+                    pin_mapping = dbo.pinnable_releases.getPinMapping(updateQuery["product"], getFallbackChannel(updateQuery["channel"]), str(version_pin))
+                    # Note that we fall back to serving the original update if the pin is not found
+                    # in the pin table, even if the version that we will serve is past the pinned
+                    # version. This is because, if there is something wrong with the update pin,
+                    # we would rather default to keeping the user up-to-date.
```
jcristau, line 169: Is that an explicit ask from product?  I'd have thought we'd never want to serve a newer version than what was pinned, so this behavior is surprising to me.

Do we know what chrome does there?

I'm a bit confused about the example, too.  How do we (balrog) know 120.15 will never be released, and therefore we point to 134, vs 120.15 is not released yet, therefore we stay on 120.14?

jcristau:
> I think we're also missing some bits re scheduled changes, e.g. we probably need admin API to list/enact scheduled changes for pinnable_releases, and corresponding changes in balrogagent.

To elaborate on this: the way scheduled changes work in balrog is they're added to the db, and then sit there until the balrog agent (code in `agent/src/balrogagent`, which is a separate daemon that acts as a balrog client by regularly polling `/scheduled_changes/{type}` endpoints, checking for each scheduled change whether its conditions (typically time and/or signoffs) are met, and calling `/scheduled_changes/{type}/{id}/enact` to tell balrog to actually make the change live.

So the bits needed are, off the top of my head:
- a ScheduledChangeView and EnactScheduledChangeView for pinnable_releases
- `/scheduled_changes/pinnable_releases` API in `src/auslib/web/admin/swagger/api.yml` to both list scheduled changes and enact a specific one
- an addition to the `V1_SCHEDULED_CHANGE_ENDPOINTS` variable in `agent/src/balrogagent/cmd.py`
- some tests
(or we may want to use api_v2 which has a bit of a different shape, I'm not sure which is appropriate, maybe @bhearsum can advise)

Unless we add this, we can add scheduled changes for pinnable_releases but they won't actually ever go anywhere.

Hope this helps.

src/auslib/AUS.py
bytesized: > Is that an explicit ask from product?

No. I just think this is definitely the right thing to do here, even if it is a bit counterintuitive.

> How do we (balrog) know 120.15 will never be released

It's not really an issue of knowing whether `120.15` will ever be released. It's an issue of doing the best thing in an ambiguous situation.

This situation is technically unsupported, since part of the requirements for enabling this feature will be to set a pin to an existing version, not a version that they think might exist one day. If they want to pin to the latest version of `120`, that is what a major pin (`120.`) is for. It is not appropriate to guess at what minor version might be desirable before it has even been released.

Despite that, we still have to decide what to do in this unsupported situation. These seem to be the two options:

 1. Treat the invalid pin as invalid and ignore it. If it later becomes valid, anyone that updated in the window when it was invalid will have updated beyond it.
 2. Treat the invalid pin as potentially being valid one day. If it never becomes valid, anyone that didn't update right away will forever be trapped on their current version.

It is my opinion that the first option is both safer (in terms of providing security fixes) and more correct and is therefore the much more preferable option.

bytesized:
For some reason, I can't respond directly to one of your comments. So I'll just respond to it here.

> We should also consider what happens on removal. My initial idea was that each release would add itself to the `pinnable_releases` table, e.g. we could have Firefox/102.1.0/esr pointing at Firefox-102.1.0esr-build1 and Firefox/102.1.1/esr pointing at Firefox-102.1.1esr-build1, and when you'd request an update with pin=102 or pin=102.1 you'd get the newest of these, and if we deleted Firefox-102.1.1esr-build1 the corresponding entry in pinnable_releases would be removed as well, and later requests with pin=102 would receive 102.1.0.

That was not my understanding at all going into this. Would you like me to rewrite it this way?

jcristau:
> For some reason, I can't respond directly to one of your comments. So I'll just respond to it here.
> 
> > We should also consider what happens on removal. My initial idea was that each release would add itself to the `pinnable_releases` table, e.g. we could have Firefox/102.1.0/esr pointing at Firefox-102.1.0esr-build1 and Firefox/102.1.1/esr pointing at Firefox-102.1.1esr-build1, and when you'd request an update with pin=102 or pin=102.1 you'd get the newest of these, and if we deleted Firefox-102.1.1esr-build1 the corresponding entry in pinnable_releases would be removed as well, and later requests with pin=102 would receive 102.1.0.
> 
> That was not my understanding at all going into this. Would you like me to rewrite it this way?

No, I think your model is fine, probably better than what I had in mind.

bytesized:
I added a patch to check that, if a pin is already set, we don't update it to point to an older version. There were a few things that I wasn't sure how to handle though, so someone should probably make sure that these choices that I made in the patch were reasonable:

 - The check just falls through and allows the pin to be added if it can't determine the version of the existing pinned release or the version of the new pinned release.
 - The API still returns 200 even if the pin was not set.
 - The JSON normally returned by the API doesn't seem like it makes a lot of sense in this case, so I added a new format that it can return if the pin was not updated.

tests/web/test_client.py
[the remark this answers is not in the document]
gbrownmozilla: Let's keep it as-is.

docs/source/pinning.rst
```diff
@@ -0,0 +1,65 @@
...
+How Pinning Works
+-----------------
+
+~~~~~~~~~~~~~~~~~~
+Release Automation
+~~~~~~~~~~~~~~~~~~
+
+The process starts when a new release is built.
+When the build has completed, Release Automation automatically submits the release to Balrog.
+The Balrog submission Taskcluster jobs specify one or more channels that the release should be pinned for (See `Bug 1762979 <https://bugzilla.mozilla.org/show_bug.cgi?id=1762979>`_ for details).
```
gbrownmozilla, line 27: I find "The Balrog submission Taskcluster jobs" a little awkward, but I'm not sure how to say it better.

bytesized: I reworded it as "The Balrog submission jobs (in Taskcluster)". Maybe that's a little bit better?

gbrownmozilla: wfm. Thanks.

jcristau (approved):
:+1:

gbrownmozilla (approved):
Thanks @bytesized !

PR #2560, opened 2022-06-13
Bump Balrog UI deps

gabrielBusta:
Bumps Balrog dependencies and node version.

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -191,7 +191,7 @@ tasks:
             routes: []
             payload:
               maxRunTime: 1200
-              image: "node:10"
+              image: "node:16"
               command:
                 - "/bin/bash"
                 - "-c"
@@ -211,7 +211,7 @@ tasks:
             routes: []
             payload:
               maxRunTime: 1200
-              image: "node:10"
+              image: "node:16"
               command:
                 - "/bin/bash"
                 - "-c"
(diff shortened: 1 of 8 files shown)


gbrownmozilla:
@gabrielBusta With such a big change to yarn.lock, I'm surprised you aren't updating package.json (other than node). Just checking, that's intentional?

gabrielBusta:
Yeah it is. The `^` in the `package.json` means update to the latest `patch` or `minor` version on `yarn install`. Here we are updating the packages to the latest `patch` version at most. I might change the `package.json` to do some major version updates (I would like to migrate our UIs to Material UI v5 and React 18). But those updates require more careful consideration to deal with breaking changes.

gbrownmozilla (changes requested):
Thanks for this.
Your changes look OK to me, but when I run them locally (on Ubuntu), I cannot load the Admin interface: https://localhost:9000 is not being served. Curiously, I don't see any failures on the console.

gabrielBusta:
Deleting `ui/node_modules` and running `yarn install` might help. Odd that it doesn't show failures on the console.

gbrownmozilla:
Aha! I cleared my browser site data for localhost...was prompted with the security warning...accepted and all's well!

gbrownmozilla (approved):

Issue #2561, opened 2022-06-16
consider only displaying "used" Releases, unless unused requested


gbrownmozilla:
There are a lot of Releases in the production db now, and new ones being added all the time. Some show "Used by: n/a", which I think means they are completely unreferenced and not in use. We may be keeping these around for auditing purposes ("have we ever shipped ..."?) But for normal UI inquiries, these are probably useless clutter (more entries to scroll through, longer wait to populate the list than necessary). We could filter out unused releases from the normal Releases list, unless requested by some new UI element.

PR #2562, opened 2022-06-17
WIP - RELENG-941: Display used Releases, unless unused requested

gabrielbusta:
Issue #2561 - Consider only displaying "used" Releases, unless unused requested

### DIFF
--- a/ui/src/views/Releases/ListReleases/index.jsx
+++ b/ui/src/views/Releases/ListReleases/index.jsx
@@ -10,6 +10,8 @@ import Radio from '@material-ui/core/Radio';
 import RadioGroup from '@material-ui/core/RadioGroup';
 import FormControl from '@material-ui/core/FormControl';
 import FormControlLabel from '@material-ui/core/FormControlLabel';
+import FormLabel from '@material-ui/core/FormLabel';
+import Switch from '@material-ui/core/Switch';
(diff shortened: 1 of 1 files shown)


escapewindow:
As mentioned in the jira ticket, we often add a new release and need to be able to view it/reference it in a rule change. So rather than hiding all unused, I’d prefer if we hid unused-and-older-than-time-threshold.

gabrielBusta:
This is a toggle switch that hides/un-hides the unreferenced releases. If we need to look at the unreferenced releases we can toggle it. Right now it defaults to hiding the unreferenced releases. But maybe it'd make more sense to show all releases by default and have that switch there to quickly filter the unreferenced releases.

escapewindow:
Is it difficult to have the toggle hide/show unused-but-older-than, say, 6 months releases?

gabrielBusta:
Yeah that’s totally doable. The toggle could hide/show unused releases older than 6 months.

gbrownmozilla:
+1 for hiding unused-and-older-than-X; the newly added release is an important consideration. I would think X=a few days would be fine, but 6 months is okay with me too.

gabrielBusta:
I may have spoken too soon. It doesn't look like the Balrog releases keep track of the date they are created. The ShipIt releases are the ones with a "created" timestamp.

PR #2602, opened 2022-07-28
Fix Mysql image base version on Dockerfile.

### LINKED ISSUE #2601: Error on docker-compose up - apt-get: not found
`docker-compose up` command results in "apt-get: not found" error.

This happens because  `mysql:5.7` is based on [oraclelinux:7-slim](https://github.com/docker-library/mysql/blob/master/5.7/Dockerfile.oracle) , we can use `mysql:5.7-debian`instead.

In addition, we have a duplication of the `links` attribute on balrogpub service in docker-compose: https://github.com/mozilla-releng/balrog/blob/5dd66b17357d8cb1c3912f5e2437fdb5af866a3c/docker-compose.yml#L65-L92

allan-silva:
- Updates mysql base image version;
- fixes links duplication on docker-compose file;

Closes #2601

### DIFF
--- a/Dockerfile.database
+++ b/Dockerfile.database
@@ -1,4 +1,4 @@
-FROM mysql:5.7
+FROM mysql:5.7-debian
 
 # netcat is used to listen to the ports.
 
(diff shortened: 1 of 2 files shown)


jcristau (approved):
Thanks!

PR #2604, opened 2022-07-29
Add optional "comment" field for emergency_shutoff

### LINKED ISSUE #1862: please provide an optional "comment" field for emergency_shutoff
Something that came up recently is it might be useful if we could document the reason for emergency shutoffs, more or less in the same way that we do for rules (or for tree closures in treestatus), typically by pointing to a bug number.  Then https://aus-api.mozilla.org/api/v1/emergency_shutoff/Firefox/nightly could be used by interested parties to know what's going on when they're not getting updates.

cc @Archaeopteryx

allan-silva:
- [x] Add comment field to emergency shutoff model
- [x] Add comment field to  api
- [x] Add comment field to Balrog UI
- [x] Do manual tests on ui
- [x] Check possible regression for SpeedDial, click fires on disabled elements. Or file a new issue...

Changes on balrogui:
image
image
image
image

Resolves #1862

### DIFF
--- a/tests/admin/views/test_emergency_shutoff.py
+++ b/tests/admin/views/test_emergency_shutoff.py
@@ -46,7 +46,18 @@ class TestEmergencyShutoff(ViewTest):
         self.assertStatusCode(resp, 404)
 
     def test_create(self):
-        data = {"product": "Thunderbird", "channel": "release", "comment": "Thunderbird panic!()"}
+        data = {"product": "Thunderbird", "channel": "release"}
+        resp = self._post("/emergency_shutoff", data=data.copy())
+        self.assertStatusCode(resp, 201)
+        emergency_shutoff_tables = [dbo.emergencyShutoffs, dbo.emergencyShutoffs.history]
+        for shutoff_table in emergency_shutoff_tables:
+            shutoffs = shutoff_table.select(where=data)
+            self.assertTrue(shutoffs)
+            for key in data.keys():
+                self.assertEqual(data[key], shutoffs[0][key])
+
+    def test_create_with_comment(self):
+        data = {"product": "Fennec", "channel": "release", "comment": "Fennec panic!()"}
         resp = self._post("/emergency_shutoff", data=data.copy())
         self.assertStatusCode(resp, 201)
         emergency_shutoff_tables = [dbo.emergencyShutoffs, dbo.emergencyShutoffs.history]
--- a/tests/admin/views/test_emergency_shutoff.py
+++ b/tests/admin/views/test_emergency_shutoff.py
@@ -9,7 +9,7 @@ class TestEmergencyShutoff(ViewTest):
     def setUp(self):
         super(TestEmergencyShutoff, self).setUp()
         dbo.emergencyShutoffs.t.insert().execute(product="Firefox", channel="nightly", data_version=1)
-        dbo.emergencyShutoffs.t.insert().execute(product="Fennec", channel="beta", data_version=1)
+        dbo.emergencyShutoffs.t.insert().execute(product="Fennec", channel="beta", comment="Fennec panic!()", data_version=1)
         dbo.emergencyShutoffs.t.insert().execute(product="Thunderbird", channel="nightly", data_version=1)
         dbo.emergencyShutoffs.scheduled_changes.t.insert().execute(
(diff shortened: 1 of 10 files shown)


ui/src/components/DialogAction/index.jsx
```diff
@@ -95,10 +95,7 @@ function DialogAction(props) {
         {body && <DialogContentText component="div">{body}</DialogContentText>}
       </DialogContent>
       <DialogActions>
-        <Button
-          disabled={actionExecuting}
-          onClick={onClose}
-          action={actions => actions && actions.focusVisible()}>
```
allan-silva, line 101: This behavior (`fucusVisible()`) makes typing the comment impossible. Seems not a big problem remove it. Let me know if has any other approach.

allan-silva:
@jcristau My last contribution was long time ago, let me know if I missed any step.

gbrownmozilla (commented):
This looks good to me. Thank you very much!

Since @jcristau originally opened the issue, I'd like him to give final approval here; he is on vacation currently - hope you don't mind waiting a couple of weeks.

allan-silva:
@gbrownmozilla no problem.
Thanks!

tests/admin/views/test_emergency_shutoff.py
```diff
@@ -38,19 +38,23 @@ def test_get_by_id(self):
...
+        self.assertIn("comment", data)
+        self.assertEqual(data["comment"], "Fennec panic!()")
 
     def test_get_notfound(self):
         resp = self._get("/emergency_shutoff/foo/bar")
         self.assertStatusCode(resp, 404)
 
     def test_create(self):
-        data = {"product": "Thunderbird", "channel": "release"}
+        data = {"product": "Thunderbird", "channel": "release", "comment": "Thunderbird panic!()"}
```
jcristau, line 49: We should keep a test for POST to /emergency_shutoff without a comment, to make sure that still works.

jcristau (commented):
A couple of nits but this looks great.  (Not approving just yet because I'd like to play with it locally, but I do want to get to it shortly.)

ui/src/components/DialogAction/index.jsx
gabrielBusta: I am not worried about it but there might be a reason it the code is calling the focus visible on that button. So I will take a closer look

gabrielBusta: @jcristau this was focusing / highlighting the cancel button on the dialogs (i.e. the signing-off pop-ups) This was probably done so pressing enter would cancel the action / close the pop-up.

I think this change an improvement. Now we can tab through the buttons in the dialog and press enter to "click" on them, which is the way they should behave. Also the button wont take away the focus from other UI elements we might want to put in the dialogs (e.g. the text box on this PR)

gabrielBusta: @jcristau this code was focusing / highlighting the cancel button on the dialogs (i.e. the signing-off pop-ups) This was probably done so pressing enter would cancel the action / close the pop-up.

I think this change is an improvement. Now we can tab through the buttons in the dialogs and press enter to "click" them (this is the way they should behave imo). Also the button wont take away the focus from other UI elements we might want to put in these dialogs (e.g. @allan-silva's text box)

gabrielBusta: @jcristau this code was focusing / highlighting the cancel button on the dialogs (i.e. the signing-off pop-ups) This was probably done so pressing enter would cancel the action / close the pop-up.

I think this change is an improvement. Now we can tab through the buttons in the dialogs and press enter to "click" them (this is the way they should behave imo). Also the button wont take away the focus from other UI elements we might want to put in the body of these dialogs (e.g. @allan-silva's TextField)

gabrielBusta (changes requested):
Looks good to me. Deferring to @jcristau for final approval here. Thank you!

gabrielBusta (approved):

ui/src/components/DialogAction/index.jsx
allan-silva: @gabrielBusta Thanks for take a close look on this!

jcristau (approved):
Looks great.  Thanks again @allan-silva!

PR #2623, opened 2022-08-31
Bug 1787125 - Fallback should use pinning logic

bytesized:


jcristau (approved):
Thanks!

Nice to get rid of an inlined copy of get_blob too :)

PR #2639, opened 2022-09-19
Update python dependencies

jcristau:
This updates the pinning script to also handle the balrog agent's requirements files.


gbrownmozilla (approved):
lgtm. Thanks for catching the agent issue.

PR #2648, opened 2022-09-26
Pass --no-deps to `pip install`

### LINKED ISSUE #2647: pip fails to install dependencies
Looks like we're running into https://github.com/pypa/pip/issues/9644:

- gcloud-aio-auth pulls in pyjwt
- auth0-python pulls in pyjwt[crypto]
- requirements/base.txt lists pyjwt[crypto], with hashes
- pip fails installing deps complaining that hashes are missing for pyjwt

jcristau:
Dependencies are resolved when generating the requirements files, so there's no need to have pip do it again at install time, and this lets us work around a pip bug.

Fixes #2647

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,7 +16,7 @@ WORKDIR /app
 # ... really speeds up building new containers
 COPY requirements/ /app/requirements/
 RUN apt-get install -q --yes gcc && \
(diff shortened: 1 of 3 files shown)


gabrielBusta:
Nice :)

gabrielBusta (approved):

PR #2657, opened 2022-10-04
Refactor get_active_release_names sql query

jcristau:
For some reason the previous query doesn't seem to work for me locally.


jcristau:
@gabrielBusta @gbrownmozilla does this make sense to you?  I don't quite get why the query appears to work in prod (since we see sensible dumps getting updated daily), but not locally.

gbrownmozilla:
I can't wrap my head around the differences and won't try debugging it on pto. Immediate thoughts: Is the difference related to some table or field being empty locally?

jcristau:
Thanks @gbrownmozilla!  The difference was indeed due to an empty `rules_scheduled_changes` locally.

gabrielBusta (approved):

Issue #2716, opened 2022-11-21
docker build jobs are failing


jcristau:
The docker image builds from the v3.21 release failed:
https://firefox-ci-tc.services.mozilla.com/tasks/groups/A7zlJHO-SFm8CBgf4lCNUw

As far as I can tell apt is failing to talk to get.docker.io, thus jq is not being installed, and we can't build the docker config file.

An easy workaround is to ignore the return code from apt-get update and hope that is enough.

That said we should also stop using a docker image from 2015, based on ubuntu 14.04 ([taskcluster/image_builder:0.1.3](https://hub.docker.com/layers/taskcluster/image_builder/0.1.3/images/sha256-25baefca49788090be9d71b5ed13eef553cd5b24b78cac6c1215a1338c841e9e?context=explore)).

jcristau:
This is fixed now after #2719 and followups, after a number of false starts v3.21 is published and deployed to stage.

PR #2717, opened 2022-11-22
Ignore errors from apt-get update in push-dockerimage script

jcristau:
The script currently runs in an environment based on ubuntu 14.04, which seems unable to talk to get.docker.io nowadays (see #2716).


gabrielBusta (approved):
We should definitely update to a more recent Ubuntu release :/

LGTM

gbrownmozilla (approved):

PR #2718, opened 2022-11-22
Don't run CI on pull-request "edited" events

jcristau:


gabrielBusta (approved):
Nice

gbrownmozilla (approved):

jcristau:
> Although, does that mean the CI won't run if I like, rebase the PR?

My understanding is that changing the source branch's HEAD commit will send a "synchronize" event, which will trigger CI.

PR #2719, opened 2022-11-23
Switch CI to taskgraph

jcristau:
Use taskgraph instead of defining tasks directly in .taskcluster.yml.

The push-dockerimage.sh scripts are replaced by a skopeo-based script stolen
from shipit, which uses the image built in a separate task and just overwrites
its version.json, instead of building a completely new image before pushing it
to docker hub.

Unfortunately this requires duplicating the Dockerfiles for backend and agent under
`taskcluster/docker/` because of the different expectations for things like
paths.

The balrog-specific taskgraph code is necessary to:
- add a `pull_request_number` decision task parameter and pass the pull request number to coveralls via environment
- add a "release" target_tasks method for `github-release` events, and filter tasks on github-release based on the git tag name (`run-on-releases`, similar to the existing `run-on-git-branches`)
- pass some params into the push-image script's environment


taskcluster/balrog_taskgraph/transforms/docker_push.py
```diff
@@ -0,0 +1,33 @@
...
+    for job in jobs:
+        version_file = os.path.join(config.graph_config.vcs_root, "version.txt")
+        version = open(version_file).read().strip()
+
+        env = job["worker"].setdefault("env", {})
+        env.update(
+            {
+                "VCS_HEAD_REPOSITORY": config.params["head_repository"],
+                "VCS_HEAD_REV": config.params["head_rev"],
+                # TODO: Figure out if we still need to add the version
```
gbrownmozilla, line 29: Want to address the TODO now?

taskcluster/ci/push-image/kind.yml
```diff
@@ -0,0 +1,59 @@
...
+        taskcluster-proxy: true
+        docker-image: {in-tree: skopeo}
+        max-run-time: 3600
+    dependencies:
+          tests-js: ui-tests
+          build-js: ui-build
+          tests-backend: test-backend
+          tests-agent: test-agent
+    run-on-tasks-for: [github-release]
+    run-on-releases: [v]
```
gbrownmozilla, line 29: Why "v"? That seems particularly terse and impossible to search for.

taskcluster/docker/skopeo/Dockerfile
```diff
@@ -0,0 +1,60 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+FROM golang:1.14 as skopeo
+
+WORKDIR /go/src/
+RUN ["/usr/bin/git", "clone", "--no-checkout", "--depth=1", "--branch=v1.1.0", "https://github.com/containers/skopeo", "."]
+RUN ["/usr/bin/git", "checkout", "63085f5bef1131aa9ec0907a5c8d66b67de7c4b2"]
+ENV GO111MODULE=on CGO_ENABLED=0
```
gbrownmozilla, line 10: I don't like all these external dependencies and odd-looking commands...but I don't have a better suggestion.

taskcluster/ci/push-image/kind.yml
jcristau: This is used as a regex, i.e. it means this task runs on release whose tag starts with the letter `v`, like today's .taskcluster.yml has `$if: 'tasks_for == "github-release" && release_tag[0] == "v"'`

taskcluster/docker/skopeo/Dockerfile
jcristau: Ack.  I should have mentioned a lot of this was stolen from https://github.com/mozilla-releng/shipit/pull/336 so that's arguably a pre-existing problem.

taskcluster/balrog_taskgraph/transforms/docker_push.py
jcristau: There's two places we have the version number for a release:
- version.txt; taskgraph also reads that and fills in the "version" param (https://github.com/taskcluster/taskgraph/blob/591768b3fd5b8aa50e003e43d29cc849cd880388/src/taskgraph/parameters.py#L121)
- the tag name, which also ends up as the "head_tag" param

I figure we should put a check in the github-release decision task to abort if the two aren't the same.  I'll remove this TODO, though, because as far as I can tell we do need to pass the version to the push-image task which uses it to generate `/app/version.json` in the image.

taskcluster/ci/push-image/kind.yml
gbrownmozilla: :face_palm

Of course!

gbrownmozilla (approved):

PR #2780, opened 2023-01-03
Dependencies

jcristau:


jcristau:
Besides running pin.sh there's a couple extra bits in there to:
- solve dependency conflicts (latest sphinx and sphinx-rtd-theme disagree on docutils version, and tox pulls in a new version of chardet that google libs don't like)
- fix up a "unused import" flake8 error about `typing.AnyStr` that started showing up
- make pin-helper print out the commands it runs
- delete the docker container created by pin.sh after it's done

gbrownmozilla (approved):
Thanks for keeping deps up to date! I like your pin-helper changes.

gabrielBusta (approved):

Issue #2785, opened 2023-01-09
rule edit pages take too long to load when there's a lot of v2 releases


bhearsum:
Currently, loads of /api/v2/releases are timing out in balrog production after 60 seconds, causing rule edit pages to fail to work. These queries are ultimately fulfilled by [this code](https://github.com/mozilla-releng/balrog/blob/f5fd742d39a40491d11c694b910ae24012161ac8/src/auslib/services/releases.py#L258), which joins to a whole bunch of tables - so it's no wonder it's so slow.

We really only need the names of the releases to fulfill the needs here - so we ought to use a different endpoint. I see two options (although maybe there's more):
1) Add a new v2 endpoint that's powered by [get_release_names](https://github.com/mozilla-releng/balrog/blob/f5fd742d39a40491d11c694b910ae24012161ac8/src/auslib/services/releases.py#L250)
2) Add an endpoint to the admin app that's powered by [release_list](https://github.com/mozilla-releng/balrog/blob/f5fd742d39a40491d11c694b910ae24012161ac8/src/auslib/web/common/releases.py#L24) -- which queries both v1 and v2 releases, and can return names only. (We already have an endpoint like this for the public API.)

In either case, we need to [update the UI code to cope with the change](https://github.com/mozilla-releng/balrog/blob/f5fd742d39a40491d11c694b910ae24012161ac8/ui/src/views/Rules/Rule/index.jsx#L169-L178).

jcristau:
Fixed by #2787

PR #2787, opened 2023-01-09
Use get_release_names to avoid timeouts on Rule view

gabrielBusta:
#2785 

cc @bhearsum

### DIFF
--- a/src/auslib/web/admin/swagger/api_v2.yml
+++ b/src/auslib/web/admin/swagger/api_v2.yml
@@ -478,8 +478,9 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/ReleaseModifyResponse'
-                $ref: '#/components/schemas/ReleasePinNotSetResponse'
+                anyOf:
+                  - $ref: '#/components/schemas/ReleaseModifyResponse'
+                  - $ref: '#/components/schemas/ReleasePinNotSetResponse'
         '400':
           description: Client error
         '403':
@@ -487,6 +488,22 @@ paths:
 
   /pins/{product}/{channel}/{version}:
     get:
+      parameters:
+        - in: path
+          name: product
+          schema:
+            type: string
+          required: true
+        - in: path
+          name: channel
+          schema:
+            type: string
+          required: true
+        - in: path
+          name: version
+          schema:
+            type: string
+          required: true
       summary: Get a JSON object representing the specified pin
       operationId: auslib.web.admin.views.pins_v2.get_pin
       responses:
--- a/src/auslib/web/admin/swagger/api_v2.yml
+++ b/src/auslib/web/admin/swagger/api_v2.yml
@@ -121,83 +121,98 @@ paths:
     get:
       summary: Get a list of basic information about Releases in Balrog's database
       operationId: auslib.web.admin.views.releases_v2.get_releases
+      parameters:
+        - name: names_only
+          in: query
+          description: "If 'names_only' is set to true, the response changes format and provides a list of all of the Release names in the database"
+          schema:
+            type: string
+          required: false
       responses:
         '200':
           description: List of Releases
           content:
             application/json:
               schema:
-                type: object
-                required:
-                  - releases
-                properties:
-                  releases:
-                    type: array
-                    items:
-                      allOf:
-                        - $ref: '#/components/schemas/Release'
-                        - type: object
-                          required:
-                            - name
-                            - rule_info
-                          properties:
-                            product_required_signoffs:
-                              $ref: '#/components/schemas/RequiredSignoffs'
-                            required_signoffs:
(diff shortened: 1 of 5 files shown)


src/auslib/web/admin/swagger/api_v2.yml
```diff
@@ -463,8 +478,9 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/ReleaseModifyResponse'
-                $ref: '#/components/schemas/ReleasePinNotSetResponse'
+                anyOf:
+                  - $ref: '#/components/schemas/ReleaseModifyResponse'
+                  - $ref: '#/components/schemas/ReleasePinNotSetResponse'
```
bhearsum, line 483: This is an unrelated fix, I assume?

ui/src/views/Rules/Rule/index.jsx
```diff
@@ -170,7 +170,9 @@ function Rule({ isNewRule, user, ...props }) {
     Promise.all([fetchReleaseNames(), fetchReleaseNamesV2()]).then(
       ([names, namesV2]) => {
         if (names.data.data && namesV2.data) {
-          setReleaseNames(names.data.data.names.concat(namesV2.data).sort());
+          setReleaseNames(
+            names.data.data.names.concat(namesV2.data.data).sort()
```
bhearsum, line 174: 3 names and 4 datas in one line, wow! It might be worth doing some of the parsing in `services/releases.js` instead (as we do in the code you're removing) -- but that's not at all a blocker.

src/auslib/web/admin/swagger/api_v2.yml
```diff
@@ -463,8 +478,9 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/ReleaseModifyResponse'
```
gabrielBusta, line 466: Not sure about this one, but having a non-unique key in an object is not valid YAML

bhearsum (approved):
Thank you for getting this together so quickly! One thing you could optionally improve below, but it's definitely not a blocker.

src/auslib/web/admin/swagger/api_v2.yml
```diff
@@ -463,8 +478,9 @@ paths:
           content:
             application/json:
               schema:
```
gabrielBusta, line 480: 🤷❓afik having a non-unique key in an object is not valid YAML

src/auslib/web/admin/swagger/api_v2.yml
gabrielBusta: Yup, a nit-picky one at that. The Swagger parser is complaining that `Map keys must be unique` I try to load the API spec in the Swagger editor. I opted for `anyOf` because it satisfies the test cases for this end-point

src/auslib/web/admin/swagger/api_v2.yml
```diff
@@ -463,15 +478,32 @@ paths:
...
+                anyOf:
+                  - $ref: '#/components/schemas/ReleaseModifyResponse'
+                  - $ref: '#/components/schemas/ReleasePinNotSetResponse'
         '400':
           description: Client error
         '403':
           description: Permission denied
 
   /pins/{product}/{channel}/{version}:
     get:
```
gabrielBusta, line 490: When I loaded the API spec into the Swagger editor it found a few semantic errors on this end-point because "Declared path parameter "foo" needs to be defined as a path parameter at either the path or operation level"

bhearsum:
Whoops, I didn't realize @jcristau had been requested to review this, heh. I still think he ought to look at it before it lands, as he's much more up to date on Balrog than me.

jcristau:
Thanks for tackling this.

I sort of feel like this is making the API spec more complex, and having a single API endpoint to get both v1 and v2 release names, using the existing code in `auslib.web.common.releases.get_releases`, or at least somehow sharing code with the public `/api/v1/releases`, would be simpler all around: the server-side code already exists since it is used by the public endpoint, and the frontend would just need to do a single xhr instead of two.  Is there a particular reason not to go that way?
Also: AIUI we're setting the swagger version to 2.0, and there's a few preexisting comments about oneOf only being supported in 3.0, and so working around that by listing the different possible properties without any of them being "required".  Likewise anyOf doesn't seem to be in the 2.0 version.

One nit: use a separate commit for fixing existing API spec bugs

And one possible followup: can we somehow catch/expose the issues you found in the swagger spec in CI?

gabrielBusta:
I ended up just sticking with the query param because it's more symmetrical with what is already there so it's a smaller refactor.

We are using openapi 3.0 on the v2 file, I am going to clean up v1 too.

I found a way to test the spec in CI https://github.com/APIDevTools/swagger-cli

CI check coming soon

PR #2803, opened 2023-01-20
tests: add a check that we can deal with encoded colon and comma in system capabilities

### LINKED ISSUE #1225: Ensure URL encoding is supported in SYSTEM_CAPABILITIES
We currently have values such as `ISET:SSE4_2,MEM:16192` in this field, which are technically invalid in an URL. https://bugzilla.mozilla.org/show_bug.cgi?id=1620327 is seeking to URL encode these to make them vaild. We need to make sure this is supported before that lands.

jcristau:
[Bug 1620327](https://bugzilla.mozilla.org/show_bug.cgi?id=1620327) wants to encode the update URL properly, so let's make sure it works.

As far as I can tell this doesn't require changes to the application code, because the CGI and wsgi specs say PATH_INFO is not URL-encoded, so we don't have to deal with this at all.

Closes #1225

### DIFF
--- a/tests/web/test_client.py
+++ b/tests/web/test_client.py
@@ -1115,6 +1115,19 @@ class ClientTest(ClientTestBase):
 """,
         )
 
+    def testVersion6GetWithLessThanEqualToMemoryEncoded(self):
+        ret = self.client.get("/update/6/f/1.0/1/p/l/a/a/ISET%3aSSE3%2cMEM%3a8000/a/a/update.xml")
+        self.assertUpdateEqual(
+            ret,
+            """<?xml version="1.0"?>
+<updates>
+    <update type="minor" version="1.0" extensionVersion="1.0" buildID="35">
+        <patch type="complete" URL="http://a.com/f" hashFunction="sha512" hashValue="34" size="33"/>
+    </update>
+</updates>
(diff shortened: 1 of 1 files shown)


tests/web/test_client.py
```diff
@@ -1112,6 +1112,19 @@ def testVersion6GetWithLessThanEqualToMemory(self):
         <patch type="complete" URL="http://a.com/f" hashFunction="sha512" hashValue="34" size="33"/>
     </update>
 </updates>
+""",
+        )
+
+    def testVersion6GetWithLessThanEqualToMemoryEncoded(self):
+        ret = self.client.get("/update/6/f/1.0/1/p/l/a/a/ISET%3aSSE3,MEM%3a8000/a/a/update.xml")
```
jcristau, line 1119: Hmm I need to quote commas as well not just colons.  Will update the patch on Monday.

gbrownmozilla (approved):

gabrielBusta (approved):

bhearsum:
@jcristau - should this get landed?

jcristau:
@bhearsum I wanted to go back and take a closer look at this, so not quite yet.

PR #2804, opened 2023-01-20
Remove UnquotingMiddleware

jcristau:
This was added in #85 to work around a bug in uwsgi, which has long since been fixed (https://github.com/unbit/uwsgi/pull/1267)


gbrownmozilla (approved):
Yeah!

PR #2812, opened 2023-01-24
Fix caching in agent tests & remove dependency on asynctest

bhearsum:
While poking at other things, I noticed that balrogagent local tests always re-install test dependencies. I don't fully understand why this was happening, but it was the `pip install` in the tiny test script causing it. The first revision here switches its test set-up to be more in line with the webapp -- using a separate Dockerfile that installs the test dependencies instead.

The second revision here gets rid of the dependency on `asynctest`, which hasn't been updated in years, and won't work with Python 3.11.


agent/Dockerfile.test
```diff
@@ -0,0 +1,21 @@
+# WARNING: keep this file in sync with taskcluster/docker/balrog-agent/Dockerfile
```
jcristau, line 1: I don't really have a good solution, but the fact that we have Dockerfile + Dockerfile.test + another Dockerfile for use in tc for each of agent and backend is a recipe for issues later when they get out of sync and tests pass but either the published images or the local dev ones are broken.  Not really sure what to do about this though.  Maybe an extra check in CI that makes sure all the differences between them are accounted for, as a poor man's replacement for actually sharing the bits?

jcristau (approved):

agent/Dockerfile.test
bhearsum: Yeah. This is my fault, and I'm not really happy about it. It looks like the recommended way to do this these days is multistage builds: https://stackoverflow.com/a/67937236 -- so maybe we ought to try this at some point.

PR #2867, opened 2023-03-21
Switch nginx image to an official one

bhearsum:
To prevent any potential issues when free Docker teams go away.


bhearsum:
I tested this out locally and the nginx container seemed to work fine still.

gbrownmozilla (approved):

PR #2869, opened 2023-05-24
Update documentation for 2023 aus5.mozilla.org cert renewal

bhearsum:
Note: this should not be merged until we successfully deploy the new cert - planned for the morning of May 25th.


jcristau (approved):
:+1:

jcristau:
If anybody else wonders about the test-backend failure in CI: it turns out tests/admin/views/test_releases_v2.py hardcodes a timestamp for some scheduled changes, and that is now in the past (Sun Apr 16 10:12:12 AM UTC 2023), causing the tests to fail.

PR #2906, opened 2023-07-05
Remove seemingly-unnecessary urldecoding

bhearsum:
I added this a Very Long Time Ago in https://github.com/mozilla-releng/balrog/commit/adb45b60b2b9b31e0a765ad3f2dd4c8f943adc45. I don't see anything in there nor the bug it references that suggests it was needed, nor do any tests fail if it is removed. There's no data in this field that _should_ be urlencoded, and even if there was, decoding is already handled by Flask/Werkzeug...so we should be able to safely remove this.

### DIFF
--- a/src/auslib/web/public/client.py
+++ b/src/auslib/web/public/client.py
@@ -12,12 +12,6 @@ from auslib.global_state import dbo
 from auslib.services import releases
 from auslib.web.public.helpers import AUS, get_aus_metadata_headers, get_content_signature_headers, with_transaction
 
-try:
-    from urllib import unquote
-except ImportError:  # pragma: no cover
-    from urllib.parse import unquote
-
-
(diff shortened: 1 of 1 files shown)


src/auslib/web/public/client.py
```diff
@@ -110,7 +104,7 @@ def getQueryFromURL(url):
     if "systemCapabilities" in query:
         query.update(getSystemCapabilities(url["systemCapabilities"]))
         del query["systemCapabilities"]
-    query["osVersion"] = unquote(query["osVersion"])
+    query["osVersion"] = query["osVersion"]
```
jcristau, line 107: Just remove the line? :)

bhearsum: errrrrrrrrm, yes

jcristau (approved):
:+1:

PR #2927, opened 2023-07-19
Update Dockerfile.test - fix netcat installation

chrmod:
Looks like netcat is provided by two different packages:
```
#0 2.852 Package netcat is a virtual package provided by:
#0 2.852   netcat-openbsd 1.219-1
#0 2.852   netcat-traditional 1.10-47
#0 2.852
#0 2.859 E: Package 'netcat' has no installation candidate
```

### DIFF
--- a/Dockerfile.test
+++ b/Dockerfile.test
@@ -13,7 +13,7 @@ MAINTAINER jcristau@mozilla.com
 # xz-utils is needed to unpack sampled ata
 # git is needed to send coverage reports
 RUN apt-get -q update \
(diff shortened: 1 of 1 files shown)


jcristau (commented):
Thanks for the patch.  I guess an alternative fix would be to use `python:${PYTHON_VERSION}-slim-bullseye` as base image.
If we're going to replace netcat we might want to stick to netcat-openbsd unless there's a reason to switch, since -openbsd is what the old netcat package used to pull in (https://packages.debian.org/bullseye/netcat)?

jcristau:
This was fixed by #2993.

PR #2989, opened 2023-09-22
Update deployment docs

bhearsum:


docs/source/infrastructure.rst
```diff
@@ -28,9 +28,7 @@ We have stage and production deployments of Balrog. Here's a quick summary:
 Support & Escalation
 --------------------
 
-RelEng is the first point of contact for issues. To contact them, follow `the standard RelEng escalation path <https://wiki.mozilla.org/ReleaseEngineering#Contacting_Release_Engineering>`_.
-
-If RelEng is unable to correct the issue, they may `escalate to CloudOps <https://mana.mozilla.org/wiki/display/SVCOPS/Contacting+Services+SRE>`_.
```
gbrownmozilla, line 33: Are we actually losing SRE support? Aren't they still valid for escalation?

docs/source/infrastructure.rst
```diff
@@ -117,27 +112,39 @@ Once the changes are deployed to stage, you should do some testing to make sure
 Pushing to Production
 ~~~~~~~~~~~~~~~~~~~~~
 
-Pushing live requires CloudOps. For non-urgent pushes, you should begin this procedure a few hours in advance to give CloudOps time to notice and respond. For urgent pushes, file the bug immediately and escalate if no action is taken quickly. Either way, you must follow this procedure to push:
+Pushing the backends live requires some button clicking in Jenkins. For each of ``balrog-admin-production``, ``balrog-production``, and ``balrog-agent-production`` in Jenkins do the following:
```
gbrownmozilla, line 115: Do we recommend waiting for one to complete before clicking the next...or does the ui force you to?

docs/source/infrastructure.rst
```diff
@@ -117,27 +112,39 @@ Once the changes are deployed to stage, you should do some testing to make sure
...
+  * Click ``Proceed``
 
-  * Make sure you substitute the version number and choose the correct options from the bug template.
+.. image:: proceed.png
 
-2. Before SRE start the deploy, notify `#releaseduty:mozilla.org` and `#sheriffs:mozilla.org` on matrix so they can both confirm no release activity is ongoing, and know to quickly escalate any fallout.
-3. Once the push has happened, verify that the code was pushed to production by checking the __version__ endpoints on `the Admin <https://aus4-admin.mozilla.org/__version__>`_ and `Public <https://aus5.mozilla.org/__version__>`_ apps.
-4. Manually delete and recreate the "production-ui" tag & release on Github to push the new UI to production:
+After this, there is nothing else to do for ``balrog-admin-production`` nor ``balrog-agent-production``. However, the public app (``balrog-production``) will first deploy a canary (meaning the new code will only be used for a small fraction of requests). You should monitor for changes in load or exceptions for at least a few minutes before proceeding. When you are ready, find the ``PROD: PROMOTE`` cell in Jenkins and click ``Proceed`` to finish with this deployment.
```
gbrownmozilla, line 123: Maybe mention how to monitor for changes in load / exceptions?

gbrownmozilla (approved):
Thanks so much for this. The screenshots are much appreciated. Just a few thoughts for possible improvement and clarification...

docs/source/infrastructure.rst
bhearsum: We are definitely not losing their support :).

I mostly removed this because I was thinking that it's same as what we do for everything else deployed through SRE. However, maybe it's more important to keep this here because this an externally facing system...

docs/source/infrastructure.rst
bhearsum: These can and should happen in parallel - I'll clarify this.

docs/source/infrastructure.rst
bhearsum: I'll add a bit of detail here, thanks for the callouts!

Issue #1146, opened 2019-09-19
Consider have a "partial match filter" on Releases page


allan-silva:
Maybe can be useful to users have a partial filter of releases. E.g: if I type "Fire 69", it will match entries like "Firefox-69.0b10-build1".

nazarenegena:
hey @bhearsum, could you assign this to me, I would like to tackle it..

PR #2997, opened 2023-10-06
partial match filter on releases

nazarenegena:
Fixes issue #1146 

Allows partial filtering of releases when a search value such as "fire 60"  is entered.

### DIFF
--- a/ui/src/components/SearchBar/index.jsx
+++ b/ui/src/components/SearchBar/index.jsx
@@ -24,7 +24,7 @@ function SearchBar(props) {
   const classes = useStyles();
 
   return (
-      <Paper className={classes.root}>
+    <Paper className={classes.root}>
       <InputBase
         className={classes.input}
         placeholder="Search..."
@@ -32,7 +32,7 @@ function SearchBar(props) {
         inputProps={{ 'aria-label': 'Search' }}
         {...rest}
       />
-      </Paper>
+    </Paper>
   );
 }
 
--- a/ui/src/components/SearchBar/index.jsx
+++ b/ui/src/components/SearchBar/index.jsx
@@ -24,7 +24,7 @@ function SearchBar(props) {
   const classes = useStyles();
 
   return (
-    <Paper className={classes.root}>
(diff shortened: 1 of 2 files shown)


gabrielBusta:
Thank you @nazarenegena

These changes are failing the linter check.

You can easily fix that by running this commands in your local repo:
```
cd ui/
yarn lint --fix
```

gabrielBusta (approved):
Thank you! The patch seems good and achieves the functionality described in #1146. In production the patch might slow down the filtering. In the development database there are about one hundred releases vs. the six thousand (and counting) releases in production. Slightly slower filtering seems like a good trade-off for the enhanced user experience (partial matching could be really useful.) I'd say we can land it, try it out and see. Dunno is trying it out in staging would be helpful (there's about six hundred releases in there)

gabrielBusta (approved):

gabrielBusta:
It would be helpful for users if the matching parts of their search is highlighted in the results. This would give visual feedback about what part of their query matched.

nazarenegena:
Thank you for the feedback.
should I go ahead and implement the highlighting and the performance check?

gabrielBusta:
Regarding the highlighting, we have two options: we can open an issue for further follow-up and merge this pull request as it is, or you can address it directly here. The choice is yours. As for performance, what specific checks were you considering? It would be beneficial to gauge the code's performance on several thousand releases before deploying to production. However, if testing proves too cumbersome, we might consider deploying it to production directly and observing its performance there. If the speed is insufficient, we can always revert. We shouldn't prematurely optimize the code without identifying an issue. Should we encounter problems, there are various solutions to explore. For instance, we could think about debouncing the search. This approach would delay the search function execution until the user pauses typing for a brief moment, like 300ms, rather than executing on every keypress.

nazarenegena:
Yes, you can go ahead and merge this request create a new issue for the highlights and assign it to me.

gabrielBusta:
OK I created #3007 as a follow-up to this issue.

Issue #2943, opened 2023-08-04
OS version is cropped when approximately more than 2 versions are provided


JohanLorenzo:
A few weeks ago, we created this Balrog rule that prevents Firefox users on deprecated Windows and macOS versions to be updated. The rule list 6 OS versions, but only 2 are visible: 

![image](https://github.com/mozilla-releng/balrog/assets/5907366/ba31dfea-c13d-4c14-aef6-18c35babb7bb)

The list should have displayed:

```
Windows_NT 6.1,Windows_NT 6.2,Windows_NT 6.3,Darwin 16,Darwin 17,Darwin 18
```

Let's change the UI to show all OS versions at once.

akankshat05:
hello, I can fix this issue, please assign this to me. thank you.

AbassKoyang:
> hello, I can fix this issue, please assign this to me. thank you.

Are you still working on it? can i take it up?

akankshat05:
> > hello, I can fix this issue, please assign this to me. thank you.
> 
> Are you still working on it? can i take it up?

Yes, I am working on it.

godplayer56:
![Screenshot from 2023-10-06 16-06-09](https://github.com/mozilla-releng/balrog/assets/121134233/d945fd59-314d-4e67-a157-a11ac8a1a32f)
Is this an issue with this function?
I think we can use text wrapping for this.

sumairq:
It is solved by removing 

```
 secondaryTypographyProps={{
                        className: classes.textEllipsis,
                      }}
``` 
from `ListItemText` 

https://github.com/mozilla-releng/balrog/blob/860e3bb34c7c0de73a9dc3d4419787bbf0810b13/ui/src/components/RuleCard/index.jsx#L560C17-L584C30

CREDO23:
It is improved by making the `osVersion` a gridItem and set it to take the full width of its parent:
`<ListItem/>` → `<Grid item xs={12}>...</Grid>`

```javascript
                   <Grid xs={12}>
                          {rule.osVersion && (
                              <ListItem className={classes.listItem}>
                                 <ListItemText
                                     title={rule.osVersion.split(',').join('\n')}
                                     primaryTypographyProps={{
                                     component: 'div',
                                     className: classes.primaryText,
                                     }}
                                     primary={
                                                   <Fragment>
                                                      OS Version
                                                      {diffedProperties.includes('osVersion') &&
                                                      rule.scheduledChange.change_type === 'update' && (
[...]

PR #2998, opened 2023-10-06
RuleCard: Remove textEllipses class from OS secondary text

sumairq:
## This PR fixes issue #2943 

We can see the complete List of OS Versions          |
:-------------------------:|
ruleCard | 

### Context
We have styling specified for `textEllipses` class that hides the overflow and shows ellipses (...) instead, to be used in the scenarios when the text is longer than the container.
A `textEllipses` class can be used wherever we need this behaviour. 

### Changes Made
Since it is preferable to see the full OS version, 
- `textEllipses` class has been removed for OS version specifically.

### DIFF
--- a/ui/src/components/RuleCard/index.jsx
+++ b/ui/src/components/RuleCard/index.jsx
@@ -565,9 +565,6 @@ function RuleCard({
                         component: 'div',
                         className: classes.primaryText,
                       }}
-                      secondaryTypographyProps={{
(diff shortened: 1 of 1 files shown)


gabrielBusta (approved):
Thanks @sumairq! The patch looks good to me.

denscholar:
Nice one @sumairq I see what you are doing. Keep it up. I am so proud of you brother.

PR #2999, opened 2023-10-06
refact(RuleCard): show all OS versions at once

### LINKED ISSUE #2943: OS version is cropped when approximately more than 2 versions are provided

CREDO23:
## This PR fixes #2943 

Now we can see all OS versions:

Screenshot from 2023-10-06 18-05-15

Screenshot from 2023-10-09 09-15-03

The RuleCard component is based on the material UI grid system, elements inside it are set in percentages, so they're always fluid and sized relative to their parent element.

### Changes made:

- Change the `osVesrion` listItem into a gridItem and give it the full width of its parent (`xs=12`) so osVesrion has enough space to be seen

### DIFF
--- a/ui/src/components/RuleCard/index.jsx
+++ b/ui/src/components/RuleCard/index.jsx
@@ -557,29 +557,7 @@ function RuleCard({
                     />
                   </ListItem>
                 )}
-                {rule.osVersion && (
-                  <ListItem className={classes.listItem}>
-                    <ListItemText
-                      title={rule.osVersion.split(',').join('\n')}
-                      primaryTypographyProps={{
-                        component: 'div',
-                        className: classes.primaryText,
-                      }}
-                      primary={
-                        <Fragment>
-                          OS Version
-                          {diffedProperties.includes('osVersion') &&
-                            rule.scheduledChange.change_type === 'update' && (
-                              <span
-                                className={classes.propertyWithScheduledChange}
-                              />
-                            )}
-                        </Fragment>
-                      }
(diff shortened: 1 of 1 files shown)


ui/src/components/RuleCard/index.jsx
```diff
@@ -692,6 +669,31 @@ function RuleCard({
                 )}
               </List>
             </Grid>
+            <Grid xs={12}>
+              {rule.osVersion && (
+                <ListItem className={classes.listItem}>
```
gabrielBusta, line 674: Should the list item be in a list? Looks OK.

docker-compose.yml
```diff
@@ -163,9 +163,9 @@ services:
       context: .
       dockerfile: Dockerfile.database
     expose:
-      - "3306"
+      - "3307"
```
gabrielBusta, line 166: Could you revert the changes on this file?

gabrielBusta (changes requested):

docker-compose.yml
CREDO23: Sure z let me do it!

PR #3000, opened 2023-10-07
Updated issue label names for frontend bugs

AMUZY:
Hi, I'm an outreachy applicant and I'd like to make my first contribution. I had an issue identifying the right issues I was to begin with as a newbie here, due to the fact that the label names were outdated. So, in this commit, I have updated the label names so that other newbies after me wouldn't have to ask the same questions about this.


gabrielBusta (approved):
Thanks! Looks good to me :)

PR #3001, opened 2023-10-08
added different background colors in header

akankshat05:
Fixes [1384](https://github.com/mozilla-releng/balrog/issues/1384).
Added different background header colors (From the theme file) for the development and staging environments.

Staging :
Screenshot 2023-10-08 203424

Local :
Screenshot 2023-10-18 152437

### DIFF
--- a/ui/src/components/Dashboard/index.jsx
+++ b/ui/src/components/Dashboard/index.jsx
@@ -16,6 +16,12 @@ const useStyles = makeStyles(theme => ({
   appbar: {
     height: APP_BAR_HEIGHT,
   },
+  staging: {
+    backgroundColor: theme.palette.info.main
+  },
+  local: {
+    backgroundColor: theme.palette.warning.main
+  },
   title: {
     textDecoration: 'none',
   },
@@ -54,7 +60,13 @@ export default function Dashboard(props) {
       <Helmet>
         <title>{title} - Balrog Admin</title>
       </Helmet>
-      <AppBar className={classes.appbar}>
+      <AppBar className={`${classes.appbar} ${process.env.NODE_ENV === 'development' ? (
+      classes.local
+    ) : (
+      process.env.NODE_ENV === 'staging' ? (
+        classes.staging
+      ) : null
+    )}`}>
         <Toolbar>
           <Typography
             className={classes.title}
--- a/ui/src/components/Dashboard/index.jsx
+++ b/ui/src/components/Dashboard/index.jsx
@@ -1,18 +1,18 @@
-import React, { Fragment } from "react";
-import { string, node, bool } from "prop-types";
-import { Helmet } from "react-helmet";
-import { makeStyles } from "@material-ui/styles";
-import AppBar from "@material-ui/core/AppBar";
-import Toolbar from "@material-ui/core/Toolbar";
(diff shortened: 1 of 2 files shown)


ui/src/components/Dashboard/index.jsx
```diff
@@ -54,7 +60,12 @@ export default function Dashboard(props) {
       <Helmet>
         <title>{title} - Balrog Admin</title>
       </Helmet>
-      <AppBar className={classes.appbar}>
+      <AppBar
+        className={`${classes.appbar} ${
+          process.env.NODE_ENV === 'development' ? classes.local : null
```
gabrielBusta, line 65: Mm is `process.env.NODE_ENV` available in the production build?

I think we'd have to do something like to make it available https://github.com/neutrinojs/neutrino/issues/82#issuecomment-417226109

ui/src/components/Dashboard/index.jsx
```diff
@@ -16,6 +16,12 @@ const useStyles = makeStyles(theme => ({
   appbar: {
     height: APP_BAR_HEIGHT,
   },
+  staging: {
+    backgroundColor: theme.palette.info.main,
+  },
+  local: {
+    backgroundColor: theme.palette.warning.main,
```
gabrielBusta, line 23: Let's choose darker tints since the text on the app bar is white (e.g. `palette.success.dark`)

ui/src/components/Dashboard/index.jsx
gabrielBusta: Mm maybe `process.env` is available, looks like we reference it in a few places like https://github.com/mozilla-releng/balrog/blob/05036afc2db59837938aae4423ec81e7eaa57b93/ui/src/utils/constants.js#L2

gabrielBusta: Do we need to add it here? https://github.com/mozilla-releng/balrog/blob/05036afc2db59837938aae4423ec81e7eaa57b93/ui/.neutrinorc.js#L64-L76

akankshat05: > Mm is `process.env.NODE_ENV` available in the production build?
> 
> I think we'd have to do something like to make it available [neutrinojs/neutrino#82 (comment)](https://github.com/neutrinojs/neutrino/issues/82#issuecomment-417226109)

Thank you for updating me. The `process.env.NODE_ENV` would not be available in the production build. I will be using the DefinePlugin plugin to create a constant for it.

ui/src/components/Dashboard/index.jsx
akankshat05: Sure, will update it.

ui/src/components/Dashboard/index.jsx
gabrielBusta: Do you need some help adding it? Once you add it I can land this to try it out in the staging instance :)

gabrielBusta (changes requested):
Once you add the environment global I can land this and try it out on our staging instance

bhearsum:
@akankshat05 - is this ready for another look? (Please make sure to press the re-request review button next to @gabrielBusta's name if so - otherwise we can't be sure.)

Issue #1145, opened 2019-09-10
don't prompt for signoff with roles that aren't relevant


bhearsum:
Examples:
- If `releng` and `relman` are required for a change, and a user holds `releng` and `tb-releng`, we currently show a prompt that allow the user to sign off with either. We should just automatically sign off with `releng` instead.
- If `releng` and `relman` are required for a change, and a user holds `releng`, `relman`, and `tb-releng`, we currently show a prompt that allow the user to sign off with any of them. We should still show a prompt in this case, but only allow `releng` or `relman` to be used.

escapewindow:
<img width="845" alt="Screen Shot 2021-04-02 at 9 41 58 AM" src="https://user-images.githubusercontent.com/826343/113435514-d0f51700-9397-11eb-88d6-4349345f52fa.png">

ruthEnanga:
@bhearsum 
I will like to work on this
Can it be assigned to me?

CREDO23:
What happens if the user holds `tb-releng` and the required are `relman` and/or `releng` ?

bhearsum:
> @bhearsum I will like to work on this Can it be assigned to me?

Hi, please feel free to work on the issue and open a PR when you feel it is ready.

CREDO23:
Okay, let me add this feature!

bhearsum:
Fixed by #3021

PR #3002, opened 2023-10-09
don't prompt for signoff with roles that aren't relevant

CREDO23:
### Changes made:
- Retreive required roles:
    ```javascript
            useEffect(() => {
                const roles =
                  requiredSignoffs.data &&
                  requiredSignoffs.data.data.required_signoffs.map(el => {
                    return el.role;
                  });

                setRequiredRoles(roles);
            }, []);
    ```
- Disable roles that are not required
    ```javascript
        const signoffDialogBody = (
            <FormControl component="fieldset">
              <RadioGroup
                aria-label="Role"
                name="role"
                value={signoffRole}
                onChange={handleSignoffRoleChange}>
                {roles.map(r => {
                  return (
                    <FormControlLabel
                      key={r}
                      value={r}
                      label={r}
                      control={<Radio />}
                      disabled={!(requiredRoles && requiredRoles.includes(r))}
                    />
                  );
                })}
              </RadioGroup>
            </FormControl>
        );
    ```
- If the user has only one role in the required role list, `signoff` directly:
    ```javascript
          const handleSignoff = async rule => {
            const userRequiredRoles = requiredRoles.filter(r => roles.includes(r));

            if (userRequiredRoles.length === 1) {
              const { error, result } = await doSignoff(userRequiredRoles[0], rule);

              if (!error) {
                updateSignoffs(result);
              }
            } else {
              setDialogState({
                ...dialogState,
                open: true,
                title: 'Signoff as…',
                confirmText: 'Sign off',
                item: rule,
                mode: 'signoff',
                handleComplete: handleSignoffDialogComplete,
                handleSubmit: handleSignoffDialogSubmit,
              });
            }
  };
    ```

### DIFF
--- a/ui/src/components/RuleCard/index.jsx
+++ b/ui/src/components/RuleCard/index.jsx
@@ -152,13 +152,13 @@ function RuleCard({
   rule,
   rulesFilter,
   onRuleDelete,
+  canSignoff,
   onSignoff,
   onRevoke,
   onViewReleaseClick,
   user,
   readOnly,
   actionLoading,
-  canUserSign,
   // We don't actually use these, but we need to avoid passing them onto
   // `Card` like the rest of the props.
   onAuthorize: _,
@@ -824,14 +824,14 @@ function RuleCard({
             (user && user.email in rule.scheduledChange.signoffs ? (
               <Button
                 color="secondary"
-                disabled={!canUserSign || !user || actionLoading}
+                disabled={!user || actionLoading}
                 onClick={onRevoke}>
                 Revoke Signoff
               </Button>
             ) : (
               <Button
                 color="secondary"
-                disabled={!canUserSign || !user || actionLoading}
+                disabled={!user || actionLoading || !canSignoff}
                 onClick={onSignoff}>
                 Signoff
               </Button>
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -163,9 +163,9 @@ services:
       context: .
       dockerfile: Dockerfile.database
     expose:
-      - "3306"
+      - "3307"
```
gabrielBusta, line 166: This doesn't look related to #1145

Could you revert the changes to this file?

You can keep them on your local, but we don't need to commit/land them.

gabrielBusta:
Thanks for your contribution @CREDO23 I will test it out soon.

docker-compose.yml
CREDO23: Sure,  let me correct this!

CREDO23:
@gabrielBusta, have you tested it ?

gabrielBusta (approved):

CREDO23:
Now:
- If user's roles are `releng`, `relman`, `tb-releng` and required roles are  `1 relman`, `1 releng` then show a popup so that the user can choose a role to sign with. Here, only  `releng` or `relman` are allowed
- If user's role is `relman` and required roles are  `1 relman`, `1 releng` then `signoff` directly upon update. No need to show a popup
- If user's role is `tb-releng` and required roles are  `1 relman`, `1 releng` then disable the `signoff` button.

bhearsum:
@gabrielBusta - it looks like this is ready for another look?

gabrielBusta:
Thanks @CREDO23!

Works for me on the local instance.

AMUZY:
I actually just tested this and I like it . Great work @CREDO23 .

PR #3005, opened 2023-10-10
Update EmergencyShutOffCard

conceptacherono:
feat: Add role selection to EmergencyShutoffCard component

- Added role selection dropdown to allow users to select their role when signing off.
- Modified the onSignoff function to accept the selected role.
- Integrated role selection into the component's UI.


bhearsum (changes requested):
I'm so sorry for the long delay here - there's been so many PRs that we lost track of this one. The change looks promising, but unfortunately it seems to cause the UI not to load in my browser. There's a number of exceptions in the Console, but the primary one seems to be:
```
Uncaught ReferenceError: PropTypes is not defined
```

conceptacherono:
okay,let me look at it

Issue #3007, opened 2023-10-10
Highlight matching search terms in the search results in the Relases page


gabrielBusta:
Now that we've added partial matching for search terms in the releases page on #1146 it would be great if the matching parts of the search query are highlighted in the results. This would give visual feedback about what part of the search query matched the release.

FatumaA:
I'll be attempting this

nazarenegena:
@FatumaA  @gabrielBusta 
i'm already working on this as a follow up on #1146 issue

PR #3009, opened 2023-10-12
implemented search highlight feature on releases

nazarenegena:
This addresses issue #3007 which is a follow up on issue #1146

### DIFF
--- a/ui/src/components/ReleaseCard/index.jsx
+++ b/ui/src/components/ReleaseCard/index.jsx
@@ -27,6 +27,7 @@ import SignoffSummary from '../SignoffSummary';
 import { withUser } from '../../utils/AuthContext';
 import Link from '../../utils/Link';
 import { release } from '../../utils/prop-types';
+import {  highlightText } from '../../utils/highlightText';
 
 const useStyles = makeStyles(theme => ({
   root: {
@@ -132,6 +133,7 @@ function ReleaseCard(props) {
     onAccessChange,
     onReleaseDelete,
     onViewScheduledChangeDiff,
+    searchValue,
     // We don't actually use these, but we need to avoid passing them onto
     // `Card` like the rest of the props.
     onAuthorize: _,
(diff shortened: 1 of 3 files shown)


gabrielBusta:
Great job @nazarenegena! This is awesome.

I tried it out locally and I ran into an issue by typing "Fi 11 s" into the search bar. The UI seems to be displaying this bit of the code:

https://github.com/mozilla-releng/balrog/blob/cd63ec5354367008e4dd04a95ace2e24282cd868/ui/src/utils/highlightText.js#L16

Screenshot 2023-10-17 at 16 24 51

gabrielBusta (changes requested):
The issue seems to be stemming from the way the `highlightText` function is highlighting the search terms.

1. The `highlightText` function takes the `text` (which in this case might be something like "Firefox-68.11.0esr-build1") and a `searchTerm` (like "Fi 11 s").
   
2. The search term is broken down into individual terms based on spaces, so "Fi 11 s" becomes `["Fi", "11", "s"]`.

3. For each term, the function attempts to replace occurrences in the `text` with a highlighted version using a regex.

Since the replacement is done for each term in the `searchTerms` array sequentially, the replacement for later terms can end up replacing parts of the HTML injected by earlier replacements.

To illustrate the problem using my specific example:
1. The term "Fi" would replace part of "Firefox" with `<span style="background-color: #804cc5; color: white;">Fi</span>refox-68.11.0esr-build1`.
2. Then, the term "11" would further modify the result to become `<span style="background-color: #804cc5; color: white;">Fi</span>refox-68.<span style="background-color: #804cc5; color: white;">11</span>.0esr-build1`.
3. Lastly, the term "s" would try to replace all occurrences of "s", including those in the previously injected `<span>` tags, leading to the result I observed.

nazarenegena:
thank you for the feedback 
let me work on a solution and update.

AMUZY:
Nice work @nazarenegena

bhearsum:
@nazarenegena - is this ready for another look? If so, please use the "Re-request review button" in the Reviewers list.

Issue #2942, opened 2023-08-04
Balrog accepts scheduled rule changes with duplicate alias'


bhearsum:
[Alias' must be unique](https://github.com/mozilla-releng/balrog/blob/036d4a95110dfdf8f3e4bd812f5ac5c6351d3912/src/auslib/db.py#L1742), yet the scheduled rule change endpoints accept the idea of a new rule being scheduled with an alias that already exists. I highly doubt these rules would successfully be enacted, but regardless we should not accept their scheduling in the first place.

![image](https://github.com/mozilla-releng/balrog/assets/49649/dc54ef5c-2ba8-4587-88e5-492d506a3118)

Stanzin7:
Hello @gabrielBusta @bhearsum, I am in the contribution phase for Outreachy. Can you please assign this task to me. Thanks )

godplayer56:
Hey @bhearsum @gabrielBusta , I have some doubts regarding this issue -
If a new rule is being scheduled, no other aliases can schedule the same rule
Am I right about this?

godplayer56:
Hey @bhearsum @gabrielBusta , I have some doubts regarding this issue -
If a new rule is being scheduled, no other aliases can schedule the same rule
Am I right about this?

bhearsum:
> Hey @bhearsum @gabrielBusta , I have some doubts regarding this issue - If a new rule is being scheduled, no other aliases can schedule the same rule Am I right about this?

Yes - only one rule can exist with the same alias at any given time. This works correctly for the `rules` table and endpoints, which reject duplicate alias'. The scheduled rules table does _not_ have any constraints about unique alias' though (this is expected). We need some safeguards at the API level (https://github.com/mozilla-releng/balrog/blob/main/src/auslib/web/admin/views/rules.py#L188) to implement this.

godplayer56:
Yes @bhearsum I think I got the fix. I am working on it

bhearsum:
This is fixed by #3010. @michellemounde - I'm happy to help you work through your own fix for it as well, if you like, but if you'd prefer to focus on something else that's not yet fixed that's fine too.

jcristau:
Reopening because #3010 is being reverted in #3048.

godplayer56:
@jcristau Should I do some changes to my commit so as to fix this issue? or should I approach it differently?

bhearsum:
[...]

PR #3010, opened 2023-10-12
a fix for duplicate alias problem

godplayer56:
This pull request tries to fix the duplicate-alias issue #2942


godplayer56:
@bhearsum @gabrielBusta I am seeing that one of the tests failed for the PR 😵‍💫. 

I had used the same function which has been used for rules. How should I change the function for the test to pass?

bhearsum:
> @bhearsum @gabrielBusta I am seeing that one of the tests failed for the PR 😵‍💫.
> 
> I had used the same function which has been used for rules. How should I change the function for the test to pass?

I can see that the failure is a bit confusing -- this is actually [a failure in the Python file formatting](https://firefox-ci-tc.services.mozilla.com/tasks/dHLmTCLRTfmAPMIsnfqdOQ/runs/0/logs/live/public/logs/live.log#L44) - not in any test.

> [task 2023-10-12T21:23:33.129Z] ERROR: InvocationError for command /builds/worker/checkouts/vcs/.tox/check/bin/black --diff --check scripts src taskcluster tests uwsgi (exited with code 1)

In addition to fixing that, it would be good to add tests for this in https://github.com/mozilla-releng/balrog/blob/main/tests/admin/views/test_rules.py (we already have a number of tests for these endpoints that should be a good model for new ones).

godplayer56:
> > @bhearsum @gabrielBusta I am seeing that one of the tests failed for the PR 😵‍💫.
> > I had used the same function which has been used for rules. How should I change the function for the test to pass?
> 
> I can see that the failure is a bit confusing -- this is actually [a failure in the Python file formatting](https://firefox-ci-tc.services.mozilla.com/tasks/dHLmTCLRTfmAPMIsnfqdOQ/runs/0/logs/live/public/logs/live.log#L44) - not in any test.
> 
> > [task 2023-10-12T21:23:33.129Z] ERROR: InvocationError for command /builds/worker/checkouts/vcs/.tox/check/bin/black --diff --check scripts src taskcluster tests uwsgi (exited with code 1)
> 
> In addition to fixing that, it would be good to add tests for this in https://github.com/mozilla-releng/balrog/blob/main/tests/admin/views/test_rules.py (we already have a number of tests for these endpoints that should be a good model for new ones).

@bhearsum How should I approach fixing the python formatting issue? How should I check if the files have formatting issues?

I am thinking of adding a test similar to the test for checking alias in rules class

bhearsum:
> > > @bhearsum @gabrielBusta I am seeing that one of the tests failed for the PR 😵‍💫.
> > > I had used the same function which has been used for rules. How should I change the function for the test to pass?
> > 
> > 
> > I can see that the failure is a bit confusing -- this is actually [a failure in the Python file formatting](https://firefox-ci-tc.services.mozilla.com/tasks/dHLmTCLRTfmAPMIsnfqdOQ/runs/0/logs/live/public/logs/live.log#L44) - not in any test.
> > > [task 2023-10-12T21:23:33.129Z] ERROR: InvocationError for command /builds/worker/checkouts/vcs/.tox/check/bin/black --diff --check scripts src taskcluster tests uwsgi (exited with code 1)
> > 
> > 
> > In addition to fixing that, it would be good to add tests for this in https://github.com/mozilla-releng/balrog/blob/main/tests/admin/views/test_rules.py (we already have a number of tests for these endpoints that should be a good model for new ones).
> 
> @bhearsum How should I approach fixing the python formatting issue? How should I check if the files have formatting issues?
> 
> I am thinking of adding a test similar to the test for checking alias in rules class

You can use [black](https://github.com/psf/black) to format the Python files correctly.

godplayer56:
I have formatted the code but some tests still fail

bhearsum:
> I have formatted the code but some tests still fail

Please have a look through [the linked log](https://firefox-ci-tc.services.mozilla.com/tasks/H3zS36u3S9emzwE6wjL8Jg/runs/0/logs/live/public/logs/live.log) for any errors that are still showing up.

godplayer56:
> > I have formatted the code but some tests still fail
> 
> Please have a look through [the linked log](https://firefox-ci-tc.services.mozilla.com/tasks/H3zS36u3S9emzwE6wjL8Jg/runs/0/logs/live/public/logs/live.log) for any errors that are still showing up.

I checked for errors but only warnings are being shown. I have formatted all the files using black, so it can be some other issue which I am not able to figure out from the logs :cry:

gbrownmozilla:
Sorry, this shouldn't be so hard!

The log shown by github is incomplete; you can see the full test log if you "View logs in Taskcluster". In the full log, I see
```
[task 2023-10-16T10:43:40.522Z] ERROR: InvocationError for command /builds/worker/checkouts/vcs/.tox/check/bin/check-manifest -v . (exited with code 1)
```
...which doesn't tell you much! I wonder if that's related to the new file you are adding? I would suggest not including `settings.json` in the PR.

godplayer56:
Hii @bhearsum @gbrownmozilla , the checks have passed, can you please review the PR? 

Should I do some changes to it?

tests/admin/views/test_rules.py
```diff
@@ -2834,3 +2834,14 @@ def testCannotRevokeOtherUsersSignoffAsNormalUser(self):
         ret = self._delete("/scheduled_changes/rules/3/signoffs", username="julie", qs={"username": "mary"})
         self.assertEqual(ret.status_code, 403, ret.get_data())
         self.assertEqual(ret.mimetype, "application/json")
+
+    def testScheduledChangeWithDuplicateAlias(self):
```
bhearsum, line 2838: This test is not quite right, unfortunately. It is testing the `/rules` view - which is already covered by https://github.com/mozilla-releng/balrog/blob/6493a9e2eae9e959fc69e7e2fd93b6e91ed16ba3/tests/admin/views/test_rules.py#L540.

You should be adding at least two tests:
- One that validates a POST to `/scheduled_changes/rules`, and one that validates a POST to `/scheduled_changes/rules/:sc_id`. The first could be modeled after https://github.com/mozilla-releng/balrog/blob/6493a9e2eae9e959fc69e7e2fd93b6e91ed16ba3/tests/admin/views/test_rules.py#L2161, the second could be modeled after https://github.com/mozilla-releng/balrog/blob/6493a9e2eae9e959fc69e7e2fd93b6e91ed16ba3/tests/admin/views/test_rules.py#L2251

In both cases the request should a result in a 400 if any of the following are true:
- The alias exists in any row in the `rules` table
- The alias exists in any row in the `rule_scheduled_changes` table that meets both the following conditions:
  - `complete` is False
  - `change_type` is `insert` or `update`

(This may require additional changes to `rules.py` as well - please let me know if you have any questions.)

bhearsum (changes requested):

tests/admin/views/test_rules.py
godplayer56: I'll start working on updating the test :+1:

godplayer56: @bhearsum I have a doubt regarding this - 
How do I select the required data in the rule_scheduled_change table? for eg-
for the rules table we have the `getAlias` function defined to get the rule with the given alias but for rule_scheduled_change I am not able to figure out how should I select the data with a given alias.

bhearsum: > @bhearsum I have a doubt regarding this - How do I select the required data in the `rule_scheduled_change` table? for eg- for the `rules` table we have the `getAlias` function defined to get the rule with the given alias but for `rule_scheduled_changes` I am not able to figure out how should I select the data with a given alias.

The `scheduled_changes` table is accessible through `dbo.rules.scheduled_changes` (You can see an example query in the tests https://github.com/mozilla-releng/balrog/blob/6d4620760ca06dc13d12c57ae7eccf6d3233705f/tests/admin/views/test_rules.py#L2012 - and there's many more example of selects in other tests and parts of the codebase).

> and also does the column `scheduled_by` denote the alias for the `rule_scheduled_changes` table?

No -- `scheduled_by` is the person who scheduled the change :). The alias is in the `base_alias` column in the scheduled changes table (apologies -- I forgot about that detail when I posted the earlier comment).

godplayer56: Ok got it! :smile:

godplayer56:
@bhearsum @gbrownmozilla I have added the tests for the code. Can you please have a look and tell me if it needs some changes?

bhearsum:
> @bhearsum @gbrownmozilla I have added the tests for the code. Can you please have a look and tell me if it needs some changes?

It looks like you're making progress. The tests that you added looks fine to me, but you need to cover some additional cases still:
* A POST to `/scheduled_changes/rules` with `update_type` set to `update` with a duplicate alias
* A POST to `/scheduled_changes/rules/:sc_id` with `update_type` set to `insert` with a duplicate alias

godplayer56:
Ok got it, so I'd be adding two more tests for it then :+1:

godplayer56:
@bhearsum are the changes fine?

src/auslib/web/admin/views/rules.py
```diff
@@ -260,6 +260,15 @@ def _post(self, transaction, changed_by):
             if what.get("fallbackMapping") is not None and len(fallback_mapping_values) != 1:
                 return problem(400, "Bad Request", "Invalid fallbackMapping value. No release name found in DB")
 
+            alias = what.get("alias", None)
+            if alias is not None and dbo.rules.getRule(alias):
+                return problem(400, "Bad Request", "Rule with alias exists.")
+
+            if alias is not None:
+                scheduled_rule_with_alias = dbo.rules.scheduled_changes.t.select().where(dbo.rules.scheduled_changes.base_alias == alias).execute().fetchall()
```
bhearsum, line 268: This (and the similar block below) are missing the conditions I called out in https://github.com/mozilla-releng/balrog/pull/3010#discussion_r1362959833:
> * complete is False
> * change_type is insert or update

tests/admin/views/test_rules.py
```diff
@@ -2238,6 +2238,124 @@ def testAddScheduledChangeMultipleConditions(self):
...
+            "when": 1234567,
+            "priority": 120,
+            "backgroundRate": 100,
+            "product": "blah",
+            "channel": "blah",
+            "mapping": "a",
+            "change_type": "insert",
+            "base_alias": "test",
+        }
+        ret1 = self._post("/scheduled_changes/rules", data=data1)
```
bhearsum, line 2258: This is returning a 400 even if I comment out the previous post:
> b'{"type": "about:blank", "title": "Bad Request", "detail": "Null/Empty Value", "status": 400, "exception": "update_type cannot be set to null/empty when scheduling insertion of a new rule"}'

This suggests that this test is not failing for the right reason. Please add an additional assert about the `exception` message from `ret1.get_data()` to make sure this fails for the right reason, and fix the test to do so. (Please do this for the other tests as well -- they're probably hitting the same issue.

bhearsum (changes requested):

tests/admin/views/test_rules.py
godplayer56: Ok got it must be I'll update the tests

bhearsum:
@godplayer56 - is this ready for another look? If so, please use the "Re-request review button" in the Reviewers list.

godplayer56:
> @godplayer56 - is this ready for another look? If so, please use the "Re-request review button" in the Reviewers list.

I am trying to resolve the post issues right now. I'll use it as soon as I am done resolving the issue :+1:

godplayer56:
@bhearsum I have added assertions that confirm that the tests are failing for the right reason :+1:

src/auslib/web/admin/views/rules.py
```diff
@@ -260,6 +260,24 @@ def _post(self, transaction, changed_by):
             if what.get("fallbackMapping") is not None and len(fallback_mapping_values) != 1:
                 return problem(400, "Bad Request", "Invalid fallbackMapping value. No release name found in DB")
 
+            alias = what.get("base_alias", None)
```
bhearsum, line 263: I think the problem is that in _this_ part of the view this needs to be referred to as `alias` (not `base_alias`). I suggest you try printing out `what` and making a web request to confirm this.

tests/admin/views/test_rules.py
```diff
@@ -2238,6 +2238,146 @@ def testAddScheduledChangeMultipleConditions(self):
...
+            "data_version": 1,
+            "rule_id": 1,
+            "priority": 100,
+            "version": "3.5",
+            "buildTarget": "d",
+            "backgroundRate": 100,
+            "mapping": "c",
+            "update_type": "minor",
+            "sc_data_version": 1,
+            "base_alias": "TestDuplicateAlias3",
```
bhearsum, line 2332: Also try simulating a request like this on https://localhost:9000/rules/364 by adding an alias and clicking the Save button - take a dlook at what the the Request JSON looks like in the Developer Tools. If it doesn't use `base` neither should you in these tests.

bhearsum (changes requested):
It looks like something is not quite complete or right here. In my local database I already have an alias of `firefox-release`, but I can still scheduled a change to add one on https://localhost:9000/rules/364, which is accepted.

tests/admin/views/test_rules.py
godplayer56: I have updated all the files to use `alias` instead of `base_alias`.

bhearsum: You should be fully authorized in your local development environment...

In any case, I can assure you that the admin API doesn't use `base` in these requests.

bhearsum (approved):
This looks good now - thank you!

PR #3013, opened 2023-10-13
Add support for cleaning up data from releases_json/release_assets

bhearsum:
This was never implemented when we added these tables. Thankfully, the schemas are identical enough that the join/conditions we're using to cleanup `releases` works just as well here.

Even with this patch, nothing will happen immediately because I discovered that we nothing that runs the cleanup in cron since we migrated Balrog to GCP. My plan is:
* Land this
* Do some extra verification with a dry run in production
* Run the cleanup by hand in production
* Set-up a cronjob to run it on a regular basis

### DIFF
--- a/scripts/manage-db.py
+++ b/scripts/manage-db.py
@@ -19,21 +19,21 @@ sys.path.append(path.join(path.dirname(__file__), ".."))
 sys.path.append(path.join(path.dirname(__file__), path.join("..", "vendor", "lib", "python")))
 
 
+RELEASES_CLEANUP_CONDITION = """
+LEFT JOIN rules rules_mapping ON (name=rules_mapping.mapping)
+WHERE name LIKE '%%nightly%%'
+AND name NOT LIKE '%%latest'
+AND rules_mapping.mapping IS NULL
+AND (STR_TO_DATE(RIGHT(name, 14), "%%Y%%m%%d%%H%%i%%S") < NOW() - INTERVAL {nightly_age} DAY);
+"""
+
+
 def cleanup_releases(trans, nightly_age, dryrun=True):
     # This and the subsequent queries use "%%%%%" because we end up going
     # through two levels of Python string formatting. The first is here,
     # and the second happens at a low level of SQLAlchemy when the transaction
     # is being executed.
-    query = (
-        """
-LEFT JOIN rules rules_mapping ON (name=rules_mapping.mapping)
-WHERE name LIKE '%%%%nightly%%%%'
-AND name NOT LIKE '%%%%latest'
-AND rules_mapping.mapping IS NULL
-AND (STR_TO_DATE(RIGHT(name, 14), "%%%%Y%%%%m%%%%d%%%%H%%%%i%%%%S") < NOW() - INTERVAL %s DAY);
-"""
-        % nightly_age
-    )
+    query = RELEASES_CLEANUP_CONDITION.format(nightly_age=nightly_age)
     if dryrun:
         todelete = trans.execute("SELECT name FROM releases" + query).fetchall()
         print("Releases rows to be deleted:")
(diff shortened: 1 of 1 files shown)


scripts/manage-db.py
```diff
@@ -45,6 +45,20 @@ def cleanup_releases(trans, nightly_age, dryrun=True):
...
+    if dryrun:
+        todelete = trans.execute("SELECT name FROM releases_json" + query).fetchall()
+        print("Releases JSON rows to be deleted:")
+        if todelete:
+            print("\n".join(itertools.chain(*todelete)))
+        else:
+            print("  - None")
+    else:
+        trans.execute("DELETE releases_json FROM releases_json" + query)
+        trans.execute("DELETE release_assets FROM release_assets" + query)
```
gabrielBusta, line 59: Does`DELETE FROM table_name WHERE condition` work? Instead of `DELETE table_name FROM table_name WHERE condition`

scripts/manage-db.py
```diff
@@ -45,6 +45,20 @@ def cleanup_releases(trans, nightly_age, dryrun=True):
         trans.execute("DELETE releases FROM releases" + query)
 
 
+def cleanup_releases_json(trans, nightly_age, dryrun=True):
+    query = RELEASES_CLEANUP_CONDITION.format(nightly_age=nightly_age)
```
gabrielBusta, line 49: suggestion (non-blocking)

We could use `text` and let sqlalchemy handle the parameter substitution

```diff
diff --git a/scripts/manage-db.py b/scripts/manage-db.py
index 4306c6c6..10528d52 100755
--- a/scripts/manage-db.py
+++ b/scripts/manage-db.py
@@ -7,6 +7,7 @@ from os import path
 from subprocess import run
 
 from sqlalchemy.engine.url import make_url
+from sqlalchemy import text
 
 from auslib.blobs.base import createBlob  # noqa: E402
 from auslib.db import AUSDatabase  # noqa: E402
@@ -24,7 +25,7 @@ LEFT JOIN rules rules_mapping ON (name=rules_mapping.mapping)
 WHERE name LIKE '%%nightly%%'
 AND name NOT LIKE '%%latest'
 AND rules_mapping.mapping IS NULL
-AND (STR_TO_DATE(RIGHT(name, 14), "%%Y%%m%%d%%H%%i%%S") < NOW() - INTERVAL {nightly_age} DAY);
+AND (STR_TO_DATE(RIGHT(name, 14), "%%Y%%m%%d%%H%%i%%S") < NOW() - INTERVAL :nightly_age DAY);
 """
 
 
@@ -33,16 +34,17 @@ def cleanup_releases(trans, nightly_age, dryrun=True):
     # through two levels of Python string formatting. The first is here,
     # and the second happens at a low level of SQLAlchemy when the transaction
     # is being executed.
-    query = RELEASES_CLEANUP_CONDITION.format(nightly_age=nightly_age)
     if dryrun:
-        todelete = trans.execute("SELECT name FROM releases" + query).fetchall()
+        query = text("SELECT name FROM releases" + RELEASES_CLEANUP_CONDITION)
+        todelete = trans.execute(query, nightly_age=nightly_age).fetchall()
         print("Releases rows to be deleted:")
         if todelete:
             print("\n".join(itertools.chain(*todelete)))
         else:
             print("  - None")
     else:
-        trans.execute("DELETE releases FROM releases" + query)
+        query = text("DELETE releases FROM releases" + RELEASES_CLEANUP_CONDITION)
+        trans.execute(query, nightly_age=nightly_age)
 
 
 def cleanup_releases_json(trans, nightly_age, dryrun=True):
diff --git a/src/auslib/db.py b/src/auslib/db.py
index d2c04767..a696899f 100644
--- a/src/auslib/db.py
+++ b/src/auslib/db.py
@@ -211,10 +211,10 @@ class AUSTransaction(object):
         if not self.conn.closed:
             self.conn.close()
 
-    def execute(self, statement):
+    def execute(self, statement, **parameters):
         try:
             self.log.debug("Attempting to execute %s" % statement)
-            return self.conn.execute(statement)
+            return self.conn.execute(statement, **parameters)
         except Exception as exc:
             self.log.debug("Caught exception")
             # We want to raise our own Exception, so that errors are easily
```

bhearsum: I think this is a good idea, but I'd prefer not to make a change to the database layer as part of this to avoid adding risk. I filed #3015 for this.

scripts/manage-db.py
jcristau: Looks like it's either `DELETE tbl_name FROM tbl_name LEFT JOIN ...` or `DELETE FROM tbl_name USING tbl_name LEFT JOIN ...`, when multiple tables are involved.
(https://dev.mysql.com/doc/refman/5.7/en/delete.html#idm46005468799856)

jcristau (approved):

scripts/manage-db.py
gabrielBusta: Oh ok

scripts/manage-db.py
gabrielBusta: Makes sense

gabrielBusta (approved):

Issue #2976, opened 2023-09-05
Update the balrog frontend to use Mozilla Protocol


gabrielBusta:
The [shipit](https://github.com/mozilla-releng/shipit) frontend was updated to use Mozilla's protocol design system with https://github.com/mozilla-releng/shipit/pull/1339

We could do the same to the [balrog](https://github.com/mozilla-releng/balrog/) ui.

nzaih1999:
Hello can you please assign me this issue, I wanna work on it

nzaih1999:
Thank you I will start working on this asap

godplayer56:
Hey @gabrielBusta , since I am done with the docker compose issue, should I work on this issue or do you want me to work on some other issue?

nzaih1999:
Hello @godplayer56 I'm working on this I'll raise the WIP PR tomorrow

gabrielBusta:
@godplayer56 feel free work on this issue. We are letting applicants work on whatever issues are available, and not worry about being assigned to things (in part because there's more contributors than bugs.) The experience is valuable to everyone even if your particular PR is not the one that ends up merged.

gabrielBusta:
@godplayer56 feel free work on this issue. We are letting applicants work on whatever issues are available, and not worry about being assigned to things (in part because there's more contributors than issues.) The experience is valuable to everyone (even if an applicant's pull-request is not the one that ends up merged.)

gabrielBusta:
@godplayer56 we are letting applicants work on whatever issues are available, and not worry about being assigned to things (in part because there's more contributors than issues.) The experience is valuable to everyone (even if an applicant's pull-request is not the one that ends up merged.)

MelakuAlehegn:
Hello, I'm an outreachy applicant and I want to work one this issue. Can I be assigned?

godplayer56:
> @godplayer56 we are letting applicants work on whatever issues are available, and not worry about being assigned to things (in part because there's more contributors than issues.) The experience is valuable to everyone (even if an applicant's pull-request is not the one that ends up merged.)

Sure @gabrielBusta I'll start working on this issue👍

PR #3014, opened 2023-10-14
Mozilla UI balrog

godplayer56:
I have updated the dashboard file by introducing the mozilla logo and have also updated the typography to use the mozilla-protocol

This PR is for the issue #2976 

I request a review and suggestions for this PR @gabrielBusta @gbrownmozilla .

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -24,6 +24,7 @@
     "@material-ui/pickers": "^3.3.10",
     "@material-ui/styles": "^4.11.2",
     "@mozilla-frontend-infra/components": "^3.0.7",
+    "@mozilla-protocol/core": "^17.0.1",
     "axios": "^0.24.0",
     "change-case": "^4.1.2",
     "classnames": "^2.2.6",
--- a/ui/src/theme.js
+++ b/ui/src/theme.js
@@ -1,5 +1,6 @@
 import { createTheme } from '@material-ui/core/styles';
 import { amber, red, indigo } from '@material-ui/core/colors';
+import '../node_modules/@mozilla-protocol/core/protocol/css/protocol-components.css';
 
 const SPACING = {
   UNIT: 8,
@@ -35,7 +36,25 @@ export default createTheme({
     },
   },
   typography: {
-    useNextVariants: true,
+    fontDisplay: 'swap',
+    fontFamily: "'Zilla Slab', 'Inter', sans-serif",
+    fontStyle: 'normal',
+    fontWeight: 'normal',
+    body1: {
+      fontFamily: "'Inter', sans-serif",
+    },
+    body2: {
+      fontFamily: "'Inter', sans-serif",
+    },
+    caption: {
+      fontFamily: "'Inter', sans-serif",
+    },
+    button: {
+      fontFamily: "'Inter', sans-serif",
+    },
+    overline: {
+      fontFamily: "'Inter', sans-serif",
+    },
   },
   mixins: {
     link: {
--- a/ui/src/Main.jsx
+++ b/ui/src/Main.jsx
@@ -48,7 +48,7 @@ function Main() {
 
   return backendError ? (
     <BrowserRouter>
-      <Dashboard title="Error" disabled>
+      <Dashboard disabled>
         <ErrorPanel fixed error={backendError} />
       </Dashboard>
     </BrowserRouter>
--- a/ui/src/components/Dashboard/index.jsx
+++ b/ui/src/components/Dashboard/index.jsx
@@ -5,6 +5,8 @@ import { makeStyles } from '@material-ui/styles';
 import AppBar from '@material-ui/core/AppBar';
 import Toolbar from '@material-ui/core/Toolbar';
 import Typography from '@material-ui/core/Typography';
+import Box from '@material-ui/core/Box';
+import ExtensionIcon from '@material-ui/icons/Extension';
 import menuItems from './menuItems';
 import Link from '../../utils/Link';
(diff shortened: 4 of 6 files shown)


godplayer56:
Hey @bhearsum @gabrielBusta . I was trying to fix the ui-build error which got resolved after I removed the group parameter but I am not able to fix the error with ui-tests. 

It shows the same error when I try to run `docker exec balrog-balrogui-1 yarn lint --fix` for fixing lint errors, but it just displays the errors without correcting them. 

How should I resolve these errors?

ui/package.json
```diff
@@ -4,7 +4,7 @@
   "main": "index.js",
   "license": "MIT",
   "engines": {
-    "node": ">=16.0.0 <17.0.0"
```
gabrielBusta, line 7: We need this to stay on CI - or we should deliberately bump the node version

ui/src/theme.js
```diff
@@ -35,7 +36,25 @@ export default createTheme({
     },
   },
   typography: {
```
gabrielBusta, line 38: Can we update the them to use the protocol pallet?

https://github.com/mozilla-releng/shipit/pull/1339/files#diff-f0f2b454dd5fb88af32df6fdc3f078409c63663d7251463a2a3c7304ef56be81R35-R46

https://protocol.mozilla.org/docs/fundamentals/color.html

ui/.eslintrc.js
```diff
@@ -1,3 +1,3 @@
 const neutrino = require('neutrino');
 
-module.exports = neutrino().eslintrc();
+module.exports = neutrino().eslintrc();
```
gabrielBusta, line 3: Could you revert the changes to this file?

gabrielBusta (changes requested):
Updating the color palette and removing `balrog.svg` from the background will go a long way in making balrog look in line with Protocol.

ui/package.json
godplayer56: I'll update this :+1:

ui/.eslintrc.js
godplayer56: I changed the file accidently in a commit :sweat_smile: , I'll revert the changes

godplayer56:
@gabrielBusta I have removed the balrog.svg file from the background and updated the pallette, do we need anymore changes?

godplayer56:
Hey @gabrielBusta just a follow up on this. 😅

ui/src/theme.js
```diff
@@ -11,31 +11,36 @@ const SPACING = {
...
-      dark: '#160065',
+      main: '#000',
     },
     secondary: {
-      light: '#678dff',
-      main: '#0061f2',
-      dark: '#0038be',
+      main: '#c50042',
     },
-    error: red,
```
gabrielBusta, line 23: Should these be updated? What changes when you delete these?

gabrielBusta (approved):
It looks good to me @godplayer56

Just one more question before I land it. I am not sure what deleting some of the pallete styles does. Should we update those to protocol colors or leave them as is?

gabrielBusta (changes requested):
I can't see the `Logo` or the pages' `{title}`. The `Logo` component should be modified for use here. `balrog` doesn't have a `group` concept like `shipit`. You may be able to display different protocol logos based on some value in the `Dashboard` component's `props`. Or you could simplify the `Logo` component to always show the Mozilla "m".

godplayer56:
> I can't see the `Logo` or the pages' `{title}`. The `Logo` component should be modified for use here. `balrog` doesn't have a `group` concept like `shipit`. You may be able to display different protocol logos based on some value in the `Dashboard` component's `props`. Or you could simplify the `Logo` component to always show the Mozilla "m".

Ok I'll update the logo component :+1:

ui/src/theme.js
godplayer56: When I delete them the theme changes to the original theme it was. for example the dashboard becomes blue when I delete the palette

godplayer56:
> It looks good to me @godplayer56
> 
> Just one more question before I land it. I am not sure what deleting some of the pallete styles does. Should we update those to protocol colors or leave them as is?

deleting them reverts the theme to what it was before, I guess it would be better if we update those to protocol colors.

What colors should we use as protocol colors?

godplayer56:
Screenshot from 2023-10-25 01-44-39
@gabrielBusta This is what I see when I start the docker containers. When I inspect the webpage I see that the logo is present but there is some issue with its rendering. Can you provide me with suggestions on what I should do to resolve this?(I have updated the logo component by removing the groups concept)

ui/src/theme.js
gabrielBusta: Oh OK I see. Can we pick similar colors from the protocol pallette?

godplayer56: Yes we do have similar colours in the protocol pallette, we can pick them if they look more lively. Should I update the pallette to include those colours?

godplayer56:
@gabrielBusta I have updated the code to include the mozilla family logo as the simple "m" logo wasn't rendering maybe because of the pallette colours we have chosen :sweat_smile: . Also should I update the pallette and to what colours?

Leo-ainy:
Enhancing Balrog's visual alignment with the Mozilla Protocol involves a significant improvement by updating the color palette and eliminating the background's `balrog.svg`. This step will contribute to a more cohesive and harmonious appearance.
https://github.com/mozilla-releng[!](https://cheapmotorcycleshipping.com/)

PR #3020, opened 2023-10-18
Fixed-error-in-console-about-duplicate-component-keys

AMUZY:
Just fixed the following error of duplicate keys among the lists of the AutoCompleteText component :
encountered_two_children


bhearsum (approved):
The fact that this error exists indicates a problem with data in the system, but nonetheless I don't see any reason we shouldn't use different keys.

PR #3021, opened 2023-10-18
Dont prompt for signoff with irrelevant roles

AMUZY:
So this is my version of a fix for the issue [don't prompt for signoff with roles that aren't relevant #1145](https://github.com/mozilla-releng/balrog/issues/1145) : 
**NOTE : In all situations, the user has three roles (relman, releng and tb-releng)**
- Where required roles are three : 
vZGRyrdfMcCueYhppxvNTUie
- Where required roles are two, user is prompted only those two roles :
KbgPpGRBLemUUBjrNVWCREQs
- Where the required role is one, user is prompted that exact role only :
dbQOJcRBJczdlnkRnUjYOMoz
- and If the user has none of the required roles, the `Signoff` button is diabled for them : 
HzdvDAiGJSqPwPOjjJKAfhmB
- and of course, a user can't sign with two roles because they would have to revoke a signoff to be able to make another signoff

### DIFF
--- a/ui/src/components/RuleCard/index.jsx
+++ b/ui/src/components/RuleCard/index.jsx
@@ -158,6 +158,7 @@ function RuleCard({
   user,
   readOnly,
   actionLoading,
+  canUserSign,
   // We don't actually use these, but we need to avoid passing them onto
   // `Card` like the rest of the props.
   onAuthorize: _,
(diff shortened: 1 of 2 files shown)


bhearsum (approved):
This looks good to me - thank you! I like the additional feature of disabling the "Signoff" button when the user has no relevant roles.

Issue #2994, opened 2023-10-05
can't create a release in the UI


jcristau:
I'm not sure if I'm doing something wrong here, but when trying to create a release I go to e.g. https://balrog.services.mozilla.com/releases/Widevine-4.10.2710.0/v2, then paste or upload the blob, but the form won't let me enter the product name, which means form submission fails.  Going into devtools to replay the request having added the product to the request body does work.

ademolaomosanya:
i'm an outreachy applicant 
and i would like to work on this issue kindly asign me

Raven49:
Kindly assign this task to me

bhearsum:
@jcristau - can you still reproduce this? I wasn't able to with the latest `main`.

jcristau:
It's possible I was on an Update page, I don't know the difference :)

jcristau:
Thanks, OK, so I guess the bug from my point of view is that `/releases/doesnotexist/v2` should give me the "new release" page and pre-fill the name, instead of giving me an "update a release that does not exist and that you can't actually update" page.  Does that sound reasonable?

bhearsum:
> Thanks, OK, so I guess the bug from my point of view is that `/releases/doesnotexist/v2` should give me the "new release" page and pre-fill the name, instead of giving me an "update a release that does not exist and that you can't actually update" page. Does that sound reasonable?

Yeah, I think that makes sense. I guess the thing that surprises me is how you're getting to a page like that though? I guess it's a manually entered URL, as I don't see how you could get there via the UI.

jcristau:
Correct. Partly because usually the new release I want to create is a variation on an existing one, but there's no "duplicate" button for releases like there is for rules. Partly because the "new release" button is on the /releases page which itself takes forever to load.

bhearsum:
I've updated the summary to match what we've discussed. It looks like the fix here is going to involve a couple of things:
* Stop setting `isNewRelease` as part of the route/input to the Release edit/create page: https://github.com/mozilla-releng/balrog/blob/ca55f9aa6aa15647370a841d8827e1e724ab8813/ui/src/views/Releases/routes.js#L27
* Adjust https://github.com/mozilla-releng/balrog/blob/ca55f9aa6aa15647370a841d8827e1e724ab8813/ui/src/views/Releases/ReleaseV2/index.jsx#L62 to stop taking this as an input, and determine for itself whether or not the release is new (presumably by querying the backend and seeing if it's a 404 or not).

PR #3022, opened 2023-10-19
allow update of product name when updating a release

nazarenegena:
fixes issue #2994 
Allows changing of product name in the form when updating a release.


bhearsum:
I'm not sure this is actually still an issue...I wasn't able to reproduce #2994. In any case, removing this line is not what we want...as it would allow existing releases to be renamed, which we explicitly do not allow.

bhearsum:
Since we don't seem to know exactly what we want to do yet, I'm going to close this while we sort it out in #2994.

Issue #1138, opened 2020-01-10
on the releases page, use aliases for rule backlinks if the rule has an alias


lundjordan:
this would be handy to quickly know if a rule that is mapped to a release is something permanent or official: 

ie `release` or `beta-cdntest` rather than some arbitrary #.

DokuaAsiedu:
Hello, mind if I try to work on this?

PR #3023, opened 2023-10-19
show release alias on rules page

nazarenegena:
This PR attempts to fix the issue #1138.
I have added the alias for the releases on the rules page. When you hover on the specific rule you can see it's alias.
however, since most of the releases' alias is 'null' I have left the '#' in its place. 
Is this the expected solution?


bhearsum (changes requested):
Unfortunately, this does not address the issue. What we'd like is for a change on the Releases page to show rule alias' instead of rule ids. For example, in this screenshot:
image

The `56` should be `firefox-release-localtest`, `145` should be `firefox-release`, and `585` should be `firefox-release-cdntest`.

This will require changes to https://github.com/mozilla-releng/balrog/blob/main/ui/src/components/ReleaseCard/index.jsx and possibly some of the data passed to it.

nazarenegena:
Cool, thanks for the update. I will work on it.

nazarenegena:
Updated the releases page to now show rule alias instead of rule ids

ui/src/components/ReleaseCard/index.jsx
```diff
@@ -254,7 +254,9 @@ function ReleaseCard(props) {
                               clickable
                               size="small"
                               icon={<LinkIcon />}
-                              label={ruleId}
+                              label={`${ruleInfo.product} ${
+                                ruleInfo.channel ? `- ${ruleInfo.channel}` : ''
```
bhearsum, line 258: This is not what we want either. The alias is a property of the `rule`. I can see why you're struggling though - the alias is not returned by the `/releases` endpoint, so the data you need is not even present.

I suggest you fetch the rules (similar to how we do on the rules page: https://github.com/mozilla-releng/balrog/blob/ca55f9aa6aa15647370a841d8827e1e724ab8813/ui/src/views/Rules/ListRules/index.jsx#L164), and use that to find the alias for each ruleId.

bhearsum (changes requested):

nazarenegena:
Cool, Noted. Will work on this.

nazarenegena:
I have updated the alias for the releases on the rules page
As for the rules that don't have an alias I have used their rule id

bhearsum (changes requested):
The changes look good - can you please fix the lint issues before we merge, though?

nazarenegena:
Working on the fixes.

nazarenegena:
I have fixed the lint issues

bhearsum (approved):
Thank you!

bhearsum:
@nazarenegena - Unfortunately, this now has conflicts with the latest `main`. Can you update your branch? Feel free to ping me on Matrix if you need help with that.

nazarenegena:
Cool, let me update my branch.

nazarenegena:
conflicts resolved.

bhearsum (approved):
This looks great now - thank you!

nazarenegena:
you're welcome.

PR #3025, opened 2023-10-20
Show required signoffs when updating rules

### LINKED ISSUE #1147: consider showing what signoffs are required when updating rules
when selecting `update` on a rule, it wasn't obvious if the rule required signoffs or who needed to sign off until the actual update was made.

e.g. here:

Screen Shot 2019-09-20 at 11 24 17 AM

it be neat if to show the signoffs required below or above the timestamp requirements in the update page:

Screen Shot 2019-09-20 at 11 26 09 AM

michellemounde:
Fixes #1147 

1. Fetch the required signoffs and check them against the rule to find matching ones
2. Create an element to display matching signoffs

No signoffs required - greyed out
Screenshot from 2023-10-20 15-22-37

Signoffs from one role required
Screenshot from 2023-10-20 15-22-57

Signoffs from two roles required
Screenshot from 2023-10-20 15-23-22

### DIFF
--- a/ui/src/views/Rules/Rule/index.jsx
+++ b/ui/src/views/Rules/Rule/index.jsx
@@ -101,7 +101,6 @@ function Rule({ isNewRule, user, ...props }) {
       : [];
   const [rule, setRule] = useState(initialRule);
   const [releaseNames, setReleaseNames] = useState([]);
-  const [ruleRequiredSignoffs, setRuleRequiredSignoffs] = useState([]);
   const [signoffSummary, setSignoffSummary] = useState('');
   const [products, fetchProducts] = useAction(getProducts);
   const [channels, fetchChannels] = useAction(getChannels);
@@ -126,7 +125,10 @@ function Rule({ isNewRule, user, ...props }) {
   const [updateSCAction, updateSC] = useAction(updateScheduledChange);
   const [deleteSCAction, deleteSC] = useAction(deleteScheduledChange);
   const isLoading =
-    fetchRuleAction.loading || products.loading || channels.loading;
+    fetchRuleAction.loading ||
+    products.loading ||
+    channels.loading ||
+    requiredSignoffs.loading;
   const actionLoading =
     releaseNamesAction.loading ||
     releaseNamesV2Action.loading ||
@@ -398,8 +400,6 @@ function Rule({ isNewRule, user, ...props }) {
             ruleMatchesRequiredSignoff(rule, rs)
           );
 
-          setRuleRequiredSignoffs(curr => curr.concat(matchingRs));
-
           if (!requiredSignoffs.length || !matchingRs.length) {
             setSignoffSummary(' Nobody');
           } else {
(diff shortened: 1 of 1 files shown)


AMUZY:
😑Took this from my code? really?

AMUZY:
Took this from my code?

michellemounde:
@AMUZY I don't know what your code looks like but I wrote my own code.

AMUZY:
It's cool though.

ui/src/views/Rules/Rule/index.jsx
```diff
@@ -401,6 +448,19 @@ function Rule({ isNewRule, user, ...props }) {
       {error && <ErrorPanel fixed error={error} />}
       {!isLoading && (
         <Fragment>
+          <div>
+            <p className={classes.signoffLabel}>
+              Required Signoffs:
```
bhearsum, line 453: Let's use a more complete sentence here: "Requires signoff from:"

This will also necessitate a different string when no signoffs are required. (Right now "None" is used, let's change that to "Nobody".)

ui/src/views/Rules/Rule/index.jsx
```diff
@@ -401,6 +448,19 @@ function Rule({ isNewRule, user, ...props }) {
...
+              <span
+                className={
+                  requiredSignoffs.data &&
```
bhearsum, line 456: I don't think we need any conditions here. Even when no signoffs are required, it's good to state that explicitly.

bhearsum (changes requested):
This looks more or less good to me. There's a couple of minor required changes, but nothing major. Nice work!

ui/src/views/Rules/Rule/index.jsx
michellemounde: Will work on this

ui/src/views/Rules/Rule/index.jsx
michellemounde: @bhearsum Not sure what you mean by stating explicitly when no signoffs are required. Does it just mean that I should remove the conditions?

bhearsum: That's right. We want to show a string here regardless of whether or not signoffs are required.

michellemounde: The conditions are for implementing CSS through the className to grey out the font for the signoffSummary if it is 'Nobody'. But to have the signoffSummary font in black when there is a list of required signoffs.

The signoffSummary string is always shown, it doesn't depend on the conditions.

So, considering this can I leave the conditions as is?

bhearsum:
@michellemounde - did you push the right thing here? It doesn't look like this branch has the changes you want anymore...

michellemounde:
> @michellemounde - did you push the right thing here? It doesn't look like this branch has the changes you want anymore...if you want a hand getting back to a good state feel free to ping me on Matrix.

Thanks for the help.

ui/src/views/Rules/Rule/index.jsx
bhearsum: It looks like you've mostly got this handled already? I do see that it now says `Nobody` when needed (great!). Are you sure you need the `class` to be conditional? It seems like the class should always be `classes.signoffs` since we always be showing this?

bhearsum (changes requested):

michellemounde:
@bhearsum I have removed the conditionals so that the classes.signoffs is always present

bhearsum (approved):
It looks good to me - thank you so much and nice work!

AMUZY:
Nice work @michellemounde

PR #3029, opened 2023-10-23
Prevent scheduling a new rule with a duplicate alias

### LINKED ISSUE #2942: Balrog accepts scheduled rule changes with duplicate alias'

michellemounde:
Fixes #2942 

1. Add duplicate alias check at API level to RuleScheduledChangesView

Before insert:
Screenshot from 2023-10-23 14-45-45

On insert:
Screenshot from 2023-10-23 14-45-54

### DIFF
--- a/src/auslib/web/admin/views/rules.py
+++ b/src/auslib/web/admin/views/rules.py
@@ -239,6 +239,10 @@ class RuleScheduledChangesView(ScheduledChangesView):
                         ext={"exception": "%s cannot be set to null/empty " "when scheduling insertion of a new rule" % field},
                     )
 
+            alias = what.get("alias", None)
+            if alias is not None and dbo.rules.getRule(alias):
+                return problem(400, "Bad Request", "Rule with alias exists.")
+
         if change_type in ["update", "insert"]:
             rule_dict, mapping_values, fallback_mapping_values = process_rule_form(what)
             what = rule_dict


bhearsum (commented):
This is an OK start, but more is needed. Specifically, we should reject duplicate aliases when:
* The alias exists in any row in the `rules` table
* The alias exists in any row in the `rule_scheduled_changes` table that meets both the following conditions:
  * `complete` is False
  * `change_type` is `insert` or `update`

We also need tests for these cases in https://github.com/mozilla-releng/balrog/blob/main/tests/admin/views/test_rules.py

michellemounde:
Ok @bhearsum will work on the changes. I have a question regarding the failing check. I've looked at the details but I am not sure how to fix the issue so that the check passes.

bhearsum:
> Ok @bhearsum will work on the changes. I have a question regarding the failing check. I've looked at the details but I am not sure how to fix the issue so that the check passes.

That error looks unrelated to your change. It will probably fix itself the next time you push - don't worry about it.

bhearsum:
This ended up getting fixed in #3010 first.

PR #3035, opened 2023-10-25
Canonical json form for signed responses

### LINKED ISSUE #1112: use canonical json form for signed responses
Currently, we don't have any guarantees that our JSON responses will be bit-for-bit the same for the same request. This means that any time the response format changes, we'll have to generate additional, possibly unnecessary, signatures (because we use the hashed, dumped-to-a-string, json as our cache key).

Something like https://github.com/Kinto/kinto-signer/blob/master/kinto_signer/canonicaljson.py or https://pypi.org/project/canonicaljson/.

Thanks to glasserc for the idea and pointers.

(Imported from https://bugzilla.mozilla.org/show_bug.cgi?id=1590609)

michellemounde:
Fixes #1112

1. Add canonicaljson dependency to requirements
2. Replace response json implementation with canonicaljson implementation


bhearsum (commented):
I know this is still just a draft, but I just want to note that we'll want one or more tests for this in https://github.com/mozilla-releng/balrog/blob/main/tests/web/test_json.py.

Also, if you need a hand with updating the requirements please let me know. Running `maintenance/pin.sh` _should_ do it though.

tests/web/test_json.py
```diff
@@ -233,6 +233,50 @@ def testGuardianResponse(client, version, buildTarget, channel, code, response):
...
+        ("0.6.0.0", "Linux_x86_64", "release", 404, {}),
+        ("0.6.0.0", "WINNT_x86_64", "beta", 404, {}),
+        # This shouldn't match because the rule on the alpha channel contains fields not used by this type of update query.
+        ("0.6.0.0", "WINNT_x86_64", "alpha", 404, {}),
+        ("0.6.0.0", "Darwin_x86_64", "evilrelease", 200, {}),
+    ],
+)
+def testGuardianResponseWithCanonicalJson(client, version, buildTarget, channel, code, response):
+    ret1 = client.get(f"/json/1/Guardian/{version}/{buildTarget}/{channel}/update.json")
+    ret2 = client.get(f"/json/1/Guardian/{version}/{buildTarget}/{channel}/update.json")
```
bhearsum, line 271: I'm not entirely certain what you're testing here. You seem to make making the same request twice, and asserting that the results are the same (which I would assume they would be...).

I think what you _should_ be doing is comparing a string version of the JSON with a string that is hardcoded in the expected, canonical order.

bhearsum (commented):
> @bhearsum I have a question. The canonical JSON tests are passing but they are also passing when I just use JSON.

I left a comment below about this.

> Should I also test for the signature in this case? I am not sure if I should do this or if it would just give the same result.

No - don't worry about the signature or any other headers here. It's out of scope for these new tests.

tests/web/test_json.py
michellemounde: I have made the requested change.

tests/web/test_json.py
```diff
@@ -233,6 +233,52 @@ def testGuardianResponse(client, version, buildTarget, channel, code, response):
...
+        ("0.6.0.0", "WINNT_x86_64", "alpha", 404, {}, "{}"),
+        ("0.6.0.0", "Darwin_x86_64", "evilrelease", 200, {}, "{}"),
+    ],
+)
+def testGuardianResponseWithCanonicalJson(client, version, buildTarget, channel, code, response, jsonString):
+    ret = client.get(f"/json/1/Guardian/{version}/{buildTarget}/{channel}/update.json")
+    assert ret.status_code == code
+    if code == 200:
+        assert ret.mimetype == "application/json"
+        assert ret.get_json() == response
```
bhearsum, line 277: This (and the `response` entries it uses) are unnecessary. Comparing the raw text to the expected JSON is all we need to do here.

(Calling `get_json` is just re-parsing the raw text into a dictionary - so we're effectively checking the expected result twice.)

bhearsum (changes requested):
One small thing below, but otherwise this is ready to merge.

tests/web/test_json.py
michellemounde: Updated this as requested

michellemounde:
I tried rebasing and resolving merge conflicts but I got an error when running `maintenance/pin.sh`. The error:
"Package platformdirs was resolved to different versions in different environments: 3.11.0 and 4.0.0
RuntimeError: Please add constraints for the package version listed above"

bhearsum:
> I tried rebasing and resolving merge conflicts but I got an error when running `maintenance/pin.sh`. The error: "Package platformdirs was resolved to different versions in different environments: 3.11.0 and 4.0.0 RuntimeError: Please add constraints for the package version listed above"

Hm...it looks like this is a general problem unrelated to this patch. I'll try to put together a fix for it.

bhearsum:
> > I tried rebasing and resolving merge conflicts but I got an error when running `maintenance/pin.sh`. The error: "Package platformdirs was resolved to different versions in different environments: 3.11.0 and 4.0.0 RuntimeError: Please add constraints for the package version listed above"
> 
> Hm...it looks like this is a general problem unrelated to this patch. I'll try to put together a fix for it.

Once https://github.com/mozilla-releng/balrog/pull/3050 is merged you should be able to rebase and update the requirements.

bhearsum (approved):
Thank you, nice work!!

PR #3037, opened 2023-10-25
Test highlightMatchedRelease

### LINKED ISSUE #3036: Add a unit test for the highlightMatchedRelease function
Add a unit test for the `highlightMatchedRelease` function introduced in #3033.

michellemounde:
Fixes #3036

1. Comment the highlightMatchedRelease function
2. Convert highlightMatchedRelease to a logic module
3. Add tests for highlightMatchedRelease when:
a. One matching substring
b. Two matching substrings
c. Release name ends with last matching substring
d. Release name starts with last matching substring
4. Modify search to remove leading and trailing whitespace to fix false positive match bug
5. Adjust highlighlightMatchedRelease to return an array instead of a React Fragment
6. Modify tests the returned array in place of the previously returned React Fragment


gabrielBusta (approved):
🎉 thanks @michellemounde!

PR #3040, opened 2023-10-30
Show required signoffs appropriately when updating rules 

### LINKED ISSUE #1147: consider showing what signoffs are required when updating rules

michellemounde:
Fixes #1147 by updating #3025 to apply the following fixes:

1. Fix unneccessary re-fetching of required signoffs by fetching them once on page load
2. Fix unneccessary re-rendering of page when signoff summary updates on channel or product value change
3. Fix signoff summary to show 'Nobody' instead of all signoffs required when no product value
4. Fix multiple renders of same role in signoff summary to only one role summary per role

### DIFF
--- a/ui/src/views/Rules/Rule/index.jsx
+++ b/ui/src/views/Rules/Rule/index.jsx
@@ -1,7 +1,7 @@
 import React, { Fragment, useState, useEffect } from 'react';
 import classNames from 'classnames';
 import { bool } from 'prop-types';
-import { defaultTo, assocPath, pick } from 'ramda';
+import { defaultTo, assocPath, pick, match } from 'ramda';
 import { stringify } from 'qs';
 import NumberFormat from 'react-number-format';
 import { makeStyles } from '@material-ui/styles';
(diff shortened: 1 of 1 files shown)


ui/src/views/Rules/Rule/index.jsx
```diff
@@ -391,34 +393,50 @@ function Rule({ isNewRule, user, ...props }) {
...
-          const matchingRs = requiredSignoffs.filter(rs =>
-            ruleMatchesRequiredSignoff(rule, rs)
-          );
-
-          if (!requiredSignoffs.length || !matchingRs.length) {
-            setSignoffSummary(' Nobody');
+    const rs =
+      requiredSignoffs.data && requiredSignoffs.data.data.required_signoffs;
+
+    if (rs) {
```
bhearsum, line 399: Although it's unlikely to be the case, we should call `setSignoffSummary` if we don't enter this if block - otherwise it will be `null`.

bhearsum (changes requested):
Just one tiny thing to address, but thank you for the quick follow-ups!

ui/src/views/Rules/Rule/index.jsx
michellemounde: Ok, lemme work on this.

bhearsum (approved):
Nice work!

PR #3041, opened 2023-10-30
Stop showing scheduled deletions on unrelated channels

### LINKED ISSUE #2312: UI shows scheduled deletions for unrelated channels
Looking at e.g. https://balrog.services.mozilla.com/rules?product=SystemAddons&channel=beta, I shouldn't see scheduled deletions for SystemAddons/esr rules.

michellemounde:
Fixes #2312:

1. Modify ruleMatchesChannel function to account for scheduled deletions
2. Rename tests for increased clarity on what is being tested
3. Create tests for scenarios that arose from rewording tests
4. Create tests for scheduled deletions matching

### DIFF
--- a/ui/src/utils/rules.js
+++ b/ui/src/utils/rules.js
@@ -22,15 +22,17 @@ const ruleMatchesChannel = (rule, channel) => {
   // we never want scheduled deletions to match any filter
   // because the scheduled change channel on deletions is null
   // and the rule channel is a truthy value that might not equal 'channel'
-  const scChannelMatches = rule.scheduledChange
-    ? (rule.channel &&
-        ruleChannelMatches &&
-        rule.scheduledChange.channel === null) ||
-      (!rule.channel && rule.scheduledChange.channel === null) ||
-      rule.scheduledChange.channel === '' ||
-      rule.scheduledChange.channel === channel ||
-      matchesGlob(rule.scheduledChange.channel, channel)
-    : false;
+  let scChannelMatches = false;
+
+  if (rule.scheduledChange) {
+    if (rule.scheduledChange.change_type !== 'delete') {
+      scChannelMatches =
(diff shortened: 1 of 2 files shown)


ui/src/utils/rules.js
```diff
@@ -18,8 +18,15 @@ const ruleMatchesChannel = (rule, channel) => {
...
+  // if a scheduled change is a deletion
+  // we never want scheduled deletions to match any filter
+  // because the scheduled change channel on deletions is null
+  // and the rule channel is a truthy value that might not equal 'channel'
   const scChannelMatches = rule.scheduledChange
-    ? rule.scheduledChange.channel === null ||
+    ? (rule.channel &&
+        ruleChannelMatches &&
+        rule.scheduledChange.channel === null) ||
+      (!rule.channel && rule.scheduledChange.channel === null) ||
```
bhearsum, line 29: This appears to work, but I'm suspicious that something subtle is off here. I was expecting `rule.scheduledChange.change_type === 'delete'` to be treated specially - not to have _all_ scheduled change types have this logic applied. For `change_type` === `update` or `insert` the channel should be matching if either the `rule` or `rule.scheduledChange` channel matches.

Could you try to make this a bit more clear? Feel free to drop the `const` on `scChannelMatches` and write this as if/else statements if it helps.

bhearsum (changes requested):

ui/src/utils/rules.js
michellemounde: I wasn't sure whether to query for the change type initially or just make the function work for all scheduled changes. I appreciate the feedback and I'll work on making it clear.

bhearsum (approved):
This looks good, and my local testing can't find any cases that are displaying incorrectly. Thank you!

PR #3044, opened 2023-11-06
added jsconfig for vscode intellisense

AMUZY:
So it was somewhat annoying that vscode's intellisense wasn't just working properly in the scripts, so to get the javascript language service running in the ui, I added a `jsconfig.json` file in the `ui` directory and it fixed the issue for me.

BEFORE : 

intellisense_before

AFTER :

Code_eOU93UMUjE
intellisense_after_2

👌👌👌As you can see...now it works!


gabrielBusta:
Thanks @AMUZY this WFM too

gabrielBusta (approved):

PR #3045, opened 2023-11-07
chore: update dependencies

ahal:


ahal:
Our pin of `pyjwt` in base.in was causing a conflict. I removed it to proceed, but I'm not sure why it was pinned in the first place (looks like Gabriel added it back in May to work around a CI failure). I guess we'll see if that failure still reproduces.

ahal:
Looks like this is also tripping over the connexion 3.0 release. I'll add a pin here as well.

gbrownmozilla (approved):

PR #3058, opened 2023-11-24
Revert #3035 for busting content signatures in JSON responses

bhearsum:
@jcristau and I discovered that we end up with a byte string from https://github.com/mozilla-releng/balrog/blob/7ee1232b714409508c3b3edf9f0557c4ad544107/src/auslib/web/public/json.py#L19. When [make_hash](https://github.com/mozilla-releng/balrog/blob/7ee1232b714409508c3b3edf9f0557c4ad544107/src/auslib/util/autograph.py#L12) formats it, we end up with a `b` prefix in the string - so we end up signing different data than we return in the body.


requirements/base.txt
```diff
@@ -408,7 +404,6 @@ cryptography==41.0.5 \
     --hash=sha256:ec3b055ff8f1dce8e6ef28f626e0972981475173d7973d63f271b29c8a2897da \
     --hash=sha256:fba1e91467c65fe64a82c689dc6cf58151158993b13eb7a7f3f4b7f395636723
     # via
-    #   auth0-python
```
jcristau, line 411: Odd.

jcristau (approved):

ahal (approved):

requirements/base.txt
bhearsum: Yeah, I'm not sure what's up with this. The version of `auth0-python` we're using _certainly_ depends on it. It seems like it's some odd artifact of using `--no-upgrade`. If I re-run `pin.sh` without that, `auth0-python` is still listed here.

PR #3064, opened 2023-11-28
Fix/silence deprecation warnings

jcristau:


tests/test_db.py
```diff
@@ -5309,7 +5309,7 @@ def testMustRemovePinToRemoveRelease(self):
             self.assertRaises(ValueError, releases.delete_release, name="Firefox-60.0-build1", changed_by="bob", trans=trans)
         self.assertEqual(dbo.releases_json.count(where=[dbo.releases_json.name == "Firefox-60.0-build1"]), 1)
         self.pinnable_releases.delete(changed_by="bob", where=[self.pinnable_releases.mapping == "Firefox-60.0-build1"], old_data_version=row["data_version"])
-        dbo.releases_json.delete(where=[dbo.releases_json.name == "Firefox-60.0-build1"], changed_by="bob", old_data_version=1)
+        dbo.releases_json.t.delete().where(dbo.releases_json.name == "Firefox-60.0-build1").execute()
```
gbrownmozilla, line 5312: Different parameters here, but I think that's okay.

src/auslib/util/versions.py
```diff
@@ -1,8 +1,157 @@
 import re
-from distutils.version import LooseVersion, StrictVersion, Version
 
 from auslib.errors import BadDataError
 
+# Version/StrictVersion/LooseVersion classes are lifted from distutils which
```
gbrownmozilla, line 5: There's `packaging.version` of course, but this approach seems safer.

gbrownmozilla (approved):

tests/test_db.py
jcristau: Right, we're bypassing the AUS db interface here and calling into sqlalchemy directly (the alternative would require setting up or mocking the GCS history bucket, which isn't really relevant to this particular test).

bhearsum (approved):
Looks totally reasonable to me. Even if we wanted to do something different than vendoring the old distutils classes, it should be done separately IMO.

PR #3065, opened 2023-11-28
Remove assumptions about release blobs from rule evaluation

### LINKED ISSUE #2985: evaluateRules throws exceptions when `pin=1` and the query maps to a DesupportBlob
When this happens, we end up at this line of evaluateRules: https://github.com/mozilla-releng/balrog/blob/0c832ed4332c7605d029a3f4bcceff31660de6d6/src/auslib/AUS.py#L148

DesupportBlob does not have a `getApplicationVersion` method, so we end up with an AttributeError:
```
AttributeError: 'DesupportBlob' object has no attribute 'getApplicationVersion'
  File "auslib/web/public/client.py", line 146, in get_update_blob
    release, update_type, eval_metadata = AUS.evaluateRules(query, transaction=transaction)
  File "auslib/AUS.py", line 148, in evaluateRules
    blob_version = blob.getApplicationVersion(updateQuery["buildTarget"], updateQuery["locale"])

AttributeError: 'DesupportBlob' object has no attribute 'getApplicationVersion'
  File "flask/app.py", line 1823, in full_dispatch_request
    rv = self.dispatch_request()
  File "flask/app.py", line 1799, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)
  File "connexion/decorators/decorator.py", line 68, in wrapper
    response = function(request)
  File "connexion/decorators/uri_parsing.py", line 149, in wrapper
    response = function(request)
  File "connexion/decorators/validation.py", line 399, in wrapper
    return function(request)
  File "connexion/decorators/produces.py", line 41, in wrapper
    response = function(request)
  File "connexion/decorators/parameter.py", line 120, in wrapper
    return function(**kwargs)
  File "auslib/db.py", line 202, in __exit__
    raise e
  File "auslib/web/public/client.py", line 146, in get_update_blob
    release, update_type, eval_metadata = AUS.evaluateRules(query, transaction=transaction)
  File "auslib/AUS.py", line 148, in evaluateRules
    blob_version = blob.getApplicationVersion(updateQuery["buildTarget"], updateQuery["locale"])
```

If possible, formalizing `getApplicationVersion` as part of the `Blob` interface would probably be good. If not, we need to be much more careful about how and when we call it.

jcristau:
Closes #2985

### DIFF
--- a/src/auslib/AUS.py
+++ b/src/auslib/AUS.py
@@ -3,11 +3,10 @@ import logging
 from random import randint
 from urllib.parse import urlparse
 
-from auslib.blobs.base import createBlob
-from auslib.errors import BadDataError
+from auslib.blobs.base import ServeUpdate, createBlob
 from auslib.global_state import cache, dbo
 from auslib.services import releases
-from auslib.util.versions import MozillaVersion, PinVersion
+from auslib.util.versions import PinVersion
 
 
 class ForceResult(object):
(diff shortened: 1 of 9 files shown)


tests/blobs/test_guardian.py
```diff
@@ -41,7 +41,8 @@ def testContainsForbiddenDomain(guardianblob, allowlistedDomains, expected):
 )
 def testShouldServeUpdateVariousVersions(guardianblob, version, expected):
     updateQuery = {"product": "Guardian", "version": version, "buildTarget": "WINNT_x86_64", "channel": "release"}
-    assert guardianblob.shouldServeUpdate(updateQuery) is expected
+    serveUpdate = {False: guardianblob.ServeUpdate.No, True: guardianblob.ServeUpdate.Yes}[expected]
```
bhearsum, line 44: question: Is this indirection necessary? If not, please replace switch out the True & False in the decorator with the enums.

src/auslib/blobs/base.py
```diff
@@ -135,6 +136,14 @@ def merge_dicts(ancestor, left, right):
 class Blob(dict):
     jsonschema = None
 
+    class ServeUpdate(enum.Enum):
```
bhearsum, line 139: suggestion (non-blocking): Move this to the top level of the file. I mainly suggest this because things like `candidate == blob.ServeUpdate.Maybe` further up comparing a return value from a blob method to a fixed property of a blob, which just seems a bit odd. `candidate == ServeUpdate.Maybe` is a bit more intuitive to me. Totally up to you though.

src/auslib/AUS.py
```diff
@@ -132,49 +132,41 @@ def get_blob(mapping):
...
-                    # This situation hides the problem from the sysadmin and leaves some
-                    # installations vulnerable.
-                    if pin_mapping is not None:
-                        mapping = pin_mapping
-                        blob = get_blob(mapping)
-                        if not blob or not blob.shouldServeUpdate(updateQuery):
-                            return None, None, eval_metadata
+        if candidate == blob.ServeUpdate.Maybe:
+            version_pin = PinVersion(updateQuery.get("pin"))
+            pin_mapping = dbo.pinnable_releases.getPinMapping(updateQuery["product"], getFallbackChannel(updateQuery["channel"]), str(version_pin))
```
bhearsum, line 143: suggestion (non-blocking): seeing as pinnable releases are only relevant to AppRelease blobs, can/should this be moved into that as well? (This could be filed away for the future...)

bhearsum (approved):
(I'm going to try using [Conventional Comments](https://conventionalcomments.org/#labels) for this review, FYI.)

I have a few non-blocking suggestions, but this looks good to me overall. Thank you for doing this!

src/auslib/blobs/base.py
jcristau: Done.

src/auslib/AUS.py
jcristau: Filed as #3067

PR #3066, opened 2023-11-28
log: avoid crashing when a log entry includes a non-JSON-serializable object

jcristau:
When enabling all debug logging we end up with fairly arbitrary objects in log messages.  Default to the object's `__str__` method in that case.

Note this is fairly arbitrary, I could be convinced to use `repr` instead.  I've seen stuff like `bytes`, `type`, `Response` in extra logging data from our dependencies, which all break with plain `json.dumps`.

### DIFF
--- a/src/auslib/log.py
+++ b/src/auslib/log.py
@@ -135,7 +135,7 @@ class JsonLogFormatter(logging.Formatter):
 
         out["Fields"] = fields
 
-        return json.dumps(out)
+        return json.dumps(out, default=str)
 
 
 def safer_format_traceback(exc_typ, exc_val, exc_tb):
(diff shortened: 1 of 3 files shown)


src/auslib/log.py
```diff
@@ -135,7 +135,7 @@ def format(self, record):
 
         out["Fields"] = fields
 
-        return json.dumps(out)
+        return json.dumps(out, default=str)
```
gbrownmozilla, line 138: I don't have a strong opinion on `str` vs `repr`. `str` seems fine.

gbrownmozilla (changes requested):
https://github.com/mozilla-releng/balrog/blob/main/agent/src/balrogagent/log.py#L105 looks very similar ... can we do the same thing there?

jcristau:
Good catch, updated to include the agent.

gbrownmozilla (approved):

PR #3068, opened 2023-12-04
Run with python 3.11

jcristau:
Client tests remain on 3.9 since that's what balrogscript uses for now.


bhearsum (commented):
> Client tests remain on 3.9 since that's what balrogscript uses for now.

Should we run them against 3.11 as well, to ensure we stay unbroken there?

bhearsum (approved):

PR #3074, opened 2023-12-05
Add new public json endpoint including OS Version

### LINKED ISSUE #3071: Consider Adding OS_VERSION to `/json/` endpoint.
Hellooo! 

Over at the Mozilla VPN project, we just had deprecated an old osx version. However, we just noticed that we're not telling balrog the users OS-version, so we could not write a rule that would not advertise the update to the now stranded user 😅 
I was experimenting with the XML endpoints, but I just hit an error (https://github.com/mozilla-releng/balrog/issues/3069, sorry 🙏 ) 

So I was wondering if it is possible to get a `/json/2` endpoint like: 
https://github.com/mozilla-releng/balrog/blob/2675386418a0c8d5ed9c12a20ed44a6fe99b2440/src/auslib/web/public/swagger/api.yml#L199
But that can also consume the`{osVersion}` field similar to the XML endpoint? 

Greetings, Basti :)

jcristau:
Fixes #3071

### DIFF
--- a/src/auslib/web/public/json.py
+++ b/src/auslib/web/public/json.py
@@ -22,3 +22,7 @@ def get_update(transaction, **parameters):
     headers.update(get_content_signature_headers(response, ""))
 
     return Response(response=response, status=200, headers=headers, mimetype="application/json")
+
+
+get_update_1 = get_update
+get_update_2 = get_update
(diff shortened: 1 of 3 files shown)


gbrownmozilla (approved):
lgtm! Thanks for the extensive tests.

jcristau:
For the record the tests are copied over from the v1 endpoint, with the addition of an extra Guardian-0.6.0.0 release and a rule that points the `obsolete` OS version at that release (with the corresponding test).

PR #3076, opened 2023-12-06
doc update: bump version after tagging, rather than after prod deploy

gbrownmozilla:
The existing docs suggest the balrog version should be bumped only after the production deployment. With our new Tuesday (staging) / Thursday (production) cadence, that means we go two days with version.txt == the current tag. I think it makes more sense to bump the version immediately after tagging and release creation -- ie, as part of the staging deployment.


jcristau (approved):
Thanks!

PR #3078, opened 2023-12-11
Rewind the rules table UI prototype

michellemounde:
Implementation of rewind the rules table UI prototype from [old Balrog UI](https://github.com/mozilla-releng/balrog-ui/pull/135).


michellemounde:
Moved to #3079

PR #3079, opened 2023-12-14
Rewind the rules table UI prototype

michellemounde:
Implementation of rewind the rules table UI prototype from [old Balrog UI](https://github.com/mozilla-releng/balrog-ui/pull/135).

### DIFF
--- a/ui/src/components/VariableSizeList/index.jsx
+++ b/ui/src/components/VariableSizeList/index.jsx
@@ -9,13 +9,19 @@ import { AutoSizer, WindowScroller, List } from 'react-virtualized';
 import { APP_BAR_HEIGHT } from '../../utils/constants';
 
 const VariableSizeList = forwardRef((props, ref) => {
-  const { scrollToRow, ...rest } = props;
+  const { scrollToRow, pathname, ...rest } = props;
   const listRef = useRef(null);
 
   useEffect(() => {
     const rowOffset = listRef.current.getOffsetForRow({ index: scrollToRow });
 
-    listRef.current.scrollToPosition(rowOffset - APP_BAR_HEIGHT);
+    if (pathname === '/rules') {
+      listRef.current.scrollToPosition(
+        rowOffset - APP_BAR_HEIGHT - rest.searchFieldHeight
+      );
+    } else {
+      listRef.current.scrollToPosition(rowOffset - APP_BAR_HEIGHT);
+    }
   }, [scrollToRow]);
 
   useImperativeHandle(ref, () => ({
--- a/ui/src/utils/getFilteredRulesInfo.js
+++ b/ui/src/utils/getFilteredRulesInfo.js
@@ -1,0 +1,33 @@
+// This utility builds the string shown in the
+// info alert when no rules match the selected filters
+export default (productChannelQueries, rewindDate, onlyScheduledChanges) => {
+  let info = 'No rules found ';
+  const rewindDateStr = `on ${rewindDate}.`;
+  const scheduledChangesStr = 'with scheduled changes.';
+
+  if (!productChannelQueries) {
+    if (rewindDate) {
+      info += rewindDateStr;
+    } else if (onlyScheduledChanges) {
+      info += scheduledChangesStr;
+    }
+  } else if (productChannelQueries[1]) {
+    info += `for the ${productChannelQueries[0]} ${productChannelQueries[1]} channel `;
+
+    if (rewindDate) {
+      info += rewindDateStr;
+    } else if (onlyScheduledChanges) {
+      info += scheduledChangesStr;
+    }
+  } else {
+    info += `for the ${productChannelQueries[0]} channel `;
+
+    if (rewindDate) {
+      info += rewindDateStr;
+    } else if (onlyScheduledChanges) {
+      info += scheduledChangesStr;
+    }
+  }
+
+  return info;
+};
--- a/ui/src/views/Rules/ListRules/index.jsx
+++ b/ui/src/views/Rules/ListRules/index.jsx
@@ -1379,17 +1379,19 @@ function ListRules(props) {
               helperText={rewindDateError}
               onDateTimeChange={handleRewindDateTimeChange}
               value={rewindDate}
-              InputProps={{
-                endAdornment: (
-                  <IconButton
-                    onClick={() => handleRewindDateTimeChange(null)}
-                    disabled={!rewindDate}
-                    style={{ order: 1 }}
-                    color="disabled">
-                    <CloseIcon />
-                  </IconButton>
-                ),
-              }}
+              InputProps={
+                rewindDate && {
+                  endAdornment: (
+                    <IconButton
+                      onClick={() => handleRewindDateTimeChange(null)}
+                      disabled={!rewindDate}
+                      style={{ order: 1 }}
+                      color="disabled">
+                      <CloseIcon />
+                    </IconButton>
+                  ),
+                }
+              }
               InputAdornmentProps={{
                 position: 'start',
                 style: { order: 2, marginLeft: 0 },
--- a/ui/src/views/Rules/ListRules/index.jsx
+++ b/ui/src/views/Rules/ListRules/index.jsx
@@ -429,7 +429,9 @@ function ListRules(props) {
       setRulesWithScheduledChanges(sortedRules);
       setRewoundRules([]);
     }
+  }, [rules.data, scheduledChanges.data, requiredSignoffs.data]);
 
+  useEffect(() => {
     if (emergencyShutoffsAction.data && scheduledEmergencyShutoffsAction.data) {
       const shutoffs = emergencyShutoffsAction.data.data.shutoffs.map(
         shutoff => {
@@ -450,13 +452,7 @@ function ListRules(props) {
 
       setEmergencyShutoffs(shutoffs);
     }
-  }, [
-    rules.data,
-    scheduledChanges.data,
-    requiredSignoffs.data,
-    emergencyShutoffsAction.data,
-    scheduledEmergencyShutoffsAction.data,
-  ]);
+  }, [emergencyShutoffsAction.data, scheduledEmergencyShutoffsAction.data]);
 
   useEffect(() => {
     fetchRules(rewindDate ? rewindDate.getTime() : null);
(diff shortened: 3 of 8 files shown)


ui/src/views/Rules/ListRules/index.jsx
```diff
@@ -457,12 +497,17 @@ function ListRules(props) {
       filteredRules = filteredRules.filter(rule => rule.scheduledChange);
     }
 
+    const rewoundFilteredRules = clone(rewoundRules);
+    // use this line for non-diff mode on rewound rules
+    let rulesToShow =
+      rewoundRules.length === 0 ? filteredRules : rewoundFilteredRules;
```
bhearsum, line 503: issue: This conditional is close, but misses the edge case of rewinding far enough in the past that no rules existed. Eg: if I rewind back to 2012, I see the current rules instead of no rules at all.

a related suggestion (non-blocking): When there are no rules to show, it might be good to show some text to confirm to the user that there are no rules matching the rewind and/or product/channel filter. (Otherwise the user may assume something has gone wrong.)

ui/src/views/Rules/ListRules/index.jsx
```diff
@@ -1285,12 +1370,33 @@ function ListRules(props) {
...
+      title={
+        rewindDate ? `Rules @ ${rewindDate.toString().split('(')[0]}` : 'Rules'
+      }>
       {isLoading && <Spinner loading />}
       {error && <ErrorPanel fixed error={error} />}
       {!isLoading && productChannelOptions && (
         <Fragment>
           <div ref={searchFieldRef} className={classes.options}>
+            <DateTimePicker
+              disableFuture
```
bhearsum, line 1383: note: This doesn't seem to disable future times - only future dates. I imagine this is a bug in the underlying component we're using - so until we upgrade to something supported there's probably nothing we can do about it.

ui/src/views/Rules/ListRules/index.jsx
```diff
@@ -1285,12 +1370,33 @@ function ListRules(props) {
...
+              clearable
+              inputVariant="outlined"
+              label="Rewind to..."
```
bhearsum, line 1386: todo: please align this control and label with the others on the page. It currently sits slightly below the others:
image

ui/src/views/Rules/ListRules/index.jsx
```diff
@@ -140,7 +144,9 @@ function ListRules(props) {
   const [rulesWithScheduledChanges, setRulesWithScheduledChanges] = useState(
     []
   );
-  const searchFieldRef = useRef();
+  const [rewoundRules, setRewoundRules] = useState([]);
```
bhearsum, line 147: nit: this would a little bit easier to read if it was above `rewindDate`.

ui/src/views/Rules/ListRules/index.jsx
```diff
@@ -413,25 +419,59 @@ function ListRules(props) {
...
+  ]);
+
+  useEffect(() => {
+    fetchRules(rewindDate ? rewindDate.getTime() : null);
+  }, [rewindDate]);
+
+  useEffect(() => {
+    Promise.all([
+      fetchScheduledChanges(),
+      fetchRules(),
```
bhearsum, line 464: question: Is this needed here? It looks to me like we're making a `GET` to `/api/rules` twice upon first page load (once because of this effect, and once because of the above). I suspect this can be removed.

ui/src/views/Rules/ListRules/index.jsx
```diff
@@ -413,25 +419,59 @@ function ListRules(props) {
...
+      fetchRequiredSignoffs(OBJECT_NAMES.PRODUCT_REQUIRED_SIGNOFF),
+      fetchEmergencyShutoffs(),
+      fetchScheduledEmergencyShutoffs(),
+      fetchProducts(),
+      fetchChannels(),
+    ]);
+  }, []);
+
+  useEffect(() => {
+    setSearchFieldHeight(searchFieldRef.current.clientHeight);
```
bhearsum, line 474: question: What is the puprose of this code?

ui/src/views/Rules/ListRules/index.jsx
```diff
@@ -1235,6 +1316,7 @@ function ListRules(props) {
             : Object.values(rule.scheduledChange).join('-')
         }
         style={style}>
+        {/* should we go read only mode if rewindDate is set instead? */}
```
bhearsum, line 1319: todo: please remove this comment. (It was an interesting thought, but I did a quick experiment locally and I don't think using readOnly mode is as nice as simply disabling the controls as you're doing already.)

ui/src/views/Rules/ListRules/index.jsx
```diff
@@ -83,10 +84,13 @@ const useStyles = makeStyles(theme => ({
     left: 0,
     padding: `${theme.spacing(4)}px ${theme.spacing(4)}px 0`,
     display: 'flex',
-    justifyContent: 'flex-end',
+    justifyContent: 'center',
```
bhearsum, line 87: praise: thank you for restoring this - I think it was changed unintentionally in a recent PR.

bhearsum (changes requested):
(I will be using [conventional commits](https://conventionalcomments.org/#labels) for this review to try to help improve clarity.)

This is really excellent work! I know there's quite a few comments I'm making, but for the most part these are small papercuts or edge cases. Overall, this is working very well, and in very good shape. You've also improved on the original prototype in a number of ways - very nice!

issue: The rules do not sort correctly when we've rewound. When I first rewind they do not appear sorted at all (I have rule 63, then 80, then 91). If I reload the page with the `timestamp` query parameter in the URL they are reverse from lowest to highest priority (it should be highest to lowest). (This happens with a filter applied, and without a filter applied).

issue: Clicking on the `rules` link at the top of the page drops the `timestamp` from the URL, but does not reset the view to show the current rules. It should either do both of these things, or neither.

todo: please update the [anchors in the rule cards to include the timestamp](https://github.com/mozilla-releng/balrog/blob/83e4fc547de428f54cf3943d836097322f3b9205/ui/src/components/RuleCard/index.jsx#L325). (This way we can share links to a particular rule at a particular point in time.)

todo: please disable or remove the "Add Rule" control on on ListRules when the table is rewound (the "Disable Updates" button can stay enabled - we want it be easily accessible at all times).

ui/src/views/Rules/ListRules/index.jsx
michellemounde: No, I checked it out and its causing the rules to be fetched twice. I will remove it.

ui/src/views/Rules/ListRules/index.jsx
bhearsum: Thank you! Please add a comment to this effect so we remember why this exists in the future.

ui/src/utils/getFilteredRulesInfo.js
```diff
@@ -0,0 +1,33 @@
+// This utility builds the string shown in the
+// info alert when no rules match the selected filters
+export default (productChannelQueries, rewindDate, onlyScheduledChanges) => {
+  let info = 'No rules found ';
```
bhearsum, line 4: praise: This function seems to handle all states that the Rules page could be in cleanly and efficiently - nice work!

bhearsum (changes requested):
I'm marking this as needing changes for the first issue noted below - but this looks very, very good overall - and is almost ready to land.

issue: The "Updates are currently disabled for the xxx yyy channel" banner is not consistently shown when updates are disabled and the rules are rewound. It seems to be shown when you first disable in rewound mode - but not shown if you refresh the page.

issue (non-blocking): I'm not sure if this is a regression or not, but loading a link with an anchor in it does not scroll to the right place. For example, loading https://localhost:9000/rules?product=Firefox&channel=beta&timestamp=1575398460000#ruleId=525 ends up with the top of the rule card cut off:
image

(It looks like the controls at the top may be floating on top of the rule card?)

ui/src/components/VariableSizeList/index.jsx
```diff
@@ -9,13 +9,19 @@ import { AutoSizer, WindowScroller, List } from 'react-virtualized';
 import { APP_BAR_HEIGHT } from '../../utils/constants';
 
 const VariableSizeList = forwardRef((props, ref) => {
-  const { scrollToRow, ...rest } = props;
+  const { scrollToRow, pathname, ...rest } = props;
```
bhearsum, line 12: praise: Thank you for fixing this!

bhearsum (changes requested):
It looks like there's still at least one issue with the "Updates are currently disabled..." state. With these steps it isn't shown:
* Select a product/channel
* Rewind the rules
* Disable updates for the product/channel
* Clear the rewind

This causes the "Updates are disabled" dialog to disappear. If I refresh the page it re-appears. I suggest doing some logging of all of the variables that control whether or not the `EmergencyShutoffCard` is shown to help debug this (eg: `productChannelFilter`, `emergencyShutoffs`, etc.).

Once you get this sorted out this will be ready to go!

bhearsum (approved):
This looks good to me now. @gabrielBusta should review it as well.

Thanks for all your hard work on this @michellemounde!!

ui/src/views/Rules/ListRules/index.jsx
```diff
@@ -1285,17 +1353,48 @@ function ListRules(props) {
...
       {isLoading && <Spinner loading />}
       {error && <ErrorPanel fixed error={error} />}
       {!isLoading && productChannelOptions && (
         <Fragment>
           <div ref={searchFieldRef} className={classes.options}>
+            <DateTimePicker
+              className={classes.rewindPicker}
+              disableFuture
+              inputVariant="outlined"
+              label="Rewind to..."
```
gabrielBusta, line 1369: What about "Rule database state as of YYYY/DD/MM HH:MM AM/PM"?

gabrielBusta (approved):
I tried it out in my local checkout and it looks good to me. Might do some light testing and land it tomorrow. Nice work!

ui/src/views/Rules/ListRules/index.jsx
michellemounde: I think rewind to might be more clear for the label when there is no rewind date selected.

PR #3081, opened 2023-12-18
Guard against non-xml releases in xml get_update_blob

### LINKED ISSUE #3069: GuardianBlob does not behave correctly when used in an XML endpoint
We end up calling `getResponseProducts`, which is not implemented. This isn't really a _real_ problem, because all MozillaVPN clients use the JSON endpoint...but we should handle this more gracefully to avoid exception spam when someone does something silly like this.

Sentry Issue: [BALROG-PROD-PUBLIC-4G](https://mozilla.sentry.io/issues/4698120466/?referrer=github_integration)

```
AttributeError: 'GuardianBlob' object has no attribute 'getResponseProducts'
  File "auslib/web/public/client.py", line 150, in get_update_blob
    response_products = release.getResponseProducts()

AttributeError: 'GuardianBlob' object has no attribute 'getResponseProducts'
(6 additional frame(s) were not displayed)
...
  File "connexion/decorators/parameter.py", line 120, in wrapper
    return function(**kwargs)
  File "auslib/db.py", line 202, in __exit__
    raise e
  File "auslib/web/public/client.py", line 150, in get_update_blob
    response_products = release.getResponseProducts()
```

jcristau:
Fixes #3069

There might be a cleaner/cleverer way to do this?

### DIFF
--- a/src/auslib/web/public/client.py
+++ b/src/auslib/web/public/client.py
@@ -7,7 +7,8 @@ from flask import current_app as app
 from flask import make_response
 
 from auslib.AUS import FORCE_FALLBACK_MAPPING, FORCE_MAIN_MAPPING
-from auslib.blobs.base import createBlob
+from auslib.blobs.base import XMLBlob, createBlob
+from auslib.errors import BadDataError
 from auslib.global_state import dbo
 from auslib.services import releases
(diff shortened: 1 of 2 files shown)


bhearsum (approved):
I suppose this could probably be "fixed" with some more enhanced type checking to make the interfaces on the blobs better...but this is totally fine.

PR #3083, opened 2023-12-22
Add filter query to url on roles page

michellemounde:
Add filter query to the url on the roles page.

Before:
Screenshot from 2023-12-22 15-41-58

Now:
Screenshot from 2023-12-22 15-42-14


bhearsum (approved):
Thank you! We'll hold off on merging this until the new year, but it looks good.

PR #3087, opened 2024-01-12
feat: enable sentry integration for manage-db.py functions

bhearsum:
This is primarily aimed at the cleanup/database dump cronjobs, which currently have no visibility, but I imagine it will also be helpful for database upgrades as well.

Unfortunately, there's no good way to test this until it hits staging...


jcristau (approved):
seems worth a try, thanks!

PR #3088, opened 2024-01-16
Admin signoffs for changes to permissions

michellemounde:
Require 2 admin signoffs for changes to permissions to solve #2194

### DIFF
--- a/src/auslib/db.py
+++ b/src/auslib/db.py
@@ -1309,19 +1309,28 @@ class ScheduledChangeTable(AUSTable):
         self._checkBaseTablePermissions(base_table_where, base_columns, changed_by, transaction)
 
     def auto_signoff(self, changed_by, transaction, sc_id, dryrun, columns):
-        # - If the User scheduling a change only holds one of the required Roles, record a signoff with it.
-        # - If the User scheduling a change holds more than one of the required Roles, we cannot a Signoff, because
-        #   we don't know which Role we'd want to signoff with. The user will need to signoff
+        # - If the User scheduling a change only holds one of the required Roles/Permissions, record a signoff with it.
+        # - If the User scheduling a change holds more than one of the required Roles/Permissions, we cannot a Signoff, because
+        #   we don't know which Role/Permission we'd want to signoff with. The user will need to signoff
         #   manually in these cases.
         user_roles = self.db.getUserRoles(username=changed_by, transaction=transaction)
+        user_permissions = self.db.getUserPermissions(username=changed_by, retrieving_as=changed_by, transaction=transaction)
         if len(user_roles):
             required_roles = set()
             required_signoffs = self.baseTable.getPotentialRequiredSignoffs([columns], transaction=transaction)
             if required_signoffs:
-                required_roles.update([rs["role"] for rs in [obj for v in required_signoffs.values() for obj in v]])
+                required_roles.update([rs["role"] for rs in [obj for v in required_signoffs.values() for obj in v] if "role" in rs])
             possible_signoffs = list(filter(lambda role: role["role"] in required_roles, user_roles))
             if len(possible_signoffs) == 1:
                 self.signoffs.insert(changed_by=changed_by, transaction=transaction, dryrun=dryrun, sc_id=sc_id, role=possible_signoffs[0].get("role"))
+        elif len(user_permissions):
+            required_permissions = set()
+            required_signoffs = self.baseTable.getPotentialRequiredSignoffs([columns], transaction=transaction)
+            if required_signoffs:
+                required_roles.update([rs["permission"] for rs in [obj for v in required_signoffs.values() for obj in v] if "permission" in rs])
+            possible_signoffs = list(filter(lambda permission: permission["permission"] in required_permissions, user_permissions))
+            if len(possible_signoffs) == 1:
+                self.signoffs.insert(changed_by=changed_by, transaction=transaction, dryrun=dryrun, sc_id=sc_id, role=possible_signoffs[0].get("permission"))
 
     def select(self, where=None, transaction=None, **kwargs):
         ret = []
@@ -3042,6 +3051,9 @@ class AUSDatabase(object):
 
     def hasPermission(self, *args, **kwargs):
         return self.permissions.hasPermission(*args, **kwargs)
+
+    def getUserPermissions(self, *args, **kwargs):
+        return self.permissions.getUserPermissions(*args, **kwargs)
 
     def hasRole(self, *args, **kwargs):
         return self.permissions.hasRole(*args, **kwargs)
--- a/src/auslib/db.py
+++ b/src/auslib/db.py
@@ -2629,14 +2629,8 @@ class Permissions(AUSTable):
         for row in affected_rows:
             if not row:
                 continue
-            # XXX: This kindof sucks because it means that we don't have great control
-            # over the signoffs required permissions that don't specify products, or
-            # don't support them.
-            if "products" in self.allPermissions[row["permission"]] and row.get("options") and row["options"].get("products"):
-                for product in row["options"]["products"]:
-                    potential_required_signoffs["rs"].extend(self.db.permissionsRequiredSignoffs.select(where={"product": product}, transaction=transaction))
             else:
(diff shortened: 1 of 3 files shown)


gabrielBusta:
```diff
diff --git a/tests/test_db.py b/tests/test_db.py
index b8477ffc..f520bd46 100644
--- a/tests/test_db.py
+++ b/tests/test_db.py
@@ -5362,7 +5362,13 @@ class TestPermissions(unittest.TestCase, MemoryDatabaseMixin):
         self.assertEqual(set(history_columns), set(expected))
 
     def testGrantPermissions(self):
-        self.permissions.insert("bob", username="cathy", permission="release", options=dict(products=["SeaMonkey"]))
+        self.permissions.insert(
+            "bob",
+            username="cathy",
+            signoffs=[{"sc_id": 1, "username": "bill", "role": "admin"}, {"sc_id": 1, "username": "zawadi", "role": "admin"}],
+            permission="release",
+            options=dict(products=["SeaMonkey"]),
+        )
         query = self.permissions.t.select().where(self.permissions.username == "cathy")
         query = query.where(self.permissions.permission == "release")
         self.assertEqual(query.execute().fetchall(), [("release", "cathy", dict(products=["SeaMonkey"]), 1)])

```

gabrielBusta:
To debug the tests, you can run [this script](https://github.com/gabrielBusta/misc-public/blob/16cd73fda5777141314c373537c25216cfcdc6bb/create-test-container.sh) in _the root of your clone_ of `balrog` to create a container using the CI's test image. Then, inside the container you can [run the tests](https://github.com/gabrielBusta/misc-public/blob/16cd73fda5777141314c373537c25216cfcdc6bb/in-test-container.sh) and use the builtin Python debugger ([here's a nice tutorial on it](https://realpython.com/python-debugging-pdb/).) To set breakpoints in the `tests` module you need to use [`pytest.set_trace` (instead of the builtin method.)](https://stackoverflow.com/a/3254078) You can also raise an exception. For example:

```diff
diff --git a/tests/test_db.py b/tests/test_db.py
index b8477ffc..84e98564 100644
--- a/tests/test_db.py
+++ b/tests/test_db.py
@@ -5362,6 +5362,7 @@ class TestPermissions(unittest.TestCase, MemoryDatabaseMixin):
         self.assertEqual(set(history_columns), set(expected))
 
     def testGrantPermissions(self):
+        raise Exception("stop here")
         self.permissions.insert("bob", username="cathy", permission="release", options=dict(products=["SeaMonkey"]))
         query = self.permissions.t.select().where(self.permissions.username == "cathy")
         query = query.where(self.permissions.permission == "release")
```

gabrielBusta:
```diff
diff --git a/src/auslib/db.py b/src/auslib/db.py
index 3861d5c6..91a46b1f 100644
--- a/src/auslib/db.py
+++ b/src/auslib/db.py
@@ -2649,7 +2649,7 @@ class Permissions(AUSTable):
             if not row:
                 continue
             else:
-                potential_required_signoffs["rs"].extend([dict(permission="admin", signoffs_required=2)])
+                potential_required_signoffs["rs"].extend(self.db.permissionsRequiredSignoffs)
         return potential_required_signoffs
 
     def assertPermissionExists(self, permission):
@@ -3016,6 +3016,7 @@ class AUSDatabase(object):
             self.setDburi(dburi, mysql_traditional_mode, releases_history_buckets, releases_history_class, async_releases_history_class)
         self.log = logging.getLogger(self.__class__.__name__)
         self.systemAccounts = []
+        self.permissionsRequiredSignoffs = []
 
     def setDburi(
         self,
@@ -3133,9 +3134,8 @@ class AUSDatabase(object):
     def productRequiredSignoffs(self):
         return self.productRequiredSignoffsTable
 
-    @property
-    def permissionsRequiredSignoffs(self):
-        return self.permissionsRequiredSignoffsTable
+    def setPermissionsRequiredSignoffs(self, permissionsRequiredSignoffs):
+        self.permissionsRequiredSignoffs = permissionsRequiredSignoffs
 
     @property
     def dockerflow(self):
diff --git a/uwsgi/admin.wsgi b/uwsgi/admin.wsgi
index 2daed094..314e23b0 100644
--- a/uwsgi/admin.wsgi
+++ b/uwsgi/admin.wsgi
@@ -46,6 +46,10 @@ if STAGING or LOCALDEV:
         }
     )
 
+if LOCALDEV:
+    PERMISSIONS_REQUIRED_SIGNOFFS = [{"permission": "admin", "signoffs_required": 1}]
+else:
+    PERMISSIONS_REQUIRED_SIGNOFFS = [{"permission": "admin", "signoffs_required": 2}]
 
 # Logging needs to be set-up before importing the application to make sure that
 # logging done from other modules uses our Logger.
@@ -148,6 +152,7 @@ else:
 dbo.setDb(os.environ["DBURI"], buckets)
 dbo.setSystemAccounts(SYSTEM_ACCOUNTS)
 dbo.setDomainAllowlist(DOMAIN_ALLOWLIST)
+dbo.setPermissionsRequiredSignoffs(PERMISSIONS_REQUIRED_SIGNOFFS)
 application.config["ALLOWLISTED_DOMAINS"] = DOMAIN_ALLOWLIST
 application.config["PAGE_TITLE"] = "Balrog Administration"
 application.config["SECRET_KEY"] = os.environ["SECRET_KEY"]
```

src/auslib/db.py
```diff
@@ -1696,14 +1711,20 @@ def __init__(self, db, metadata, dialect, baseName):
         super(SignoffsTable, self).__init__(db, dialect, versioned=False, historyClass=HistoryTable)
 
     def insert(self, changed_by=None, transaction=None, dryrun=False, **columns):
-        if "sc_id" not in columns or "role" not in columns:
+        if "sc_id" not in columns and "role" not in columns:
```
gabrielBusta, line 1714: We do we need to update `or` to `and` here?

src/auslib/db.py
```diff
@@ -168,7 +168,8 @@ def verify_signoffs(potential_required_signoffs, signoffs):
     for signoff in signoffs:
         signoffs_given[signoff["role"]] += 1
     for rs in potential_required_signoffs:
-        required_signoffs[rs["role"]] = max(required_signoffs.get(rs["role"], 0), rs["signoffs_required"])
+        signoff_verifier = "role" if "role" in rs else "permission"
```
bhearsum, line 171: Unless this function needs to handle verifying permission _and_ non-permission signoffs at the same time, I think it would be better to have this passed as an argument.

It may also be worth considering moving this into a method on the `RequiredSignoffsTable`, and having `ProductRequiredSignoffsTable` and `PermissionsRequiredSignoffsTable` wrap it to set `signoff_verifier` correctly, as we already have those two classes that specialize signoffs along the exact lines we're doing here.

src/auslib/db.py
```diff
@@ -168,7 +168,8 @@ def verify_signoffs(potential_required_signoffs, signoffs):
...
+        required_signoffs[rs[signoff_verifier]] = max(required_signoffs.get(rs[signoff_verifier], 0), rs["signoffs_required"])
     for role, signoffs_required in required_signoffs.items():
         if signoffs_given[role] < signoffs_required:
             raise SignoffRequiredError("Not enough signoffs for role '{}'".format(role))
```
bhearsum, line 175: `role` is not always accurate in this error message anymore. (Passing `signoff_verifier` into the function would make it easy to deal with this.)

bhearsum (commented):
Just to make sure I understand: this patch is adding the _general_ ability to require signoffs based on permissions (even though we're only specifying one `admin` signoff for permission changes at the moment).

Assuming that's right, does this mean that we could theoretically set up required signoffs from roles _and_ permissions in the same place?

src/auslib/db.py
michellemounde: We do not need to update, that was an error on my part

src/auslib/db.py
michellemounde: Not sure I understand what you mean by wrap it.

 Does this mean setting the signoff_verifier in the `ProductRequiredSignoffsTable` and `PermissionsRequiredSignoffsTable` and then passing the signoff_verifier to `verify_signoffs`?

Issue #3094, opened 2024-02-26
Switch from WSGI to ASGI to update to connexion 3.0+


gabrielbusta:
To close #3046 we should switch to ASGI as per [the connexion docs](https://connexion.readthedocs.io/en/latest/v3.html#running-the-application).

> You now MUST run the Connexion application instead of the underlying Flask application.
You should use an ASGI server instead of a WSGI server.

PR #3095, opened 2024-02-26
Switch from WSGI to ASGI 

gabrielbusta:
#3094


bhearsum:
Looks like https://github.com/mozilla-releng/balrog/pull/3095 is a more recent effort to do this.

PR #3103, opened 2024-03-20
Configure webpack production build to avoid using devtools that rely on eval

gabrielbusta:
Looks like Neutrino was doing this for us.
This should take care of the eval calls in the transpiled production code.

### DIFF
--- a/ui/webpack.config.js
+++ b/ui/webpack.config.js
@@ -17,241 +17,243 @@ const eslintConfig = {
   baseConfig,
 };
 
-module.exports = {
-  mode: 'development',
-  devtool: 'cheap-module-eval-source-map',
-  target: 'web',
-  context: __dirname,
-  stats: {
-    children: false,
-    entrypoints: false,
-    modules: false,
-  },
-  node: {
-    Buffer: false,
-    fs: 'empty',
-    tls: 'empty',
-  },
-  output: {
-    path: `${__dirname}/build`,
-    publicPath: '/',
-    filename: 'assets/[name].js',
-    globalObject: 'this',
-  },
-  resolve: {
-    alias: {
-      'react-native': 'react-native-web',
-      'react-dom': '@hot-loader/react-dom',
-    },
-    extensions: [
-      '.web.jsx',
-      '.web.js',
-      '.wasm',
-      '.mjs',
-      '.jsx',
-      '.js',
-      '.json',
-    ],
-  },
-  devServer: {
-    port,
-    hot: true,
-    historyApiFallback: {
-      disableDotRule: true,
-    },
-    overlay: true,
+module.exports = env => {
+  return {
+    mode: env,
+    devtool: env === 'production' ? false : 'cheap-module-eval-source-map',
+    target: 'web',
+    context: __dirname,
     stats: {
-      all: false,
-      errors: true,
-      timings: true,
-      warnings: true,
+      children: false,
+      entrypoints: false,
+      modules: false,
     },
-    host: process.env.HOST || DEFAULT_HOST,
-    https: true,
-    headers: {
-      'Content-Security-Policy':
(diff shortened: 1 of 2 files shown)


ui/webpack.config.js
```diff
@@ -17,241 +17,243 @@ const eslintConfig = {
...
-    globalObject: 'this',
-  },
-  resolve: {
-    alias: {
-      'react-native': 'react-native-web',
-      'react-dom': '@hot-loader/react-dom',
+module.exports = env => {
+  return {
+    mode: env,
+    devtool: env === 'production' ? false : 'cheap-module-eval-source-map',
```
gabrielBusta, line 23: Don't use `'cheap-module-eval-source-map'` in production builds.

ui/webpack.config.js
```diff
@@ -17,241 +17,243 @@ const eslintConfig = {
...
-    path: `${__dirname}/build`,
-    publicPath: '/',
-    filename: 'assets/[name].js',
-    globalObject: 'this',
-  },
-  resolve: {
-    alias: {
-      'react-native': 'react-native-web',
-      'react-dom': '@hot-loader/react-dom',
+module.exports = env => {
```
gabrielBusta, line 20: There's a lot of noise around here because the indentation changed.

bhearsum (commented):
I don't really understand all of the changes to `webpack.config.js` -- is this a generated file?

ui/webpack.config.js
```diff
@@ -17,241 +17,243 @@ const eslintConfig = {
...
-                        'last 2 Edge versions',
-                        'last 2 Opera versions',
-                        'last 2 Safari versions',
-                        'last 2 iOS versions',
-                      ],
+      host: process.env.HOST || DEFAULT_HOST,
+      https: true,
+      headers: {
+        'Content-Security-Policy':
+          "default-src 'none'; script-src 'self' 'unsafe-eval'; img-src 'self' https://*.gravatar.com https://*.githubusercontent.com https://i1.wp.com; style-src 'self' 'unsafe-inline' https://fonts.googleapis.com; font-src https://fonts.gstatic.com; connect-src https://localhost:8010 'self' https://balrog-localdev.auth0.com https://www.googleapis.com/; frame-src https://balrog-localdev.auth0.com; frame-ancestors 'none'; base-uri 'none'; form-action 'none'",
```
gabrielBusta, line 74: I see `'unsafe-eval'` here. I wonder why this is an issue now. It is what Neutrino was generating...

ui/webpack.config.js
```diff
@@ -17,241 +17,243 @@ const eslintConfig = {
...
+                          'last 2 iOS versions',
+                        ],
+                      },
                     },
-                  },
+                  ],
+                  [
+                    '@babel/preset-react',
+                    {
+                      development: env === 'development',
```
gabrielBusta, line 128: Don't set `development` to true in production builds.

ui/webpack.config.js
```diff
@@ -17,241 +17,243 @@ const eslintConfig = {
...
                 ],
-                [
-                  '@babel/preset-react',
-                  {
-                    development: true,
-                    useSpread: true,
-                  },
+                plugins: [
+                  '@babel/plugin-syntax-dynamic-import',
+                  ...(env === 'development' ? ['react-hot-loader/babel'] : []),
```
gabrielBusta, line 135: Don't use `react-hot-loader/babel` in production builds.

bhearsum:
> I don't really understand all of the changes to `webpack.config.js` -- is this a generated file?

Oh - nevermind. Hiding whitespace gets rid of most of it.

ui/webpack.config.js
bhearsum: This is probably worth a comment in the code explaining why not.

ui/webpack.config.js
bhearsum: Same thing here.

gabrielBusta:
Yes. The reason for this is that we are now reading the env and configuring the build based on that. Because the export is now a function instead of a plain object the indentation changed and it looks like more changes than it is.

bhearsum (approved):

PR #3110, opened 2024-04-17
Bug 1882147 - Support legacy keys for signing json responses

hneiva:


hneiva:
I felt like `semver` was the right option for calculating version numbers, but I'm not 100% it will cover all our cases.
I'm happy to re-implement with a different method if it makes more sense.

JohanLorenzo:
I think using `semver` makes sense. Not blocking: I'd recommend moving the `semver` call to https://github.com/mozilla-releng/mozilla-version/. This way, `mozilla-version` remains the one source of truth for parsing versions of Mozilla's products. That'll prevent any duplication next time we have to deal with a Mozilla VPN version number.

src/auslib/web/public/json.py
```diff
@@ -1,11 +1,36 @@
...
 
 from auslib.AUS import FORCE_FALLBACK_MAPPING, FORCE_MAIN_MAPPING
 from auslib.web.public.helpers import AUS, get_aus_metadata_headers, get_content_signature_headers, with_transaction
 
+# Map of products and version.
+# Uses legacy key if requested version is < value set here.
+LEGACY_PRODUCT_VERSION_MAPPING = {"FirefoxVPN": "2.22.0"}
+
+
+def check_for_legacy_key(parameters):
```
bhearsum, line 15: a couple of nits here:
- `use_legacy_key` would be a better name (`check` sounds like it should return nothing, to me)
- you're only using `product` and `version` from parameters - just pass those in rather than everything (it makes it more clear what using a legacy key or not depends on)

bhearsum (changes requested):
This looks good overall - I believe this change alone has the necessary functionality in it.

I'd like to see the `cloudops-infra` change as well before approving this, to see how everything will interact, and I'd really like to see some tests for this - ideally various some tests involved requests for updates (https://github.com/mozilla-releng/balrog/blob/main/tests/web/test_json.py and https://github.com/mozilla-releng/balrog/blob/main/tests/web/test_client.py).

There's some examples of how to mock out the autograph bits:
- https://github.com/mozilla-releng/balrog/blob/main/tests/web/test_client.py#L911
- https://github.com/mozilla-releng/balrog/blob/main/tests/web/test_json.py#L18

You'll probably need slightly more complex mocks that return a different x5u based on what's in the args.

hneiva:
Cloudops-infra portion is merged in https://github.com/mozilla-services/cloudops-infra/pull/5561

maintenance/pin.sh
```diff
@@ -6,4 +6,4 @@
 set -e
 set -x
 
-docker run --rm -t -v $PWD:/src -w /src python:3.11 maintenance/pin-helper.sh $@
+docker run --platform=linux/x86_64 --rm -t -v $PWD:/src -w /src python:3.11 maintenance/pin-helper.sh $@
```
bhearsum, line 9: This is necessary for M1 (or other non-x86_64 machines), I guess?

tests/web/test_helpers.py
```diff
@@ -0,0 +1,42 @@
+from unittest.mock import MagicMock
+
+import auslib.web.public.helpers
+import auslib.web.public.json
+
+
+def test_get_content_signature_headers(monkeypatch):
+    mockapp = MagicMock()
+    mockapp.config = {
+        "AUTOGRAPH_product_URL": "AUTOGRAPH_product_URL",
```
bhearsum, line 10: Just for readability, can you change the values here to not match the keys, please? I spent a good few minutes wondering why `mock_sign_hash` was looking at keys before realizing it was actually receiving the values :laughing:

bhearsum (approved):
Please change the mocked values as noted below, but I'm happy with this once that's done. Thank you!

maintenance/pin.sh
hneiva: That's right. Just to make sure we're always compiling with the same platform.

tests/web/test_helpers.py
hneiva: Oops :) That was indeed confusing. Fixed!

PR #3111, opened 2024-04-23
chore: update dependencies

ahal:


bhearsum (approved):
mozilla-version major upgrade is just dropping python 3.7

PR #3113, opened 2024-05-07
chore: update dependencies

ahal:


ahal:
The major bumps are `sentry-sdk` and `coveralls`.

For the former I scanned through https://docs.sentry.io/platforms/python/migration/1.x-to-2.x#scope-pushing and grepped for all the backwards incompatible things, but didn't find anything. So I think we'll be ok.

I couldn't find a migration guide for coveralls, but I think we can deal with any fallout there after the fact.

jcristau (approved):

PR #3116, opened 2024-05-21
chore: update dependencies

gbrownmozilla:


bhearsum (approved):

gbrownmozilla:
There's a major bump of setuptools with several deprecation removals, but they don't look applicable to us.

PR #3129, opened 2024-07-30
chore: bump python dependencies

bhearsum:


ahal (approved):

bhearsum:
cryptography [is a major bump](https://cryptography.io/en/latest/changelog/#v43-0-0), but it looks harmless:
> BACKWARDS INCOMPATIBLE: Support for OpenSSL less than 1.1.1e has been removed. Users on older version of OpenSSL will need to upgrade.
> BACKWARDS INCOMPATIBLE: Dropped support for LibreSSL < 3.8.

I haven't confirmed, but I highly doubt we're using either of these, and CI would almost certainly fail if we were.

PR #3208, opened 2024-10-17
chore(deps): upgrade dependencies

ahal:


bhearsum (approved):

ahal:
Changelog for markupsafe major bump:
https://markupsafe.palletsprojects.com/en/3.0.x/changes/#version-3-0-0

Though it's not a direct dependency, so unless we want to audit the chain of packages from our directs to make sure they all did their due diligence, I don't think there's much for us to test here.

PR #3225, opened 2024-10-31
chore: bump python dependencies

bhearsum:


bhearsum:
The only major bump is `pytest-cov`.

ahal (approved):

PR #3252, opened 2024-11-26
Bug 1933266 - Update domain allowlist for Widevine

aosmond:


uwsgi/admin.wsgi
```diff
@@ -34,6 +34,10 @@ DOMAIN_ALLOWLIST = {
         "Widevine",
         "Widevine-L1",
     ),
+    "www.google.com": (
```
bhearsum, line 37: This is a bit of a gaping hole in that Balrog would allow anything accessible through https://www.google.com to be served as a Widevine plugin... Is there any we can get a more scoped down domain for the mirrors? Or would that just end up with the same problem of things getting blocked?

aosmond: I could see how that might get abused, but part of the point is to use a domain that is unlikely to be blocked. The URL does have a set form though, perhaps I could extend the allowlisting to support partial URLs, e.g. https://www.google.com/dl/release2/chrome_component

aosmond: This seems to be the only place that uses the result directly, so I could change the format a bit to provide both domains and URLs:
https://github.com/mozilla-releng/balrog/blob/b2d6bd30a05cc430e6067c7375797234267ec3ca/src/auslib/AUS.py#L38

If it doesn't match a domain, it can check to see if the URL matches one of the starting fragments instead. Change DOMAIN_ALLOWLIST to something like
```
DOMAIN_ALLOWLIST = {
  "domains": {
    // ... existing list
  },
  "urlFragments": {
    "https://www.google.com/dl/release2/chrome_component": (
        "Widevine",
        "Widevine-L1",
    )
  },
}
```

aosmond: I did a more advanced version of the latter and updated the PR. Now it can take in an optional fullmatch regex for the path to ensure more generic domains don't get abused. The regex for www.google.com should only allow us to access files from the chrome component update service.

src/auslib/AUS.py
```diff
@@ -34,14 +35,29 @@ def isSpecialURL(url, specialForceHosts):
 def isForbiddenUrl(url, product, allowlistedDomains):
     if allowlistedDomains is None:
         allowlistedDomains = []
-    domain = urlparse(url)[1]
+    parsedUrl = urlparse(url)
+    domain = parsedUrl[1]
```
bhearsum, line 39: I realize that you're inheriting this, but it would be a good time to update this to use [the named attributes that urlparse returns](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlparse) (it looks like you want `.netloc` and `.path` specifically).

src/auslib/AUS.py
```diff
@@ -34,14 +35,29 @@ def isSpecialURL(url, specialForceHosts):
...
         return True
-    if product not in allowlistedDomains[domain]:
+    allowlistedDomain = allowlistedDomains[domain]
+    if isinstance(allowlistedDomain, tuple):
+        if product in allowlistedDomain:
+            return False
         logging.warning("Forbidden domain for product %s: %s", product, domain)
-        return True
-    return False
+    elif isinstance(allowlistedDomain, object):
```
bhearsum, line 48: You want `isinstance(allowlistedDomain, dict)` here. (`object` is the base class of the class hierarchy, so this happens to work, but if there was another `elif` block afterwards there'd be a problem.)

bhearsum (changes requested):
Thank you so much for implementing this! This needs a small revision for the `isinstance` bug noted below. The other one is up to you.

src/auslib/AUS.py
aosmond: Thanks! Corrected.

src/auslib/AUS.py
aosmond: Sure, I noticed that from the docs, but chose to follow the convention set in the code. Fixed!

bhearsum (approved):

PR #3267, opened 2024-12-12
ci: use taskgraph 12.1.0

jcristau:


bhearsum (approved):
Thanks for doing this!

PR #3281, opened 2025-01-23
fix(docker): Switch to `pip install` and remove `repoze-lru` workaround

hneiva:
- Replaces `python setup.py install` with `pip install` in the Docker image
- Removes the hard-coded `repoze.lru` (`-` to `.`) workaround in `pin-helper`

The issue arose because `pip-compile` switches `.` to `-` when following [PEP-0503](https://peps.python.org/pep-0503/#normalized-names) package name normalization rules. While `pip` seamlessly handles the aliased name (installing the package with `.`), `setuptools` does not manage the mismatch well. By moving to `pip install`, we avoid the naming conflict entirely and prevent future mishaps.


bhearsum (approved):
Excellent sleuthing! Thank you for fixing the root cause!

PR #3381, opened 2025-05-01
fix: pin to working versions of `pip` and `pip-compile-multi`; repin dependencies

bhearsum:
There are a few major bumps here (`mozilla-version`, `packaging`, `setuptools`). I'm not concerned about them as long as tests pass.


hneiva (approved):
Maybe we should file a bug to migrate pinning to `uv`?

LGTM for now

bhearsum:
> Maybe we should file a bug to migrate pinning to `uv`?
> 
> LGTM for now

https://github.com/mozilla-releng/balrog/issues/3382

PR #3435, opened 2025-06-30
web: rename app to flask_app

jcristau:
We have a Connexion app which wraps a Flask app, so having something called `app` can get confusing.


jcristau:
> Any reason not to keep `flask_app` in tests to be consistent?

 Not really, other than I split this out of the port to connexion 3 and that'll have a bunch more changes to the tests so it was simpler this way.

PR #3437, opened 2025-06-30
public: be more careful when cleaning up avast query param

jcristau:


src/auslib/web/public/client.py
```diff
@@ -92,9 +92,10 @@ def getCleanQueryFromURL(url):
 
         query["force"] = force_split[0]
 
-        avast_parameter = force_split[1]
-        avast_split = avast_parameter.split("=")
-        query[avast_split[0]] = int(avast_split[1])
+        if len(force_split) == 2:
+            avast_parameter = force_split[1]
+            avast_split = avast_parameter.split("=")
```
Eijebong, line 97: Since we're here, wouldn't `?force=1?avast=` be an issue here? `?force=1?avast=foo` also?

I'm not sure if it's worth the effort to fix this but 🤷

Eijebong (approved):
Looks fine to me

PR #3438, opened 2025-06-30
tests: fix disabling of flask error handlers

jcristau:
As far as I can tell this has been broken since
https://github.com/pallets/flask/commit/fd62210f583e90764be6e8d9f295f37f33a65e48


bhearsum:
> I didn't actually add a test here :) It's a good idea though.

Oh, heh. For some reason I read the `test_json.py` hunk as a test :facepalm:

PR #3439, opened 2025-06-30
fix: use flask_app in wsgi files

bhearsum:
Follow-up to #3435 where `app` was renamed to `flask_app`.


jcristau (approved):
Oops, sorry about that.

PR #3440, opened 2025-06-30
fix: reduce the number of requests needed to populate the the Users UI view

bhearsum:
Over in https://github.com/mozilla-releng/balrog/issues/1043#issuecomment-3020506402 I realized that the rate limiting we see in this part of the UI is actually caused by refetching info for the same user over and over again, which happens because we make one request to the backend for every user in the system (plus a couple more). We can reduce this to 1 or 2 total for the page by returning more complete data from the backend, which should sidestep the rate limit.


src/auslib/web/admin/views/permissions.py
```diff
@@ -28,8 +28,11 @@
 ]
 
 
-def get_users():
+@requirelogin
+def get_users(changed_by):
```
jcristau, line 32: Looking at this function, and the swagger spec for its response, they're clearly disagreeing (even before this).
This might be worth updating (and noting the new login requirement) while we're making changes?

PR #3449, opened 2025-07-04
Don't emit warnings more than once when running tests

Eijebong:
Locally this makes tests go from 45s to 38s
We don't need to see every warning more than once to know why it's there...


Eijebong:
Apparently I got lucky with **every** single test I did with this because the testing times has crazy variability. And the number of warnings emitted has nothing to do with.

PR #3451, opened 2025-07-07
tests: set app.testing instead of poking through flask internals

jcristau:
We can tell flask to let exceptions propagate by setting the testing flag.  So do that consistently, and for the one test class that wants to test the error handlers, we already explicitly set PROPAGATE_EXCEPTIONS to False.

### DIFF
--- a/tests/admin/views/conftest.py
+++ b/tests/admin/views/conftest.py
@@ -24,6 +24,7 @@ def api():
 
     connexion_app = create_app()
     app = connexion_app.app
+    app.testing = True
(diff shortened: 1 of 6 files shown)


bhearsum (approved):
Thanks for the continued cleanup/improvement here.

> for the one test class that wants to test the error handlers, explicity set PROPAGATE_EXCEPTIONS to False.

Is this referring to the existing https://github.com/mozilla-releng/balrog/blob/a83ccf49c69ec83fcf62534127faada1af7f63b4/tests/web/test_client.py#L2189, or is there somewhere else you intended to set it?

PR #3453, opened 2025-07-07
docker-compose: reduce balrogdb health check interval

jcristau:
Check for mysql status more frequently so we don't have to wait 30s to start the app after the db is up.


Eijebong (approved):
I had the exact same patch locally at some point and forgot about it

PR #3458, opened 2025-07-09
feat: use uv for dependency resolution in pin-helper.sh

bhearsum:
pip-compile-multi gained support for this a few weeks ago. It looks like there's no downsides to using it?


Eijebong (approved):
I'd like to switch the whole project to uv at some point, but in the meantime, this avoids waiting 10mn every time we want to update deps so that's an easy win, thanks

PR #3459, opened 2025-07-09
feat: add request timing and response code stats for the public app

bhearsum:
The request timing stats are inspired by [what Autograph does](https://github.com/search?q=repo%3Amozilla-services%2Fautograph%20signerRequestsTiming&type=code), and allows us to keep track of the most expensive part(s) of serving requests. We may want to make this more granular in the future if we're looking at optimizations -- but it's a starting point.

Also included here is an addition to the development environment, adding a container that runs statsd and graphite. This is not strictly necessary (and I could be convinced to remove it), but it was useful to have the UI available when validating the less trivial Timer metrics.


src/auslib/web/public/client.py
```diff
@@ -159,81 +163,88 @@ def get_update_blob(transaction, **url):
...
-
-        # Appending Header
-        # In case of superblob Extracting Header form parent release
-        xml.append(release.getInnerHeaderXML(query, update_type, app.config["ALLOWLISTED_DOMAINS"], app.config["SPECIAL_FORCE_HOSTS"]))
-        for response_blob in response_blobs:
-            xml.extend(
-                response_blob["response_release"].getInnerXML(
-                    response_blob["product_query"], response_blob["response_update_type"], app.config["ALLOWLISTED_DOMAINS"], app.config["SPECIAL_FORCE_HOSTS"]
+    with statsd.timer("client.make_response"):
+        if release:
```
Eijebong, line 206: That broke my brain a little.

We're going from:

```py
if release:
    # block a
    # block b
else:
    # ...
```

To:

```py
if release:
    # block a
with statsd(...):
    if release:
        # block b
    else:
        # ...
```

Which looks correct to me despite it being very confusing to follow

src/auslib/web/public/client.py
```diff
@@ -159,81 +163,88 @@ def get_update_blob(transaction, **url):
...
-    response.mimetype = "text/xml"
+            # Appending Footer
+            # In case of superblob Extracting Header form parent release
+            xml.append(release.getInnerFooterXML(query, update_type, app.config["ALLOWLISTED_DOMAINS"], app.config["SPECIAL_FORCE_HOSTS"]))
+            xml.append(release.getFooterXML())
+            # ensure valid xml by using the right entity for ampersand
+            xml = re.sub("&(?!amp;)", "&amp;", "\n".join(xml))
+
+            # Bug 1517743 - remove newlines and 4 space indents
+            if squash_response:
```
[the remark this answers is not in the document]
bhearsum: Ah, yes, it is technically a behaviour change...just not one that causes problems (because this workaround only matters for a particular Firefox version where updates are present.

In any case, taking a behaviour change like this is probably not great...so I'll adjust it.

src/auslib/web/public/client.py
Eijebong: At least the diff without whitespaces shows that something changed there, and then I emptied the condition branches to make sense of it. I don't know if there's a better way of writing this 🤷

bhearsum: I've refactored this a bit in a separate commit, which should improve the readability (both of the diff and the code in general). I'll push it shortly.

Eijebong (approved):

PR #3464, opened 2025-07-14
Update axios to 1.10.0

Eijebong:


bhearsum:
I assume you did some clicky-clicky in the UI to sanity check this?

PR #3477, opened 2025-07-21
fix: don't log non-GET requests nor dockerflow requests in the public app

bhearsum:
The latter in particular is pointless, and causes a lot of noise in the stats.

### DIFF
--- a/tests/web/test_dockerflow.py
+++ b/tests/web/test_dockerflow.py
@@ -4,7 +4,8 @@ from .test_client import ClientTestBase
 
 
 class TestDockerflowEndpoints(ClientTestBase):
-    def testVersion(self):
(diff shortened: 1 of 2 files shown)


tests/web/test_dockerflow.py
```diff
@@ -4,7 +4,8 @@
 
 
 class TestDockerflowEndpoints(ClientTestBase):
-    def testVersion(self):
+    @mock.patch("auslib.web.admin.base.statsd.incr")
```
jcristau, line 7: `s/admin/public/`?

jcristau (changes requested):

tests/web/test_dockerflow.py
bhearsum: Sorry..forgot to push.

PR #3480, opened 2025-07-23
Update webpack, related dependencies, node and bring config in line with shipit

Eijebong:
We had a few plugins in the dependencies that were not even used... This brings the webpack config in line with what shipit does with the following differences:

- We don't need the assert polyfill because we don't have a dependency on taskcluster-libs
- We don't need the process plugin because reasons (I think it's because we only ever use process.env in jsx templates instead of using it outside and for some reason that works by default)
- I had to disable the overlay for errors otherwise going to the releases page on dev was painful as it always shows some error because required_signoffs/read_only 404s for 3 products. That error is not new but the previous overlay config just never worked...

I'd recommend using a diff tool to diff the shipit/balrog configs to review this.

### DIFF
--- a/ui/webpack.config.js
+++ b/ui/webpack.config.js
@@ -1,10 +1,8 @@
 const webpack = require('webpack');
-
-const DEFAULT_HOST = 'localhost';
-const DEFAULT_PORT = 9000;
-const port = process.env.PORT || DEFAULT_PORT;
+const MiniCssExtractPlugin = require('mini-css-extract-plugin');
 const HtmlWebpackPlugin = require('html-webpack-plugin');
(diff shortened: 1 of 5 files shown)


ui/webpack.config.js
```diff
@@ -56,22 +51,22 @@ module.exports = env => {
...
+      fallback: {
+        Buffer: false,
+        fs: false,
+        tls: false,
+      },
     },
     devServer: {
+      host: process.env.HOST || DEFAULT_HOST,
       port,
-      hot: true,
```
bhearsum, line 62: Is hot reloading the default now?

bhearsum (approved):

PR #3481, opened 2025-07-23
Replace clipboard-copy with native javascript

Eijebong:
I don't think we need to support firefox < 63 or IE11... Pulling in a library for this is just unnecessary


ui/src/components/Dashboard/UserMenu.jsx
```diff
@@ -34,7 +33,7 @@ function UserMenu(props) {
   const handleCopyAccessToken = async () => {
     const accessToken = await auth0.getAccessTokenSilently();
 
-    copy(accessToken);
+    navigator.clipboard.writeText(accessToken);
```
jcristau, line 36: Is this missing `await`?

Eijebong: It was missing one before and it's still missing it and yet it works... Gotta love this language

jcristau (approved):

PR #3488, opened 2025-07-24
Version bump to 3.83

jcristau:


jcristau:
3.82 was never tagged on main but that number has been burned for staging-only releases.

PR #3509, opened 2025-08-08
Replace `is-absolute-url` with pure JS

Eijebong:
https://developer.mozilla.org/en-US/docs/Web/API/URL/canParse_static this does the exact same thing when not passed a second argument


bhearsum (approved):
What is the world coming to...next you'll be removing a dependency on leftpad?! :laughing:

PR #3515, opened 2025-08-08
Remove the engine limit in packages.json

Eijebong:
I'm tired of removing it every time I want to run yarn locally. IMO this doesn't provide anything as this isn't a library and the version that matters is the one that builds it in the docker container, and that isn't going to magically become incompatible


ahal (approved):
\o/

PR #3517, opened 2025-08-12
Replace react-virtualized by tanstack/react-virtual

Eijebong:
react-virtualized had some bug with anchors that prevented us from updating it. That bug has never been fixed, and the author seems to have moved onto a third rewrite of it since we last updated. If we were going to replace it by a rewrite, we might as well try something else, tanstack/react-virtual seems to be what people use these days and it seems to work very well with our stuff. The update wasn't exactly straightforward as tanstack forces you to handle a bit more yourself that react-virtualized ever did, but in the end it works as well as before and even before in some cases:

- the releases list used to have the gap triple in the middle for some reason, this does not happen anymore.
- We're now using a table for tables instead of the div soup that react-virtualized was using. While the visual output doesn't change, I feel a lot better about that...

### After / Before

image
image
image
image
image
image

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -14,8 +14,6 @@
     "lint:ci": "biome ci"
   },
   "comment": "IMPORTANT! date-fns is pinned to 1.3.x due to https://github.com/mui-org/material-ui-pickers/issues/1440",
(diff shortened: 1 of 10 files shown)


ahal (approved):
I assume you know this is pretty much just a rubberstamp. Thanks for doing this though!

PR #3521, opened 2025-08-13
Update codemirror and react to 17

Eijebong:
We have to switch from react-codemirror to uiw/react-codemirror for this because react-codemirror isn't maintained anymore and does not support react > 16. While I was touching this code I also fixed an issue that was bothering me where the code editor was overflowing the page even when it was empty. I also made it not overflow and scroll after it's filled up although I'd be ok with reverting that change if people think that the extra 4 lines you get by scrolling the page are important. Given that I doubt people use this much to manually edit releases, I didn't spend too much time on matching color themes with the previous one and just picked the material dark one which I would've expected to be the same (it's not...).

The slew of dependencies added is because codemirror split its package and I had to add explicit dependencies on them to make yarn shut up about peer dependencies not being declared (I **hate** this concept).

### Before / After

image
image


ahal (approved):
Thanks!

PR #3522, opened 2025-08-14
Remove the dependency on typescript

Eijebong:
This is a relic from the neutrino removal, we don't use typescript in here (we probably should but that's another topic and a project on its own...)


hneiva (approved):
I used to like hacking front-ends in typescript..
LGTM

PR #3529, opened 2025-08-20
Switch dependency management to uv

Eijebong:
The python 3.9 requirements on the workspace stems from balrogclient. Since a workspace can only declare one required python version and has to be the minimal one. However that only influences what the lockfile contains and the actual python version used is decided by the user (or in our case the version of python in our containers).
We can probably bump that version to 3.11 but I'd rather do that in a separate PR.


Dockerfile
```diff
@@ -1,41 +1,41 @@
-# WARNING: keep this file in sync with taskcluster/docker/balrog-backend/Dockerfile
+# WARNING: keep this file in sync with agent/Dockerfile
+ARG PYTHON_VERSION=3.13
+FROM ghcr.io/astral-sh/uv:python${PYTHON_VERSION}-bookworm-slim AS builder
```
ahal, line 3: Nice to see more multi-stage builds! We should probably document some best practices for our docker images at some point.

agent/tox.ini
```diff
@@ -1,30 +1,13 @@
...
-    PYTHON_VERSION=3.13
-    TOXENV=check,py313
-deps =
-usedevelop = false
-depends =
-skip_install = true
-commands =
-    docker build --build-arg PYTHON_VERSION={env:PYTHON_VERSION} -t balrog_agent-{envname}-test -f Dockerfile.test .
-    docker run --rm -v {toxinidir}:/app -v balrog_agent-{envname}-tox:/app/.tox -e TOXENV={env:TOXENV} balrog_agent-{envname}-test
+envlist = py313
```
ahal, line 2: Would love to stop using tox, especially if we're only testing against one Python version.. but that's definitely follow-up fodder.

maintenance/pin.sh
```diff
@@ -1,9 +1,8 @@
...
 #
-# Usage: maintenance/pin.sh [<extra-pip-compile-multi-arguments>]
+# Usage: maintenance/pin.sh [<extra-uv-arguments>]
 #
-set -e
-set -x
+set -ex
 
-docker run --platform=linux/x86_64 --rm -t -v $PWD:/src -w /src python:3.13 maintenance/pin-helper.sh $@
+uv lock -U $@
```
ahal, line 8: Personally I'd prefer we delete this file and document / get more comfortable using uv directly.. but I guess there's not really any harm in leaving it.

ahal (approved):
Nice, thank you very much!

maintenance/pin.sh
hneiva: +1 - we should just have an entry in README.md, there's already a maintenance section there:
>To update the python dependencies, use ./maintenance/pin.sh, do not run pip-compile-multi manually.

Eijebong: I did this to keep consistency but I'm more than fine with removing more stuff :D

PR #3532, opened 2025-08-20
docker: use python:3.13-slim-bookworm

jcristau:
The python:3.13-slim image is now based on debian 13 (trixie) instead of 12 (bookworm); go back to the older distro for now to prevent breakage.


Eijebong (approved):
SGTM

PR #3534, opened 2025-08-20
tox: ignore coveralls failures

jcristau:
Let's not block PRs on coveralls infra issues.


Eijebong (approved):
That works I guess yeah

PR #3536, opened 2025-08-21
Bump version to 3.88

Eijebong:


jcristau (approved):
boo at having to maintain this in 2 places now :/

Eijebong:
Well the second one is by doing `uv lock` but yeah... That's the issue of having a proper lockfile, and I don't think there's any way to escape that

PR #3539, opened 2025-08-21
Drop support for python 3.9 for balrogclient

Eijebong:
AFAIK this is only ever used in scriptworker-scripts which has a minimum python version of 3.11


pyproject.toml
```diff
@@ -7,7 +7,7 @@ authors = [
     { name = "Mozilla Release Engineering", email = "release+python@mozilla.com" }
 ]
 readme = "README.rst"
-requires-python = ">=3.9"
+requires-python = ">=3.11"
```
jcristau, line 10: Shouldn't this be in client/pyproject.toml?

jcristau: (and remove black's target-version there)

Eijebong: Oh I didn't think about checking if it was published or not... Will bump

jcristau, line 10: nit: Still not sure this change belongs in this PR

jcristau (approved):

jcristau (approved):

PR #3548, opened 2025-09-02
Pass the state correctly to the `Link` component

### LINKED ISSUE #3547: Navigating back from updating a rule doesn't put the filters back on
This is a regression, most likely from updating react-router stuff. I checked on 
09ae01e99f7a1fb9eaf7d06c5af7ddf6f8f80d34 and it worked there.

To repro:

- Go to rules, select a product channel
- Update a rule
- On save, it redirects to `https://localhost:9000/rules?product=Firefox&channel=esr#scId=6990` before, `https://localhost:9000/rules#scId=6990` after

Eijebong:
This fixes a regression from c3f865c7229f128fb99de73e66891f76d918ff4e where updating react-router silently broke keeping some state across navigation. The issue is that react router's API changed the state property from being nested within the `to` object to being directly on the routing components and the `to` component becoming a string. We didn't see this at all because of the `Link` wrapper that was doing some massaging on the `to` property when it wasn't a string, so we were just silently discarding the state parameter.

I removed said massaging to since we're passing strings to `to` everywhere now as we should.

Fixes #3547

### DIFF
--- a/ui/src/components/RuleCard/index.jsx
+++ b/ui/src/components/RuleCard/index.jsx
@@ -338,10 +338,8 @@ function RuleCard({
           action={
             !readOnly ? (
               <Link
-                to={{
-                  pathname: `/rules/${rule.rule_id}/revisions`,
-                  state: { rulesFilter },
-                }}
+                to={`/rules/${rule.rule_id}/revisions`}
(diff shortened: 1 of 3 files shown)


jcristau (approved):
Thanks!

PR #3555, opened 2025-09-24
chore: remove unused AUSConfig

bhearsum:
This seems like it's been unused for years, if not a decade. When the `.wsgi` files moved into their own directory in https://github.com/mozilla-releng/balrog/commit/0bffa8c08768c6402afc82bb199f5d200b1cae81, we dropped references to these configs. The old configs were then removed at some point later that I haven't bothered to track down.


jcristau (commented):
You'll want to drop the reference from docs/source/autodocs.rst

PR #3556, opened 2025-09-26
feat: add support for using redis for public app instead of in memory caches

bhearsum:
This will support fixing https://github.com/mozilla-releng/balrog/issues/2767; for now it just makes it possible to turn it on.

One thing I'm not entirely certain of is how exactly things will go when the redis cache expires. Because the LRU cache expirations are set up to match redis, it's possible that all active pods may go try to refresh themselves from the database at once, and cause a load spike. I'm hoping to test out this scenario in stage (either by generating fake load on it, or mirroring production traffic to it for a period of time) to get a better sense. If this ends up being a problem (either in stage testing, or when enabled in production) I expect that it can easily be worked around by adding some jitter to the in-memory cache to stagger the expirations. (I don't particularly want to design for this unless it proves to be an issue in practice though.)

Other detailed comments and explanations are in the commit messages & comments.


src/auslib/util/cache.py
```diff
@@ -98,3 +120,48 @@ def invalidate(self, name, key):
...
+        return f"{self._name}-{key}"
+
+    def get(self, key, default=None):
+        self.lookups += 1
+        if value := self._redis.get(self.fullkey(key)):
+            if value == default:
+                self.misses += 1
+            else:
+                self.hits += 1
+            return pickle.loads(value)
```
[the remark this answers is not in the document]
bhearsum: I did some basic local testing on pickling viability by making a single request before anything was cached, and then one request/second for a few minutes afterwards.

Results with the existing cache:
image

Results with the new RedisCache:
image

Clearly using pure Redis is a notable regression in response time (and presumably also CPU usage...but it's difficult to say that with any certainty outside of production).

With that in mind, I also hacked up a two layer cache, where we look in-memory, and fallback to Redis:
image

Unsurprisingly, that brings perf back inline. It seems like this should be fairly straightforward to clean up, so I'll pursue it further.

src/auslib/util/cache.py
```diff
@@ -98,3 +121,99 @@ def invalidate(self, name, key):
...
+        self.lookups += 1
+        value = self._lru_cache.get(key, default)
+        if value == default:
+            value = self._redis_cache.get(key, default)
+            # ensure the LRU cache timeout matches the one in redis
+            # this is important to ensure that when new pods spin up that
+            # they will not cache keys for longer than redis
+            # this ensures that we don't have to wait for multiple caches
+            # to expire to refresh data.
+            self._lru_cache.put(key, value, self._redis_cache.remaining_timeout(key))
```
jcristau, line 199: I guess there's a small window here where the key might be expired in redis but not the lru cache (since `time.time()` is called first in `remaining_timeout` and then in `ExpiringLRUCache.put`). We're also assuming clocks are synchronized.
(I think that's OK, just figured I'd call it out.)

bhearsum: That's a good point, yeah. In an ideal world I guess we'd want to be able to pass in an absolute timestamp for the timeout to avoid this, but `ExpiringLRUCache` doesn't support that.

Presumably these will be close enough in the vast majority of cases; I'll add a comment to this effect though.

bhearsum:
> Could we s/pickle/json/, or is that even worse?

I originally benchmarked this, and pickle was significantly faster. However, `orjson` is faster for dumping and loading...so we should switch to that (probably everywhere in Balrog):
```
~/tmp/2025-10-01 ❯ cat test.py                                                                           10s
import pickle
import orjson
import timeit
import sys

pickled = pickle.dumps(sys.argv[1])
jsoned = orjson.dumps(sys.argv[1])

def pickledump():
    pickle.dumps(sys.argv[1])

def pickleload():
    pickle.loads(pickled)

def jsondump():
    orjson.dumps(sys.argv[1])

def jsonload():
    orjson.loads(jsoned)

print("pickledump: ", end="")
print(timeit.timeit(stmt="pickledump()", setup="from __main__ import pickledump", number=10000000))
print("pickleload: ", end="")
print(timeit.timeit(stmt="pickleload()", setup="from __main__ import pickleload", number=10000000))
print("jsondump: ", end="")
print(timeit.timeit(stmt="jsondump()", setup="from __main__ import jsondump", number=10000000))
print("jsonload: ", end="")
print(timeit.timeit(stmt="jsonload()", setup="from __main__ import jsonload", number=10000000))
~/tmp/2025-10-01 ❯ python3 test.py ~/repos/balrog/tests/data/Firefox-100.0-build1.json 
pickledump: 3.5769623639998827
pickleload: 2.6630689870003152
jsondump: 1.392012985999827
jsonload: 1.8235570860001644
```

bhearsum:
> I've got the redis parts sorted out now...but while testing that I discovered that blobs are not loaded correctly from redis (they end up as dicts instead of Blob objects). I'm putting this back into draft state until I sort that out.

This was caused by switching to `orjson` for serialization. Blobs are dict subclasses, so they serialize as them. I decided to revert back to pickling because of this. There's no technical reason why we can't support `orjson`, but I don't think the wins are big enough to worry about it (most of the time, we'll be fetching already unpickled data from the in-memory cache anyways).

I did add a new test that will catch this issue should we ever try switching to `orjson` again in the future.

PR #3606, opened 2026-01-12
test: unpin hypothesis and bump version

ahal:


ahal:
The issue that forced the pin was fixed in:
https://github.com/HypothesisWorks/hypothesis/pull/4639

However, the maintainers pointed out that we're setting things up poorly:

> Thanks Andrew. To clarify, the issue was the json strategy passes lambda x: useful_list | useful_dict as the extension strategy to st.recursive, but this doesn't actually use x and therefore doesn't recurse more than one level.
>
> We started assuming this in the new min_leaves feature, and have now both relaxed this assumption so this code works again, and added a warning when we detect this class of miss-stated extend.
>
> I recommend adjusting this strategy to recurse on x regardless, as its test distribution is probably weaker than expected at the moment!

I'll attempt to also fix up the test.

jcristau:
Superseded by #3773 and #3775.  Closing.

PR #3747, opened 2026-04-08
chore: update uv.lock

jcristau:
Missed in the last version bump


Eijebong (approved):
How... `runner = uv-venv-lock-runner`, tests shouldn't even have run on that bump PR

PR #3750, opened 2026-04-10
Run tox with `uv run --locked`

Eijebong:
We're using the locked tox runner but calling tox via uv without `--locked` was making it completely useless since the outer command was updating the lockfile before it even reached tox. Use `--locked` on that one too.


Eijebong:
https://firefox-ci-tc.services.mozilla.com/tasks/E5G6yV0iQQGJxE-PQb2WTw/runs/0/logs/public/logs/live.log#L67

jcristau (approved):

PR #3755, opened 2026-04-16
Bug 1934514 - Assert autoincrement primary keys are not set explicitly on insert

jcristau:
Followup to #3754 to catch any such issues at the db layer even if they make it through the web layer.

### DIFF
--- a/src/auslib/db.py
+++ b/src/auslib/db.py
@@ -1092,7 +1092,7 @@ class ConditionsTable(AUSTable):
 
         self.enabled_condition_groups = {k: v for k, v in self.condition_groups.items() if k in conditions}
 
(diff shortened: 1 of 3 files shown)


bhearsum (approved):
lgtm. Theoretically I guess we could drop the checks from https://github.com/mozilla-releng/balrog/pull/3754, but keeping them will maintain better error messages? (This is more like a backstop to prevent similar bugs in new tables/endpoints?)

jcristau:
> lgtm. Theoretically I guess we could drop the checks from #3754, but keeping them will maintain better error messages? (This is more like a backstop to prevent similar bugs in new tables/endpoints?)

Yes exactly.

PR #3773, opened 2026-05-05
Remove hypothesis pin

jcristau:
The incompatibility was fixed in
https://github.com/HypothesisWorks/hypothesis/pull/4639

We should still fix the code on our side but don't have to block updates to hypothesis in the mean time.


jcristau:
Followup in #3775.

bhearsum (approved):

PR #3774, opened 2026-05-05
tests: consolidate pytest config in pyproject.toml

jcristau:
Having split config caused this warning:

> configfile: pyproject.toml (WARNING: ignoring pytest config in tox.ini!)


jcristau:
We may want to revisit some of the pytest config from tox.ini, that's been unused since #3283...

ahal (approved):

PR #3781, opened 2026-05-08
chore: update UI deploy action with proper values for prod and stage

bhearsum:
* Fill in the TBDs
* Temporarily change BALROG_ROOT_URL to an internal DNS name that points at the MozCloud prod
* Adjust history bucket names (including for stage, which I should've done prior to cutting that over to MozCloud)


hneiva (approved):
LGTM

PR #3782, opened 2026-05-08
docs: add ui deploy instructions to stage deploy

bhearsum:


ahal (approved):
nit: if you're following conventional commits, the prefix would be `docs:`

PR #3789, opened 2026-05-20
feat: support specifying a ref when building and deploying the ui

bhearsum:
This allows easier rollbacks (we don't need to revert the code to do one).


bhearsum:
Tested this in https://github.com/mozilla-releng/balrog/actions/runs/26179580180/job/77018657736

jcristau (approved):

PR #3792, opened 2026-05-28
ui: show wildcard channels in product/channel dropdown

jcristau:
Rules with wildcard channels (e.g. `nightly*`) were invisible in the product/channel filter dropdown, making it appear a product (e.g. Widevine) had no channel-specific rules. Strip the trailing `*` when building dropdown options so `esr*` is shown as `Widevine : esr`.


bhearsum (approved):
Thank you!

PR #3795, opened 2026-05-30
H1 Balrog redacted Taskcluster secret validation

MathCarv:
Controlled HackerOne validation PR.

This PR is intended to validate whether the public pull request Taskcluster role can access only redacted metadata for the configured secret scopes through taskclusterProxy.

Safety notes:
- does not print full secret values
- does not upload to third-party services
- does not mutate Mozilla services
- writes only redacted proof metadata to public/build/h1-balrog-redacted-secret-proof.json

This can be closed after validation.

### DIFF
--- a/scripts/h1_redacted_secret_probe.py
+++ b/scripts/h1_redacted_secret_probe.py
@@ -1,0 +1,167 @@
+#!/usr/bin/env python3
+import hashlib
+import json
+import os
+import sys
+from datetime import datetime, timezone
+from pathlib import Path
+from urllib.error import HTTPError, URLError
+from urllib.parse import quote
+from urllib.request import Request, urlopen
+
+
+DEFAULT_SECRET_NAMES = (
+    "repo:github.com/mozilla-releng/balrog:coveralls,"
+    "project/taskcluster/gecko/hgmointernal,"
+    "project/taskcluster/gecko/hgfingerprint"
+)
+DEFAULT_ARTIFACT_PATH = (
+    "/builds/worker/artifacts/build/h1-balrog-redacted-secret-proof.json"
+)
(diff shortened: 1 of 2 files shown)


scripts/h1_redacted_secret_probe.py
```diff
@@ -0,0 +1,167 @@
...
+from datetime import datetime, timezone
+from pathlib import Path
+from urllib.error import HTTPError, URLError
+from urllib.parse import quote
+from urllib.request import Request, urlopen
+
+
+DEFAULT_SECRET_NAMES = (
+    "repo:github.com/mozilla-releng/balrog:coveralls,"
+    "project/taskcluster/gecko/hgmointernal,"
```
bhearsum, line 15: These two are not actually secrets; just data that happens to be stored in the secrets service.

scripts/h1_redacted_secret_probe.py
```diff
@@ -0,0 +1,167 @@
...
+import sys
+from datetime import datetime, timezone
+from pathlib import Path
+from urllib.error import HTTPError, URLError
+from urllib.parse import quote
+from urllib.request import Request, urlopen
+
+
+DEFAULT_SECRET_NAMES = (
+    "repo:github.com/mozilla-releng/balrog:coveralls,"
```
bhearsum, line 14: This is kindof a secret, but not one that we care to protect in any way. We'd rather allow PRs from untrusted contributors to be allowed to send coverage reports and with rotation of this token if needed, than not have coverage from these PRs.

bhearsum (commented):
This is technically correct, I guess, but there's no real risks or attack here for the reasons described below.

PR #3812, opened 2026-06-10
Re-enable refresh tokens in the UI

Eijebong:
I had disabled them in #3468 because there were a problem due to re-requesting MFA when the client was trying to use it. The issue got fixed recently (in https://mozilla-hub.atlassian.net/browse/IAM-1935) so we should be able to use refresh tokens now which will fix the issue we're currently seeing where balrog sessions are expiring after 15mn (the authorization token expiration time).

I left `useRefreshTokensFallback` off because AFAIU it'd try to login the "normal" way (but in a hidden iframe) if the refresh token fails, which would end up requiring MFA anyway and thus break.

### DIFF
--- a/ui/src/App.jsx
+++ b/ui/src/App.jsx
@@ -38,6 +38,7 @@ const App = () => {
             scope: process.env.AUTH0_SCOPE,
           }}
           leeway={30}
+          useRefreshTokens
           cacheLocation="localstorage"
           sessionCheckExpiryDays={7}
         >


bhearsum (approved):
Presumably we can verify this in stage? (We have a prod deploy scheduled for tomorrow, so preferably you'd deploy this to stage today and test it.)

Eijebong:
Hopefully. I think worst case is that it's going to show the refresh failing and log you out anyway, which is the current behavior.

PR #3830, opened 2026-06-18
releases: avoid mutating a cached object

jcristau:
In get_release, copy the base blob before merging assets into it, to avoid modifying a shared object.
This was causing extra memory usage
(https://bugzilla.mozilla.org/show_bug.cgi?id=2014991#c6), because the releases cache ended up storing unwanted release assets.  And would cause a data race if/when we set UWSGI_THREADS to >1, with multiple request handlers modifying that object
(https://bugzilla.mozilla.org/show_bug.cgi?id=2003549#c7).

### DIFF
--- a/src/auslib/services/releases.py
+++ b/src/auslib/services/releases.py
@@ -383,7 +383,11 @@ def get_release(name, trans, include_sc=True):
         # the time we're serving old release data
         if base_row["data_version"] < base_data_version:
             base_row = get_base_row(name, trans)
-        base_blob = base_row["data"]
+        # Copy the base blob before merging assets into it below. In the public
+        # app base_row["data"] is a shared object handed back to every request.
+        # Merging assets in place would mutate the cached object, corrupting it
+        # for other callers and racing with concurrent readers.
+        base_blob = deepcopy(base_row["data"])
         data_versions["."] = base_row["data_version"]
 
     # same thing here for the assets -- if any of the full asset data versions
--- a/tests/web/api/test_releases.py
+++ b/tests/web/api/test_releases.py
@@ -1,3 +1,8 @@
+from copy import deepcopy
+
+from auslib.global_state import cache, dbo
+from auslib.services import releases
+
 from .base import CommonTestBase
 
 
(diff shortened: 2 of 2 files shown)


tests/web/api/test_releases.py
```diff
@@ -73,6 +78,34 @@ def test_get_release_new_tables(self):
...
+    def test_get_release_does_not_mutate_cached_blob(self):
+        # The public app runs the cache with make_copies=False, so the cached
+        # base blob is a single object shared across every request (and, with
+        # UWSGI_THREADS>1, across concurrent threads). get_release assembles the
+        # full release by merging the separately-cached assets into the base, and
+        # it must NOT do that merge in place on the shared cached object.
+        # Reproduce the public-app caching setup and assert the cached row is
+        # left untouched after a merge.
+        release = "Firefox-56.0-build1"
+        cache.reset()
```
Eijebong, line 90: Hrmf. Looking at the tests for cache related things did not spark joy.

Because `cache` is the real global object here. I don't know if xdist allows tests from different modules to run in the same process, but if it does, you'll want to force `cache.make_copies = False` here? Otherwise the test is kinda moot since the cache itself would deepcopy.

```
tests/test_db.py
3970:        cache.make_copies = True
4189:        cache.make_copies = True

tests/admin/views/base.py
48:        cache.make_copies = True

tests/util/test_cache.py
61:        cache.make_copies = True
71:        cache.make_copies = True
```

It might be nothing, but also if we ever add one of those `make_copies = True` in this module, the test becomes useless. Although honestly we might want to make a proper fixture that resets the cache settings every time

src/auslib/services/releases.py
```diff
@@ -383,7 +383,11 @@ def get_release(name, trans, include_sc=True):
         # the time we're serving old release data
         if base_row["data_version"] < base_data_version:
             base_row = get_base_row(name, trans)
-        base_blob = base_row["data"]
+        # Copy the base blob before merging assets into it below. In the public
+        # app base_row["data"] is a shared object handed back to every request.
+        # Merging assets in place would mutate the cached object, corrupting it
+        # for other callers and racing with concurrent readers.
+        base_blob = deepcopy(base_row["data"])
```
Eijebong, line 390: I don't think this is good enough re thread safety? I think the memory impact is null after this since we're not reinserting the computed blob in the cache but in `set_by_path(base_blob, path, asset["data"])` below, `asset` comes from the `release_assets` cache, which doesn't copy either AFAICT. So you're keeping `release_assets` references in blobs.

Eijebong: Looking at it some more I don't think it matters here because we're only ever reading from it. But it's a footgun

Eijebong (approved):
Looks good enough to me. Should at least help with memory. Hard to judge the impact on requests timing but we'll see I guess...

src/auslib/services/releases.py
jcristau: The base_blob object is per-request, the assets within it are shared, I think this is OK, we don't write to them.

(insert something here about how there's programming languages whose type system let you do this safely)

Eijebong: Yeah, we don't write to them now. I'm not expecting this to change but it spooked me during review and I wanted to point at it.

tests/web/api/test_releases.py
jcristau: Updated to explicitly set it to false in this test (and revert it on teardown/cleanup)

PR #3836, opened 2026-06-19
fix: require signoff when modifying only a base release or only adding a new asset

bhearsum:
We received a report that demonstrated that assets can be inserted or modified for protected releases could be inserted without signoff. It turns out that the check being fix here was broken in two ways: 1) It only required signoff if the base blob _and_ and an asset were being modified 2) It didn't compare assets correctly at all, so `current_assets != new_assets` was always true (this meant that #1 was mitigated...but only because of this bug).

This addresses both issues, and fixes https://bugzilla.mozilla.org/show_bug.cgi?id=2047458.

### DIFF
--- a/src/auslib/services/releases.py
+++ b/src/auslib/services/releases.py
@@ -606,8 +606,16 @@ def set_release(name, blob, product, old_data_versions, when, changed_by, trans)
     current_assets = get_assets(name, trans)
     base_blob, new_assets = split_release(blob, blob["schema_version"])
 
+    # munge the data to get it in a consistent format; current_assets is a dict,
+    # but contains non-blob columns from the database
+    # new_assets is a list of tuples of (path, asset)
+    # `path` is a tuple of the keys used to reach the `asset`, eg:
+    # ("platforms", "linux", "locales", "de") corresponds to
+    # blob["platforms"]["linux"]["locales"]["de"] in a reconstructed full blob
+    current_asset_data = {path: asset["data"] for path, asset in current_assets.items()}
+    new_asset_data = {"." + ".".join(path): asset for path, asset in new_assets}
     # if current == new, we will just be cancelling a scheduled change
-    if current_base_blob != base_blob and current_assets != new_assets and not when and any([v for v in live_on_product_channels.values()]):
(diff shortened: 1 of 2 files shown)


bhearsum:
ok, clearly i didn't test well enough locally first, sigh

src/auslib/services/releases.py
```diff
@@ -606,8 +606,13 @@ def set_release(name, blob, product, old_data_versions, when, changed_by, trans)
     current_assets = get_assets(name, trans)
     base_blob, new_assets = split_release(blob, blob["schema_version"])
 
+    # munge the data to get it in a consistent format; current_assets is a dict,
+    # but contains non-blob columns from the database
+    # new_assets is a list of tuples of (path, asset)
+    current_asset_data = {path: asset["data"] for path, asset in current_assets.items()}
+    new_asset_data = {"." + ".".join(path): item for path, item in new_assets}
```
ahal: Something seems off here, probably the comment.

`(path, asset)` makes me think the type is `Tuple[str, ...]`. But the `".".join(path)` makes it seem like `path` is itself a list?

Also small nit:
You refer to the value as `asset` everywhere except this line where it is called `item` for some reason.

ahal (approved):
Thanks, just double check my comment to make sure there's not a mistake or anything.

src/auslib/services/releases.py
bhearsum: Yeah, the path is _indeed_ an iterable (a tuple actually, it comes indirectly from [here](https://github.com/mozilla-releng/balrog/blob/02d1612d75eb7bdc69be917489f67e5f6c909006/src/auslib/services/releases.py#L120)). I admit this is not exactly clear, especially when you're not familiar with the wider context. The real fix for this is probably to stop calling this `path` (or perhaps, don't store it as a tuple of the top to bottom keys to get to the associated `asset`); those are obviously out of scope for this.

I'll add a comment here, although there's probably many other places in this file that this same confusion exists.

>  You refer to the value as asset everywhere except this line where it is called item for some reason.

oops. Yeah, I'll fix this.

PR #3840, opened 2026-06-24
fix: preserve `-sysaddon` in partner distribution fallback

ahal:
We have a fallback so partner distribution channels map to the official channels we use in Balrog, so e.g `release-cck-ubuntu` is translated to `release`.

But if you change your update url to something like `%CHANNEL%-sysaddon` for testing purposes, the current logic continues to translate it to `release`. Instead it should be `release-sysaddon`.

This gotcha caused me some confusion while trying to validate a new system addon release on my Firefox distributed by Canonical.

### DIFF
--- a/src/auslib/AUS.py
+++ b/src/auslib/AUS.py
@@ -61,7 +61,10 @@ def isForbiddenUrl(url, product, allowlistedDomains):
 
 
 def getFallbackChannel(channel):
-    return channel.split("-cck-")[0]
+    fallback = channel.split("-cck-")[0]
+    if channel.endswith("-sysaddon"):
+        fallback += "-sysaddon"
+    return fallback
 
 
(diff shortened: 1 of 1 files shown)


ahal:
I guess this fallback isn't system addon specific.. so hardcoding `-sysaddon` isn't great.. But it also doesn't seem harmful. I'm open to other suggestions

jcristau (changes requested):
I'd rather we adjust the testing instructions to work for the distribution case on the client side than add systemaddon specific hacks in here, if possible.

ahal:
Sounds good, I updated:
https://mozilla-hub.atlassian.net/wiki/spaces/RelEng/pages/1052999723/System+Addon+Releases#Testing-Live-Updates

PR #3841, opened 2026-06-25
initdb_and_run.sh: exit on error

jcristau:
Prevent the server from starting if the db upgrade fails.


Eijebong (approved):

jcristau:
Actually this is broken, the first thing we do is run a command that can fail (`grep`) and check its exit code, expecting it not to have terminated the script.

PR #3874, opened 2026-09-02
Bug 2063240 - avoid xml injection

jcristau:


src/auslib/web/public/client.py
```diff
@@ -241,8 +241,12 @@ def construct_response(release, query, update_type, response_blobs, squash_respo
         # In case of superblob Extracting Header form parent release
         xml.append(release.getInnerFooterXML(query, update_type, app.config["ALLOWLISTED_DOMAINS"], app.config["SPECIAL_FORCE_HOSTS"]))
         xml.append(release.getFooterXML())
-        # ensure valid xml by using the right entity for ampersand
-        xml = re.sub("&(?!amp;)", "&amp;", "\n".join(xml))
+        # ensure valid xml by using the right entity for ampersand. This is a
+        # safety net for any value not already escaped by the blob layer; the
```
bhearsum: So to be clear: this is a guard against us forgetting to escape earlier in the future? (If not, why not ensure we escape things before they get here?)

bhearsum (approved):
This looks pretty sensible and clear to me.

src/auslib/web/public/client.py
jcristau: Yeah this is superfluous, I'll drop it.

PR #3878, opened 2026-09-15
ui: add a Cancel Scheduled Change button for rules

jcristau:
The UI could only remove a scheduled *insert* (via the Delete button); a pending scheduled update or delete on an existing rule had no cancel affordance. Clicking Delete on such a rule instead tried to schedule a second change and was rejected with ChangeScheduledError, so a stale or unwanted scheduled change (as in #3875) could only be cleared through the API.

Add a Cancel Scheduled Change button on rule cards that have a pending update/delete, which deletes the scheduled change (leaving the rule itself untouched) after a confirmation dialog. The button's visibility rule is extracted to canCancelScheduledChange() in utils/rules and unit tested.

Refs #3875


bhearsum (approved):
Looks correct to me; I'll try it out in stage.

PR #3899, opened 2026-09-15
public: return 400 instead of a 500 when the JSON API serves a non-Guardian blob

### LINKED ISSUE #3871: AttributeError: 'DesupportBlob' object has no attribute 'getResponse'
Sentry Issue: [BALROG-PROD-PUBLIC-A0](https://mozilla.sentry.io/issues/7687469431/?referrer=github_integration)

```
AttributeError: 'DesupportBlob' object has no attribute 'getResponse'
(3 additional frame(s) were not displayed)
...
  File "auslib/web/public/json.py", line 23, in get_update
    response = json.dumps(release.getResponse(parameters, app.config["ALLOWLISTED_DOMAINS"]))
```

niukanen1:
Fixes #3871.

`/json/*` update endpoints assumed any blob matched by a rule provides `getResponse()`. `DesupportBlob` and other `XMLBlob` subclasses don't, so a rule pointing at a desupport release made `/json/1/<product>/...` crash with `AttributeError: 'DesupportBlob' object has no attribute 'getResponse'` and surface as a 500 in production (Sentry: BALROG-PROD-PUBLIC-A0).

The XML endpoint already guards against this with an explicit blob type check (`if not isinstance(release, XMLBlob): raise BadDataError("Wrong blob type")` in `auslib/web/public/client.py`), which the Flask error handler turns into a 400. This adds the mirror check for the JSON endpoint: only `GuardianBlob` provides the JSON response shape, so anything else is now a 400 with `Wrong blob type` instead of an unhandled exception.

**Testing**

- New regression test `testJSONForDesupportBlob` in `tests/web/test_json.py`: a rule mapping to a DesupportBlob served via `/json/1/...` must return 400 (previously: AttributeError → 500).
- `pytest tests/web/test_json.py` — 36 tests pass locally.


jcristau (approved):
Thanks!

PR #3902, opened 2026-09-17
tests: actually run the async ReleasesJSON tests

jcristau:
TestReleasesJSON derives from unittest.TestCase, and pytest-asyncio's marker has no effect there: each `async def test...` was collected, called, and its coroutine discarded without ever being awaited, so the bodies never executed.

Derive from unittest.IsolatedAsyncioTestCase so the coroutines are awaited. testInsertCreatesCorrectHistory, testUpdateCreatesCorrectHistory and testDeleteCreatesCorrectHistory now run, and pass.


bhearsum (approved):
oops

PR #3904, opened 2026-09-18
Bug 2065063 - fix required signoff bypass

jcristau:
We were allowing changes to releases without signoffs when:
- editing/adding a release that a rule maps to but that didn't already exist in its table (releases or releases_json)
- editing a release that is indirectly referenced via a SuperBlob


tests/admin/views/test_releases_v2.py
```diff
@@ -623,6 +637,65 @@ def test_put_of_new_asset_fails_when_signoff_required(api, firefox_56_0_build1):
...
+    blob["name"] = "leaf-child"
+
+    ret = api.put("/v2/releases/leaf-child", json={"blob": blob, "product": "Firefox"})
+    assert ret.status_code == 400, ret.data
+    assert not dbo.releases_json.select(where={"name": "leaf-child"})
+
+
+@pytest.mark.usefixtures("releases_db")
+def test_put_of_unwritten_mapped_release_fails_when_signoff_required(api, firefox_56_0_build1, mock_verified_userinfo):
+    # Rule 7 maps to a name that exists in neither release table. Writing it for the
```
bhearsum, line 684: I'm surprised this is allowed...obviously `mapping` can't be a foreign key because of the split-brain release table situation at the moment, but it seems like the preferable fix for this is to refuse to allow rules to map to releases that don't exist.

bhearsum (commented):
I'm not sure I'm totally sold on the entire new table to track references. It seems like most of the potential performance issues being solved with it are from `getPotentialRequiredSignoffsForReleaseNames`. Can we solve that with indexes on the Releases tables rather than the new (almost single purpose) table?

If we need it we need it, of course...but it would be great to make sure other options have been explored, if they haven't already.

jcristau:
> I'm not sure I'm totally sold on the entire new table to track references. It seems like most of the potential performance issues being solved with it are from `getPotentialRequiredSignoffsForReleaseNames`. Can we solve that with indexes on the Releases tables rather than the new (almost single purpose) table?

The references are in the releases.data / releases_json.data columns, an index on the Releases table is basically what this new table is.  On mysql we could potentially build an index that parses the json `data` column, but on sqlite AFAIK we'd need the shadow table anyway.  Plus we'd have to keep the index expression in sync with the getResponseBlobs method.

bhearsum:
> > I'm not sure I'm totally sold on the entire new table to track references. It seems like most of the potential performance issues being solved with it are from `getPotentialRequiredSignoffsForReleaseNames`. Can we solve that with indexes on the Releases tables rather than the new (almost single purpose) table?
> 
> The references are in the releases.data / releases_json.data columns, an index on the Releases table is basically what this new table is. On mysql we could potentially build an index that parses the json `data` column, but on sqlite AFAIK we'd need the shadow table anyway. Plus we'd have to keep the index expression in sync with the getResponseBlobs method.

Right, OK, that's a pretty solid reason to keep the table.

tests/admin/views/test_releases_v2.py
jcristau: We already refuse them at the web layer (https://github.com/mozilla-releng/balrog/blob/30d59922dcdd126c26590a66b06bef4f4fcdeb7a/src/auslib/web/admin/views/rules.py#L58-L62) but at the db layer we can't really do that because of the split-brain, so I'm not confident there's not some kind of path to that situation (e.g. delete the release as there's a scheduled rule change that points at it, or something like that).

bhearsum (approved):

